Issue #229, opened 2025-07-14
Issues Testing Session 14-JUL-25


ovitrif:
- [x] 1. Create wallet - issue for JC having to tap 3 times (slow UI update?!)
  `cant-repro`
- [x] 2. Receive sheet - issue with navigation bar showing behind the UI - e8cf15a7
- [x] 3. "Copied" system notification seen after sending - d94c688
- [x] 4. min send limit for onchain is 601, should be 547
- [x] 5. JC: tx fee same as tx value, payment = 0
  `duplicate of 7`
- [x] 6. cpfp boost - Piotr - on init - bitkit was unable to boost error
  `non-issue` - ldk-node logic - correct accounting-wise
- [x] 7. send to self address - amount of funds seen in activity is not correct
  `non-issue` - ldk-node logic - correct accounting-wise
- [x] 8. LN pay notification - delayed OS notification received while in background
  `regtest-notifications-server-bug`, fixed, should work now
- [x] 9. send LN routing fees - hardcoded, need implementation
  `missing-ldk-node` need ldk-node feature
- [x] 10. menu > widgets > should add toggle to enable widgets (same as settings)
  `waiting-for-figma-update`
- [x] 11. backup sheet > last > PL lang uses `<brand>`
- [x] 12. LN connections > close channel > too many notifications
- [x] 13. No warning when attempting to send > 50% of the funds
- [x] 14. pin code enabled settings should not be restored on wallet restore
- [x] 15. empty wallet view reappears after tapping on `x` to close it
- [x] 16. receive > share - should share QR code image, not only text
- [x] 17. bip21 should not have empty parameter when bolt11 is missing: `?lightning=`
- [x] 18. JC can't transfer to spending
  #137

___
see [Bitkit Android Testing Sheet Report](https://docs.google.com/spreadsheets/d/10_DufEPwKUCExkXF-jTY7njE0mh5ZspwhRw42NBBd-o/edit?gid=750686172#gid=750686172)

coreyphillips:
Accounting-wise, ldk-node is handling this correctly. A self-send should have an amount of 0 along with the fee paid. If we want to abstract the self-send amount for the UX we'll need to do-so outside of ldk-node.

PR #260, opened 2025-07-28
Adapt BG image of SpendingConfirmScreen to small screens

jvsena42:
Replated to #229 

### Description

The Swipe component was hidden on small screens because the coin image was using too much space

### Preview

image


ovitrif:
@jvsena42 I'm preparing a PR to deal with those irrelevant alerts from detekt

PR #262, opened 2025-07-28
Fix build dependency resolution when cache is missing

ovitrif:
This PR fixes failing builds on new machines which has no dependencies cache.

### Description

This command should work after removing all local gradle/maven dependency caches:
```
./gradlew --refresh-dependencies clean assembleDevDebug
```

It was failing for our own dependencies served via jitpack:
- bitkit-core
- ldk-node (synonym fork)

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -12,7 +12,6 @@ dependencyResolutionManagement {
         mavenLocal()
         google()
         mavenCentral()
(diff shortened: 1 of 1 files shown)


ovitrif:
Thanks @piotr-iohk for finding this issue

jvsena42 (approved):

PR #281, opened 2025-08-05
Send Custom Speed

ovitrif:
`todo`

### Description
`todo`

### Preview
`todo`

### QA Notes
`todo`


ovitrif:
The issue where we are seeing different fees is coming from `ldk-node`, we probably have another bug which causes different fee value returned when calculating fee for the same inputs values:

Android Studio 2025-08-06 000079

ovitrif:
The issue where we are seeing different fees is coming from `ldk-node`.

Android Studio 2025-08-06 000079

We get different return value each time we call `node.onchainPayment().calculateTotalFee()` with `utxosToSpend=null`, even if all the other params are the same.

The underlying issue seems to be related to which UTXOs are selected. If we pass the same list of UTXOs we do get the same fee returned each time.

If we don't pass UTXOs, it's probably almost always estimating for different utxos.

I noticed the same issue when calling `node.onchainPayment().selectUtxosWithAlgorithm()` with an `BRANCH_AND_BOUND` (didn't check with other algorithms).
Very often it returns a different number of UTXOs.

This makes developing a reliable fee estimation system quite difficult in the native app, bloating the codebase with hacks to manage state, all for not having to re-estimate fees unless absolutely necessarily.
I already had to re-refactor many times as I was discovering more clearly the root cause of the bugs I saw on UI.

Even after all the patch fixing, I still encounter an issue which I can't properly solve:
🐞  When user selects another fee rate, I _must_ re-run auto UTXOs selection using the current algorithm, because the old list of UTXOs could be invalid.
Currently this is what I am doing, but because of it, when selecting another fee rate, the fee estimate on the review screen gets changed, so it's not displaying the same value we selected on the previous screen.

cc @coreyphillips, would appreciate your input 🙏🏻

coreyphillips:
Hey @ovitrif! I may need to consider adding a few comments or updating the documents around fee estimation.

The randomness around fees when passing `null` to `utxosToSpend` is due to how BDK selects UTXO's. Effectively, When `calculate_total_fee()` is called without specifying UTXOs, it delegates to BDK's `build_tx()` which uses random selection, resulting in different UTXOs and thus different fees each time. So if we want to remove this randomness we'll need to pass it specific UTXO's.

Otherwise, any time we adjust the desired fee-rate, value and/or UTXO set we'll need to re-run the fee-estimation method since adjusting these values in any way may result in a change to the resulting fee that the user will need to pay.

ovitrif:
> Hey @ovitrif! I may need to consider adding a few comments or updating the documents around fee estimation.
> 
> The randomness around fees when passing `null` to `utxosToSpend` is due to how BDK selects UTXO's. Effectively, When `calculate_total_fee()` is called without specifying UTXOs, it delegates to BDK's `build_tx()` which uses random selection, resulting in different UTXOs and thus different fees each time. So if we want to remove this randomness we'll need to pass it specific UTXO's.
> 
> Otherwise, any time we adjust the desired fee-rate, value and/or UTXO set we'll need to re-run the fee-estimation method since adjusting these values in any way may result in a change to the resulting fee that the user will need to pay.

Thank you @coreyphillips 🙏🏻 

Makes sense to me, would've helped to know in advance so I could've avoided the trial-and-error approach I had to take to figure out how to optimise this so I can get less bugs on the UI and a more-deterministic fee calculation.

Updating the documentation + comments in code if needed (especially if we want to open a PR with upstream) is a good solution IMHO 🙏🏻 

cc. @pwltr - knowing this in advance will help get it right from the start on iOS side 🙏🏻

PR #284, opened 2025-08-07
Send Set Custom Fee

ovitrif:
Related: #49 
Design: [Figma v57 Send - Set Custom Fee](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=33495-112240&t=ySVfs1PZaWFEhIeT-4)

This PR adds the feature to set a custom fee rate in satPerVb during the onchain send flow.

### Description
- Send custom fee screen UI & logic
- Use pixel perfect sizing for sheet topbar and drag handle
- Fix to navigate back on speed click in settings
- Show spinner while loading fee amount on review screen
- Validate against max and min fee limits and disable invalid speeds
  - Fee must not exceed 50% of available balance and it must be less than the remaining funds after send

### Preview
https://github.com/user-attachments/assets/335523fd-5107-4c20-99e3-1a49e1697c35

### QA Notes
Test Custom Fee Screen:
- test with custom fee already set via settings
- test by setting different custom speeds
- test by switching from fee rate preset to custom
- test validation by using amount close to max

### DIFF
--- a/app/src/main/java/to/bitkit/ui/screens/transfer/SettingUpScreen.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/transfer/SettingUpScreen.kt
@@ -1,13 +1,11 @@
 package to.bitkit.ui.screens.transfer
 
 import androidx.compose.foundation.Image
-import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
 import androidx.compose.material3.CenterAlignedTopAppBar
 import androidx.compose.material3.ExperimentalMaterial3Api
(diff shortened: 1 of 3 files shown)


jvsena42:
I see a tendency for wallets to allow a fee rate smaller than 1. I wonder when we should start thinking about this to Bitkit

app/src/main/java/to/bitkit/ui/screens/wallets/send/SendFeeCustomScreen.kt
```diff
@@ -53,17 +76,33 @@ private fun Content(
...
+            LargeRow(
+                prefix = null,
+                text = input.ifEmpty { "0" },
+                symbol = BITCOIN_SYMBOL,
+                showSymbol = true,
+            )
+
+            if (isValid) {
+                VerticalSpacer(28.dp)
+                BodyM(totalFeeText, color = Colors.White64)
```
jvsena42, line 93: This text could have an ellipsis middle to handle big fee values, but it can be done in another PR

image

jvsena42 (approved):

PR #294, opened 2025-08-11
chore: use bitkit-core from github packages

ovitrif:
Related: 
- https://github.com/synonymdev/bitkit-core/pull/32
- https://github.com/synonymdev/bitkit-core/pull/33
- https://github.com/synonymdev/bitkit-core/pull/34

### Description

This PR updates the build config to consume the `bitkit-core` dependency from GitHub packages instead of Jitpack.

**Important** Check the [updated bindings README](https://github.com/synonymdev/bitkit-core/tree/master/bindings/android#readme), you will need to [Setup your GitHub credentials](https://github.com/synonymdev/bitkit-core/tree/master/bindings/android#1-setup-your-github-credentials) in order to be able to build locally using the dependency from the GitHub packages maven repository.

### Preview

`n/a`

### QA Notes

Test that building succeeds and the new app runs without errors (existing wallet).


ovitrif:
Fixed the issue with missing credentials in the GH workflows via 059ed83

ovitrif:
> `Could not GET 'https://maven.pkg.github.com/synonymdev/bitkit-core/com/synonym/bitkit-core-android/0.1.10/bitkit-core-android-0.1.10.pom'. Received status code 401 from server: Unauthorized`

My bad, fixed the wrong README instructions in e7aa04f 🙏🏻

jvsena42 (dismissed):
Tested

- Set credentials in local.properties
- Sync and build the app

ovitrif:
Nice enhancement to also support local.properties!

Was thinking though...
Can you pls use the same name for the keys?

It's a bit confusing now, with different prop keys in local.properties vs gradle.properties.

That would also align with the readme, both in this repo and the repos of the dependencies.

ovitrif:
> > Nice enhancement to also support local.properties!
> > 
> > Was thinking though... Can you pls use the same name for the keys?
> > 
> > It's a bit confusing now, with different prop keys in local.properties vs gradle.properties.
> > 
> > That would also align with the readme, both in this repo and the repos of the dependencies.
> 
> Do you mean like this?
> 
> ```
> username = System.getenv("GITHUB_USER") ?: providers.gradleProperty("gpr.user").orNull
> password = System.getenv("GITHUB_KEY") ?: providers.gradleProperty("gpr.key").orNull
> ```

No, I mean in local.properties, should also use if possible the same key names:
- gpr.user
- gpr.key 🙏🏻

pwltr (approved):

Issue #149, opened 2025-05-19
[FEATURE] Suggestion Cards


jvsena42:
[FIGMA](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=31760-203721&t=E5H5HCNBHpeWkaMf-4)

- What suggestions are shown to the user depends on the wallet state
- Suggestions can be removed by user
- Suggestions can be reset by user (respawn)
- Progress cards are shown temporarily during a particular process.

### Todo
- [ ] Card for "In Progress" transactions

PR #300, opened 2025-08-14
Progress cards

jvsena42:
[FIGMA](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=33495-112450&t=Bly0N2Var5Fhb5vQ-4)
Related to #149 

### Description
This PR implements the progress cards in the suggestion flow

- [x] Create Progress Suggestion types
- [x] Create logic to display
  - [x] Channel force close
  - [x] Chanel coop close
  - [x] Transfer to spending
  - [x] Transfer ready
- [x] Collect changes in the Home and update the suggestion list
- [x] Implement radial gradient animation on Suggestion cards
- [x] Implement logic to replace lightning card with progress cards

### Preview

https://github.com/user-attachments/assets/53ca61f8-b5c2-40b8-a2fc-6dab2a95a800

[transfer_android.webm](https://github.com/user-attachments/assets/6f302b65-789a-4cd8-8a6e-4e2dc9f1f8d5)

### QA Notes

- Coop close and force closed are not being cached yet. It will be done on an issue related to transfer
- The navigation will be implemented in other PR

How to reproduce:

1. Purchase a channel - > navigate to home -> The transfer in progress card should be displayed
2. Confirm the in-progress transfer -> The lighting ready card should be displayed


app/src/main/java/to/bitkit/models/Suggestion.kt
```diff
@@ -10,62 +10,98 @@ enum class Suggestion(
...
     ),
     QUICK_PAY(
         title = R.string.cards__quickpay__title,
         description = R.string.cards__quickpay__description,
-        color = Colors.Green,
+        color = Colors.Green24,
         icon = R.drawable.fast_forward
     ),
+
+    /**Replaces SPEND when a LN channel is being force closed*/
```
ovitrif, line 72: nit: `SPEND` got renamed to `LIGHTNING`

app/src/main/java/to/bitkit/models/Suggestion.kt
```diff
@@ -10,62 +10,98 @@ enum class Suggestion(
...
+    TRANSFER_CLOSING_CHANNEL(
+        title = R.string.cards__transferClosingChannel__title,
+        description = R.string.cards__transferClosingChannel__description,
+        color = Colors.Red24,
+        icon = R.drawable.transfer,
+        dismissible = false
+    ),
+
+    /**Replaces LIGHTNING when the transfer to spending balance is in progress*/
+    LIGHTNING_SETTING_UP(
```
ovitrif, line 91: The `LIGHTNING_SETTING_UP` name will not work with
```kt
modifier = Modifier.testTag("Suggestion-${item.name.lowercase()}")`
```
from the `HomeScreen.kt` to generate the testTag expected by e2e tests:
- `Suggestion-lightningSettingUp`

Either update the enum constant name (using `@Suppress("EnumEntryName")`) or update this entry in the `e2e-test-ids.md` to reveal that it maps to another testTag:
```md
| Suggestion-lightningSettingUp | ❌ |
<!-- change into: -->
| Suggestion-lightningSettingUp | Suggestion-lightning_setting_up |
```

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
```diff
@@ -278,20 +284,33 @@ class ActivityRepo @Inject constructor(
...
-                                transferTxId = activityMetaData.transferTxId
-                            )
+                            v1 = onChainActivity
                         )
 
                         updateActivity(
                             id = updatedActivity.v1.id,
                             activity = updatedActivity
                         ).onSuccess {
+                            if (onChainActivity.isTransfer && onChainActivity.doesExist) {
```
ovitrif, line 302: Wondering what is the reason for the `onChainActivity.doesExist` in the condition?

I noticed we never assign `false` to it in `CoreService` 🤷🏻 .

PS. Not saying it isn't needed, just trying to figure out how we could manage the `doesExist` value properly.

app/src/main/java/to/bitkit/ui/components/SuggestionCard.kt
```diff
@@ -38,14 +54,57 @@ fun SuggestionCard(
...
+
+                        else -> Triple(
+                            gradientColor,
+                            gradientColor,
+                            gradientColor.copy(alpha = MIN_ALPHA_GRADIENT)
+                        )
+                    }
+
+                    Modifier
+                        .gradientRadialBackground(gradientSelectedColor, glowAlpha)
```
ovitrif, line 104: nit: we only use `gradientRadialBackground` in `SuggestionCard.kt`, it could've been a private modifier as well, scoped to this file.

That would've avoided the need to rename `gradientModifier` which is used a lot, into `gradientLinearBackground` 🙃

app/src/main/java/to/bitkit/models/Suggestion.kt
```diff
@@ -10,62 +10,98 @@ enum class Suggestion(
...
     BUY(
         title = R.string.cards__buyBitcoin__title,
         description = R.string.cards__buyBitcoin__description,
-        color = Colors.Brand,
-        icon = R.drawable.b_emboss
+        color = Colors.Brand24,
+        icon = R.drawable.b_emboss,
     ),
     LIGHTNING(
+        // Lightning ready from RN
```
ovitrif, line 23: nit: this is NOT `lightningReady` from RN, `LIGHTNING_READY` is 👍🏻

app/src/main/java/to/bitkit/ui/screens/wallets/HomeViewModel.kt
```diff
@@ -255,11 +258,20 @@ class HomeViewModel @Inject constructor(
     private fun createSuggestionsFlow() = combine(
```
ovitrif, line 258: nit: We should refactor `createSuggestionsFlow` soon to use only one `listOfNotNull` builder, it's becoming too confusing having to maintain 3 different lists

ovitrif (approved):
LGMT 🎉 

added comments for some cleanup suggestions but nothing blocking

Mainly the comment about the e2e test id `Suggestion-lightningSettingUp` should be addresses ASAP in a new PR.

app/src/main/java/to/bitkit/ui/components/SuggestionCard.kt
jvsena42: Good point, it could be a single modifier with a `isLinear=true` parameter

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
jvsena42: I thought in extremely rare cases, but not impossible, like setting a fee too low so the transfer is removed from the mempool, or using RBF

app/src/main/java/to/bitkit/ui/screens/wallets/HomeViewModel.kt
jvsena42: agree

app/src/main/java/to/bitkit/models/Suggestion.kt
jvsena42: Thanks! I'll handle it in the next PR

jvsena42: Fixed in #303

PR #311, opened 2025-08-20
Wait for node starts in SpendingAmountScreen

jvsena42:
### Description
Some Blocktank methods need  the node running to work properly

- [x] Display SyncNodeView on the SpendingAmountScreen when the node is initializing
- [x] Wait for the node run on the required methods
- [x] Removed the now unnecessary retry logic

### Preview

https://github.com/user-attachments/assets/ba02f8ee-82c3-4f8a-9fe8-15b94a2aa216

### QA Notes

Tested:

- [x] Open the transfer amount screen before the node starts
- [x] Perform the full transfer flow


ovitrif (changes requested):
LGTM, just waiting for that one UI regression fix from prev PR

**Tested** with new transfer while node wasn't ready yet.

ovitrif (approved):

PR #320, opened 2025-08-25
Make tagInputTestTag and addButtonTestTag visible

piotr-iohk:
### Description

The test tags (i.e. `resource-ids`) where not visible in the add tag screen until explicitely adding `.semantics { testTagsAsResourceId = true } ` in this particular component function, that is used on the screen. Probably this component not created/rendered from the main scope where `testTagsAsResourceId = true` is set globally. Maybe there is a better way to do this, but this fixes the problem and `resource-id` is now rendered.


jvsena42 (changes requested):
Only missing the lint update. On Windows or Linux you can do `ctrl+alt+L`

jvsena42:
Probably the component test tag was merged
https://github.com/appium/appium/issues/15138

Issue #268, opened 2025-07-30
Issues Testing Session 30-JUL-25


ovitrif:
- [ ] 1. require pin when idle - didn't get asked, Ovi: got it only after disabling and re-enabling the feature
- [ ] 2. send - onchain - receiver sees no confetti — `missing-in-ldk-node` missing onchain payment event
- [ ] 3. send - onchain - receiver sees no push notification — `missing-in-ldk-node` missing onchain payment event
- [ ] 4. send - self-send - activity shows amount=fee and payment=0 — `non-issue` - correct accounting-wise
- [ ] 5. boost - Ovi: activity item didn’t update after rbf boost, Piotr: cpfp boosted the same tx Ovi tried to rbf.
- [ ] 6. transfer to spending - doesn’t work, loading never ends — missing call to `onTransferToSpendingConfirm` , order never watched
- [ ] 7. transfer to spending - api error when trying to transfer 10 000 - should say something like: amount is too small
- [ ] 8. transfer to spending - continue using bitkit - order never finalizes — app doesn’t keep watching order for status updates after restart
- [ ] 9. bitrefill - menu > shop doesn’t show intro screen
- [ ] 10. bitrefill - webview not picking language of phone/app
- [ ] 11. pin - red text after wrong pin missing space in PL language
- [ ] 12. onboarding - instant payments slide - bottom sentence should only show when geo-blocked

Biggest bug: Transfer to spending broken - couldn’t test lightning.

---
see [Bitkit Android Testing Sheet - 30 Jul 2025](https://docs.google.com/spreadsheets/d/10_DufEPwKUCExkXF-jTY7njE0mh5ZspwhRw42NBBd-o/edit?gid=1366894933#gid=1366894933)

Main blocker (regression bug): 6. Transfer to spending broken - couldn’t test lightning.

jvsena42:
> 1. require pin when idle - didn't get asked, Ovi: got it only after disabling and re-enabling the feature

Couldn't reproduce this one. Discussion about this feature
https://synonymworkspace.slack.com/archives/C07BJ7DNPCG/p1754647706979969

ovitrif:
> > 1. require pin when idle - didn't get asked, Ovi: got it only after disabling and re-enabling the feature
> 
> Couldn't reproduce this one. Discussion about removing this feature https://synonymworkspace.slack.com/archives/C07BJ7DNPCG/p1754647706979969

[...]

PR #332, opened 2025-08-29
Remove pin on idle

jvsena42:
Related to #268
[FIGMA](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=33495-111106#1379134822)
### Description
This PR removes the "Require pin on idle" feature
[Slack thread](https://synonymworkspace.slack.com/archives/C07BJ7DNPCG/p1754647706979969)

### Preview

https://github.com/user-attachments/assets/1b01ce05-9dba-4dd4-a45b-ee98fc1bf730

### QA Notes
Tested:

- [x] On old install -> Enable Require pin on idle" and pin -> Install this version -> Wait 90 second in iddle -> shouldn,'t require pin
- [x] On fresh install -> Enable pin ->  Wait 90 second in iddle -> shouldn,'t require pin

### DIFF
--- a/app/src/test/java/to/bitkit/repositories/ActivityRepoTest.kt
+++ b/app/src/test/java/to/bitkit/repositories/ActivityRepoTest.kt
@@ -19,10 +19,12 @@ import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import org.mockito.kotlin.wheneverBlocking
 import to.bitkit.data.AppCacheData
+import to.bitkit.data.AppDb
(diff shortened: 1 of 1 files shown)


ovitrif (commented):
@jvsena42 

Wasn't this feature supposed to be changed into a different implementation?

I still see the feature on iOS and it works as such:

When switching to another app (ie. Bitkit app gets backgrounded), then returning to Bitkit, pin is required to be re-entered.

AFAIK we should do the same instead of removing the feature altogether.

PR #335, opened 2025-09-01
Make ScanPrompt, QRDialog, QRInput visible

piotr-iohk:
### Description

Make ScanPrompt, QRDialog, QRInput visible as resource-id. Similar case as in https://github.com/synonymdev/bitkit-android/pull/320.


ovitrif (approved):
Approved but ideally we should find a general fix, instead of filling the codebase with this modifier.

piotr-iohk:
@ovitrif Yes, I've been seing this every now and then on some dialogs etc. like:
 - https://github.com/synonymdev/bitkit-android/pull/320
 - https://github.com/synonymdev/bitkit-android/pull/326

I'm not sure where some global modifier should be placed such that it fixes this generally though.

ovitrif:
@piotr-iohk Makes perfect sense. It's because the sheets and dialogs are not in the same UI tree, they're added separately by the Android framework.

> I'm not sure where some global modifier should be placed such that it fixes this generally though.

Will look into a fix myself 👍🏻 , it's not that easy, there won't be just a single place, but more.

PR #339, opened 2025-09-02
Set testTagsAsResourceId on AppAlertDialog

piotr-iohk:
### Description

Set testTagsAsResourceId on AppAlertDialog.


ovitrif:
Nice, this is the way 🚀

PR #345, opened 2025-09-04
Simplify requires pin on idle logic

jvsena42:
closes #332 
Related to #268 

### Description

This PR replaces the "require pin on idle" logic with one that requires it every time the app is resumed
[Slack thread](https://synonymworkspace.slack.com/archives/C07BJ7DNPCG/p1754647706979969)

### Preview

https://github.com/user-attachments/assets/c86cd227-aea6-4e05-9a28-36ced8d1fb92

### QA Notes
Tested:

- Enable the feature -> but the app in the background -> should require the pin
- Disables the feature -> but the app in the background -> should not require the pin


ovitrif (approved):
LGTM :shipit: 

### Tests
- pin on idle on:
  - switch to home > resume app 🟢 
  - switch to another app > resume 🟢 
  - open app drawer > resume > expect not having to enter pin 🟢

PR #363, opened 2025-09-09
fix: App crash if phone is ofline on startup

jvsena42:

### DIFF
--- a/app/src/main/java/to/bitkit/async/ServiceQueue.kt
+++ b/app/src/main/java/to/bitkit/async/ServiceQueue.kt
@@ -6,11 +6,11 @@ import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
-import kotlinx.io.IOException
 import to.bitkit.ext.callerName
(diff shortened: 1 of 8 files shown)


jvsena42:
changed to draft to fix `@settings_10 - Can enter wrong Electrum server and get an error message` E2E test

ovitrif (changes requested):
Ran both test cases successfully with a caveat on test case 2

> Phone online -> open app -> should display successfull status -> phone offline -> should update failure status -> online again -> resync and update status again -> send bitcoin with success

ended up with logs spam after restoring internet connection then trying swipe-to-refresh (all coming from `lightingRepo.sync()`:
```
ERROR❌️: ServiceQueue.LDK error [AppError='TxSyncTimeout='Syncing transactions timed out.'']
```

The error also shows up as toast after each pull-to-refresh:


But otherwise send and receive works both for onchain and LN, with one problem for LN send: `estimateRoutingFeesForAmount` takes a lot of waiting time, freezing the send flow for a while.

---

### Remarks
- Suggesting to keep retrying in a loop, instead of stopping after 5 attempts.

If I wait long enough during test case 1, nothing works anymore even if I get back online.

jvsena42:
Set as draft to apply changes and retest

PR #377, opened 2025-09-15
Set Copylot code review instructions

jvsena42:
### Description
Instructions to Copilot perform code reviews
[Reference](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review)

### Preview

`N/A`

### QA Notes
`N/A`

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -8,7 +8,7 @@ When performing a code review, ensure ViewModels are never injected as dependenc
 
 When performing a code review, verify that all async operations use `viewModelScope.launch` instead of `GlobalScope.launch` in ViewModels.
 
-When performing a code review, ensure Repository pattern is followed with proper data source abstraction.
+When performing a code review, ensure Repository pattern is followed
 
 When performing a code review, verify StateFlow is used for reactive state management in ViewModels with proper `_uiState` and `uiState` pattern.
 
@@ -17,10 +17,6 @@ When performing a code review, verify StateFlow is used for reactive state manag
 When performing a code review, flag any use of the not-null assertion operator (`!!`) and suggest safe calls (`?.`) with proper null handling.
 
 When performing a code review, ensure all coroutine operations use proper error handling with `runCatching`
-
-When performing a code review, verify that all Logger calls include the TAG as context parameter, e.g., `Logger.warn("message", e = e, context = TAG)`.
-
-When performing a code review, check that methods exist before they are called, especially after refactoring.
 
 ## UI & Compose Best Practices
 
@@ -34,7 +30,7 @@ When performing a code review, ensure proper state management patterns with `Mut
 
 ## Code Quality & Readability
 
-When performing a code review, focus on readability and avoid nested ternary operators.
+When performing a code review, focus on readability and avoid nested if-else, replacing with early return wherever possible.
 
 When performing a code review, ensure unused code is removed after refactoring.
 
@@ -48,11 +44,7 @@ When performing a code review, ensure services contain business logic and don't 
 
 ## Build & Testing
 
-When performing a code review, ensure proper build variant usage (dev for regtest, tnet for testnet).
-
-When performing a code review, verify that unit tests follow the pattern `./gradlew testDevDebugUnitTest --tests ClassName`.
-
-When performing a code review, suggest Unit Test com composable components and business logic covering the most important cases
+When performing a code review, suggest Unit Test for composable components and business logic covering the most important cases
 
 ## Lightning & Bitcoin Specific
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -48,8 +48,6 @@ When performing a code review, suggest Unit Test for composable components and b
 
 ## Lightning & Bitcoin Specific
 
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+When performing a code review, respond in English.
+
+## Architecture & Patterns
+
+When performing a code review, ensure ViewModels are never injected as dependencies into services or repositories. Only Android activities and composable functions should use ViewModels.
+
+When performing a code review, verify that all async operations use `viewModelScope.launch` instead of `GlobalScope.launch` in ViewModels.
+
+When performing a code review, ensure Repository pattern is followed with proper data source abstraction.
```
ovitrif, line 11: > Repository pattern is followed with proper data source abstraction

nit: pointless over engineering IMHO, we're not going to create an interface for a single class, just to make it happy.

PS. In fact interfaces were not even supposed to be used this way, SOLID principles' take on interface segregation is more valuable.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Error Handling & Safety
+
+When performing a code review, flag any use of the not-null assertion operator (`!!`) and suggest safe calls (`?.`) with proper null handling.
+
+When performing a code review, ensure all coroutine operations use proper error handling with `runCatching`
+
+When performing a code review, verify that all Logger calls include the TAG as context parameter, e.g., `Logger.warn("message", e = e, context = TAG)`.
+
+When performing a code review, check that methods exist before they are called, especially after refactoring.
```
ovitrif, line 23: > When performing a code review, check that methods exist before they are called, especially after refactoring.

This is only relevant for local development. Are we intending to use copilot to review PRs in Github, or it can also be used to self-review locally?!

PS. I'm not really up-to-date with GH Copilot's latest capabilities, hence asking such questions.

PPS. Will probably give GH copilot a test-drive very soon, the FOMO for some of its new features is real.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+When performing a code review, verify that expensive Compose computations are wrapped in `remember` blocks.
+
+When performing a code review, check that Material3 design guidelines are followed for UI components.
+
+When performing a code review, ensure proper state management patterns with `MutableStateFlow` and `StateFlow`.
+
+## Code Quality & Readability
+
+When performing a code review, focus on readability and avoid nested ternary operators.
```
ovitrif, line 37: > When performing a code review, focus on readability and avoid nested ternary operators.

I like the idea to focus on readability, but we're not dealing with nested ternary in Kotlin because it's not possible. Maybe instead, ask it to return early and avoid nested if-else wherever possible.

Example:

```kt
// do this:
fun decideCodepath(byDependency: Boolean) {
  if (!byDependency) return callAlternative()
  return callCodePath()
}

// instead of:
fun decideCodepath(byDependency: Boolean) {
  if (byDependency) {
    callCodePath()
  } else {
    callAlternative()
  }
}
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Dependency Injection & Services
+
+When performing a code review, verify proper Hilt dependency injection patterns are followed.
+
+When performing a code review, ensure services contain business logic and don't directly depend on ViewModels.
+
+## Build & Testing
+
+When performing a code review, ensure proper build variant usage (dev for regtest, tnet for testnet).
```
ovitrif, line 51: nit: moot, we don't need it really

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+When performing a code review, verify that unit tests follow the pattern `./gradlew testDevDebugUnitTest --tests ClassName`.
```
ovitrif, line 53: nit: moot, we don't need it in GH reviews

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+When performing a code review, suggest Unit Test com composable components and business logic covering the most important cases
```
ovitrif, line 55: nit:
"suggest Unit Test `com` composable" -> "suggest Unit Test `for` composable"

.github/copilot-instructions.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Lightning & Bitcoin Specific
+
+When performing a code review, ensure proper usage of LDK Node and bitkitcore library patterns.
```
ovitrif, line 59: > When performing a code review, ensure proper usage of LDK Node and bitkitcore library patterns.

While it's nice, I found it too hard (still as of today) for AI to understand what those patterns are.

ovitrif (changes requested):
Nice one 👏🏻 , almost ready IMHO, just a few remarks

.github/copilot-instructions.md
jvsena42: Agree, it would force us to implement unnecessary abstractions

> When performing a code review, ensure Repository pattern is followed

And like this?

.github/copilot-instructions.md
ovitrif: sounds good, can keep it 👍🏻

PR #380, opened 2025-09-15
fix(receive): geo block issues with channels to external node

ovitrif:
This PR fixes all known issues related to geo blocking when node has external (non-LSP) peer connection.

### Description
Changes:
- fix(wallet): hide transfer to spending when geoblocked
- refactor: replace hasChannelsReady with canReceive
- fix(receive): block LN receive when when geoblocked if having no lsp channels ready
- fix(receive): enable LN receive and QR only when having channels ready
- fix(receive): refresh geoblock state on each receive sheet open
- fix(receive): enable LN toggle when having external node even if geoblocked

### Preview

| Before | After |
| - | - |
| <video src="https://github.com/user-attachments/assets/77b46365-fc51-49ef-8e44-75fb7438722a"> | <video src="https://github.com/user-attachments/assets/dfa9aec8-3931-45f7-b5eb-3310e894dcea"> |

### QA Notes

**Test:**
1. Use VPN to set location to a geoblocked region (or force return true in `CoreService.isGeoBlocked()`)
2. Use [bitkit-docker: External Node manual setup](https://github.com/ovitrif/bitkit-docker?tab=readme-ov-file#external-node-manual-setup) guide to setup channel to external node
3. Go to Receive > **expect receive on LN switch ON & unified QR**
4. Go back, tap Spending > Transfer to savings > complete flow
5. Go to Receive > **expect receive on LN switch OFF & onchain QR**
6. Tap 'Receive on Spending' switch > expect to see the geoblock screen in the receive sheet

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/WalletRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/WalletRepo.kt
@@ -102,9 +102,9 @@ class WalletRepo @Inject constructor(
     suspend fun refreshBip21(force: Boolean = false): Result<Unit> = withContext(bgDispatcher) {
         Logger.debug("Refreshing bip21 (force: $force)", context = TAG)
 
-        val shouldBlockLightning = coreService.checkGeoBlock().second
+        val receiveOnSpendingBalance = !coreService.checkGeoBlock().second && lightningRepo.hasChannelsReady()
         _walletState.update {
-            it.copy(receiveOnSpendingBalance = !shouldBlockLightning)
+            it.copy(receiveOnSpendingBalance = receiveOnSpendingBalance)
         }
 
         // Reset invoice state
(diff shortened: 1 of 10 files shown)


app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -383,9 +382,8 @@ class WalletRepo @Inject constructor(
             updateBip21AmountSats(amountSats)
             updateBip21Description(description)
 
-            val hasChannels = lightningRepo.hasChannels()
-
-            if (hasChannels && _walletState.value.receiveOnSpendingBalance) {
```
jvsena42, line 388: Now I see, I forgot to test this case

jvsena42 (commented):
Tested on emulator including pasting an invoice while the node is initializing

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
ovitrif: indeed it's quite a lot of tests :)

the roundtrip test flow from the 2nd video was broken here and there, so all the other fixes that I did was to basically make it work

but the main fix for the e2e tests was actually setting `receiveOnSpendingBalance` to true (again if needed) in the `refreshBip21()` method

PR #387, opened 2025-09-17
fix: High Balance warning time out calculation for fresh wallets

jvsena42:
### Description
 isTimeOutOver didn't pass on fresh wallet because the value is initialized as zero

### Preview

https://github.com/user-attachments/assets/1cfb2604-5448-436d-acc4-711d127c99e7

### QA Notes
- New wallet -> receive $500 USD > should display warning


ovitrif (approved):
LGTM 👍🏻 (didn't test but it's simple and clear from code + preview)

PR #444, opened 2025-10-21
Add balance and transfer context to copilot instructions

jvsena42:
### Description

Improve copilot instructions with transfer and balance context

### Preview

`N/A`

### QA Notes

`N/A`

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -70,5 +70,6 @@ When performing a code review, check that new features integrate well with the e
 
 ## Business Logic
 
-When performing a code review on balance management logic, get context from `/docs/balance.md` file
-When performing a code review on transfer logic, get context from `/docs/transfer.md` file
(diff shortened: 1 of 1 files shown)


ovitrif (approved):
Maybe we can also try to be more generalistic, just asking to review against `*.md` files from `docs/` or something new like `docs/ai/`, see how that goes 🙃

ovitrif:
> > Maybe we can also try to be more generalistic, just asking to review against `*.md` files from `docs/` or something new like `docs/ai/`, see how that goes 🙃
> 
> I thought the same, but implemented like this because the doc only used files as reference
> https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review#customizing-copilots-reviews-with-custom-instructions

Oh, very nice then 🙌🏻
Let's go 🚀🚀🚀

PR #445, opened 2025-10-27
Activity list polish

### LINKED ISSUE #386: Select tag screen has unnecessary Apply and Clear buttons
The screen is displayed when filtering activities by tags.

**To reproduce:**
 - Go to "All Activities"
 - Hit "tag" icon

**Actual:**

The screen has "Apply" and "Clear" buttons. One needs to select tag and explicitely "Apply" it.
Also, after "Applying" the tag should be placed in the Search input, which is not the case.

Image

**Expected:**

In Bitkit RN the buttons are missing. The behavior is also that one only needs to tap a tag and the filtering applies.

Image

See also: https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=36310-190961&t=xjRruqCah5BU6831-4

jvsena42:
Closes #386 

[FIGMA](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=36310-191626&t=2uEytzAEjlKEBuG8-4)
[ROADMAP](https://docs.google.com/spreadsheets/d/1NBtx1y-_ZcODo8L0LsbudBNV_9x7cF6MyckR_W1xxq4/edit?pli=1&gid=0#gid=0&range=C6)
### Description

This PR implements the Figma v58 changes on the Activity List screen, except from the drawer menu

### Preview

[Screen_recording_20251028_070651.webm](https://github.com/user-attachments/assets/4d6211a7-f475-431a-acae-49730ce8e7eb)

### QA Notes
- Used "generate test activities" from developer settings to facilitate testing
- The Material3 DateRange calendar doesn't allow display a single month. I'll try to create a custom component on the next PR

### DIFF
--- a/app/src/main/java/to/bitkit/ui/screens/wallets/activity/TagSelectorSheet.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/wallets/activity/TagSelectorSheet.kt
@@ -40,7 +40,10 @@ fun TagSelectorSheet() {
     Content(
         availableTags = availableTags,
         selectedTags = selectedTags,
-        onTagClick = { activity.toggleTag(it) },
+        onTagClick = {
+            activity.toggleTag(it)
+            app.hideSheet()
+        },
     )
 }
 


app/src/main/java/to/bitkit/ui/screens/wallets/activity/TagSelectorSheet.kt
```diff
@@ -42,13 +41,6 @@ fun TagSelectorSheet() {
         availableTags = availableTags,
         selectedTags = selectedTags,
         onTagClick = { activity.toggleTag(it) },
```
ovitrif: we should also close sheet, that's what happens in RN app and it's more intuitive:

you tap the tag icon to open the sheet → you tap the tag → filter applies & sheet hides

Just needs a `app.hideSheet()` at the end of `onTagClick`

ovitrif (changes requested):
Tested ACK 💚 

Added one suggestion for better UX IMHO, which is also in RN Bitkit.

And added a few nits about moving modifiers last in args list, if ok 🙏🏻 .

Issue #337, opened 2025-09-02
Rapid Gossip Sync Server in regtest


piotr-iohk:
Observed while testing on the backend from e2e tests (see: [bitkit-e2e-tests/docker](https://github.com/synonymdev/bitkit-e2e-tests/tree/main/docker)).

Currently the Rapid Gossip Sync Server in regtest shows P2P while the previous e2e test scenario [Can connect to different Rappid Gosip Sync Server](https://github.com/synonymdev/bitkit/blob/master/e2e/settings.e2e.js#L577-L616) assumes there is a link there.

<img width="405" height="473" alt="Image" src="https://github.com/user-attachments/assets/44c8c324-7ae3-440e-9a67-4dac4e70bb7b" />

Also to note, there is no real validation if it is a real RGS Server, one can 'successfuly' connect to any host it seems:

<img width="405" height="473" alt="Image" src="https://github.com/user-attachments/assets/cc7d3853-ebbf-4778-9c7a-ca0699f7b077" />

PR #451, opened 2025-10-31
Test tags to connect to RGS 

piotr-iohk:
Rel #337, https://github.com/synonymdev/bitkit-e2e-tests/pull/40 

### Description

In order to enable `@settings_11` e2e test.


ovitrif:
> I suppose, as far as the formatting style, it would be good to enforce it via CI.

IIRC it used to work, don't know if this rule is not applied or it doesn't work anymore, it's been some time since I last saw any `detekt` lint issues on PRs.

I could look into a way to see if there's a rule or plugin for this rule to add to the `detekt` linter.

jvsena42:
> > I suppose, as far as the formatting style, it would be good to enforce it via CI.
> 
> IIRC it used to work, don't know if this rule is not applied or it doesn't work anymore, it's been some time since I last saw any `detekt` lint issues on PRs.
> 
> I could look into a way to see if there's a rule or plugin for this rule to add to the `detekt` linter.

Lets see if copilot detects

PR #466, opened 2025-11-12
e2e-status for branch protection

piotr-iohk:
### Description

Adds an `e2e-status` job that aggregates the results of all E2E test shards.
This job provides a single unified status check (`e2e-status`) that can be used in branch protection rules to ensure all E2E shards pass before merging.


ovitrif (approved):
Nice, TY 🙏🏻

PR #471, opened 2025-11-18
Show Dev Settings by default in debug builds

piotr-iohk:
### Description

In order to be consistent with iOS [[see](https://github.com/synonymdev/bitkit-ios/blob/master/Bitkit/Views/Settings/MainSettings.swift#L6)] and given the fact that both builds for e2e tests are done in debug configuration the PR sets Dev Settings as enabled in debug mode. E2E tests updated accordingly.

### QA Notes

https://github.com/synonymdev/bitkit-e2e-tests/pull/51


ovitrif (approved):
ACK ✅

PR #478, opened 2025-11-24
fix: LightningNodeService startup

### LINKED ISSUE #134: Review logic to show notification about Bitkit running in BG
See https://github.com/synonymdev/bitkit-android/pull/124#pullrequestreview-2828824452

Having the notification show up without wallet info, (ie. on fresh install with no wallet or after wallet wipe) feels a bit strange from the user’s perspective IMHO. Maybe we should look if there’s an optimal solution to run & stop it in connection to whether there is a wallet or not.

jvsena42:
Closes #134 

### Description

This PR improves the LightningNodeService startup. Now it is only called if the wallet exists, notifications are enabled and the app is not in recovery mode

### Preview

[recovery_mode.webm](https://github.com/user-attachments/assets/65983f10-56e1-4123-bfb8-065668b2b9c5)

[create_allow_notifications.webm](https://github.com/user-attachments/assets/93d61000-f8f8-408b-a017-713b84d6bbe1)

[node_working_without_service.webm](https://github.com/user-attachments/assets/d6752640-b9f1-4c57-b383-87813d5b28b4)

[create_dont_allow.webm](https://github.com/user-attachments/assets/01f991a9-0657-4ab7-a95e-c10b051cd9db)

### QA Notes
Note: Notification permission is only required for API 33+
- [x] Device with and without google services
   - [x] Create / recover wallet -> Allow notifications -> should start service
   - [x] Create / recover wallet -> DON'T Allow notifications -> should NOT start service
   - [x] Wallet not created -> DON'T start service
- [x]  Recovery mode ->  DON'T start service
- [x] Node should work even with service disabled

### DIFF
--- a/app/src/main/java/to/bitkit/ui/screens/wallets/activity/DateRangeSelectorSheet.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/wallets/activity/DateRangeSelectorSheet.kt
@@ -120,7 +120,6 @@ fun DateRangeSelectorSheet() {
         initialEndDate = endDate,
         onClearClick = {
             activityListVM.clearDateRange()
-            app.hideSheet()
         },
         onApplyClick = { start, end ->
             activityListVM.setDateRange(


app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -170,6 +188,27 @@ class MainActivity : FragmentActivity() {
...
     }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        if (!settingsViewModel.notificationsGranted.value) {
+            runCatching {
+                stopService(Intent(this, LightningNodeService::class.java))
+            }
+        }
+    }
```
jvsena42: This is necessary to the case:
Foreground service running -> User disable notifications -> service notification hidden but it service is not destroyed

PR #529, opened 2025-12-17
Fix/lnurl minSendable edge case when msats are not divisable by 1000

piotr-iohk:
### Description

 - LNURL-pay/withdraw endpoints express minSendable / maxSendable (and withdraw bounds) in millisatoshis (msat), while Bitkit’s amount entry and validation are in whole sats.
 - Previously we converted msats→sats using integer division (floor). This is usually fine when values are multiples of 1000, but it can understate the true minimum when the server returns non-round msat values (e.g. 100500 msat), letting the user enter 100 sats even though the server requires at least 101 sats.
 - This PR fixes the conversion by rounding minimum bounds up (ceil) and keeping maximum bounds down (floor), preventing “amount too low” errors after the user already satisfied the UI minimum.
 - additionally adding Boost toast ids (not related to fix)
 - related iOS change: https://github.com/synonymdev/bitkit-ios/pull/276

### QA Notes

 - Added unit tests that document the rounding rules and guard against regressions.
 - E2e test verifies the edge case on lnurl-pay (https://github.com/synonymdev/bitkit-e2e-tests/pull/72)

### DIFF
--- a/app/src/test/java/to/bitkit/ext/LnurlExtTest.kt
+++ b/app/src/test/java/to/bitkit/ext/LnurlExtTest.kt
@@ -38,6 +38,7 @@ class LnurlExtTest : BaseUnitTest() {
         )
 
         assertEquals(100u, data.minSendableSat())
+        assertEquals(0u, data.copy(minSendable = 0u).minSendableSat())
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


ovitrif (approved):
utAck

I trust the unit test and the easily readable logic 🤞🏻

PR #552, opened 2025-12-23
fix: max sendable lightning calc doesn't account for routing fee

### LINKED ISSUE #533: Send: Lightning ‘Available/Max’ doesn't account for routing fees, causes `Failed to send the given payment`
Android’s “Available” on the Lightning amount screen ([`SendAmountScreen.kt`](https://github.com/synonymdev/bitkit-android/blob/master/app/src/main/java/to/bitkit/ui/screens/wallets/send/SendAmountScreen.kt#L190C13-L193)) shows `maxSendLightningSats` without reserving anything for routing fees, but the confirm screen does estimate routing fees ([`estimateLightningRoutingFeesIfNeeded()`](https://github.com/synonymdev/bitkit-android/blob/master/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt#L1440)). That mismatch can make “Max”/available misleading (especially for amountless invoices), and it’s exactly why iOS can show 8 997 while Android shows 9 000, in case of total spending balance = 10 000.

**Expected:** available/max invoice amount ≈ maxSendLightningSats - estimatedRoutingFee (optionally with a small buffer) iOS uses additional 2 sats buffer -> https://github.com/synonymdev/bitkit-ios/blob/master/Bitkit/Views/Wallets/Send/SendAmountView.swift#L264
**Actual:** shows raw maxSendLightningSats; applying max available funds ends with error
**Impact:** user can pick an amount that fails or is confusing; cross-platform inconsistency

See iOS left/Android right (both having 10 000 spending balance and trying to send max available to each other; Android fails):
⚠️ additionally we see that sending over 50% of lightning funds triggers a 50% balance warning sheet on iOS, which is not triggered on Android. iOS is consistent with Bitkit RN in that, so Android should also display this 50% balance warning.

https://github.com/user-attachments/assets/45219877-535d-41e7-a56b-ef65be250310

jvsena42:
Closes #533 
### Description

This PR adds a round fee accounting to the max sendable lightning amount, with a 2 sats buffer. Also changes the 50% total balance sanity check to use individual unchain/lightning balance instead of total balance

### Preview

[lightning.webm](https://github.com/user-attachments/assets/9fc2bc02-59c4-4e22-abe5-dbefee5afa55)

[onchain.webm](https://github.com/user-attachments/assets/396e9c23-750d-48f8-98d0-31dd27924f4d)

### QA Notes

- Transfer full available lightning
- Check Lighning and on-chain 50% balance sanity check

### DIFF
--- a/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
+++ b/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
@@ -83,6 +83,7 @@ import to.bitkit.models.NewTransactionSheetType
 import to.bitkit.models.Suggestion
 import to.bitkit.models.Toast
 import to.bitkit.models.TransactionSpeed
+import to.bitkit.models.safe
 import to.bitkit.models.toActivityFilter
 import to.bitkit.models.toTxType
 import to.bitkit.repositories.ActivityRepo
@@ -1479,11 +1480,7 @@ class AppViewModel @Inject constructor(
         }
 
         val buffer = 2uL
-        val amountToEstimate = if (maxSendLightning > buffer) {
-            maxSendLightning - buffer
-        } else {
-            maxSendLightning
-        }
+        val amountToEstimate = maxSendLightning.safe() - buffer.safe()
 
         val feeResult = lightningRepo.estimateRoutingFeesForAmount(
             bolt11 = bolt11,


app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1458,6 +1465,41 @@ class AppViewModel @Inject constructor(
...
+
+        val maxSendLightning = walletRepo.balanceState.value.maxSendLightningSats
+        if (maxSendLightning == 0uL) {
+            _sendUiState.update { it.copy(estimatedRoutingFee = 0uL) }
+            return
+        }
+
+        val buffer = 2uL
+        val amountToEstimate = if (maxSendLightning > buffer) {
+            maxSendLightning - buffer
```
ovitrif: think this one needs the `.safe()` too ext on the ULongs participating in the subtraction

ovitrif (commented):
utAck, one remark then it LGTM's

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
jvsena42: indeed, I forgot

jvsena42: if both are positive, with safe, the min minimum output would be zero, but without safe, the minimum would be maxSendLightning

but they are so small values that makes no difference for this.

Still better apply `safe()` for the negative checks

ovitrif (approved):
utAck

PR #554, opened 2025-12-23
refactor: migrate navigation to nav3

### LINKED ISSUE #180: Decrease bottom sheet swipe-to-dismiss sensitivity
Find a way to decrease the sensitivity of swipe-to-dismiss in bottom sheets (make it harder to dismiss), at least on screens which are scrollable vertically.

**Actual:**
Sheet swipe-to-dismiss fires too fast, triggering accidental dismisses when trying to scroll down a screen rendered inside the sheet.

**Expected:**
It should be harder to dismiss bottom sheet when trying to scroll the screen inside it.
Ideally a swipe of 50% of the sheet's height should be needed.

See:
> PS: I dismissed the dialog twice by accident while selecting the words
> _Originally posted by @jvsena42 in https://github.com/synonymdev/bitkit-android/issues/179#issuecomment-2943729909_

### LINKED ISSUE #480: refactor: use Nav3, pass rust types to screens and add navigation viewmodel
note: Now that all rust types from our libraries are serializable, we should be able to do a great refactoring where we're just passing the type.

Also Navigation3 is finally stable, so this means we should soon refactor all navigation logic, which is very bad, but we didn't have any better choice given Compose's navigation was too poorly designed.

I will plan this refactoring.

_Originally posted by @ovitrif in https://github.com/synonymdev/bitkit-android/pull/476#discussion_r2557113781_

ovitrif:
Resolves #480
Closes #180

---
> [!CAUTION]
> This PR will be split into many separate PRs focusing on single-responsibility refactoring with the goal of fully migrating all navigation to `nav3`

---

### Description

This PR migrates the entire navigation system from legacy Compose Navigation (NavController) to the newly stable Navigation 3 (Nav3).

**Key changes:**
- Updated all navigation usage with Nav3 primitives
- Moved all navigation code to its dedicated `nav` package
- Moved all type-safe route definitions to `Routes.kt` sealed interface
- Introduced `Navigator` class centralising navigation logic
- Added `SheetSceneStrategy` for custom bottom sheets
- Replaced material3 `BottomSheetScaffold` with custom sheet implementation
  - Decreases sheet swipe-to-dismiss sensitivity #180
  - Allows to slightly move sheet up making it more alike the RN sheet impl.
- Defines all entry handlers in `nav/entries/` for all screen and sheet routes
- Removed legacy code and old sheet implementations
- Updated 55+ screens from `NavController` to `Navigator` pattern
- Net reduction of +3k lines of code
- TabBar now slides out below the viewport when showing sheets or non-wallet screens

### Preview
https://github.com/user-attachments/assets/2d7a1a5b-4ecc-4201-a94b-7d6fa09bacba

### QA Notes

- [ ] Verify all screen navigation works (back buttons, forward navigation)
- [ ] Test all bottom sheets open and close correctly
- [ ] Verify sheet dismiss behavior (swipe down, back button, overlay tap)
- [ ] Test deep link handling
- [ ] Test timed sheets (backup reminder, app update, etc.)
- [ ] Verify drawer menu navigation
- [ ] Test QR scanner navigation flow
- [ ] Verify settings screens navigation hierarchy
- [ ] Verify app with pin on launch enabled
- [ ] Verify auto-read clipboard enabled


jvsena42 (changes requested):
Send amount is not being reseted

[Screen_recording_20251229_082736.webm](https://github.com/user-attachments/assets/f16343fd-fff5-46cc-8954-d829615848a2)

piotr-iohk:
It seems that it breaks the timesheets behavior somehow. I don't see High Balance timed sheet appearing now when expected.

piotr-iohk:
Re-adding `ReceiveScreen` test tag as it was removed -> [e6a9de6](https://github.com/synonymdev/bitkit-android/pull/554/commits/e6a9de665f6ca05d89a1d5328c6838b689b1affb)

ovitrif:
@piotr-iohk The issues we get are mostly due to critical changes in UX which have long been desired.

I understand your frustration and totally resonate with your request to make changes more contained, but I don't think it's targeted in this specific case, entirely accurate 🙃.

The funny thing is that most of the relevant issues are stemming from targeted changes, and then there's the little flaws here and there which claude did, those need simple fixes like the one in this commit: [e6a9de6](https://github.com/synonymdev/bitkit-android/pull/554/commits/e6a9de665f6ca05d89a1d5328c6838b689b1affb).

---

At least so far my understanding is that many issues with e2e stem from the way sheets are getting dismissed now, which needs more swipe-down or a different API call to swipe down.

This was fully my own intent to chage the way sheets are getting dismissed, because the default material3 implementation from Google/Android is too rigid and not flexible enough for our UX needs.

Maybe there's other reasons which we can get to once we polish this more 🤷🏻 

PS. The changes in the e2e were never meant to be pushed even, I was only playing with AI and trying to get it to understand how to run the suite so I don't need CI to figure out if my changes are breaking the e2e or not.

It's normal that most of it could be slop, I can even intuitively tell because it fails in over 50% of times when it tries to run them locally. So there's clearly lack of understanding for Claude on what needs to be done and how.

ovitrif:
@piotr-iohk IDK if it makes sense but I would love to have more readily-available setup for AI to be able to work with the 2 repos nicer. Right now it's too bad, it can't really reliably run e2e for some big changes, and it doesn't understand what is broken and whether it's because of the changes in Android or because of some other setup issues; it keeps defaulting to hallucinations about the e2e not working correctly ever, LOL.

I'd very much appreciate if you can take time one of these days to setup a proper AGENTS.md file where you'd put some of your "secret knowledge" so that we can get more success rates with adjusting the e2e properly by other non-test engineer devs like myself 🙏🏻

ovitrif:
> There's unnecessary Backup sheet after doing "Backup your wallet" from Settings (shouldn't be there according to [design](https://www.figma.com/proto/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=37570-215676&p=f&t=hdFCJCdNQzHqKaoZ-0&scaling=scale-down&content-scaling=fixed&page-id=37570%3A212422&starting-point-node-id=37570%3A216502&show-proto-sidebar=1)). Also the wallet shows 24-word phrase consisting from only `secret` words. Tap to reveal state is not reseted as well.
> 

There is a bug at the foundational layer of in-sheet navigation which is the root of all of the issues of this type.
I'm working on a fix 👍🏻

piotr-iohk:
Receive -> Edit takes whole screen (should be contained in the sheet) and cannot go back.

https://github.com/user-attachments/assets/f02682a7-064c-4abd-975a-c70b4bfc7b2b

piotr-iohk:
> > There's unnecessary Backup sheet after doing "Backup your wallet" from Settings (shouldn't be there according to [design](https://www.figma.com/proto/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=37570-215676&p=f&t=hdFCJCdNQzHqKaoZ-0&scaling=scale-down&content-scaling=fixed&page-id=37570%3A212422&starting-point-node-id=37570%3A216502&show-proto-sidebar=1)). Also the wallet shows 24-word phrase consisting from only `secret` words. Tap to reveal state is not reseted as well.
> 
> There was a bug at the foundational layer of in-sheet navigation which is the root of all of the issues of this type.
> 
> The fix has landed in [f8ccbbb](https://github.com/synonymdev/bitkit-android/commit/f8ccbbb0f412e1ecd2136780ccfe4413661ab0a3)

This doesn't fix the problem. Issue still observed. 
It's a bit worse now, as the tap to reveal expands on the whole screen and one cannot swipe down to fold it...


https://github.com/user-attachments/assets/e35c3dde-b8a5-4f88-9816-3514cfe2b750

piotr-iohk:
> > I'm not seeing it that way. A lot of different issues surfaced just from the e2e runs, and I haven't had time to analyze every failure yet - so there may be more. Also, there could be regressions e2e won't catch. Given the size of the refactor, I'd expect to see unit tests added or updated alongside it. I'm not seeing much of that, which makes me worry we're leaning heavily on e2e coverage.
> 
> This is all UI changes, I won't add unit tests for UI.

Fair enough.
Although, given the scope of the refactor, perhaps it would warrant to revive and expand `ui-tests.yml`. I think that some targeted nav ui tests (without the full infra of e2e) would be good to have. I get that full unit testing doesn't make sense here but having a thinner UI layer of checks could still help.

ovitrif:
> Although, given the scope of the refactor, perhaps it would warrant to revive and expand `ui-tests.yml`. I think that some targeted nav ui tests (without the full infra of e2e) would be good to have. I get that full unit testing doesn't make sense here but having a thinner UI layer of checks could still help.

On the contrary, I want to remove them.
They're only giving ME work, because I build the Android app differently.

Countless times again and again changes that were not done by myself warranted more work that nobody sees which only I have to do.

So unless we teach all devs making implementation changes on Android to build this way, I personally I am better off without that code. It's been > 6 months and it's not adding any value, but on the contrary it's adding more work ONLY for me.

ovitrif:
The issues related to in-sheet navigation and sheets overall might be fixed in 3ad4a884, at least it appears so in my brief manual testing session.

ovitrif:
Should be done in a split fashion and using the new nav3 migration skill from android-cli.

The branch can stay though, for reference the PR is enough context also closed.

PR #559, opened 2025-12-26
fix: qr code decoding error message

### LINKED ISSUE #502: Different error message in Toast when pasting (vs. iOS)
_Originally posted by @ovitrif in https://github.com/synonymdev/bitkit-android/pull/490#pullrequestreview-3552364460_

jvsena42:
Fixes #502 
### Description

This PR updates the QR code decode message, like iOS

OBS: I noticed that iOS uses the same message to all decode cases. It could be used a generic message to when user paste the invoice 

### Preview

N/A

### QA Notes

Read invalid QR code


ovitrif (approved):
utAck

PR #570, opened 2026-01-07
Migrate to-do cards

ben-kaufman:
Migrates the todo cards in the suggestions below balance to hide the dismissed ones after RN app migration.

Note: iOS has Get Paid and Shop which Android and React Native don't have and we should likely fix this to match.


jvsena42:
> Note: iOS has Get Paid and Shop which Android and React Native don't have and we should likely fix this to match.

Android has a shop card, only RN doesn't have it

image

PR #571, opened 2026-01-07
fix: prevent force close channels of trusted peers

### LINKED ISSUE #528: Force close transactions with Blocktank not being confirmed
Same issue from iOS https://github.com/synonymdev/bitkit-ios/issues/275

ben-kaufman:
This PR is to skip force close of channels with trusted peers (Blocktank). If only trusted peers channels are being closed the app will not show the force close option and will just return with an error. If both trusted and not trusted channels are forced closed, the trusted peers channels will be skipped.

Fix #528


jvsena42 (approved):
There is still an edge case of when the force close is triggered directly by ldk-node

PR #593, opened 2026-01-09
fix(migration): pickup pending orders for fulfillment

ovitrif:
Fixes scenario: User pays for a channel order in RN, migrates to Native before the channel opens → App doesn't know the order is paid & pending fulfillment → channel never opens.

> [!NOTE]
> This was a bug I found by being more creative than the crew during the mainnet testing session of this week when we were testing the migration. Would love more creativity from everyone especially wrt. mainnet testing.

### Description

- Store paid orders with their transaction IDs during migration from RN backup
- The `paidOrders` backup data is a map of `orderId -> txId`, not just a list of IDs
- By calling `cacheStore.addPaidOrder(orderId, txId)`, the native app can now recognize pending paid orders and fulfill them when block confirmation is received

### Refactorings

- Adopt Result API for consistent error handling in migration code
- Remove suspend modifier from pure functions that don't need it
- Replace reflection with direct property access for signing key derivation
- Remove unused methods and dead code
- Extract magic numbers to named constants
- Move private data classes to bottom of file per convention
- Fix lint warnings and simplify conditional expressions

### Preview

N/A - No UI changes

### QA Notes

**Test: Migrating pending channel order from RN to Native**

1. Uninstall Bitkit app
2. Install RN (React Native) Bitkit
3. Load a fresh wallet → fund it
4. Start a transfer to spending and complete the flow without waiting
5. Immediately after (before the order is fulfilled):
   - Install native Bitkit build of this branch over the RN app
6. Open app and wait for migration to complete
7. Go to **Settings → Advanced → Lightning Connections** to see the pending order
8. Wait with the app open and active until the order is fulfilled (after block confirmation is received)
9. **Expected:** The app fulfills the channel order and the channel opens, with the app showing the toast for it and the Lightning Connections screen showing it as open
10. Verify Send and Receive LN funds work correctly

### DIFF
--- a/app/src/main/java/to/bitkit/services/MigrationService.kt
+++ b/app/src/main/java/to/bitkit/services/MigrationService.kt
@@ -51,6 +51,7 @@ import to.bitkit.models.TransactionSpeed
 import to.bitkit.models.TransferType
 import to.bitkit.models.WidgetType
 import to.bitkit.models.WidgetWithPosition
+import to.bitkit.models.safe
 import to.bitkit.models.widget.BlocksPreferences
 import to.bitkit.models.widget.FactsPreferences
 import to.bitkit.models.widget.HeadlinePreferences
@@ -1215,7 +1216,7 @@ class MigrationService @Inject constructor(
                 else -> TransferEntity(
                     id = txId,
                     type = TransferType.TO_SPENDING,
-                    amountSats = (order.clientBalanceSat + order.feeSat).toLong(),
+                    amountSats = (order.clientBalanceSat.safe() + order.feeSat.safe()).toLong(),
                     channelId = null,
                     fundingTxId = null,
                     lspOrderId = orderId,


ben-kaufman:
Yes, will do

ovitrif:
> Yes, will do

Dunno if iOS has the same mechanism, but if not it should:
if there are pendingOrders in caches, it should keep checking if their status is signalling they're ready to be fulfilled.

Then the actual migration only needs to ensure the `pendingOrders` map in caches is filled with the data picked-up from React Native's remote backups.

jvsena42:
The restore without migration worked. I'll test now on RN side

jvsena42:
> The restore without migration worked. I'll test now on RN side

Also worked, so the issue is in RN restore mapping

jvsena42:
> > The restore without migration worked. I'll test now on RN side
> 
> Also worked, so the issue is in RN restore mapping

worked on #595

piotr-iohk:
> Worked for me on both scenarios. For the restore from mnemonic, before deleting the app please go to backup in settings to make sure it synced, as in case the app is deleted before backup is synced there's nothing we can do.

Ok, worked after making sure it synced. 👍

PR #675, opened 2026-01-19
fix: unified invoice validation with zero on-chain balance

jvsena42:
This PR fixes unified invoice validation to accept Lightning payment when on-chain balance is zero.

### Description

When typing or pasting a unified invoice (BIP21 with embedded Lightning), the validation incorrectly rejected it if on-chain balance was zero, even when the user had sufficient Lightning balance. The fix adds unified invoice detection to `validateOnChainAddress()`, mirroring the existing logic in `onScanOnchain()`.

Changes:
1. Extracts a reusable `extractViableLightningInvoice()` helper that checks if a BIP21 URI contains a valid, non-expired Lightning invoice with sufficient sending capacity
2. Updates `validateOnChainAddress()` to check for Lightning payment viability before falling back to on-chain balance validation
3. Refactors `onScanOnchain()` to use the same helper, eliminating code duplication
4. Fixes on-chain balance checks to only trigger when payment method is on-chain

### Preview

[zero-amount-invoice.webm](https://github.com/user-attachments/assets/a8a28e1d-f066-4e60-a570-3d43f42e3f48)

[with-amount.webm](https://github.com/user-attachments/assets/f3899d4a-5f85-4f1c-a883-c065ccc757de)

### QA Notes
OBS: iOS doesn't have this bug
#### 1. Unified invoice with Lightning balance only
1. Create wallet with 0 on-chain balance but funded Lightning channel
2. Paste a unified invoice
3. Verify: Address should be accepted
4. Before fix: Shows "Insufficient Savings" error

#### 2. Regression - Pure on-chain address
1. Paste a plain Bitcoin address
2. Verify on-chain balance validation still works as expected

#### 3. Regression - Scan flow
1. Scan a unified QR code
2. Verify it still routes correctly to Lightning payment

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,71 +1,38 @@
 name: Bug Report
 description: Report a bug in Bitkit Android
-title: ''
 type: bug
 body:
   - type: textarea
     id: description
     attributes:
-      label: Description
-      description: Brief description of the issue or additional context
+      label: What happened?
+      placeholder: Describe the bug and what you expected instead
     validations:
-      required: true
+      required: false
 
   - type: textarea
     id: steps
     attributes:
       label: Steps to Reproduce
-      value: |
-        1. 
-        2. 
-        3. 
+      placeholder: |
+        1. Go to...
+        2. Tap on...
+        3. See error
     validations:
-      required: true
+      required: false
 
   - type: textarea
-    id: actual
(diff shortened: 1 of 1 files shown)


ovitrif (approved):
utAck

LGTM but will test just in case.

EDIT: tested and validated using fresh wallet with 0 onchain funds and LN got via CJIT🟢 :
- Pay unified invoice with amount
- Pay unified invoice, zero amount
- Quickpay unified invoice

Issue #553, opened 2025-12-23
crash with MnemonicNotFound exception after reseting app


jvsena42:
App crash after reseting wallet

Steps to reproduce:

1. Wallet created
2. Click on reset and restore
3. Stay a few seconds on the onboarding screeen

```
2025-12-23 09:28:01.945  6177-6315  AndroidRuntime          to.bitkit.dev                        E  FATAL EXCEPTION: DefaultDispatcher-worker-20
                                                                                                    Process: to.bitkit.dev, PID: 6177
                                                                                                    MnemonicNotFound
                                                                                                    	at to.bitkit.data.backup.VssStoreIdProvider.getVssStoreId(VssStoreIdProvider.kt:22)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2$1.invokeSuspend(VssBackupClient.kt:38)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2$1.invoke(Unknown Source:8)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2$1.invoke(Unknown Source:4)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndspatched(Undispatched.kt:66)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturnIgnoreTimeout(Undispatched.kt:50)
                                                                                                    	at kotlinx.coroutines.TimeoutKt.setupTimeout(Timeout.kt:149)
                                                                                                    	at kotlinx.coroutines.TimeoutKt.withTimeout(Timeout.kt:44)
                                                                                                    	at kotlinx.coroutines.TimeoutKt.withTimeout-KLykuaI(Timeout.kt:72)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2.invokeSuspend(VssBackupClient.kt:34)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2.invoke(Unknown Source:8)
                                                                                                    	at to.bitkit.data.backup.VssBackupClient$setup$2.invoke(Unknown Source:4)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndspatched(Undispatched.kt:66)
                                                                                                    	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:43)
[...]

PR #678, opened 2026-01-20
fix: mnemonic not found crash

piotr-iohk:
Fixes: https://github.com/synonymdev/bitkit-android/issues/553
Fixes: #691
OBS: https://github.com/synonymdev/bitkit-android/pull/657#issuecomment-3772867286

### Description

This PR fixes a crash that occurs when `VssBackupClient.setup()` is called before the mnemonic is available in the keychain. This race condition can happen during wallet restoration when `BackupRepo.startObservingBackups()` is triggered (on `NodeLifecycleState.Running`) before the mnemonic has been saved.

**Root cause:**
```
FATAL EXCEPTION: DefaultDispatcher-worker-6
to.bitkit.utils.ServiceError$MnemonicNotFound: Mnemonic not found
    at to.bitkit.data.backup.VssStoreIdProvider.getVssStoreId(VssStoreIdProvider.kt:23)
    at to.bitkit.data.backup.VssBackupClient$setup$2$1$1.invokeSuspend(VssBackupClient.kt:38)
```

**Changes:**
- `VssBackupClient.setup()` now returns `Boolean` (true = success, false = mnemonic not available yet)
- Added early mnemonic availability check before attempting VSS initialization
- Added idempotency check to return immediately if already set up
- `BackupRepo` now retries `setup()` with linear backoff (1s, 2s, 3s... up to 10 attempts) when mnemonic is not yet available

### Preview

N/A - crash fix, no UI changes

### QA Notes

1. **To reproduce the original crash:** note: it is intermittent.
   - https://github.com/synonymdev/bitkit-android/issues/553
   - https://github.com/synonymdev/bitkit-android/pull/657#issuecomment-3772867286

2. **Unit tests added:**
   - `setup returns false when mnemonic is not available`
   - `setup does not call vssStoreIdProvider when mnemonic is not available`
   - `setup checks mnemonic before proceeding with vss initialization`
   - `setup can be called multiple times when mnemonic not available`

   Run with: `./gradlew testDevDebugUnitTest --tests "to.bitkit.data.backup.VssBackupClientTest"`

### DIFF
--- a/app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
+++ b/app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
@@ -11,6 +11,8 @@ import com.synonym.vssclient.vssStore
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import to.bitkit.data.keychain.Keychain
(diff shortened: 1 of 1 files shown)


app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
```diff
@@ -59,10 +64,44 @@ class VssBackupClient @Inject constructor(
...
+        var onSuccess: (attempt: Int) -> Unit = {}
+        var onRetry: (attempt: Int, maxAttempts: Int, delayMs: Long) -> Unit = { _, _, _ -> }
+        var onExhausted: (maxAttempts: Int) -> Unit = {}
+    }
+
+    suspend fun setupWithRetry(
+        maxAttempts: Int = 10,
+        baseDelayMs: Long = 1000L,
+        logger: SetupRetryLogger.() -> Unit,
+    ): Result<Boolean> = withContext(ioDispatcher) {
```
jvsena42: same here, could return `Result<Unit>`

app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
```diff
@@ -29,8 +29,15 @@ class VssBackupClient @Inject constructor(
 ) {
     private var isSetup = CompletableDeferred<Unit>()
 
-    suspend fun setup(walletIndex: Int = 0) = withContext(ioDispatcher) {
+    suspend fun setup(walletIndex: Int = 0): Result<Boolean> = withContext(ioDispatcher) {
```
jvsena42: nit: could add a check to prevent concurrent calls, there was a bug on this on iOS

There is a similar logic on `LightningRepo.sync()`

jvsena42 (approved):
Just left two suggestion that can be solved here or in other PR

jvsena42 (approved):
Tested:
- Lightning channels
- Install over RN
- recover from RN
- Reset and restore

Issue #663, opened 2026-01-19
Electrum Connection issue in App Status after receiving onchain


ovitrif:
[bitkit_logs_1768812658798.zip](https://github.com/user-attachments/files/24711150/bitkit_logs_1768812658798.zip)

PR #679, opened 2026-01-20
chore: add defensive check to skip retry when node is starting

jvsena42:
Related to #663

This PR adds a defensive check to the node start retry logic to prevent unnecessary retries when the node is already starting or running.

### Description

Adds a guard in the `.onFailure` block of the `start()` function that checks if `nodeLifecycleState.isRunningOrStarting()` before triggering a retry. If the node is in a Starting or Running state when an error occurs, the function returns success instead of retrying, which could cause issues like `NodeException.AlreadyRunning`.

### Preview

[restart.webm](https://github.com/user-attachments/assets/df0755a0-5ba0-4e49-a9f4-afc00218c968)

[restore.webm](https://github.com/user-attachments/assets/89555fa1-437b-43f5-9e65-84104e1d1f4a)

[migration.webm](https://github.com/user-attachments/assets/023389c9-4950-4ae6-bb11-b7a615683d0b)

### QA Notes

1. Run unit tests
   - `./gradlew testDevDebugUnitTest --tests LightningRepoTest`
   - Verify `start should not retry when node lifecycle state is Starting` test passes

2. Verify the fix
   - The defensive check prevents retry loops when errors occur while the node is starting
   - This complements the existing fix that made sync non-fatal


ovitrif (approved):
utAck

PR #689, opened 2026-01-20
fix: improve gift code flow

### LINKED ISSUE #680: Easter Egg Marketing Campaign – Gift code flow behaves inconsistently
**Description:**
While testing the Easter Egg marketing campaign flow (gift code redemption), the behavior is inconsistent and confusing for the user.

**Steps to reproduce:**
	1.	Insert a gift code shared by Denys and/or Sev.
	2.	Click Send.
	3.	Click Paste invoice.
	4.	Observe the redemption flow.

**Observed behavior:**
	•	The flow initially worked, but in a very unclear way.
	•	The gift amount was displayed as 0 instead of the expected value.
	•	The app spun/loading for a while, then returned an error saying the gift could not be claimed and to contact support.
	•	Retrying the same code then resulted in a message saying the code had already been used.
	•	Only after this, a channel with balance appeared in the wallet.

**Expected behavior:**
	•	The gift amount should be clearly displayed.
	•	The redemption flow should either succeed or fail clearly, not both.
	•	The code should not be marked as “used” if the UI reports a failure.
	•	Channel creation and balance availability should be explicit and immediate once the gift is successfully claimed.

**Notes:**
This works technically in the end, but the UX is very confusing and could easily make users think the gift was lost or invalid. This is especially risky for a marketing-facing feature.

ovitrif:
Fixes #680

Improves the UX of the Easter Egg gift code redemption flow by adding better logging, error handling, and validation.

### Description

- Add comprehensive logging throughout gift claim operations for better debugging
- Fix amount=0 handling to allow backend validation instead of rejecting upfront
- Improve error messages with full error chain context
- Add operation guards to prevent duplicate claim attempts
- Validate that channel funding transaction exists
- Better error classification (Used/UsedUp/Error)

### QA Notes

Test scenarios:
1. Test with a valid gift code (happy path)
2. Test with a gift code that has amount=0 in QR
3. Test with an already-used gift code
4. Test with an exhausted/depleted gift code
5. Test in both liquidity scenarios (with/without existing channels)
6. Monitor logs to verify detailed debug information is captured

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
@@ -473,7 +473,9 @@ class BlocktankRepo @Inject constructor(
 
         return GiftClaimResult.SuccessWithLiquidity(
             paymentHashOrTxId = giftResponse.bolt11PaymentId ?: giftResponse.id,
-            sats = amount.toLong(),
+            sats = giftResponse.bolt11Payment?.paidSat?.toLong()
+                ?: giftResponse.appliedGiftCode?.giftSat?.toLong()
+                ?: amount.toLong(),
             invoice = invoice,
             code = code,
         )


app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -436,14 +444,25 @@ class BlocktankRepo @Inject constructor(
...
+                        Logger.debug("Amount unknown (0), defaulting to channel opening path", context = TAG)
+                    } else {
+                        Logger.debug("Insufficient liquidity, opening new channel", context = TAG)
+                    }
                     Result.success(claimGiftCodeWithoutLiquidity(code, amount))
                 }
             }.getOrThrow()
         }.onFailure {
-            Logger.error("Failed to claim gift code", it, context = TAG)
+            Logger.error("Failed to claim gift code: ${it.message}", it, context = TAG)
```
ovitrif: `: ${it.message}` is redundant here, the Logger.error internals are already doing something similar, but better. Adding it again will pollute the log record with duplication.

app/src/main/java/to/bitkit/ui/sheets/GiftViewModel.kt
```diff
@@ -113,12 +123,28 @@ class GiftViewModel @Inject constructor(
     }
 
     private suspend fun handleGiftClaimError(error: Throwable) {
-        Logger.error("Gift claim failed: $error", error, context = TAG)
+        val errorMessage = buildString {
+            append("Gift claim failed: ")
+            append(error.message ?: error.toString())
+            error.cause?.let {
```
ovitrif: remove this as well, again. NEVER print error details in logs.

PR #703, opened 2026-01-22
ci: cache emulator AVD for faster E2E boots

ovitrif:
This PR adds AVD (Android Virtual Device) caching to E2E workflows to enable emulator "warm boots" instead of cold boots, reducing CI run times.

### Description

- Add `actions/cache@v4` step to cache `~/.android/avd/*` and `~/.android/adb*` directories
- Add snapshot generation step that runs only on cache miss to create the initial AVD snapshot
- Change emulator option from `-no-snapshot` to `-no-snapshot-save` to load cached snapshots without overwriting them
- Applied to both `e2e.yml` (main E2E tests) and `e2e_migration.yml` (migration tests) workflows

**Expected impact:**
- Cold boot: ~60-90 seconds
- Warm boot with cached snapshot: ~15-30 seconds
- Savings: ~45-60 seconds per emulator start
- With 6 shards, this could save 4-6+ minutes per E2E run on cache hits

### Preview

N/A - CI workflow changes only

### QA Notes

1. Trigger E2E workflow manually on this branch
2. Check logs for "AVD cache" step - should show cache miss on first run
3. Verify "Create AVD and generate snapshot for caching" step executes
4. Trigger E2E workflow again
5. Verify "AVD cache" step shows cache hit
6. Compare emulator boot times between runs
7. Verify E2E tests still pass with the new configuration

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -344,7 +344,7 @@
     <string name="onboarding__passphrase_header">Secure with &lt;accent&gt;Passphrase&lt;/accent&gt;</string>
     <string name="onboarding__passphrase_text">You can add a secret passphrase to the 12-word recovery phrase. If you do, make sure you don’t forget.</string>
     <string name="onboarding__pp_checkbox">Privacy Policy</string>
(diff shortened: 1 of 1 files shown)


piotr-iohk (approved):
ok

PR #729, opened 2026-01-26
feat: use battery saving sync intervals

ovitrif:
This PR:
2. Adds battery saving sync intervals when the app goes to the background

### Description

When the app is backgrounded, the Lightning node now switches to battery-saving sync intervals using the new `RuntimeSyncIntervals` API from ldk-node. When the app returns to the foreground, normal sync intervals are restored. This reduces battery drain while keeping the node active for receiving payments.

The SDK update also adds a new `claimableOnCloseSats` property to channel details, which is now included in mock data and extensions.

### Preview

N/a - No UI changes.

### QA Notes

1. **Test background behavior**
   - Open the app and wait for the node to start
   - Background the app by pressing the home button
   - Check logcat for: `Sync intervals entered battery saving mode`
   - Foreground the app
   - Check logcat for: `Sync intervals exited battery saving mode`

2. **Test normal operation**
   - Verify payments can still be sent and received
   - Verify wallet balance updates correctly
   - Verify channel information displays properly

### DIFF
--- a/app/src/main/java/to/bitkit/App.kt
+++ b/app/src/main/java/to/bitkit/App.kt
@@ -1,10 +1,7 @@
 package to.bitkit
 
 import android.annotation.SuppressLint
-import android.app.Activity
 import android.app.Application
-import android.app.Application.ActivityLifecycleCallbacks
-import android.os.Bundle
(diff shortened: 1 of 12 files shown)


ovitrif:
stale for too long

PR #747, opened 2026-01-29
fix: skip backup warning after wallet restore

### LINKED ISSUE #669: Wallet Backup sheet shown after restoring wallet
The 'Wallet Backup' bottom sheet shows after successfully restoring wallet. This is not necessary because I just used my phrase to restore.

Same as on iOS ⬆️

jvsena42:
Fixes #669

This PR sets `backupVerified` to `true` when a wallet is successfully restored, preventing the unnecessary backup warning sheet from appearing after restore.

### Description

When a user restores a wallet using their recovery phrase, they clearly have access to their backup. The app was incorrectly showing the "Wallet Backup" bottom sheet after restore because `backupVerified` remained `false`. This PR adds an `onSuccess` handler to the `restoreWallet` function that marks the backup as verified.

### Preview

[Screen_recording_20260129_072648.webm](https://github.com/user-attachments/assets/d14330af-557b-4f38-bede-7a14d771b29b)

#### 1. Restore wallet flow
1. Wipe the app data or use a fresh install
2. Tap "Restore wallet"
3. Enter a valid 12-word recovery phrase
4. Complete the restore process
5. Verify the "Wallet Backup" bottom sheet does NOT appear

#### 2. New wallet flow (regression)
1. Wipe the app data or use a fresh install
2. Create a new wallet
3. Verify the "Wallet Backup" suggestion still appears (expected behavior for new wallets)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,8 +55,8 @@ android {
         applicationId = "to.bitkit"
         minSdk = 28
         targetSdk = 36
-        versionCode = 174
-        versionName = "2.0.0"
+        versionCode = 175
+        versionName = "2.0.1"
         testInstrumentationRunner = "to.bitkit.test.HiltTestRunner"
         vectorDrawables {
             useSupportLibrary = true


app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -300,6 +300,8 @@ class WalletRepo @Inject constructor(
                 keychain.saveString(Keychain.Key.BIP39_PASSPHRASE.name, bip39Passphrase)
             }
             setWalletExistsState()
+        }.onSuccess {
+            settingsStore.update { it.copy(backupVerified = true) }
```
ovitrif: nit: don't we have in BackupRepo callbacks or post-restore ops room where to add this logic?!

As a general rule, logic that is tied together should stay together. IMO it feels like it's not really the responsibility of the wallet to handle post-backup restore concerns 🤷🏻

jvsena42: solved in #754

PR #751, opened 2026-01-29
fix: show tab bar on savings and spending screens

### LINKED ISSUE #696: Send and Receive buttons are missing after one taps on spending or savings balance
### What happened?

Send and receive buttons are missing after one taps on spending or savings balance.

### Steps to Reproduce

 - Have wallet with no lightning channel
 - tap on "Spending" balance

Android on the right:
_The video only shows spending balance screen, but it is the same for savings._

https://github.com/user-attachments/assets/2d486567-0b1d-45e7-b11b-75c34a460eb3

### Environment

Bitkit Version: 2.0.0 170, mainnet, regtest, local

jvsena42:
Fixes #696

This PR fixes the missing Send and Receive buttons on the Savings and Spending activity screens.

### Description

When navigating to the Savings or Spending balance screens, the tab bar (containing Send and Receive buttons) was not visible. This was because the tab bar visibility logic only included the Home and AllActivity routes. This PR adds the Savings and Spending routes to the list of screens where the tab bar should be displayed.

### Preview

[Screen_recording_20260129_124425.webm](https://github.com/user-attachments/assets/32379315-8e4a-4b62-9299-0b47c10d6110)

#### 1. Savings screen tab bar
1. Open the wallet
2. Tap on Savings balance
3. Verify the tab bar with Send and Receive buttons is visible at the bottom

#### 2. Spending screen tab bar
1. Open the wallet
2. Tap on Spending balance
3. Verify the tab bar with Send and Receive buttons is visible at the bottom

#### 3. Regression - Home and Activity screens
1. Verify tab bar still appears on Home screen
2. Verify tab bar still appears on All Activity screen

### DIFF
--- a/app/src/main/java/to/bitkit/ui/ContentView.kt
+++ b/app/src/main/java/to/bitkit/ui/ContentView.kt
@@ -477,6 +477,8 @@ fun ContentView(
                     val showTabBar = currentRoute in listOf(
                         Routes.Home::class.qualifiedName,
                         Routes.AllActivity::class.qualifiedName,
+                        Routes.Savings::class.qualifiedName,
+                        Routes.Spending::class.qualifiedName,
                     )
 
                     if (showTabBar) {


ovitrif (approved):
utAck

nit: We may not need `.qualifiedName`, `.simpleName` might be enough but it's definitely not as safe mid-term, until we get the nav3 migration done.

PR #759, opened 2026-01-30
fix: disable Details button until activity is available

### LINKED ISSUE #673: Error toast "Activity not found"  after Send swipe-to-pay then tap Details in confetti sheet
### Repro
- Send onchain
- Tap on 'Details' fast after the send
- 🔴 Notice error toast

### Remark
AFAICR this button isn't even shown in the react native app. Maybe we should hide it for as long as the activity info isn't yet received, and enable it after a LDK event confirms the payment is recorded in ldk-node's store.

### Media

Image

jvsena42:
Fixes #673

This PR fixes the "Activity not found" error that appeared when tapping the Details button too quickly after sending a payment.

### Description

When sending a payment, the success confetti sheet would show with the Details button enabled before the activity data was fully synced. Tapping Details prematurely would cause an error toast. This fix:

1. Shows the Details button in a disabled state until the activity is available
2. For lightning payments, waits for the `PaymentSuccessful` event to confirm the activity is synced (already implemented)
3. For onchain payments, waits for the activity sync to complete before enabling the button

### Preview

[Screen_recording_20260130_131737.webm](https://github.com/user-attachments/assets/4f4e397f-27d0-4928-a506-5333c313b312)

#### 1. Lightning send - Details button loading state
1. Send a lightning payment
2. Observe the confetti success sheet
3. Tap Details and verify it navigates to the activity detail screen

#### 2. Onchain send - Details button loading state
1. Send an onchain payment
2. Observe the confetti success sheet
3. Verify the Details button shows a disabled state initially
4. Wait for the button to become enabled
5. Tap Details and verify it navigates to the activity detail screen

#### 3. Regression - Close button
1. Send any payment (LN or onchain)
2. Verify the Close button works immediately
3. Verify the sheet can be dismissed normally

### DIFF
--- a/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
+++ b/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
@@ -1642,11 +1642,8 @@ class AppViewModel @Inject constructor(
                             direction = NewTransactionSheetDirection.SENT,
                             paymentHashOrTxId = actualPaymentHash,
                             sats = paymentAmount.toLong(), // TODO Add fee when available
-                            isLoadingDetails = true,
                         ),
                     )
-                    activityRepo.syncActivities()
-                    _successSendUiState.update { it.copy(isLoadingDetails = false) }
                 }.onFailure { e ->
                     // Delete pre-activity metadata on failure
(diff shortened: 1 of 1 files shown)


ovitrif (approved):
utAck

PR #760, opened 2026-02-02
fix: add retry and logging for channel monitor migration

jvsena42:
This PR adds error tracking and retry logic for channel monitor retrieval during React Native to native Android migration, preventing silent channel loss.

### Description

During migration from the RN app, channel monitors are fetched from the remote backup. Previously, if any monitor retrieval failed (network timeout, decryption error, etc.), it was silently dropped via `.mapNotNull()`, potentially causing permanent channel loss without any indication to the user or in logs.

This PR:
1. Tracks which specific channel monitors fail to retrieve
2. Logs failed channel IDs for debugging
3. Adds a retry mechanism (3 attempts with linear backoff: 1s, 2s, 3s)
4. Warns when monitor count doesn't match expected

### Preview

N/A - Backend migration logic, no UI changes

#### 1. Migration with all monitors available
1. Perform an RN to native migration with a wallet that has channels
2. Verify migration completes successfully
3. Verify channels are available after migration

#### 2. Log verification (requires simulated failure)
1. If a monitor fails transiently: should see "Retrying channel monitor retrieval..." debug logs
2. If a monitor fails permanently: should see "Failed to retrieve X/Y channel monitors after retries" error log
3. Should see "Channel monitor count mismatch" warning if any monitors are lost

#### 3. Regression - Normal operation
1. Verify normal wallet operation is unaffected
2. Verify lightning payments work as expected


jvsena42:
The force close wasn't displayed as Transfer Activity because the LDK node couldn't track the channel

jvsena42:
> It looks like Github closed this PR by mistake

probably because it interpreted `fix` as keyword on iOS PR

ovitrif (approved):
utAck

logic looks solid 👍🏻

PR #763, opened 2026-02-03
fix: add network error retry for transfer polling

### LINKED ISSUE #608: Spending balance remains 0 after opening the channel from the app, despite transfer to spending confirmed (Android)
I observed the same situation as in iOS from here: https://github.com/synonymdev/bitkit-ios/issues/338, (no logs, as the channel eventually opened and we thought there was some blocktank hiccup)

Pasting AI analysis on the potential root cause.

---

### Root Cause

The bug is in `TransferViewModel.kt` in the `pollUntil` function. When `getOrder()` fails due to a network error, `.getOrNull()` returns `null`, which is misinterpreted as "order not found" and causes the polling loop to exit permanently.

**Relevant code** (`TransferViewModel.kt:267-283`):
```kotlin
private suspend fun pollUntil(orderId: String, condition: (IBtOrder) -> Boolean): IBtOrder? {
    while (true) {
        val order = blocktankRepo.getOrder(orderId, refresh = true).getOrNull()
        if (order == null) {
            Logger.error("Order not found: '$orderId'", context = TAG)
            return null  // BUG: Network errors also cause null, exits loop permanently
        }
        if (order.state2 == BtOrderState2.EXPIRED) {
            Logger.error("Order expired: '$orderId'", context = TAG)
            return null
        }
        if (condition(order)) {
            return order
        }
        delay(POLL_INTERVAL_MS)
    }
}
```

### Bug Flow

1. `onTransferToSpendingConfirm()` launches coroutine calling `watchOrder()`
2. `watchOrder()` calls `pollUntil()` to wait for order state changes
3. `pollUntil()` calls `blocktankRepo.getOrder()` which returns `Result<IBtOrder>`
4. **On network error**: `Result.failure` → `.getOrNull()` returns `null`
5. Code logs "Order not found" (misleading - it's actually a network error)
6. `pollUntil()` returns `null`
7. `watchOrder()` returns `Result.failure(Exception("Order not found or expired"))`
8. Coroutine completes, **polling stops permanently**
9. Channel opens on Blocktank's side, but app never detects it

### Key Issue

The code conflates two different error conditions:
- **Actual "order not found"**: Should stop polling (order doesn't exist)
- **Network error**: Should retry (transient failure)

Using `.getOrNull()` loses the distinction between these cases.

### Recommended Fix

1. **Distinguish between network errors and actual "not found"**:

```kotlin
private suspend fun pollUntil(orderId: String, condition: (IBtOrder) -> Boolean): IBtOrder? {
    var consecutiveErrors = 0
    val maxConsecutiveErrors = 5
    
    while (true) {
        val result = blocktankRepo.getOrder(orderId, refresh = true)
        
        result.fold(
            onSuccess = { order ->
                consecutiveErrors = 0  // Reset on success
                
                if (order.state2 == BtOrderState2.EXPIRED) {
                    Logger.error("Order expired: '$orderId'", context = TAG)
                    return null
                }
                if (condition(order)) {
                    return order
                }
            },
            onFailure = { error ->
                consecutiveErrors++
                Logger.warn("Failed to fetch order (attempt $consecutiveErrors): ${error.message}", context = TAG)
                
                if (consecutiveErrors >= maxConsecutiveErrors) {
                    Logger.error("Too many consecutive errors, giving up", context = TAG)
                    return null
                }
                // Continue polling on transient errors
            }
        )
        
        delay(POLL_INTERVAL_MS)
    }
}
```

2. **Resume watching pending transfers on app foreground** - When the app returns to foreground, check for any pending `toSpending` transfers and resume watching them.

3. **Consider adding a "refresh" button** - Allow users to manually trigger a sync of pending transfers if automatic recovery fails.

jvsena42:
Fixes #608

This PR adds network error resilience to the transfer-to-spending polling mechanism by distinguishing between transient network failures and permanent order errors.

### Description

- Uses `.fold()` instead of `.getOrNull()` to properly handle `Result` failures
- Implements a retry mechanism that tolerates up to 5 consecutive network errors before giving up
- Resets the error counter on successful fetches, allowing recovery from intermittent issues
- Preserves immediate exit behavior for permanent failures (order not found, order expired)

### Preview

https://github.com/user-attachments/assets/19e01fb7-7d00-401a-a0ec-676de4d00552

### QA Notes

To facilitate the simulation, you can put a breakpoint in the line and then disable internet
```
 val paidOrder = pollUntil(orderId) { order ->
```

#### 1. Normal transfer flow (regression)
1. Start a transfer to spending
2. Verify the setup steps progress (0 → 1 → 2 → 3)
3. Confirm the channel opens successfully

#### 2. Network error recovery
1. Start a transfer to spending
2. Enable airplane mode briefly during polling
3. Disable airplane mode before 12.5 seconds (5 retries × 2.5s)
4. Verify polling resumes and completes successfully

#### 3. Persistent network failure
1. Start a transfer to spending
2. Enable airplane mode and keep it enabled
3. After ~12.5 seconds, verify the transfer fails gracefully with appropriate logging

### DIFF
--- a/app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
+++ b/app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
@@ -1,0 +1,17 @@
+package to.bitkit.ext
+
+import android.content.Context
+import org.lightningdevkit.ldknode.PaymentFailureReason
+import to.bitkit.R
+
+fun PaymentFailureReason?.toUserMessage(context: Context): String = when (this) {
+    PaymentFailureReason.RECIPIENT_REJECTED ->
+        context.getString(R.string.wallet__toast_payment_failed_recipient_rejected)
+    PaymentFailureReason.RETRIES_EXHAUSTED ->
+        context.getString(R.string.wallet__toast_payment_failed_retries_exhausted)
+    PaymentFailureReason.ROUTE_NOT_FOUND ->
+        context.getString(R.string.wallet__toast_payment_failed_route_not_found)
+    PaymentFailureReason.PAYMENT_EXPIRED ->
+        context.getString(R.string.wallet__toast_payment_failed_timeout)
+    else -> context.getString(R.string.wallet__toast_payment_failed_description)
+}
--- a/app/src/main/java/to/bitkit/fcm/WakeNodeWorker.kt
+++ b/app/src/main/java/to/bitkit/fcm/WakeNodeWorker.kt
@@ -23,6 +23,7 @@ import to.bitkit.data.CacheStore
 import to.bitkit.data.SettingsStore
 import to.bitkit.di.json
 import to.bitkit.ext.amountOnClose
+import to.bitkit.ext.toUserMessage
 import to.bitkit.models.BITCOIN_SYMBOL
 import to.bitkit.models.BlocktankNotificationType
 import to.bitkit.models.BlocktankNotificationType.cjitPaymentArrived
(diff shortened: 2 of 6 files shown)


jvsena42:
draft to apply suggestions

PR #765, opened 2026-02-05
fix: auto-recover stale network graph cache

jvsena42:
This PR adds automatic validation and recovery for stale RGS network graph caches that cause Lightning payment routing failures.

### Description

When the app uses RGS (Rapid Gossip Sync) delta updates, nodes that are missing from a stale cached graph won't be restored by incremental syncs. This caused `RouteNotFound` errors when trying to pay certain destinations like Blink wallet.

1. Adds `validateNetworkGraph()` to check if trusted peers (Blocktank LSP nodes) are present in the graph
2. Adds `resetNetworkGraph()` to delete the cached graph file when stale
3. Validates the graph automatically on node startup and triggers a fresh full sync if trusted peers are missing
4. Skips validation when graph is empty to avoid false positives on first launch

### Preview

https://github.com/user-attachments/assets/31b50a89-305f-4186-852f-ea53c2e4c080

#### 1. Verify stale graph auto-recovery
1. Launch app with existing wallet that has a stale network graph cache
2. Check logs for: `Network graph missing X trusted peers`
3. Verify logs show: `Network graph is stale, resetting and restarting...`
4. After restart, verify: `Network graph validated: all X trusted peers present`

#### 2. Verify normal startup (no stale cache)
1. Launch app with valid network graph
2. Verify logs show: `Network graph validated: all X trusted peers present`
3. No restart should occur

#### 3. Verify payment routing works
1. After graph validation passes, attempt a Lightning payment to Blink wallet
2. Payment should route successfully without `RouteNotFound` error

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.withTimeoutOrNull
 import org.lightningdevkit.ldknode.Address
 import org.lightningdevkit.ldknode.BalanceDetails
 import org.lightningdevkit.ldknode.BestBlock
+import org.lightningdevkit.ldknode.Bolt11Invoice
 import org.lightningdevkit.ldknode.ChannelConfig
 import org.lightningdevkit.ldknode.ChannelDataMigration
 import org.lightningdevkit.ldknode.ChannelDetails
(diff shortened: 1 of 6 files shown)


jvsena42:
related LDK PR
https://github.com/lightningdevkit/ldk-node/pull/692

ovitrif (approved):
tAck

**Test results:**
1. Verify stale graph auto-recovery: success
2. Verify normal startup (no stale cache): success
3. Verify payment routing works: success

PR #769, opened 2026-02-08
refactor: secure sensitive strings

ovitrif:
This PR introduces a `Secret` model that wraps sensitive data in a wipeable `CharArray`, replacing immutable `String` usage for mnemonics, passphrases, and PINs across the entire app.

### Description

1. Adds a `Secret` model with `use`/`peek`/`wipe` lifecycle and ergonomic extensions (`useAsString`, `splitWords`, `wipeAll`)
2. Adds secure `loadSecret`/`saveSecret`/`upsertSecret` methods to `Keychain`
3. Replaces `String`-based mnemonic handling with `Secret` across all services, repositories, and ViewModels
4. Eliminates `String` from the entire PIN lifecycle by accumulating input as `CharArray` via a new `mutableSecretOf()` Compose utility, and removes PIN from navigation route serialization
5. Adds `ByteArray.wipe()` extension and wipes derived encryption keys in `RNBackupClient` after use
6. Clears restore wallet state on ViewModel destroy

### Preview

IMAGE_1

#### 1. Wallet creation with PIN
- Create a new wallet
- Set a PIN during onboarding
- Confirm the PIN matches
- Verify wallet is created successfully

#### 2. PIN validation flows
- Lock the app and unlock with PIN
- Send a payment that requires PIN confirmation
- Verify PIN validation works correctly in both flows

#### 3. Change PIN
- Go to Settings > Security > Change PIN
- Enter current PIN
- Enter new PIN and confirm
- Verify the new PIN works for subsequent unlocks

#### 4. Wallet restore
- Start wallet restore flow
- Paste or type a 12-word mnemonic
- Optionally add a passphrase
- Verify restore completes successfully

#### 5. Backup verification
- Go to Settings > Backups
- View mnemonic and passphrase screens
- Verify they display correctly
- Complete the mnemonic confirmation flow

#### 6. Regression
- `./gradlew testDevDebugUnitTest` passes
- `./gradlew detekt` clean


ovitrif:
Thorough re-review of the scope of these changes reveals the harsh truth that, while it is a nice attempt, it is far from delivering on its promise because we are forced to default to `Strings` anyways at the Rust/UniFFI boundary.

So, while `codex` was trying to be nice in its review, I am of the impression that this PR would actually introduce more harm than good, because it forces additional boxing/cloning, while for no real gain; the use of Strings for secrets cannot be eliminated, the vulnerability in its original shape is still there, but now we would end up with more bloat.

While there are some valuable parts, there's not much ROI in extracting them.

Closed.

PR #775, opened 2026-02-10
fix: delete vss network graph using new ffi client

ovitrif:
Closes #765

This PR:
1. Adds a new VssBackupClientLdk for direct LDK-specific VSS operations via FFI bindings
2. Switches VSS graph deletion during stale-graph reset to use the new FFI client
3. Improves graph validation logging with node IDs, graph stats, and RGS timestamp
4. Extracts the network graph debug dump from LightningService

### Description

Builds on #740. The stale graph auto-recovery added in #765 attempted to deleted the graph from VSS using the app-level backup client but due to different key obfuscation namespacing in ldk-node, the graph backup was never deleted.

This PR switches to a dedicated LDK-specific FFI client implementation that points operations to the correct obfuscated file names backed up by ldk-node on VSS. This new VSS client implementation adjusted for supporting OPs on ldk-node backups is found in PR:
- synonymdev/vss-rust-client-ffi/pull/8

In practice this translate to successful LN payments to nodes that would otherwise fail, one example being payments to Blink wallet if having a stale graph in your wallet's VSS backups.

LDK-node already has the required mechanism to trigger redownload of the network graph if attempts to retrieve it from its backup on VSS fails. The cleanup mechanism is triggered when the graph is found to miss any of the expected trusted peers, which in turn triggers ldk-node to kick-off its re-download of the graph from the latest RGS server snapshot.

> [!NOTE]
> After more research this issue proved to reproduce only on wallets created with the early release candidate versions of the app, when caching of network graph was still done on VSS.

#### Test 1️⃣ - Payment to Blink
https://github.com/user-attachments/assets/5b89c08e-ae84-4d05-b952-d9c9cc53e4a4

#### Test 2️⃣ - VSS app data preserved across `vss-rust-client` update
| Before (`0.4.0`) | After (`0.5.5`) |
| - | - |
| before | after |

### QA Notes

> [!TIP]
> No need to go all the way to replicate a stale graph on your VSS backup. To validate, simply test on mainnet some random LN payments to a couple wallets you have on other apps like Blink, Phoenix, WoS. If payments succeed, PR is good to be merged.

#### 1. Verify stale graph reset clears VSS via FFI client
1. Launch app with a wallet that has a stale network graph
2. Check logs for: `Network graph is stale, resetting and restarting...`
3. Verify logs show: `Cleared stale network graph from VSS (first delete)`
4. After restart, verify graph validates successfully
1. After graph validation passes, send a Lightning payment to Blink wallet
2. Payment should route successfully

#### 2. Verify VSS app data preserved across versions
1. Install a previous version of the app (e.g. latest release from `master`)
2. Create or restore a wallet and let it fully sync
3. Personalize the wallet, eg.
   - Change display currency
   - Set a custom denomination (e.g. sats)
   - Toggle biometric authentication
   - Update widgets on home screen
4. Make note of all customized values
5. Uninstall the app so wallet data lives only on VSS
6. Install the build from this PR
7. Restore the same wallet using the seed phrase
8. After sync completes, verify all personalization is preserved

### DIFF
--- a/app/src/main/java/to/bitkit/services/LightningService.kt
+++ b/app/src/main/java/to/bitkit/services/LightningService.kt
@@ -48,10 +48,11 @@ import to.bitkit.ext.totalNextOutboundHtlcLimitSats
 import to.bitkit.ext.uByteList
 import to.bitkit.ext.uri
 import to.bitkit.models.OpenChannelResult
+import to.bitkit.utils.AppError
 import to.bitkit.utils.LdkError
 import to.bitkit.utils.LdkLogWriter
-import to.bitkit.utils.LogDumperLdk
 import to.bitkit.utils.Logger
+import to.bitkit.utils.LoggerLdk
 import to.bitkit.utils.ServiceError
 import to.bitkit.utils.jsonLogOf
 import java.io.File
@@ -70,7 +71,7 @@ class LightningService @Inject constructor(
     private val keychain: Keychain,
     private val vssStoreIdProvider: VssStoreIdProvider,
     private val settingsStore: SettingsStore,
-    private val logDumperLdk: LogDumperLdk,
+    private val loggerLdk: LoggerLdk,
 ) : BaseCoroutineScope(bgDispatcher) {
 
     companion object {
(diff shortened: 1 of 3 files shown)


jvsena42 (approved):
Performed Lightning operation on main net

PR #788, opened 2026-02-18
feat: support all address types

ben-kaufman:
This PR adds support for multiple Bitcoin address types in the iOS app, allowing users to select and monitor different address formats. Users can select their preferred address type from: Legacy (P2PKH), Nested SegWit (P2SH-P2WPKH), Native SegWit (P2WPKH), and Taproot (P2TR).

This requires extensive testing across all functionalities, send, receive, channel open, channel close (funds should always go to native segwit address except when selected address when channel opened was Taproot, in which case it should be a Taproot address), cpfp, rbf. All should be checked for each address type, and for multiple address types combined (ie. send with utxos of different address types, open channel with inputs of different address types, boost txs with inputs of different address types, etc.)

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -81,7 +81,7 @@ jobs:
           name: bitkit-e2e-apk_${{ github.run_number }}
           path: app/build/outputs/apk/dev/debug/bitkit_e2e.apk
 
-  build-regtest:
+  build-staging:
     needs: detect-changes
     if: github.event.pull_request.draft == false && needs.detect-changes.outputs.code == 'true'
     runs-on: ubuntu-latest
@@ -295,10 +295,10 @@ jobs:
         if: failure()
         uses: jwalton/gh-docker-logs@v2
 
-  e2e-tests-regtest:
+  e2e-tests-staging:
     if: github.event.pull_request.draft == false && needs.detect-changes.outputs.code == 'true'
     runs-on: ubuntu-latest
-    needs: [detect-changes, build-regtest, e2e-branch]
+    needs: [detect-changes, build-staging, e2e-branch]
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -299,6 +306,12 @@ class WalletRepo @Inject constructor(
             if (bip39Passphrase != null) {
                 keychain.saveString(Keychain.Key.BIP39_PASSPHRASE.name, bip39Passphrase)
             }
+            settingsStore.update {
+                it.copy(
+                    selectedAddressType = "nativeSegwit",
```
jvsena42: Could get from `AddressType.P2WPKH.toSettingsString()` to prevent typos

ben-kaufman:
Same https://github.com/synonymdev/bitkit-ios/pull/431#issuecomment-3975870243

We could offer to reduce the amount sent to the recipient to have more for fees but this is not a currently supported feature

app/src/main/java/to/bitkit/services/LightningService.kt
```diff
@@ -911,6 +924,51 @@ class LightningService @Inject constructor(
         }
     }
 
+    suspend fun getBalanceForAddressType(addressType: AddressType): org.lightningdevkit.ldknode.AddressTypeBalance =
```
jvsena42: extract import

ben-kaufman:
@piotr-iohk couldn't reproduce but made a small fix and added logs which should help understand the issue if it happens again

.github/workflows/e2e_migration.yml
```diff
@@ -85,7 +85,6 @@ jobs:
           - { name: migration_1-restore, grep: "@migration_1" }
           - { name: migration_2-migration, grep: "@migration_2" }
           - { name: migration_3-with-passphrase, grep: "@migration_3" }
-          - { name: migration_4-with-sweep, grep: "@migration_4" }
```
piotr-iohk: Not needed, migration tests updated in related e2e branch.

PR #798, opened 2026-02-21
fix: update LND4 IP and bump ldk-node to rc.26

ben-kaufman:
This PR:
1. Updates the hardcoded LND4 fallback peer IP from `34.65.186.40` to `34.65.153.174` after the node migration
2. Bumps ldk-node from v0.7.0-rc.18 to v0.7.0-rc.26 which includes a fix for `PeerStore::add_peer` silently ignoring address updates for existing peers

Without the ldk-node fix, existing installs would keep retrying the stale cached IP indefinitely even after the app pushes the updated address via `Node::connect`.

### Description

LND4's IP changed during a node migration. The hardcoded fallback address in Env is updated, and the ldk-node dependency is bumped to include the peer address upsert fix ([synonymdev/ldk-node#53](https://github.com/synonymdev/ldk-node/pull/53)).

### Preview

N/A — no UI changes.

### QA Notes

1. Update from a previous build where LND4 had the old IP
2. Verify on startup that LND4 connects successfully (check logs for the new IP `34.65.153.174`)
3. Fresh install should also connect to LND4 without issues

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -170,7 +170,7 @@ android {
             reset()
             // Only architectures supported by native libs (ldk-node, bitkit-core)
             // x86 not supported; x86_64 only for debug/emulator
-            include("armeabi-v7a", "arm64-v8a")
+            include("armeabi-v7a", "arm64-v8a", "x86_64")
             isUniversalApk = true
         }
     }


ovitrif (approved):
tAck

PR #800, opened 2026-02-23
feat: peer cards with labels + use ldk-node from GH

### LINKED ISSUE #682: [FR]: Lightning node details
Align with iOS app with lightning node details

**iOS**

Image

**android**

Image

ovitrif:
Fix #682

This PR:
1. Define `NodePeer` data model to enrich known peers with names
2. Adds peer cards with labels on the node info screen
3. Shows all node info by default, regardless of dev mode on/off value (iOS parity)
4. Consumes `ldk-node-android` dependency GitHub Packages instead of JitPack
5. Updates detekt config for magic numbers to include common design tokens values
6. Updates AI rules for view model file conventions
7. Add padding between ldk debug screen peer input and buttons
8. Bumps app version

### Description

Polishes peer card UI and makes Node Info screen show all details from the start. Introduces a dedicated `NodeInfoViewModel` and `NodePeer` model, and sources peer labels from `Env` configuration if fetching from LSP API fails.

Also switches the `ldk-node-android` dependency from JitPack to GitHub Packages, and includes minor housekeeping for detekt and AI rules.

### Preview

peerCards

#### 1. Peer cards visibility
1. Build and run a dev debug build
2. Navigate to the Node Info screen
3. Verify peer cards display with labels (e.g. LSP, LND nodes)
5. Confirm cards show peer alias, public key, and disconnect button
6. Tap on a card → expect to see copy toast
9. Tap on disconnect → expect peer to disconnect (it should reconnect on node restart)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,8 +55,8 @@ android {
         applicationId = "to.bitkit"
         minSdk = 28
         targetSdk = 36
-        versionCode = 176
-        versionName = "2.0.2"
+        versionCode = 177
+        versionName = "2.0.3"
         testInstrumentationRunner = "to.bitkit.test.HiltTestRunner"
         vectorDrawables {
             useSupportLibrary = true


piotr-iohk:
Created a separate bug ticket https://github.com/synonymdev/bitkit-android/issues/801. Managed to make it work after couple of retries (i.e. setting up wallet again and funding with 100 000 sats - eventually worked with no 0 cap when transfer to spending, it didn't work with initial balance 100 000 000, but not sure if it is balance related, b2b I think I also seen it once or twice before but couldn't reproduce later)

PR #853, opened 2026-03-18
fix: comment out orphaned channel monitor recovery

jvsena42:


ovitrif:
why does PR title references something that I can't find in code if I search for it?

I guess it should be: `checkForOrphanedChannelMonitorRecovery`

PS. These little details are quite critical for proper work with AI.

PR #903, opened 2026-04-23
fix: align TOS padding with iOS

jvsena42:
This PR polishes the Terms of Use onboarding screen to match bitkit-ios and align with the next screen (Intro).

### Description

1. Increases the top spacer above the "Terms of use" header from 16dp to 48dp, matching iOS's `.padding(.top, 48)` so the heading has the same breathing room below the status bar
2. Reduces the footer bottom padding from 24dp to 16dp so the Continue button sits 16dp above the navigation bar inset — identical to the Get Started / Skip Intro buttons on the following Intro screen, preventing the visual jump when advancing through onboarding

The Scaffold's `contentPadding` already provides the `systemBars` inset (Android equivalent of iOS safe area), so no structural changes were needed.

### Preview

[android.webm](https://github.com/user-attachments/assets/6df7fd31-6b5b-458d-ac3d-43ed555552a3)

[android-nav-bar.webm](https://github.com/user-attachments/assets/6c53badd-5e93-40a6-bf56-42257e9a8c1d)

https://github.com/user-attachments/assets/d57aea15-5ea3-4dc7-94af-d1bbc70fd658

### QA Notes

1. Launch a fresh onboarding flow and open the Terms of Use screen
   - Verify the "Terms of use" header has visible breathing room below the status bar (larger than before)
   - Verify the scrollable TOS content starts below the header as expected
2. Scroll to the bottom of the TOS content and tap Continue
   - Verify the Continue button sits close to the navigation bar area with a 16dp gap
   - Verify there is no vertical jump when transitioning to the Intro screen — the Continue button position should match the Get Started / Skip Intro buttons
3. Test on a device with gesture navigation and a device with 3-button navigation
   - Verify the button respects the system bar inset on both (no overlap with nav bar, no excess whitespace)
4. Verify the `TermsPreview` still renders correctly in Android Studio

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
-- Polish Terms of Use screen padding to match iOS
+- Polish Terms of Use screen padding to match iOS #903
 
 ## [2.2.0] - 2026-04-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Polish Terms of Use screen padding to match iOS
+
 ## [2.2.0] - 2026-04-07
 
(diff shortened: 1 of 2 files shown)


piotr-iohk (approved):
LGTM, tested on sim.

PR #915, opened 2026-04-28
feat: add LDK debug action to export pathfinding scores

ovitrif:
This PR adds an "Export Scorer" action to the LDK Debug screen that dumps the running node's merged pathfinding scorer to a file and triggers an Android share sheet, so it can be moved off-device for inspection. It also nulls the regtest scorer URL since the staging endpoint serves a 5-byte minimal payload.

Depends on:
- https://github.com/synonymdev/rust-lightning/pull/2 — sibling work in the LDK fork (not a hard dependency for this PR; this branch uses `node.exportPathfindingScores()` which is already in `ldk-node 0.7.0-rc.36`)
- https://github.com/synonymdev/ldk-node/pull/79 — the consumer that reads the bytes this debug action exports

### Description

Mirrors the existing `exportNetworkGraph` flow exactly:

- New "Export Scorer" button under a `PATHFINDING SCORER` section in `LdkDebugScreen`.
- `LdkDebugViewModel.exportScorer()` writes to `<context.cacheDir>/exports/pathfinding_scores.bin` and surfaces the file via `FileProvider` so a share sheet pops up — same UX as Export Network Graph.
- `LightningService.exportPathfindingScoresToFile()` is a thin wrapper around `node.exportPathfindingScores()` (existing UDL binding).
- `LoggerLdk.exportPathfindingScoresToFile()` writes the raw bytes; logs the byte count.
- `Env.ldkScorerUrl` for `Network.REGTEST` is set to `null`. The staging endpoint (`https://api.stag0.blocktank.to/scorer`) serves 5 bytes — minimal/empty data — so wiring it through accomplishes nothing. Making this explicit matches "we don't use the scorer on regtest".

Dev-only screen — no localized strings (per CLAUDE.md "NEVER add string resources for strings used only in dev settings screens").

### QA Notes

- Build & install: `./gradlew installDevDebug` (or `installTnetDebug` / `installMainnet` per network).
- Open Settings → Dev → LDK Debug → tap "Export Scorer". A toast should show the byte count, then the system share sheet opens.
- Save the `.bin` somewhere off-device.
- (Optional) Pull via adb: `adb shell "run-as to.bitkit.dev cat files/cache/exports/pathfinding_scores.bin" > /tmp/scorer_local.bin` (adjust app id per flavor).
- The exported bytes are `ChannelLiquidities`-encoded — same wire format as the served scorer at `https://api.blocktank.to/scorer-prod`. They can be inspected with the `scorer-inspect` CLI from synonymdev/ldk-node#79 once that lands.
- Regression-check: the existing "Export Network Graph" button still works.
- Verified: `./gradlew compileDevDebugKotlin testDevDebugUnitTest detekt` — all pass; only pre-existing detekt warnings (none introduced by this change).

### DIFF
--- a/app/src/main/java/to/bitkit/env/Env.kt
+++ b/app/src/main/java/to/bitkit/env/Env.kt
@@ -76,7 +76,6 @@ internal object Env {
     val ldkScorerUrl
         get() = when (network) {
             Network.BITCOIN -> "https://api.blocktank.to/scorer-prod"
-            Network.REGTEST -> "https://api.stag0.blocktank.to/scorer"
             else -> null
         }
 
--- a/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
@@ -1360,6 +1360,11 @@ class LightningRepo @Inject constructor(
         executeWhenNodeRunning("exportNetworkGraphToFile") {
             lightningService.exportNetworkGraphToFile(outputDir)
         }
+
+    suspend fun exportPathfindingScoresToFile(outputDir: String): Result<File> =
+        executeWhenNodeRunning("exportPathfindingScoresToFile") {
+            lightningService.exportPathfindingScoresToFile(outputDir)
+        }
     // endregion
 
     // region probing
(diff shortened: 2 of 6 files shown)


ovitrif:
Closing as superseded by scorer-kit in synonymdev/scorer#20. This draft was useful as an investigation artifact, but the current testing path should use a hosted merged scorer binary or a purpose-built debug scorer URL override instead. It also contains stale references and an unrelated regtest scorer URL change, so a fresh focused PR would be cleaner if we still need Android-side debug support.

PR #929, opened 2026-05-05
fix: await ln payment before gift card confetti

jvsena42:
This PR:
1. Fixes a false-positive confetti animation in the gift card redemption flow when the user already has an open Lightning channel.
2. Fixes the gift sheet re-opening after an app language change re-delivers the launching intent.

#### 1. False-positive confetti

`BlocktankRepo.claimGiftCodeWithLiquidity` previously returned success as soon as Blocktank's `giftPay` HTTP RPC returned — but that response only confirms the LSP accepted the request, not that the LN payment actually landed. If Blocktank's background payer subsequently failed to route the HTLC, Bitkit still played confetti and surfaced a "received" sheet with the requested amount, while no incoming payment ever arrived.

The repo now subscribes to `LightningRepo.nodeEvents` for the `Event.PaymentReceived` matching the freshly-created invoice's payment hash *before* triggering `giftPay`, then awaits that event with a 45 second timeout. On timeout it throws `ServiceError.GiftClaimPaymentNotReceived`, which the existing error handler in `GiftViewModel` routes to the standard `GiftRoute.Error` screen. The success result reports the actual received sat amount instead of the requested amount.

The without-liquidity path (channel open + `openChannel`) is unchanged — that path already blocks on the funding transaction.

#### 2. Intent re-delivery on Activity recreate

When the app language is changed, Android destroys and recreates the Activity, re-delivering the original launching intent. `MainActivity.onCreate` was unconditionally calling `handleDeeplinkIntent(intent)`, so any deeplink-driven flow (gift sheet, send sheet, pubky auth, etc.) re-fired on the locale switch.

`onCreate` now only processes the launching intent on a fresh start (`savedInstanceState == null`). New deeplinks that arrive while the app is running are still handled via `onNewIntent`.

### Preview

N/A

#### Reproducing the failure path

The originally-reported `gUCdzXfm12V71xztGzDoiQ-3000` is now exhausted and returns `GIFT_CODE_USED_UP` (a separate, already-handled path), so it can no longer reproduce the false-positive. To exercise case 1, use one of:

- Coordinate with the Blocktank team to provide a code whose LSP-side LN payer is in an unpayable state (accepts the HTTP request but cannot route the HTLC).
- Issue a valid code via the regtest blocktank harness, then disconnect the LSP peer / cut network between scanning the QR and the HTLC settling, so the inbound HTLC never arrives. `giftPay` returns success but `Event.PaymentReceived` never fires.

#### Regtest helpers

Fire the gift deeplink directly without scanning a QR (replace the package id per flavor — `to.bitkit.dev` for dev/regtest, `to.bitkit.tnet` for testnet, `to.bitkit` for mainnet):

```sh
adb shell am start -a android.intent.action.VIEW -d "bitkit://gift-gUCdzXfm12V71xztGzDoiQ-3000" to.bitkit.dev
```


jvsena42:
Looking for a reliable way to reproduce it

jvsena42:
@ovitrif the success paths worked for me, but couldn't reproduce the bug from master. we could do an internal release for @dzdidi

app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -205,6 +206,15 @@ class MainActivity : FragmentActivity() {
...
     }
 
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        intent?.data?.toString()?.let { outState.putString(KEY_CONSUMED_DEEPLINK_URI, it) }
+    }
+
+    private companion object {
+        const val KEY_CONSUMED_DEEPLINK_URI = "consumed_deeplink_uri"
+    }
```
ovitrif: nit: could be moved to top, right under class name, to respect coding convention.

ovitrif (dismissed):
utAck

ovitrif:
> @ovitrif the success paths worked for me, but couldn't reproduce the bug from master. we could do an internal release for @dzdidi

@jvsena42 The failed `multi_address_local` e2e check doesn't look related, I retriggered it; but please double check.

jvsena42:
> > @ovitrif the success paths worked for me, but couldn't reproduce the bug from master. we could do an internal release for @dzdidi
> 
> @jvsena42 The failed `multi_address_local` e2e check doesn't look related, I retriggered it; but please double check.

it is happening in all branches

piotr-iohk:
Tested the gift false-positive failure path on regtest. ✅ 

Setup:
- Wallet had an existing Lightning channel with enough inbound liquidity (`maxInbound=744191`, gift amount `3000 sats`).
- Blocked the LSP peer `34.65.86.104:9400` with a local `pf` rule while keeping HTTPS/Blocktank reachable.

> Set up rule:
> 
> ```
> # 1. Block only the regtest LSP peer, keep HTTPS working
> printf 'block drop out quick proto tcp to 34.65.86.104 port 9400\n' | sudo tee /tmp/bitkit-lsp-block.pf
> sudo pfctl -a com.apple/bitkit-lsp-block -f /tmp/bitkit-lsp-block.pf
> sudo pfctl -E
> 
> # 2. Verify rule is active
> sudo pfctl -a com.apple/bitkit-lsp-block -sr
> ```
> Remove after test:
> 
> ```
> sudo pfctl -a com.apple/bitkit-lsp-block -F rules
> ```



- Claimed a fresh valid gift code via deeplink.

Observed:
- App used the with-liquidity path.
- Invoice was created.
- `giftPay` returned successfully:
  `Gift payment request completed ... awaiting LDK PaymentReceived`
- LSP peer connection failed due to the firewall rule.
- No `PaymentReceived` arrived within the timeout.
- App failed with `GiftClaimPaymentNotReceived` and showed the error state, with no confetti / received success sheet.

After removing the firewall rule, the delayed payment arrived, which confirms the test blocked settlement long enough to exercise the timeout path.

Also regression tested:
- Fresh valid gift code with an existing channel and with no channel - payment complete ✅ 
- Depleted gift code - no confetti, appropriate error message ✅

piotr-iohk (approved):

ovitrif (approved):

PR #935, opened 2026-05-07
fix: OS widget cleanup crash and ordering

jvsena42:
This PR:
1. Fixes an intermittent NPE crash when an OS widget is removed or its configuration is cancelled.
2. Reorders the widget config activity's provider lookup so it matches the in-app `AddWidgetsScreen` ordering.

#### 1. Receiver crash on widget removal

All five Glance receivers (`Price`, `Headlines`, `Blocks`, `Facts`, `Weather`) overrode `onDeleted` to unregister the widget entry from the preferences store. The override called `super.onDeleted()` first and then `goAsync()` itself. Glance's `GlanceAppWidgetReceiver.onDeleted` already consumes the broadcast `PendingResult` via its internal `goAsync` extension, so the second call returned `null` and `pendingResult.finish()` in our `finally` block threw a `NullPointerException`. Per Glance's own KDoc:

> if you override any of the methods of this class, you must call the super implementation, and you must not call `AppWidgetProvider.goAsync`, as it will be called by the super implementation.

The unregister logic now lives on each `GlanceAppWidget.onDelete(context, glanceId)` override, which Glance invokes inside its own broadcast-scoped coroutine. The `onDeleted` overrides on every receiver are removed, so there is no longer a duplicate `goAsync` call to NPE on.

#### 2. Provider lookup ordering

`AppWidgetConfigActivity.resolveWidgetType` was listing the providers in `Headlines, Price, Blocks, Weather` order. Reordered to `Price, Headlines, Blocks, Weather` so it matches the `AppWidgetType` enum, the manifest receivers, the refresh worker, and the in-app `AddWidgetsScreen` listing. Pure code-style change with no behavior difference.

#### A note on the system widget picker order

The Android system widget picker (long-press home → Widgets) is rendered by the launcher, not by us. The Pixel launcher groups widgets within an app by their `targetCellWidth × targetCellHeight` area, in descending order, then by manifest declaration order within each size bucket. Because `appwidget_info_facts.xml` declares `targetCellHeight="1"` while every other widget declares `targetCellHeight="2"`, Facts ends up alone in the smaller bucket and appears last in the picker. That positioning is launcher-controlled and intentional given the widget's existing 1-cell-tall design — this PR does not change widget sizes.

### Preview

[Screen_recording_20260507_120327.webm](https://github.com/user-attachments/assets/fc5a2fa3-f0c1-4716-92b0-19cff5d55a46)

#### Manual Tests
- [x] **1.** Add a Bitcoin Headlines OS widget with Bitkit running in the background → tap the back arrow on the configure screen → no crash, widget is removed, app stays alive. Repeat several times to cover the intermittent timing.
- [x] **2.** Repeat step 1 for Price, Blocks, Facts, and Weather OS widgets.

### DIFF
--- a/app/src/main/java/to/bitkit/appwidget/config/AppWidgetConfigActivity.kt
+++ b/app/src/main/java/to/bitkit/appwidget/config/AppWidgetConfigActivity.kt
@@ -85,8 +85,8 @@ class AppWidgetConfigActivity : ComponentActivity() {
         val providerClass = AppWidgetManager.getInstance(this)
             .getAppWidgetInfo(appWidgetId)?.provider?.className
         return when (providerClass) {
-            HeadlinesGlanceReceiver::class.java.name -> AppWidgetType.HEADLINES
             PriceGlanceReceiver::class.java.name -> AppWidgetType.PRICE
+            HeadlinesGlanceReceiver::class.java.name -> AppWidgetType.HEADLINES
             BlocksGlanceReceiver::class.java.name -> AppWidgetType.BLOCKS
             WeatherGlanceReceiver::class.java.name -> AppWidgetType.WEATHER
             else -> {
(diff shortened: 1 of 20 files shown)


ovitrif:
> Facts ends up alone in the smaller bucket and appears last in the picker. 

Not really but irrelevant that it's not accurate, what matters is the conclusion that we can't alter the ordering in the OS picker programmatically:

ovitrif (approved):
Tested fixes with success 😎

PR #957, opened 2026-05-20
fix: gate paykit ui

### LINKED ISSUE #954: Conditional flag for Paykit Features
### Problem or use case

Add a conditional flag (preferred a simple build flag/ dev settings toggle on UI) or even just reuse dev-mode flag to hide Paykit features (Profile, Contacts, and everything related to Pay2Contacts Public/Private) on UI.

### Proposed solution

Build flag so we can either turn it completely off or expose it thru Dev Settings, initial implementation proposal defaulting to exposure via the Dev Settings screen, with an additional prompt dialog warning users that this functionality may not work reliably yet, to account for delays deploying the supporting changes in homeserver.

ben-kaufman:
Closes #954

### Description

This PR:
1. Adds build-level and developer-setting gates for the experimental Paykit UI.
2. Gates Profile, Contacts, Payment Preference, activity contact, Pubky route, and Paykit background sync surfaces while Paykit UI is disabled.
3. Keeps Send -> Contact visible as the production placeholder path and routes it to Coming Soon while Paykit UI is disabled.
4. Sanitizes existing/restored Paykit state and retries endpoint cleanup safely when Paykit UI is disabled.
5. Fixes stale public cleanup markers so they cannot remove endpoints after sharing is re-enabled.
6. Fixes fallback address derivation when LDK address infos are unavailable by using the account derivation path expected by bitkit-core.

### Preview

N/A

#### Manual Tests
- [ ] **1.** Dev Settings -> enable Paykit UI -> Profile / Contacts / Send Contact: Paykit surfaces are available.
- [ ] **2.** Dev Settings -> disable Paykit UI -> Profile / Contacts: Coming Soon or hidden states are shown.
- [ ] **3.** Send -> Contact with Paykit UI disabled: Coming Soon opens.
- [ ] **4.** `regression:` Paykit previously enabled -> update with Paykit UI disabled -> app startup/foreground: published endpoints are cleaned up or left pending for retry.
#### Automated Checks
- `git diff --check`
- `./gradlew testDevDebugUnitTest --tests to.bitkit.viewmodels.AppViewModelSendFlowTest --tests to.bitkit.viewmodels.SettingsViewModelTest --tests to.bitkit.repositories.PublicPaykitRepoTest --tests to.bitkit.repositories.WalletRepoTest --tests to.bitkit.repositories.BackupRepoTest --tests to.bitkit.repositories.PubkyRepoTest`
- `./gradlew compileDevDebugKotlin testDevDebugUnitTest detekt`
- `PAYKIT_UI_DISABLED=true ./gradlew compileDevDebugKotlin`


ben-kaufman:
@piotr-iohk paykit in e2e now fails since it's disabled by default, should we enable it for all e2e runs or just for the specific one of paykit tests?

piotr-iohk:
> @piotr-iohk paykit in e2e now fails since it's disabled by default, should we enable it for all e2e runs or just for the specific one of paykit tests?

~I think we should enable Paykit for E2E builds the same way dev settings are - default on in debug builds (Env.isDebug), off in release, still toggleable in Dev Settings.~

On second thought, I will handle on e2e side so we have in e2e tests the version that is closest to the prod.
Will need to push some test ids for that.

piotr-iohk:
On Android `isPaykitEnabled` is inside SettingsData, so it gets backed up/restored with the wallet. On iOS the equivalent `paykitUiEnabled` flag is not backed up, so restore resets it to the default.

I'm wondering if this is a concern for later: since this is a temporary feature flag, maybe it should not be backed up, otherwise old backup state could override whatever default we choose in the future.

piotr-iohk (dismissed):
LGTM functionally. Same behavior as iOS.

Only small concern is that this setting is backed up/restored on Android: https://github.com/synonymdev/bitkit-android/pull/957#issuecomment-4507780503
If that is intentional then all good. Otherwise it may be worth making both apps consistent, either backing it up on both or not backing it up on either.

ben-kaufman:
Yes thanks, I'll remove that from backup

app/src/main/java/to/bitkit/flags/PaykitFeatureFlags.kt
```diff
@@ -0,0 +1,11 @@
+package to.bitkit.utils
+
+import to.bitkit.BuildConfig
+
+object PaykitFeatureFlags {
```
ovitrif, line 5: nit: could sit into existing `.models` package instead, or even a new `.flags`

---
PS. I don't really like utils, it's way too generic, always ends up as a dump bin, usually I would even hesitate adding such a dir in any project.

app/build.gradle.kts
```diff
@@ -164,6 +164,7 @@ android {
         buildConfigField("boolean", "TREZOR_BRIDGE", trezorBridgeEnv)
         buildConfigField("String", "TREZOR_BRIDGE_URL", "\"$trezorBridgeUrlEnv\"")
         buildConfigField("boolean", "GEO", System.getenv("GEO")?.toBoolean()?.toString() ?: "true")
+        buildConfigField("boolean", "PAYKIT_UI_DISABLED", System.getenv("PAYKIT_UI_DISABLED")?.toBoolean()?.toString() ?: "false")
```
ovitrif: nit: could add a prefix to foresee easier reasoning about these when it grows, suggesting `FEATURE_PAYKIT_OFF` or even `FEAT_PAYKIT`, but do note that would need it to be `ON` by default, thus the usage path needs to revert the condition.

ovitrif (dismissed):
tAck with some nits.

Special thanks for:
- naming `PaykitRouteGuard` like this 🙌🏻 
- adding `BackupRepoTest` and `SettingsViewModelTest`

ovitrif (approved):
re-reviewed utAck

PR #958, opened 2026-05-21
fix: hide widget size label

jvsena42:
This PR temporarily hides the widget size label in the widget size carousel.

### Description

While only the wide widget size is available, the carousel shows a single page, so the "WIDE"/"SMALL" caption below it adds no value and is now commented out alongside the existing single-page setup. The label and its string lookup will be restored once the small size widget variants are implemented.

### Preview

[Screen_recording_20260521_085359.webm](https://github.com/user-attachments/assets/ee06607a-d6f5-4e14-aa1a-7ae4c0581f60)

#### Manual Tests
- [x] **1.** Home → add/edit any widget → widget size carousel: single wide preview shows with no "WIDE"/"SMALL" caption below it.

#### Automated Checks
- N/A — UI-only change commenting out the size label; no automated coverage added or changed.


piotr-iohk (dismissed):
LGTM, makes sense - and consistent with iOS I believe.

ovitrif (approved):
utAck

Issue #953, opened 2026-05-19
v2.2.0: Native libraries not reproducible (libbitkitcore.so, libdatastore_shared_counter.so,    libjnidispatch.so, libpubky_app_specs)


xrviv:
We attempted to reproduce the Play Store release of Bitkit v2.2.0 (`to.bitkit`) as part of
WalletScrutiny's automated verification pipeline.

Build succeeded. The build command `./gradlew bundleMainnetRelease --no-build-cache 
--no-configuration-cache` completed in ~7 minutes using JDK 17 (Zulu), Android SDK build-tools 35/36, and bundletool v1.17.2 to extract split APKs from the produced AAB.

Commit: e6cd61e4c3896cbf37fc386690024994eec24a30 (tag `v2.2.0`)

Split APK Comparison Results

| Split | Result |
|----------------------------|-----------------------------------------------------|
| `base.apk` | ✅ Identical (modulo Play Store signing artifacts) |
| `split_config.xxhdpi.apk` | ✅ Identical (modulo Play Store signing artifacts) |
| `split_config.arm64_v8a.apk` | ❌ 4 native libraries differ |

Non-reproducible files in split_config.arm64_v8a.apk

- `lib/arm64-v8a/libbitkitcore.so`
- `lib/arm64-v8a/libdatastore_shared_counter.so`
- `lib/arm64-v8a/libjnidispatch.so`
- `lib/arm64-v8a/libpubky_app_specs-90a3a8b6be1a7a22.so`  ← new in v2.2.0

The same three libraries also differed in our `v2.1.2` verification. `libpubky_app_specs` is new in `v2.2.0`.

All `AndroidManifest.xml`, `resources.arsc`, `META-INF/`, and `stamp-cert-sha256` differences are Play Store injection artifacts and are excluded from this finding.

Likely Causes

- `libbitkitcore.so` and `libpubky_app_specs` are Rust/native libraries distributed via GitHub Packages AARs (`com.synonym:bitkit-core-android`, etc.). If the AARs themselves were not built reproducibly, the embedded .so files will differ regardless of how the app build is reproduced.
- `libdatastore_shared_counter.so` and `libjnidispatch.so` may differ due to how Play Store packages .so files vs. what `bundletool` extracts from a locally built AAB.

We have not yet run diffoscope on the differing libraries. If the developers can point to
reproducible-build documentation or upstream CI for the native AARs, that would help narrow the
root cause.

Environment

[...]

Issue #867, opened 2026-03-26
v2.1.2 reproducibility: native libraries in split_config.arm64_v8a differ from Play Store release


xrviv:
Hi Bitkit team,

  WalletScrutiny independently verifies that Bitcoin wallet apps on the Play Store can be reproduced 
  from their published source code. We attempted to reproduce the v2.1.2 Play Store release and found
   that most of the app reproduces cleanly — but three native libraries in split_config.arm64_v8a.apk
   differ between our local build and the official release.                       

  Build method

  ./gradlew bundleMainnetRelease --no-build-cache --no-configuration-cache                           
                                                                                                     
  - Tag: v2.1.2 (commit 5e711291e11aa4ac23ff24039c5158a8c8818bc0)                                    
  - JDK: Zulu 17                                                                                     
  - Android SDK build-tools: 36.0.0                                                                  
  - Split APKs extracted from built AAB using bundletool with device-spec: arm64-v8a, xxhdpi, SDK 36 
  - Official APKs sourced from Play Store                                                            
                                                                                                     
  Findings                                                                                           
                                                                                                     
  base.apk and split_config.xxhdpi.apk match the Play Store release (only expected signing and Play  
  Store injection artifacts differ).
                                                                                                     
  split_config.arm64_v8a.apk has three differing native libraries:                                   
  
  lib/arm64-v8a/libbitkitcore.so        *** differs ***                                              
  lib/arm64-v8a/libdatastore_shared_counter.so  *** differs ***                                      
  lib/arm64-v8a/libjnidispatch.so       *** differs ***
                                                                                                     
  All other files in this split are explained by Play Store signing artifacts (META-INF/,            
  stamp-cert-sha256, com.android.vending.derived.apk.id in manifest).                                
                                                                                                     
  Questions                                                                                          
  
  1. Is libbitkitcore.so built from source during bundleMainnetRelease, or is it a pre-built         
  artifact? If built, is the Rust/C++ build expected to be reproducible?          
  2. Are libdatastore_shared_counter.so and libjnidispatch.so pre-compiled AAR/JAR dependencies? If  
  so, are they pinned to specific versions and hashes?                                               
  3. Do you have reproducible build instructions or a known-good build environment (Docker image, NDK
   version, Rust toolchain version) that produces byte-for-byte matching .so files? 

Thanks for building an open-source Bitcoin wallet.                                                 
  
  — WalletScrutiny

ovitrif:
[...]

PR #959, opened 2026-05-21
ci: add release workflows

### LINKED ISSUE #1006: Add Release Workflows
Task to add 2 release workflows:
1. Store Release - builds mainnet artifacts (bundle for store, apks for github releases)
2. Internal Release - builds mainnet apk artifacts for any other reason

ovitrif:
Closes #1006
Related: #953
Supersedes: #867 (closed as outdated in favor of #953)

This PR:
1. Adds protected manual release workflows for signed mainnet store and internal artifacts.
2. Adds debug-only Firebase placeholders so fresh clones can build dev and testnet debug variants without private Firebase files.
3. Adds a WalletScrutiny-oriented manual reproducible release workflow, local reproduction script, and documentation.

> [!CAUTION]
> If you already have a private debug Firebase config at `app/google-services.json`, move it before checking out this PR or pulling `master` after this PR merges:
>
> ```sh
> mkdir -p app/src/debug
> mv app/google-services.json app/src/debug/google-services.json
> ```
>
> `app/google-services.json` is now the tracked placeholder for fresh clones. Keep `app/src/mainnetRelease/google-services.json` unchanged for release builds.

### Description

The core change is adding protected manual `release` and `release-internal` GitHub Actions workflows. Both workflows build signed mainnet release artifacts from protected environment secrets. `release` keeps both APK and AAB outputs for Play/GitHub release handling, while `release-internal` produces the mainnet APK signed with the internal keystore. These workflows do not run automatically on pushed tags; `/release` remains the release-process orchestrator until it is deliberately updated to consume workflow-produced artifacts.

The reproducibility support adds a manual `Reproducible Release` workflow and `scripts/reproduce-release.sh` to build `bundleMainnetRelease`, recreate APK splits with bundletool, extract `arm64-v8a` native libraries, and upload checksum evidence. When a comparison artifact is provided, `diffoscope` differences now fail the workflow while still uploading the generated report.

GitHub Actions setup:

- Created the `release` and `release-internal` GitHub environments for the protected release workflows.
- Configured the required `release` environment secrets:
  - `MAINNET_RELEASE_GOOGLE_SERVICES_JSON_BASE64`,
  - `BITKIT_KEYSTORE_BASE64`,
  - `BITKIT_KEYSTORE_PASSWORD`,
  - `BITKIT_KEY_ALIAS`, and
  - `BITKIT_KEY_PASSWORD`.
- Configured the required `release-internal` environment secrets:
  - `MAINNET_RELEASE_GOOGLE_SERVICES_JSON_BASE64`,
  - `INTERNAL_KEYSTORE_BASE64`,
  - `INTERNAL_KEYSTORE_PASSWORD`,
  - `INTERNAL_KEY_ALIAS`, and
  - `INTERNAL_KEY_PASSWORD`.
- `GPR_USER` and `GPR_TOKEN` remain optional fallback secrets; the workflows prefer the built-in GitHub Actions token with `packages: read`.

Remaining upstream work for #953:

- Update `/release` so it can trigger/download the `release` workflow artifacts instead of relying on local APK creation.
- Make `bitkit-core-android` / Rust-native AAR builds reproducible upstream.
- Pin Rust toolchain, Android NDK, `Cargo.lock`, build paths, `SOURCE_DATE_EPOCH`, path remapping, stripping, and published native `.so` checksums.
- Continue investigating third-party native outputs from `androidx.datastore:datastore-core` and `net.java.dev.jna:jna` with diffoscope evidence.

### Preview

N/A

#### Manual Tests
- [x] **1.** 2026-05-25 → Firebase config setup: confirmed real debug Firebase config belongs in the ignored shared debug path (`app/src/debug/google-services.json`) so one file can override the checked-in `app/google-services.json` placeholder without dirtying Git; `app/src/mainnetRelease/google-services.json` remains the release-only path.
- [x] **2.** 2026-05-25 → Release workflow review: confirmed store release keeps APK + AAB outputs, internal release keeps APK output, and both run post-build signature verification before artifact upload.

#### Automated Checks
- `bash -n scripts/reproduce-release.sh`
- YAML parse for `.github/workflows/release.yml`, `.github/workflows/release-internal.yml`, and `.github/workflows/reproducible-release.yml`
- YAML parse for all workflows after moving debug Firebase secrets to the shared `app/src/debug/google-services.json` path while keeping release secrets at `app/src/mainnetRelease/google-services.json`
- `bash -n` for the release signature verification shell blocks
- `git diff --check`
- `git check-ignore -v app/src/debug/google-services.json app/src/mainnetRelease/google-services.json`
- Google Services plugin lookup order inspected from local 4.4.4 sources: `src/debug/google-services.json` resolves before the checked-in root `app/google-services.json` placeholder, while `src/mainnetRelease/google-services.json` remains the release override. The root placeholder intentionally omits the production `to.bitkit` client.
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/reproducible-release.yml .github/workflows/release.yml .github/workflows/release-internal.yml`
- GitHub checks on `4aa52470d`: lint and detekt passed; build/E2E jobs were skipped while the PR was draft.
- Workflow behavior must be verified in GitHub Actions (after merge).


.github/workflows/reproducible-release.yml
```diff
@@ -0,0 +1,102 @@
...
+          umask 077
+          keystore_path="$RUNNER_TEMP/bitkit.keystore"
+          printf '%s' "$BITKIT_KEYSTORE_BASE64" | base64 --decode > "$keystore_path"
+          echo "KEYSTORE_FILE=$keystore_path" >> "$GITHUB_ENV"
+
+      - name: Download comparison artifact
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.comparison_artifact_name != '' }}
+        uses: actions/download-artifact@v6
+        with:
+          name: ${{ inputs.comparison_artifact_name }}
```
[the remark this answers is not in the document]
ovitrif: Resolved in 5d368ac

scripts/reproduce-release.sh
```diff
@@ -0,0 +1,156 @@
...
+rm -rf "$artifacts_dir" "$checksums_dir" "$extracted_dir" "$native_dir"
+mkdir -p "$artifacts_dir" "$checksums_dir" "$extracted_dir" "$native_dir"
+
+if [[ "${SKIP_GRADLE_BUILD:-false}" != "true" ]]; then
+    ./gradlew bundleMainnetRelease --no-daemon --stacktrace
+fi
+
+aab_path=${AAB_PATH:-}
+if [[ -z "$aab_path" ]]; then
+    aab_path=$(find app/build/outputs/bundle/mainnetRelease -name 'bitkit-mainnet-release-*.aab' | LC_ALL=C sort | tail -n 1)
```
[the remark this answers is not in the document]
ovitrif: Resolved in 3b24fd1

.github/workflows/reproducible-release.yml
```diff
@@ -0,0 +1,121 @@
...
+          set -euo pipefail
+          test -n "$COMPARISON_RUN_ID"
+
+      - name: Download comparison artifact
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.comparison_artifact_name != '' }}
+        uses: actions/download-artifact@v6
+        with:
+          name: ${{ inputs.comparison_artifact_name }}
+          path: ${{ runner.temp }}/comparison
+          github-token: ${{ github.token }}
```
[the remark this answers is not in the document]
ovitrif: Resolved in 26f4f24

.github/workflows/reproducible-release.yml
```diff
@@ -0,0 +1,121 @@
...
+          GPR_TOKEN: ${{ secrets.GPR_TOKEN || github.token }}
+          GITHUB_TOKEN: ${{ secrets.GPR_TOKEN || github.token }}
+          KEYSTORE_PASSWORD: ${{ secrets.BITKIT_KEYSTORE_PASSWORD }}
+          KEY_ALIAS: ${{ secrets.BITKIT_KEY_ALIAS }}
+          KEY_PASSWORD: ${{ secrets.BITKIT_KEY_PASSWORD }}
+          OUTPUT_DIR: ${{ runner.temp }}/reproducible-release
+        run: |
+          set -euo pipefail
+          if [ -d "$RUNNER_TEMP/comparison" ]; then
+            export DIFFOSCOPE_COMPARE_DIR="$RUNNER_TEMP/comparison"
```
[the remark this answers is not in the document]
ovitrif: Resolved in 26f4f24

README.md
```diff
@@ -10,11 +10,13 @@ This repository contains the **native Android app** for Bitkit.
 
 #### 1. Firebase Configuration
 
-Download `google-services.json` from the Firebase Console for each of the following build flavor groups,:
+Debug builds include placeholder Firebase configuration so a fresh clone can compile without a private `google-services.json` files.
+
+Download `google-services.json` from the Firebase Console when you need real Firebase integration for push notifications testing:
 - dev/tnet/mainnetDebug: Place in `app/google-services.json`
 - mainnetRelease: Place in `app/src/mainnetRelease/google-services.json`
```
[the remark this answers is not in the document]
ovitrif: Resolved in 56cc5b3

scripts/reproduce-release.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
+cd "$repo_root"
+
+output_dir=${OUTPUT_DIR:-.ai/reproducible-release}
+bundletool_version=${BUNDLETOOL_VERSION:-1.18.1}
+bundletool_sha256=${BUNDLETOOL_SHA256:-a73341a7945abcb0e6b8971c7b1b2801bd765006447ca0d2437a4260d572ceac}
+bundletool_url=${BUNDLETOOL_URL:-https://repo1.maven.org/maven2/com/android/tools/build/bundletool/${bundletool_version}/bundletool-${bundletool_version}.jar}
```
ben-kaufman: The default `bundletool_url` points at Maven Central, but this artifact returns 404 there:

`https://repo1.maven.org/maven2/com/android/tools/build/bundletool/1.18.1/bundletool-1.18.1.jar`

A fresh run without `BUNDLETOOL_JAR` fails in `download_bundletool` before creating the APK set. Could we switch it to the Google Maven URL instead?

`https://dl.google.com/dl/android/maven2/com/android/tools/build/bundletool/1.18.1/bundletool-1.18.1.jar`

README.md
```diff
@@ -10,11 +10,15 @@ This repository contains the **native Android app** for Bitkit.
 
 #### 1. Firebase Configuration
 
-Download `google-services.json` from the Firebase Console for each of the following build flavor groups,:
-- dev/tnet/mainnetDebug: Place in `app/google-services.json`
+Debug builds include placeholder Firebase configuration so a fresh clone can compile without private `google-services.json` files.
+
+Download `google-services.json` from the Firebase Console when you need real Firebase integration for push notifications testing:
+- devDebug: Place in `app/src/devDebug/google-services.json`
```
ben-kaufman: These debug Firebase paths are tracked placeholder files, so replacing them with real configs makes Git show secret-bearing changes. Could we document a safe local workflow, or adjust the config strategy so real Firebase files stay ignored?

scripts/reproduce-release.sh
ovitrif: Resolved in 02bd93f7a

README.md
ovitrif: Resolved in 02bd93f7a

.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+          set -euo pipefail
+          artifact_dir="$RUNNER_TEMP/release"
+          mkdir -p "$artifact_dir"
+          find app/build/outputs/bundle/mainnetRelease -name 'bitkit-mainnet-release-*.aab' -print0 |
+            xargs -0 -I {} cp {} "$artifact_dir/"
+          find app/build/outputs/apk/mainnet/release -name 'bitkit-mainnet-release-*.apk' -print0 |
+            xargs -0 -I {} cp {} "$artifact_dir/"
+          (cd "$artifact_dir" && sha256sum *.aab *.apk > SHA256SUMS.txt)
+          echo "artifact_dir=$artifact_dir" >> "$GITHUB_OUTPUT"
+
```
jvsena42, line 117: nit: could add a post-build signature verification `apksigner verify / jarsigner -verify ` as defensive check for a misconfigured release

jvsena42 (dismissed):

.github/workflows/release.yml
ovitrif: Resolved in 6fbe9ce87

jvsena42 (dismissed):

ovitrif:
> The debug files above are tracked placeholders. Before replacing a debug placeholder with a real Firebase config, hide that local-only change from Git:
> 
> ```sh
> git update-index --skip-worktree app/src/devDebug/google-services.json
> cp /secure/path/google-services.json app/src/devDebug/google-services.json
> ```

Reworking this from above, I don't really like this approach.

---
EDIT: done and force-pushed squashed work in 1 commit.

PR description updated to inform about the "migration" requirement to the new setup which ignores all but placeholder, and requires local clones to move current `app/google-services.json` to the git-ignored `app/debug/google-services.json`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+          while IFS= read -r -d '' apk_path; do
+            apk_count=$((apk_count + 1))
+            "$apksigner_path" verify --verbose --print-certs "$apk_path"
+          done < <(find app/build/outputs/apk/mainnet/release -name 'bitkit-mainnet-release-*.apk' -print0)
+          test "$apk_count" -gt 0
+
+          bundle_count=0
+          while IFS= read -r -d '' bundle_path; do
+            bundle_count=$((bundle_count + 1))
+            jarsigner -verify -verbose -certs "$bundle_path"
```
ben-kaufman: `jarsigner -verify` can return 0 for an unsigned bundle and only print `jar is unsigned.` The Gradle build should normally sign the AAB when the release signing config is valid, but this post-build check would not catch an unsigned AAB if that ever regressed. Could we add `-strict` or otherwise fail when the output says the bundle is unsigned?

app/google-services.json
```diff
@@ -0,0 +1,67 @@
+{
+  "project_info": {
+    "project_number": "000000000000",
+    "project_id": "bitkit-debug-placeholder",
+    "storage_bucket": "bitkit-debug-placeholder.appspot.com"
+  },
+  "client": [
+    {
```
ben-kaufman: The root placeholder can satisfy `mainnetRelease` too, because it is the module fallback and includes the `to.bitkit` client. I verified `processMainnetReleaseGoogleServices` succeeds without `app/src/mainnetRelease/google-services.json` and parses `app/google-services.json`. Could we add a release guard so release variants fail unless the real `app/src/mainnetRelease/google-services.json` exists?

ovitrif: IDK if needed, I would treat fallbacks the way they work in this case, because devs involved in actual releases are supposed to know what they're doing, wdyt?

ben-kaufman: Yes the protected workflow should be fine since it writes `app/src/mainnetRelease/google-services.json` first. My concern is just that this changes the fallback behavior: `mainnetRelease` now succeeds with the placeholder if that file is missing. For release builds I think it’s better to fail loudly than rely on noticing the fallback.

ovitrif: Oh, I think I get what changed now, which shouldn't have changed and stay as it used to be, in my view: placeholder shouldn't cover `to.bitkit`…
Former fallback wasn't supposed to cover for it either, AFAICR.

In either case, it was a different "app" in Firebase even if it had `to.bitkit`, totally different environment useful for testing maybe, not the actual PROD one. Real PROD sits in the mainnetRelease `google-services.json`.

Gonna fix it, thanks 🙏🏻

.github/workflows/release.yml
ovitrif: Resolved in 13528fa0f

app/google-services.json
ovitrif: Resolved in 13528fa0f

ben-kaufman (dismissed):

ovitrif:
> - Created the `release` and `release-internal` GitHub environments for the protected release workflows.
> - Required `release` environment secrets:
	- `MAINNET_RELEASE_GOOGLE_SERVICES_JSON_BASE64`,
	- `BITKIT_KEYSTORE_BASE64`,
	- `BITKIT_KEYSTORE_PASSWORD`,
	- `BITKIT_KEY_ALIAS`, and
	- `BITKIT_KEY_PASSWORD`.
> - Required `release-internal` environment secrets:
	- `MAINNET_RELEASE_GOOGLE_SERVICES_JSON_BASE64`,
	-  `INTERNAL_KEYSTORE_BASE64`,
	- `INTERNAL_KEYSTORE_PASSWORD`,
	- `INTERNAL_KEY_ALIAS`, and
	- `INTERNAL_KEY_PASSWORD`.
> 
> These will be added AFTER the PR is approved.

Environments are now configured with the secrets enumerated above (excerpt from PR description)

Sample view of `release` repo environment:
|  |
| - |

We don't need `GPR_USER` and `GPR_TOKEN`, verified locally with:
```sh
GPR_USER='ovitrif' \
GPR_TOKEN='***' \
./gradlew compileDevDebugKotlin --refresh-dependencies --no-daemon

To honour the JVM settings for this build a single-use Daemon process will be forked. For more on this, please refer to https://docs.gradle.org/8.13/userguide/gradle_daemon.html#sec:disabling_the_daemon in the Gradle documentation.
Daemon JVM discovery is an incubating feature.
Daemon will be stopped at the end of the build 

> Task :app:compileDevDebugKotlin
w: ATTENTION!
This build uses unsafe internal compiler arguments:

-XXLanguage:+PropertyParamAnnotationDefaultTargetMode

This mode is not recommended for production use,
as no stability/compatibility guarantees are given on
compiler or generated code. Use it at your own risk!

w: file:///Volumes/ssd/r/synonym/bitkit-android-3/app/src/main/java/to/bitkit/models/PubkyProfile.kt:88:13 Redundant creation of Json format. Creating instances for each usage can be slow.
w: file:///Volumes/ssd/r/synonym/bitkit-android-3/app/src/main/java/to/bitkit/ui/screens/trezor/TrezorScreen.kt:93:34 'fun <reified VM : ViewModel> hiltViewModel(viewModelStoreOwner: ViewModelStoreOwner = ..., key: String? = ...): VM' is deprecated. Moved to package: androidx.hilt.lifecycle.viewmodel.compose.

BUILD SUCCESSFUL in 3m 34s
21 actionable tasks: 13 executed, 8 from cache
```

.github/workflows/release-internal.yml
```diff
@@ -0,0 +1,102 @@
+name: Release Internal
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
```
piotr-iohk: If I understand correctly the intention of the flow is to create internal mainnet build - not officially signed - e.g. for ad-hoc testing purposes? In that case I don't think this one should be pushed on tag.

.github/workflows/reproducible-release.yml
```diff
@@ -0,0 +1,119 @@
...
+        description: Optional artifact name to compare against with diffoscope
+        required: false
+        default: ''
+      comparison_run_id:
+        description: Workflow run id that produced the comparison artifact
+        required: false
+        default: ''
+  push:
+    tags:
+      - 'v*'
```
piotr-iohk: I don't think this one should be pushed on tag either.
In fact, it is not really clear to me why we need `reproducible-release.yml` in our pipeline.

.github/workflows/release.yml
```diff
@@ -0,0 +1,126 @@
+name: Release
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*'
```
piotr-iohk: If I understand correctly the main goal of `release.yml` is to create artifact for release.
Not sure, if this should be pushed on tag automatically. I think /release command need to be adjusted to account for that. Currently also release process (driven by the /release command) allows re-tagging (e.g. after some changes to release branch) - not sure how pushing on tag would behave in such cricumstance - additional release artifact created? This should be checked.

scripts/reproduce-release.sh
```diff
@@ -0,0 +1,182 @@
...
+
+checksum_tree "$artifacts_dir" "$checksums_dir/release-artifacts.sha256"
+checksum_tree "$extracted_dir" "$checksums_dir/extracted-apks.sha256"
+checksum_tree "$native_dir" "$checksums_dir/arm64-native-libs.sha256"
+
+if [[ -n "${DIFFOSCOPE_COMPARE_DIR:-}" && -d "$DIFFOSCOPE_COMPARE_DIR" ]]; then
+    if command -v diffoscope >/dev/null 2>&1; then
+        diffoscope "$DIFFOSCOPE_COMPARE_DIR" "$extracted_dir" \
+            --html "$output_dir/diffoscope.html" \
+            > "$output_dir/diffoscope.txt" || true
```
piotr-iohk: Should this `diffoscope` comparison be non-blocking? With `|| true`, the workflow passes even when artifacts differ, so it’s evidence-only rather than a reproducibility check.

piotr-iohk (dismissed):
Not clear about the `reproducible-release.yml`. Also I think that `/release` command and process should be adjusted such that it takes into account `release.yml` producing release artifacts. Please see individual comments.

.github/workflows/release-internal.yml
ovitrif: Resolved in 117d56eb5

.github/workflows/reproducible-release.yml
ovitrif: Resolved in 117d56eb5

.github/workflows/release.yml
ovitrif: Resolved in 117d56eb5

scripts/reproduce-release.sh
ovitrif: Resolved in 117d56eb5

scripts/reproduce-release.sh
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+repo_root=$(git rev-parse --show-toplevel 2>/dev/null || pwd)
+cd "$repo_root"
+
+output_dir=${OUTPUT_DIR:-.ai/reproducible-release}
+bundletool_version=${BUNDLETOOL_VERSION:-1.18.1}
+bundletool_sha256=${BUNDLETOOL_SHA256:-a73341a7945abcb0e6b8971c7b1b2801bd765006447ca0d2437a4260d572ceac}
+bundletool_url=${BUNDLETOOL_URL:-https://dl.google.com/dl/android/maven2/com/android/tools/build/bundletool/${bundletool_version}/bundletool-${bundletool_version}.jar}
```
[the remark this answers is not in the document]
ovitrif: Resolved in be71ce1

jvsena42:
### Local test: is the now-blocking diffoscope comparison safe?

Since the diffoscope compare is now blocking, I checked locally whether two honest builds with the same keystore yield byte-identical signed APKs, or whether signature non-determinism could cause false failures.

This only needs to exercise the script's signing path (`bundletool build-apks` + `apksigner`), so no Gradle build / GPR creds / Firebase config required — just an existing `mainnetRelease` AAB intermediate and a throwaway keystore.

**Setup**

```sh
WORK=.ai/repro-det-test; mkdir -p "$WORK"
# any mainnetRelease bundle works; I reused the on-disk intermediate
cp app/build/intermediates/intermediary_bundle/mainnetRelease/packageMainnetReleaseBundle/intermediary-bundle.aab \
   "$WORK/bitkit-mainnet-release-999.aab"
export BUNDLETOOL_JAR="$PWD/$WORK/bundletool.jar"   # shared across runs

keytool -genkeypair -keystore "$WORK/rsa.keystore" -storepass test1234 -keypass test1234 \
  -alias testkey -keyalg RSA -keysize 2048 -validity 30 -dname "CN=test, O=test, C=US"
keytool -genkeypair -keystore "$WORK/ec.keystore"  -storepass test1234 -keypass test1234 \
  -alias testkey -keyalg EC -groupname secp256r1 -validity 30 -dname "CN=test, O=test, C=US"
```

**Test A — run `reproduce-release.sh` twice with the same RSA key, compare checksums**

```sh
export SKIP_GRADLE_BUILD=true
export AAB_PATH="$PWD/$WORK/bitkit-mainnet-release-999.aab"
export KEYSTORE_FILE="$PWD/$WORK/rsa.keystore" KEYSTORE_PASSWORD=test1234 KEY_ALIAS=testkey KEY_PASSWORD=test1234

OUTPUT_DIR="$WORK/rsa-run1" scripts/reproduce-release.sh
OUTPUT_DIR="$WORK/rsa-run2" scripts/reproduce-release.sh
diff "$WORK/rsa-run1/checksums/extracted-apks.sha256" "$WORK/rsa-run2/checksums/extracted-apks.sha256"
# -> no diff: byte-identical extracted APKs across runs ✅
```

**Test B — sign the same APK twice with `apksigner` (v2/v3), RSA vs EC**

```sh
APKSIGNER=$ANDROID_HOME/build-tools/36.0.0/apksigner
base=$(find "$WORK/rsa-run1/extracted-apks" -name '*.apk' | head -1)
for alg in rsa ec; do
  for f in a b; do
    cp "$base" "$WORK/$alg-$f.apk"
    "$APKSIGNER" sign --ks "$WORK/$alg.keystore" --ks-pass pass:test1234 --key-pass pass:test1234 \
      --ks-key-alias testkey --v1-signing-enabled false --v2-signing-enabled true \
      --v3-signing-enabled true --min-sdk-version 24 "$WORK/$alg-$f.apk"
  done
  echo "$alg:"; shasum -a 256 "$WORK/$alg-a.apk" "$WORK/$alg-b.apk"
done
```

**Results**

| Test | Key | Outcome |
|------|-----|---------|
| A — script ×2, same key | RSA 2048 | `extracted-apks.sha256` **byte-identical** ✅ |
| B — apksigner ×2, same key | RSA 2048 | both signatures **identical** ✅ |
| B — apksigner ×2, same key | EC secp256r1 | hashes **differ** ❌ (ECDSA random nonce) |

**Conclusion**

The blocking diffoscope step is safe **iff the release/internal keystores use RSA keys** (RSA PKCS#1 v1.5 signatures are deterministic → two honest builds are byte-identical). This is the common case for Android signing keys. With an ECDSA key the signature block is non-deterministic and would cause false reproducibility failures.

Side note: bundletool `build-apks --mode=default` actually **rejects** an EC key outright (`ECDSA signatures only supported for minSdkVersion 18 and higher`, because v1/JAR signing is enabled without a minSdk), so an EC release key would break the workflow regardless.

So this reduces to confirming the key algorithm once:

```sh
keytool -list -v -keystore <release.keystore> -alias <alias> | grep -i 'Signature algorithm'
```

ovitrif:
@jvsena42 @ben-kaufman wdyt of splitting this into 2:
1. for release workflows (stays here)
2. for reproducibility

I keep seeing it is getting dragged by reproducibility concerns, while release workflow could've been already useful for building `2.3.0` artifacts.

ovitrif:
Closing this combined PR in favor of the split replacements:

- #1011
- #1012

This keeps release artifact automation reviewable separately from WalletScrutiny/reproducibility work, per https://github.com/synonymdev/bitkit-android/pull/959#issuecomment-4681157475.

PR #971, opened 2026-05-26
fix: localize copyright year

piotr-iohk:
### Description

Updates the Support page copyright text so it no longer hardcodes `2025`.

- Changes the English source string to use the shared Transifex `{year}` placeholder.
- Replaces `{year}` with the current calendar year at runtime.
- Keeps a temporary `2025` replacement fallback so existing translated strings continue to show the current year until Transifex syncs updated translations.
- Adds a changelog fragment for the user-visible Support page copy fix.

Transifex note: push the source string to Transifex, then pull translated strings once translations are ready. The runtime fallback keeps current localized strings correct during the transition.

### Preview

TBD

### QA Notes

- Open Settings > Support and verify the copyright line shows the current year.
- Verify at least one translated locale still displays the current year through the legacy `2025` fallback until Transifex translations are updated.
- Verified `git diff --check` passes.
- Verified Cursor lints reported no errors.
- Not run: Gradle build or detekt.


changelog.d/next/971.fixed.md
ovitrif, line 1: Not sure but I noticed this naming issue and it does somehow self-explains itself*… IIRC these fragments were supposed to have the `{PR_ID}` as prefix before the first persiod in their name, but they ain't getting it by AI.

\* Agent would need instructions to "compute" the next PR id while work is being done, or to read it after PR creation then amend PR with PR id as file name prefix.

ovitrif (dismissed):
utAck

ovitrif (approved):
utAck

Thanks, this is much better now, both code and more hopeful for the `/pr` command to behave as expected 🙏🏻

PR #974, opened 2026-05-27
Add legacy RN close recovery flow

ben-kaufman:
## Summary
- Bumps bitkit-core to v0.1.63 for legacy RN native SegWit close-output recovery support.
- Adds a Dev Settings recovery flow with a configurable scan index limit, defaulting to 10,000.
- Scans for affected close outputs, shows the found amount/output count, prepares a sweep preview, and broadcasts only after explicit user confirmation.

## Notes
- Recovery sweeps to the current wallet on-chain address using the app's current network and Electrum config.
- The flow is intentionally in Dev Settings so support can start it manually instead of auto-sending user funds.
- The UI reuses existing Bitkit settings/navigation/button/text input components to stay aligned with the app style.

## Validation
- `./gradlew detekt` (passes; remaining printed findings are pre-existing import-order findings in unrelated files)
- `./gradlew testDevDebugUnitTest`


pwltr (commented):
tACK

pwltr (dismissed):

app/src/main/java/to/bitkit/ui/screens/settings/LegacyRnRecoveryScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+        color = Colors.White64,
+    )
+    Spacer(modifier = Modifier.height(12.dp))
+    TextInput(
+        value = state.indexLimit,
+        onValueChange = onIndexLimitChange,
+        placeholder = "10000",
+        singleLine = true,
+        keyboardOptions = KeyboardOptions(keyboardType = KeyboardType.Number),
+        modifier = Modifier.fillMaxWidth(),
```
ovitrif: nit: shouldn't have comma, but at this point I give up on tweaking agent rules, they're never getting this; plus, it's less and less important to optimise the codebase for hand-written code, so the weight of this "custom rule" for our codebase is less relevant. Still, for consistency reasons we might need at one point to go over all the code and just add the trailing comma after all modifiers, then we can drop these agent instructions altogether.

app/src/main/java/to/bitkit/ui/screens/settings/LegacyRnRecoveryScreen.kt
ovitrif, line 1: nit: should add a private `@Preview` composable for IDE previews.

gradle/libs.versions.toml
```diff
@@ -21,7 +21,7 @@ activity-compose = { module = "androidx.activity:activity-compose", version = "1
 appcompat = { module = "androidx.appcompat:appcompat", version = "1.7.1" }
 barcode-scanning = { module = "com.google.mlkit:barcode-scanning", version = "17.3.0" }
 biometric = { module = "androidx.biometric:biometric", version = "1.4.0-alpha05" }
-bitkit-core = { module = "com.synonym:bitkit-core-android", version = "0.1.58" }
+bitkit-core = { module = "com.synonym:bitkit-core-android", version = "0.1.63" }
```
ovitrif: Should update to `*.64` which includes new bindings with the changes of both these PRs that were open simultaneously:
- synonymdev/bitkit-core#91
- synonymdev/bitkit-core#92

ovitrif (changes requested):
LGTM all except the `bitkit-core` version pin which needs to be updated.

ovitrif (approved):
utAck

PR #979, opened 2026-05-31
fix: handle samrock bitkit links

ovitrif:
### Description

This PR:
1. Adds support for the Bitkit-owned SamRock wrapper deeplink so BTCPay setup links can open Bitkit reliably without relying on unverified wildcard web links.
2. Keeps raw SamRock setup QR and paste handling for BTCPay URLs, including installs served from a base path, while rejecting public HTTP payloads.
3. Surfaces BTC method-level setup failures from SamRock responses so failed BTCPay registration shows the useful server error.

#### Motivation

The previous manifest filter registered Bitkit for raw BTCPay `http` and `https` setup URLs on any host matching `/plugins/.../samrock/protocol`. That was both too broad and unreliable for the tap-to-open case: BTCPay setup URLs are usually served from arbitrary merchant or self-hosted domains, and Bitkit cannot verify all of those domains as Android App Links. On newer Android versions, unverified web links are not a dependable app-launch contract.

This PR keeps raw SamRock URLs supported where they are reliable, inside Bitkit scanner and paste flows, and moves external app launch to the Bitkit-owned `bitkit://btcpay/samrock?url=...` wrapper.

### Preview

N/A

#### Manual Tests
- [x] **1.** BTCPay setup QR → Bitkit scanner: BTCPay connection sheet opens for Bitcoin on-chain setup.
- [x] **2.** Bitkit BTCPay deeplink → open `bitkit://btcpay/samrock?url=...`: BTCPay connection sheet opens and query values are not exposed in logs.
- [x] **3.** `regression:` BTCPay setup URL under base path → scan or paste: setup is recognized.
- [x] **4.** `regression:` Public HTTP BTCPay setup link → scan or paste: invalid setup toast appears without opening send flow.
- [x] **5.** BTCPay descriptor rejection → Connect: error message shows the BTC method failure reason.

#### Automated Checks
- Unit tests updated: cover Bitkit SamRock wrapper parsing and redaction, base-path payloads, public HTTP rejection, ViewModel routing, and method-level failure messages.
- `git diff --check`
- `./gradlew testDevDebugUnitTest`
- `./gradlew detekt`


ben-kaufman (approved):

ovitrif:
@piotr-iohk Can you pls give me status on whether I should wait for `pubky_paykit` e2e to succeed and retrigger?

piotr-iohk:
The failure seems to be unrelated but consistently failing in the same place. Backup seems to be failing - on a check before restoring the wallet, which makes one of the pubky-profile tests to fail. Adjusted tests to wait longer for it.
Screenshot 2026-06-01 at 15 36 51

PR #985, opened 2026-06-03
feat: v61 compact widgets support

### LINKED ISSUE #965: [Feature]: add compact widget support in app
## Problem or use case

The 2.3.0 widget launch comms and release assets refer to both wide and compact widget layouts. Android currently ships wide in-app widget previews/layout views only, which creates a mismatch between the product experience and planned release messaging.

## Proposed solution

Add compact in-app preview/view support for the Android widgets included in the 2.3.0 widget release.

Scope:
- Show compact previews anywhere the app presents widget previews or layout choices.
- Ensure compact and wide variants match the v61 widget designs.
- Keep OS widget configuration and in-app widget discovery/edit screens aligned with release screenshots/video.
- Include all widgets referenced in the release comms for this launch.

## Alternatives considered

n/a

## Additional context

Slack thread: https://synonymworkspace.slack.com/archives/C07BJ7DNPCG/p1779299477247879

jvsena42:
Closes #965

This PR:

1. Adds a per-widget size (compact/wide) chosen from the preview sheet's size carousel, driving a new two-column home grid where wide widgets span the full width and compact widgets pair side by side.
2. Redesigns the home widgets edit mode as an inline grid with blurred cards, a dashed brand border, centered name and delete/settings/reorder actions, and location-based drag-to-reorder with spring placement animations matching iOS.
3. Adds compact calculator support with the number pad rendered as a bottom overlay, so both compact and wide calculators stay interactive and the focused card lifts above the pad in any position.
4. Persists the chosen size, includes it in backup (serialized to the iOS-compatible `small`/`wide` values), and updates the default widget set to the v61 layout.

### Description

Ports the iOS v61 widgets revamp to Android. The per-widget small card variants and size carousel already existed but were disabled; this wires them end to end:

- Size carousel in each preview sheet now swipes between compact and wide, opens on the saved size, and the selection is saved with the widget.
- The home screen renders both display and edit modes through a shared two-column flow layout. Suggestions is always wide; the calculator forces an interactive editor.
- Edit mode is an inline grid: cards stay in place, blurred under a scrim with a dashed border and the same 24dp/32dp action icons (cardinal-move reorder icon) across every widget including Suggestions, with the settings icon dimmed where it has no editor.
- Drag reordering only reorders when the target slot changes and resets when the finger returns over the dragged card, so it settles instead of flickering; the page still scrolls and taps on the blurred cards are inert.
- The calculator number pad moved out of the card into a screen-bottom overlay; the page scrolls so the focused calculator clears the pad.
- Size persists through the widgets store and rides along in the existing widgets backup payload; the default set is now suggestions, price, blocks, facts, weather, calculator, news.

### Preview

[fresh-wallet.webm](https://github.com/user-attachments/assets/01e07e4c-8aa2-4d0e-a6d5-0f319ec426fb)

[grid-edit.webm](https://github.com/user-attachments/assets/6824c1a8-812f-437f-b8ac-2da3795cfe62)

[restore.webm](https://github.com/user-attachments/assets/d605e850-726d-4804-81ab-8c4b5e117399)

[calculator-widget.webm](https://github.com/user-attachments/assets/5aa1d19d-4a0b-41ad-aef0-74b8550a8a6d)

#### Manual Tests

- [x] **1.** Add Widgets sheet → pick a widget → swipe size carousel to Small → Save: widget shows compact on home, paired side by side with another compact widget; wide widgets span full width; Suggestions always wide.
- [x] **2.** Re-open a saved widget's preview: carousel opens on the saved size; change size → Save: home updates.
- [x] **3.** Fresh wallet → home widgets: default set is suggestions, price, blocks, facts, weather, calculator, news, with blocks/facts/weather/calculator compact.
- [x] **4a.** Edit mode (top-right check icon) → each widget shows blurred content, dashed border, name, and delete/settings/reorder icons at the same size (incl. Suggestions, with its settings icon dimmed).
  - [x] **4b.** Drag the reorder handle across slots: reorders once per slot crossing and springs into place, no flicker.
  - [x] **4c.** Drag vertically on a card body (not the handle): the widgets area scrolls.
  - [x] **4d.** Tap a blurred card body: nothing happens (card inert); exit edit: order persists.
- [x] **5a.** Calculator set to compact → tap a row: bottom number pad overlay appears and edits the value.
  - [x] **5b.** `regression:` Calculator set to wide → tap a row: same bottom number pad, values stay in sync.
  - [x] **5c.** Calculator positioned at the bottom of the list → tap a row: card scrolls up and sits just above the number pad (not covered).
- [x] **6.** Backup round-trip: change a few sizes → wipe → restore: sizes are preserved.

#### Automated Checks

- Unit tests added: `app/src/test/.../components/WidgetGridSlotsTest.kt` covers the two-column flow geometry (pairing, lone trailing compact, totals) and the nearest-slot drag targeting incl. gap bias.
- Unit tests added: `app/src/test/.../models/WidgetSizeTest.kt` covers `WidgetSize.default`, iOS-compatible `small`/`wide` serialization, v60 size-less back-compat, the v61 default widget set, and backup-payload size round-trip.


jvsena42:
@codex review

jvsena42:
cc
https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-130669&m=dev#1785339966

piotr-iohk:
> cc https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-130669&m=dev#1785339966

Makes sense for facts and calc, but suggestions doesn't have compact type, so should be disabled for it, no?

Screenshot 2026-06-04 at 19 24 06

ovitrif:
e2e for widgets might succeed if the lost test-tag is restored:

P1: Missing home test tag for Calculator breaks the PR’s own E2E path. The new default widget set includes Calculator, and CI repeatedly waits for `CalculatorWidget`, but this PR never applies that tag to the calculator card in `WidgetCardContent`; other widgets keep `PriceWidget`, `BlocksWidget`, etc.
`Add Modifier.testTag("CalculatorWidget")` to the shared `calcModifier` so both compact and wide calculator cards expose it. See `HomeScreen.kt:1148`.

piotr-iohk:
When adding widgets iOS by default proposes small option, on Andorid it is different for different widgets - would be good to make it consistent. Also the position is not preserved when going back - contrary to iOS (try add facts and go back) - other than that LGTM:


https://github.com/user-attachments/assets/5cb83b7d-108f-4549-8f78-e77dcb8793e6


- added missing tag in [3d22482](https://github.com/synonymdev/bitkit-android/pull/985/commits/3d224825dd83e5d8509f19ad2b1adcec9956a23d)

ovitrif (commented):
All code changes LGTM. I was a bit concerned about the amount of code removed from `CalculatorCard`, but after looking through it more carefully I don’t think any parts of the calculator UI/UX logic were removed by accident.

Most of it looks like an architecture move for the compact/wide widget work: `CalculatorCard` is now mostly the visual card, `CalculatorNumberPadBar` owns the bottom number pad, and `HomeScreen` owns the active input / overlay positioning because the pad now needs to work outside both the compact and wide card. I also checked this against the older calculator PRs, especially the placeholder / active-input / formatting fixes, and the important data + formatting behavior still seems covered by the existing unit tests.

The main thing I still want to call out is numpad UX polish. In the previous calculator widget work I spent a lot of time tuning the NumPad show/dismiss behavior so it had smooth transitions with the numpad sliing, tab bar, and page layout. With this PR the pad is now conditionally displayed as an external bottom overlay, so the previous transition choreography is no longer really there; it feels more like the keypad pops in/out rigidly. I understand why the architecture changed for compact calculator support, but this is still a visible behavior change and I think it's a UX regression and might be nice to follow up after the merge/release, when there might be time for UX polishing focus.

One other thing I’d be careful with is dismiss/cleanup behavior. The old `CalculatorCard` owned more of that lifecycle and would clean up the active input state when the editor/card went away. Now that this is split between `HomeScreen`, `CalculatorCard`, and `CalculatorNumberPadBar`.

For next review pass when I'll test, I plan to specifically QA check the edge cases where the calculator input is active and then the user navigates away, switches between wallet/widgets pages, opens another sheet, enters edit mode, deletes the widget, or taps outside. Those are the areas where stale active input / stale keypad state could show up if any dismiss path was missed. These checks are planned after testing the actual new features added by this PR.

ovitrif (approved):
Tested and validated

https://github.com/user-attachments/assets/6b7208d5-8870-45d3-991b-a38ada59efc8

Didn't find many issues with Calculator widget, except that now there's a space between the calculator and its numpad when the calculator is placed right at the top of the widgets list, which looks quite odd. Hopefully not many users will have reasons to move it there; since the widgets list is already populated at the start. I also noticed the same regression and also regression on iOS regarding the initial slide-up/down transitions for calculator numpad reveal/dismiss; at least we're consistent across both platforms.

A few small glitches here and there, one is the constant jumping when user is undecided where to place the widget, but I don't consider it blocking, and another one which I didn't manage to capture in the recording was that the draggable widget card got shrinked when I moved it to the very top.

More concerned about changing widget settings which if triggered then loses the updated widget position -- see video -- but that shouldn't be blocking either.

Approved 👏🏻 

Other remarks: 
- on iOS the calculator numpad doesn't scroll when scrolling the page up, but it does actually look nicer on Android and there are no bugs around this behavior.
- on Android if I tap the toolbar (not any icon of it) while the calculator numpad shows, the pad doesn't hide. On iOS it does.
- small nit/ UX feedback (perhaps for @aldertnl) I'm not sure users will know they have to tap "Preview" to then be able to change the widget size. I myself got confused and was about to add a comment that other widgets except calculator can't be resized after being added; because the calculator shows the preview directly when the user taps the ⚙️ in edit view, due to it not having any other settings, while the other widgets show the settings first and only then they show the preview with the compact/wide carousel.

jvsena42:
> small nit/ UX feedback (perhaps for @aldertnl) I'm not sure users will know they have to tap "Preview" to then be able to change the widget size. I myself got confused and was about to add a comment that other widgets except calculator can't be resized after being added; because the calculator shows the preview directly when the user taps the ⚙️ in edit view, due to it not having any other settings, while the other widgets show the settings first and only then they show the preview with the compact/wide carousel.

@ovitrif  I also had the same impression https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=43100-273317&m=dev#1777777033

https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-130669&m=dev#1785339966

PR #1004, opened 2026-06-10
fix: request notification permission on enable

### LINKED ISSUE #566: Add a delay for Android notification permission system dialog
> > Do you mean we should deny it now? 🤔
> > exactly
> 
> Ok, got it. So in order for bg payments sheet to appear we need to do "Don't allow" on this system alert. (Verified it is happening indeed) I think I will keep it as is (Allow), because I see the alert re-appering during the test... which is a bit suboptimal and will surely cause flakiness to the tests.
> 
> So, we'll just ignore bg payments for andorid.

Good point, currently the system is fully deciding when stop displaying this dialog, but I could add a delay too.

_Originally posted by @jvsena42 in https://github.com/synonymdev/bitkit-android/issues/556#issuecomment-3702005583_

jvsena42:
Closes #566

This PR stops the Android notification permission system dialog from appearing automatically on the home screen, requesting it only when the user taps "Enable" on the background payments prompt.

### Description

The home screen previously asked `RequestNotificationPermissions` to launch the `POST_NOTIFICATIONS` system dialog on every entry while notifications were ungranted, so a user who dismissed it kept getting re-prompted on each app open. Issue #566 asked to gate how often this appears; this aligns Android with the iOS flow, where the dialog is only triggered from the background payments intro sheet's "Enable" button (`NotificationsSheet` → `requestPermission()`).

- The home-screen `RequestNotificationPermissions` now runs with `showPermissionDialog = false`, so it only keeps the granted/denied state in sync (including when the user changes it in system settings); it no longer launches the dialog.
- Tapping "Enable" on the background payments prompt requests the OS notification permission in place and dismisses the sheet; "Later" just dismisses it. "Enable" no longer navigates to the Background Payments settings screen, matching iOS.
- The runtime request is guarded behind Android 13+ (`TIRAMISU`); on older versions, where there is no runtime notification permission, "Enable" simply dismisses.

The background payments prompt itself is unchanged: it still appears only when notifications are not yet granted and the wallet has a Lightning spending balance.

### Preview

[before.webm](https://github.com/user-attachments/assets/05ea790e-139e-4b99-930c-a14bd656f07a)

[allow-flow.webm](https://github.com/user-attachments/assets/8314d9e7-3013-4d7e-870d-477f484e6433)

[deny-flow.webm](https://github.com/user-attachments/assets/3b7de43b-b872-44dc-aeaa-0cebf65fa006)

[suggestion-intro.webm](https://github.com/user-attachments/assets/b642af20-39dd-4598-aadf-797994764653)

[android-11-enable-flow.webm](https://github.com/user-attachments/assets/f3b5d42f-e017-4ab0-9c09-8100e71e1fd4)

[cjit.webm](https://github.com/user-attachments/assets/5fea7cda-299f-436f-a784-d0658e157cdc)

[transfer.webm](https://github.com/user-attachments/assets/404183b2-5e54-41ec-a306-eb29c87c6d51)

#### Manual Tests

- [x] **1.** LN spending balance + notifications not granted → open app to Home: no system permission dialog appears automatically.
- [x] **2.** Background payments prompt → tap Enable: OS permission dialog appears and the sheet dismisses with no navigation.
- [x] **3.** OS dialog → tap Allow: notifications enabled and the node keeps running in the background.
- [x] **4.** `regression:` Background payments prompt → tap Later: sheet dismisses, no OS dialog.
- [x] **5.** `regression:` Android 12 or below → tap Enable: navigate to payment settings (no runtime permission).

OBS: On Android 12 or bellow, the notifications are enabled by default

#### Automated Checks

- N/A — UI wiring change in `ContentView` only; no business logic added. `just compile` and `just lint` pass.

### DIFF
--- a/app/src/main/java/to/bitkit/ui/ContentView.kt
+++ b/app/src/main/java/to/bitkit/ui/ContentView.kt
@@ -1382,10 +1382,18 @@ private fun NavGraphBuilder.generalSettingsSubScreens(
     }
 
     composableWithDefaultTransitions<Routes.BackgroundPaymentsIntro> {
+        val notificationPermissionLauncher = rememberLauncherForActivityResult(
+            ActivityResultContracts.RequestPermission()
+        ) { granted ->
+            settingsViewModel.setNotificationPreference(granted)
+        }
         BackgroundPaymentsIntroScreen(
             onBack = { navController.popBackStack() },
             onLater = { navController.popBackStack() },
             onEnable = {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                    notificationPermissionLauncher.launch(Manifest.permission.POST_NOTIFICATIONS)
+                }
                 navController.navigateTo(Routes.BackgroundPaymentsSettings)
             },
         )
--- a/app/src/main/java/to/bitkit/ui/ContentView.kt
+++ b/app/src/main/java/to/bitkit/ui/ContentView.kt
@@ -200,6 +200,7 @@ import to.bitkit.ui.utils.AutoReadClipboardHandler
 import to.bitkit.ui.utils.RequestNotificationPermissions
 import to.bitkit.ui.utils.composableWithDefaultTransitions
 import to.bitkit.ui.utils.navigationWithDefaultTransitions
+import to.bitkit.ui.utils.rememberRequestNotificationPermission
 import to.bitkit.utils.Logger
 import to.bitkit.viewmodels.ActivityListViewModel
 import to.bitkit.viewmodels.AppViewModel
@@ -241,11 +242,10 @@ fun ContentView(
     val notificationsGranted by settingsViewModel.notificationsGranted.collectAsStateWithLifecycle()
     val walletExists = walletUiState.walletExists
 
-    val notificationPermissionLauncher = rememberLauncherForActivityResult(
-        ActivityResultContracts.RequestPermission()
-    ) { granted ->
-        settingsViewModel.setNotificationPreference(granted)
-    }
+    val requestNotificationPermission = rememberRequestNotificationPermission(
+        onPermissionResult = { granted -> settingsViewModel.setNotificationPreference(granted) },
+        onPreTiramisu = { navController.navigateTo(Routes.BackgroundPaymentsSettings) },
+    )
 
     // Effects on app entering fg (ON_START) / bg (ON_STOP)
     DisposableEffect(lifecycle) {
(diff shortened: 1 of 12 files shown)


app/src/main/java/to/bitkit/viewmodels/SettingsViewModel.kt
```diff
@@ -64,6 +67,21 @@ class SettingsViewModel @Inject constructor(
         }
     }
 
+    val shouldRequestNotificationPermission: StateFlow<Boolean> = combine(
+        settingsStore.data,
+        walletRepo.balanceState,
+    ) { settings, balance ->
+        !settings.notificationsGranted &&
+            !settings.notificationDialogShown &&
+            balance.totalLightningSats > 0UL
```
jvsena42: still not sure if it is a good idea gating it under LN balance, because the notification could be used for CJIT flow

jvsena42:
I prioritized matching the iOS flow, because the current behavior is really annoying for a user that don't want to enable notifications.

Some UX improvements could be addressed in follow-up PRs:

- Request permission once even without LN balance because of CJIT and channel opening, or at least in these specific flows
- Make the FG service optional, since it became less vital after ldk-node caching improvements
  - Could be an option in [background payments settings](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=40364-134589&m=dev). Would need some design work

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -896,8 +914,10 @@ private fun NavGraphBuilder.home(
         val isRecoveryMode by walletViewModel.isRecoveryMode.collectAsStateWithLifecycle()
         val hazeState = rememberHazeState()
 
+        // Only keep notification permission state in sync; the system dialog is requested
+        // from the background payments intro sheet, not automatically on the home screen.
         RequestNotificationPermissions(
-            showPermissionDialog = !isRecoveryMode,
+            showPermissionDialog = false,
```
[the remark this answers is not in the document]
jvsena42: fixed in https://github.com/synonymdev/bitkit-android/pull/1004/commits/f8f20a30c172343ad6c32f224f818c96d58f3e96

ovitrif:
> this aligns Android with the iOS flow, where the dialog is only triggered from the background payments intro sheet's "Enable" button

Should've been this way from native v1 🤝 , it's part of how Google suggests to implement these in their Android guidelines

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -501,8 +511,16 @@ fun ContentView(
...
                                             settingsViewModel.setBgPaymentsIntroSeen(true)
+                                            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                                                notificationPermissionLauncher.launch(
+                                                    Manifest.permission.POST_NOTIFICATIONS
+                                                )
+                                            } else {
+                                                // Pre-13 has no runtime permission dialog; open the
+                                                // in-app background payments settings instead.
+                                                navController.navigateTo(Routes.BackgroundPaymentsSettings)
+                                            }
```
ovitrif: nit: this belongs to a reusable top level fn, not inline in a handler, especially because other entrypoints should be handled the same way…

ovitrif (changes requested):
tAck

I would expect this to work in same way also in:
- Transfer to Spending > Toggle BG Payments ON
- CJIT Flow > same Toggle ON

Any specific reasons why it doesn't? Even If iOS doesn't do the same, I kinda still think Android should.

### Rec - Transfer to Spending
https://github.com/user-attachments/assets/109d6960-3581-4a80-a596-b446bf40a68d

jvsena42:
changes implement. re-testing after BT staging fix

jvsena42:
Videos updated

ovitrif (approved):
utAck final updates. Planning to retest.

PR #1013, opened 2026-06-15
fix: set transfer amount to max on cap error

jvsena42:
This PR makes the Transfer to Spending amount screen fill in the maximum transferable amount when you try to enter a value above the cap, instead of only flashing an error toast.

### Description

Previously, entering an amount above the spending maximum on the Transfer to Spending screen blocked the input and showed the "Spending Balance Maximum" toast while leaving the field unchanged. Now, when the cap is exceeded, the amount snaps to the current maximum allowed to send and the toast still appears, so it is a single step to continue with the largest possible transfer.

This ports the max-exceeded behavior from the iOS fix in synonymdev/bitkit-ios#595. The two-pass LSP max-client-balance clamping that the iOS PR also introduced already exists on Android in the transfer view model, so this change is limited to the input-snapping behavior.

### Preview

[Screen_recording_20260615_082049.webm](https://github.com/user-attachments/assets/4c7de922-4abd-4d77-99a7-982917a78583)

#### Manual Tests
- [ ] **1.** On-chain balance above the spending cap → Transfer → Spending Amount → type an amount above the maximum: input snaps to the maximum and the "Spending Balance Maximum" toast appears.
- [ ] **2.** `regression:` Spending Amount → tap MAX: amount fills to the maximum.
- [ ] **3.** `regression:` Spending Amount → tap 25%: amount is the quarter capped to the maximum.
- [ ] **4.** Spending Amount → amount at maximum → Continue: proceeds to the next step.

#### Automated Checks
- Local: `just compile` passes.

### DIFF
--- a/app/src/main/java/to/bitkit/ui/screens/transfer/SpendingAdvancedScreen.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/transfer/SpendingAdvancedScreen.kt
@@ -112,13 +112,16 @@ fun SpendingAdvancedScreen(
     LaunchedEffect(Unit) {
         amountInputViewModel.effect.collect {
             when (it) {
-                AmountInputEffect.MaxExceeded -> app.toast(
-                    type = Toast.ToastType.WARNING,
-                    title = context.getString(R.string.lightning__spending_advanced__error_max__title),
-                    description = context.getString(R.string.lightning__spending_advanced__error_max__description)
-                        .replace("{amount}", currentMaxLspBalance.formatToModernDisplay()),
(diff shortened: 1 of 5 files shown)


jvsena42:
I didn't implement on send flows because could cause issues on services that require exact amounts like Boltz if the user don't see that the amount was auto-capped

ovitrif (approved):
tAck

Very nice and intuitive, wanted to suggest this behavior for long time :)

ovitrif:
Updated branch via github web app because an e2e shard reported failure already, so waiting was warranted anyways.

PR #1016, opened 2026-06-15
fix: reset suggestions widget when re-added empty

### LINKED ISSUE #1010: Suggestion widget bug when empty and re-adding
### What happened?

Same as https://github.com/synonymdev/bitkit-ios/issues/596

### Expected behavior

_No response_

### Steps to Reproduce

_No response_

_No response_

### Bitkit Version

2.3.0 (182)

### Device / OS

Samsung s22 Android 16

### Reproducibility

Always

### Additional context

- Re-tested on Android after iOS report in [#596](https://github.com/synonymdev/bitkit-ios/issues/596); same repro on both platforms.

jvsena42:
Fixes #1010

This PR fixes the Suggestions widget disappearing after every card is dismissed, both when the widget is re-added and while editing the home grid. It is a port of the iOS fix in synonymdev/bitkit-ios#598.

### Description

When every suggestion card is dismissed with the [x] button, the Suggestions widget hides from the home grid, which is expected. Two follow-on problems remained, both caused by the dismissed cards being persisted so the widget kept resolving to an empty set:

1. Re-adding the widget (after removing it in edit mode) never brought it back, because the suggestions stayed empty. Re-adding now clears the dismissed cards when the widget would otherwise show nothing, restoring the default set. This only runs when the suggestions are empty, so re-adding while some cards are still visible leaves those dismissals untouched.

2. With all cards dismissed, editing the home grid showed a blank placeholder for the widget. When editing an empty Suggestions widget it now falls back to the static preview set (the same cards shown in the widget preview), so it stays visible and removable.

The suggestion-visibility logic was extracted from the home screen into a new shared repository, so the home grid and the widget preview flow resolve the visible cards the same way.

### Preview

[Screen_recording_20260615_114634.webm](https://github.com/user-attachments/assets/8c2291a3-027c-4c16-b731-5d9aefc8d7bb)

#### Manual Tests
- [x] **1.** Home → dismiss every Suggestions card with [x]: widget hides once the last card is dismissed.
- [x] **2.** edit mode → remove the Suggestions widget → re-add via Widgets list → Suggestions → Save: widget reappears showing the default suggestion cards.
- [x] **3.** dismiss every card (do not remove widget) → enter edit mode: widget shows the static preview set and can be reordered/removed, instead of a blank placeholder.
- [x] **4a.** `regression:` dismiss only some cards → remove and re-add the Suggestions widget: remaining cards stay as-is, dismissed ones are not restored.
  - [x] **4b.** `regression:` some cards still visible → enter edit mode: live cards show as normal (no preview fallback).

#### Automated Checks
- Unit tests added: cover the reset-when-empty behaviour and dismissed filtering in `app/src/test/java/to/bitkit/repositories/SuggestionsRepoTest.kt`.
- Local: `just test`, `just compile`, and `just lint` pass.


piotr-iohk (approved):
tACK

ovitrif (approved):
utAck

Great refactor too! 🙌🏻

PR #1037, opened 2026-06-22
fix: restore explore button on transfers

jvsena42:
This PR restores access to the on-chain details for transfer activities, including force closes. Ports [synonymdev/bitkit-ios#597](https://github.com/synonymdev/bitkit-ios/pull/597).

### Description

A force close, like any channel close, settles on-chain, so Bitkit already stores it as an on-chain activity with a transaction id, inputs, outputs, and a block explorer link. Those details had become unreachable from the activity detail screen: the transfer layout replaced the Explore button with a Connection button, and Connection only opens the channel, not the transaction.

- Keeps the Explore button in the shared Boost/Explore row for every activity, so the on-chain transaction details stay reachable.
- Moves the transfer-only Connection button to its own full-width row below, shown only for transfers.
- Leaves regular on-chain and Lightning activities unchanged. The Explore row applies to every transfer because a stored on-chain activity does not record whether a close was cooperative or forced, and both are real on-chain transactions with a transaction id worth surfacing.

Figma Designs:
- [Activity Item](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-129368&m=dev)

### Preview

[connection-oppened.webm](https://github.com/user-attachments/assets/459c22b6-332e-46da-b218-69df2573937e)

[regular-ln.webm](https://github.com/user-attachments/assets/e637e17f-3c16-484d-9181-51cac52edc65)

[regular-receive.webm](https://github.com/user-attachments/assets/9ab7d913-4956-44e6-b174-cd7bbe0a11ed)

[connection-closed.webm](https://github.com/user-attachments/assets/936e4854-da98-4298-a8ca-288eb0086fb7)

#### Manual Tests
- [x] **1.** Open a connection → close it → Activity → tap the From Spending transfer → Activity Detail: shows Boost and Explore in the top row, plus a new full-width Connection button below.
- [x] **2.** Activity Detail (transfer) → Explore → Activity Explorer: shows the transaction id, inputs, outputs, and Open Block Explorer.
- [x] **3.** Activity Detail (transfer) → Connection: still opens the Channel Detail.
- [x] **4.** `regression:` received on-chain activity → Activity Detail: Explore appears in its usual slot, with no extra Connection row.
- [x] **5.** `regression:` LN payment → Activity Detail: unchanged; Explore opens the preimage and invoice details.

#### Automated Checks
- `ActivityDetailViewModelTest.kt` is unaffected (no button-layout assertions); it still passes.
- Ran locally: `just compile` and `just lint`.


ovitrif (approved):
tAck

PR #1040, opened 2026-06-24
refactor: integrate paykit sdk

ben-kaufman:
This PR:
1. Replaces Bitkit's custom Paykit private/public payment plumbing with the native Paykit SDK.
2. Moves Pubky profile, contact, public endpoint, private endpoint, and SDK backup state handling through SDK APIs.
3. Keeps Bitkit responsible for wallet execution, payment-request mapping, contact attribution, endpoint rotation, and public fallback behavior.
4. Pins Paykit to the published `com.synonym:paykit-android:0.1.0-rc23` artifact.
5. Adds hardening for auth approval capabilities, canceled Ring auth cleanup, profile label fallback, pending private drain retries, and best-effort profile delete/sign-out cleanup.

### Description

- Adds a Paykit SDK service wrapper for session bootstrap, Ring auth, profile/avatar publishing, contact records, public endpoint sync, private payment list sync, and SDK backup state import/export.
- Refactors public and private Paykit repositories to resolve and publish payment endpoints through SDK APIs while preserving Bitkit's existing endpoint preference order and local payability checks.
- Moves private contact link and recovery state into the SDK backup string, while keeping Bitkit-owned address reservations and payment attribution in app storage.
- Updates Pubky profile/contact loading, profile edits, sign-out/delete cleanup, backup/restore, and wallet wipe flows for the SDK-backed state model.

### Preview

N/A - SDK integration; no UI layout change.

#### Manual Tests

- [ ] **1.** Profile -> create/edit profile -> add contact by Pubky key: contact profile resolves and remains visible after app restart.
- [ ] **2.** Send -> Contact -> select contact -> complete payment: private Paykit is attempted first and public endpoints remain available as fallback when private capability is unavailable.
- [ ] **3.** Backup/restore -> restore a wallet with Pubky state -> open contacts/pay contact: SDK state restores and contact payment preparation works.
- [ ] **4.** Settings -> Payment Preference -> toggle public/private contact payments: endpoint publication state follows the selected preferences.
- [ ] **5.** Profile -> Sign Out / Delete Profile / Disconnect Profile: remote endpoint cleanup runs first, then local Pubky and SDK state clear on success.

#### Automated Checks

- `./gradlew compileDevDebugKotlin` passed.
- `./gradlew testDevDebugUnitTest --tests to.bitkit.repositories.PrivatePaykitRepoTest` passed.
- `./gradlew detekt` passed.
- `git diff --check` passed.


app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -0,0 +1,713 @@
...
+                code = "revision_conflict",
+                context = "SDK state revision changed",
+            )
+        }
+
+        val nextRevision = UUID.randomUUID().toString()
+        val snapshot = SdkStateBlobSnapshot(blob = blob, revision = nextRevision)
+        keychain.upsert(Keychain.Key.PAYKIT_SDK_STATE.name, encodeSdkStateBlobSnapshot(snapshot))
+        nextRevision
+    }
```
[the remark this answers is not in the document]
ben-kaufman: Agree this is worth cleaning up. I will move the SDK blob write away from blocking DataStore work inside the monitor, or otherwise keep the synchronized section limited to revision checks.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -0,0 +1,713 @@
...
+
+    override fun publicStorageAvailable(): Boolean = true
+
+    override fun clearSessionAccess() {
+        runBlocking {
+            clearLiveSessionAccess()
+            keychain.delete(Keychain.Key.PAYKIT_SESSION.name)
+            keychain.delete(Keychain.Key.PUBKY_SECRET_KEY.name)
+        }
+    }
```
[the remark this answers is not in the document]
ben-kaufman: Agree, will clean this up with an explicit IO path or synchronous helper so the SDK callback does not block an arbitrary thread.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -1021,6 +972,90 @@ class PubkyRepo @Inject constructor(
...
+        }
+        resolveContactProfile(prefixedKey).getOrNull()?.let {
+            return it
+        }
+        return PubkyProfile.forDisplay(
+            publicKey = prefixedKey,
+            name = label,
+            imageUrl = null,
+        )
+    }
```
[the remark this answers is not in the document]
ben-kaufman: Good catch, will fix this and mirror it on iOS. If SDK profile display data is empty, we should fall back to the saved contact label instead of showing a blank name.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -262,54 +263,70 @@ class PubkyRepo @Inject constructor(
...
-        return runCatching {
-            withContext(ioDispatcher) {
-                val sessionSecret = pubkyService.completeAuth()
-                ensureAuthAttemptActive(attemptId)
-                val pk = pubkyService.importSession(sessionSecret).ensurePubkyPrefix()
-                ensureAuthAttemptActive(attemptId)
+        return try {
+            runSuspendCatching {
+                withContext(ioDispatcher) {
+                    pubkyService.completeAuth()
```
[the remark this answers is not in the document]
ben-kaufman: Good catch, will fix. If auth is canceled or superseded after completeAuth activates the SDK session, we should clear that session before restoring the UI state.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -0,0 +1,713 @@
...
+        operationMutex.withLock {
+            activeAuthRequest = null
+        }
+    }
+
+    suspend fun approveAuth(authUrl: String, secretKeyHex: String) {
+        isSetup.await()
+        PubkySessionBootstrap().approveAuth(
+            authUrl = authUrl,
+            expectedCapabilities = requiredCapabilities(),
```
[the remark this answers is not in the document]
ben-kaufman: Agree. This should use the capabilities from the incoming auth URL, and I will make the same fix on iOS for parity.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -482,1239 +381,440 @@ class PrivatePaykitRepo @Inject constructor(
...
-                val invoice = StoredInvoice(
-                    bolt11 = bolt11,
-                    paymentHash = decoded.paymentHash.toHex(),
-                    expiresAt = expiresAt,
-                )
-                ensureState().contacts.getOrPut(publicKey) { ContactState() }.localInvoice = invoice
-                persistState()
-                invoice
+    private fun schedulePendingPrivateMessageDrainRetries(reason: String, publicKeys: List<String>) {
+        pendingMessageDrainRetryJob?.cancel()
```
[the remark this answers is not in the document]
ben-kaufman: Good catch. I will merge pending retry keys instead of replacing the current batch, and make the same fix on iOS.

ben-kaufman:
For the legacy backup migration note: this is intentional for this PR. The old private Paykit link backup format never shipped, so there is no production data to migrate. Treating it as if it never existed keeps the restore path simpler.

ovitrif (commented):
Left one inline comment.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -502,62 +478,48 @@ class PubkyRepo @Inject constructor(
...
         avatarBytes: ByteArray?,
-    ): Result<Unit> = runCatching {
+    ): Result<Unit> = runSuspendCatching {
         withContext(ioDispatcher) {
             val (publicKeyZ32, secretKeyHex) = deriveKeys().getOrThrow()
 
             val homegate = fetchHomegateSignupCode()
 
-            val session = runCatching {
+            runSuspendCatching {
```
ovitrif, line 541: I think this path still needs to bump `pubkyRepo.backupStateVersion` after the sign-up/sign-in state is persisted. The SDK emits its own backup version here, but `BackupRepo` maps that to `WALLET`; the Pubky session snapshot is only included in `METADATA` when `pubkyRepo.backupStateVersion` changes. Without that, a newly created identity can back up SDK wallet state without backing up the Pubky session needed to restore it.

piotr-iohk:
That is not necessarily due to this change, because I saw it on other PR also - however e2e tests here failed partially because of this. The failure is intermittent and most of the time tests pass after re-runs.

To reproduce:
 - create a profile.
 - delete profile
 - recreate profile

Result after hitting "Continue" on the following screen:
Screenshot 2026-06-25 at 14 03 04

Attaching logs from e2e run where this happened:
[bitkit_2026-06-24_17-37-36.log](https://github.com/user-attachments/files/29336148/bitkit_2026-06-24_17-37-36.log)
[logcat.txt](https://github.com/user-attachments/files/29336149/logcat.txt)

ben-kaufman:
Fixed now in `041548681`.

Root cause was Android public Paykit publishing only refreshed the reusable on-chain address if the cached address was already reserved/unavailable. In the delete profile -> recreate profile flow, Lightning receive could be unavailable and the cached reusable on-chain address could still be blank, so endpoint sync concluded there were no supported endpoints and showed the toast.

I changed public Paykit endpoint sync to ensure a reusable on-chain address exists before deciding there is no publishable endpoint, and added regression coverage for the blank-address case. Also merged latest `master` and resolved the version-catalog conflict by keeping `bitkit-core 0.1.75` from `master` plus Paykit `0.1.0-rc21` from this PR.

Checked:
- `./gradlew testDevDebugUnitTest --tests to.bitkit.repositories.PublicPaykitRepoTest --tests to.bitkit.repositories.WalletRepoTest`
- `./gradlew detekt`
- `git diff --check`

GitHub now reports the PR as mergeable.

jvsena42:
#### ⚠️ Ring sign-in crashes: `there is no reactor running, must be called from the context of a Tokio 1.x runtime`

Reproduced when tapping "Sign in with Pubky Ring":

[Screen_recording_20260701_095809.webm](https://github.com/user-attachments/assets/5cabac62-9454-4a9e-9e99-8011f4c94895)

```
ERROR [PubkyChoiceViewModel.kt:101] Starting Ring auth failed
[AppError='there is no reactor running, must be called from the context of a Tokio 1.x runtime']
```

**Call chain**

```
PubkyChoiceViewModel.startRingAuth()
  → PubkyRepo.startAuthentication()          (PubkyRepo.kt:268)
    → PubkyService.startAuth()               (PubkyService.kt:88)
      → PaykitSdkService.startAuth()         (PaykitSdkService.kt:201)
        → PubkySessionBootstrap().startSignInAuth(...)   ← panics here
```

**Root cause (SDK binding, not app code)**

Decompiled `paykit-android:0.1.0-rc21` to confirm:

- `startSignInAuth` / `startSignUpAuth` / `resumeAuth` are exported as **synchronous** FFI calls (`uniffiRustCallWithError`). UniFFI does **not** enter a Tokio runtime around blocking calls.
- The bootstrap functions we use elsewhere — `signIn`, `signUp`, `importSession`, `complete`, `approveAuth` — are `suspend`, driven through UniFFI's async scaffolding on the SDK's Tokio runtime, so a reactor is present.

The Rust impl of `startSignInAuth` needs a Tokio reactor (builds the relay/network client for the Ring flow), but because it's a blocking export it runs on our `core-queue` thread with no runtime entered → panic. Pure-crypto sync functions in the same SDK (`derivePubkySecretKey`, `pubkyPublicKeyFromSecret`, `parsePubkyAuthUrl`) work fine because they touch no reactor.

The Ring `startSignInAuth` API did not exist in `rc8` — it's new in `rc21`.

**No clean app-side fix**

Kotlin can't enter a Tokio reactor for a blocking UniFFI call, and there is no `suspend` alternative for *starting* the flow (only sync `startSignInAuth`/`startSignUpAuth`/`resumeAuth` exist), so `withContext(ioDispatcher)` / `ServiceQueue.CORE` don't help.

**Fix belongs in `paykit-rs`:** export the start-auth bootstrap functions as `async`, or have the Rust side enter/hold a runtime (`Handle::enter()`) inside them. Also worth checking whether a newer `paykit-android` rc already makes these async before pinning.

jvsena42 (changes requested):
See https://github.com/synonymdev/bitkit-android/pull/1040#issuecomment-4855275219

piotr-iohk:
## Manual regression — Paykit / contact payments

**Environment:** regtest, staging  
**Pair tested:** Android (`pubkyraoz…`) ↔ iOS (`pubkytrb4ja…`)  
**Logs attached:**
ios: [bitkit_logs_2026-07-01_13-20-03.zip](https://github.com/user-attachments/files/29553248/bitkit_logs_2026-07-01_13-20-03.zip)
android: [logs.zip](https://github.com/user-attachments/files/29553257/logs.zip)


---

### Test setup

| Device | Platform | Profile (pubky) | LN node ID |
|--------|----------|-----------------|------------|
| A | Android | `pubkyraozwuopbt5pa3e8ki4kqeec8rmw7giruqicw53zehk3uef71agy` | `02f2dc5c…` |
| B | iOS | `pubkytrb4ja4aorm19xsiouw5hmq6ecfp1xprbdkh8x9jqe9edmrwtz1o` | `021714b0…` |

---

### Session 1 — fresh profiles (smoke)

| Test | Android | iOS |
|------|:-------:|:---:|
| Create Pubky profile | ✅ | ✅ |
| Paykit session / identity | ✅ | ✅ |
| Add contact (scan pubky) | ✅ | ✅ |
| Open LN channel (Blocktank) | ✅ | ✅ |
| On-chain send | ✅ | ✅ (`9a042478…`) |
| LN send to/from contact | ✅ | ✅ |
| Activity sync | ✅ | ✅ |
| **Incoming activity shows “Received from [contact]”** | ✅ | ✅ |
| **`RestoreReplayError` in logs** | Not seen | Not seen |

Session 1 looked good for basic contact + payment flows cross-platform.

**Private Paykit in session 1:** Incoming activity showing **“Received from [contact]”** indicates the receive path worked — that label is only set when the payment matches a **private Paykit invoice/address** (not a generic public profile invoice). There are no private Paykit link errors in session 1 logs on either platform. Send-side logs showing `Handling decoded scan data: OnChain(…?lightning=lnbcrt1…)` do **not** by themselves prove public vs private; that is how the send flow represents the payment request.

---

### Session 2 — profile delete, re-create, re-add contacts, second delete blocked

| Step | Android | iOS |
|------|:-------:|:---:|
| Delete profile (1st) | ✅ ~12:23 | ✅ ~12:22 (`Deleted all contacts`, keychain cleared) |
| Re-create profile (same pubky key) | ✅ 409 → sign-in retry | ✅ 409 → sign-in retry |
| Re-add contact | ✅ ~12:26 | ✅ |
| Contact LN send A → B | ✅ | ✅ |
| Contact LN send B → A | ✅ | ✅ |
| **Incoming activity shows “Received from [contact]”** | ❌ | ❌ |
| **Private Paykit link (no `RestoreReplayError`)** | ❌ | ❌ |
| **Delete profile again (2nd attempt)** | ❌ ~13:17–13:18 | ❌ ~13:17 UTC |
| **2nd delete error** | “Private Paykit is not available.” | “Private Paykit is not available.” |

Delete profile:

https://github.com/user-attachments/assets/948948b4-f4ff-4c98-a276-50fe15954be5



---

### Regression — private Paykit broken after profile reset

**Session 1:** Private Paykit appears to work (receive-side “Received from contact” + no link errors).  
**Session 2:** After deleting/re-creating profiles (same pubky keys) and re-adding contacts, contact LN sends still succeed but **private Paykit does not recover**. Incoming activity **no longer shows “Received from [contact]”** — consistent with payments hitting **public** endpoints instead of private ones. Public fallback is **by design** (`includePublicEndpoints = true`); no in-app warning is expected for payments.

Later in the same session, a **second profile delete also failed** on both platforms — private Paykit cleanup runs before delete and throws `PrivateUnavailable`, blocking sign-out entirely.

#### Private Paykit errors (identical on both platforms)

Every private Paykit attempt (`prepare`, `channel usable` / `refresh`, `foreground`, contact payment) logs:

```
Failed to prepare private Paykit link for '<contact>'
  → RestoreReplayError: pubky-noise handshake restore failed

Failed to queue private Paykit endpoints …
  → Encrypted Link recovery is required for counterparty <pubky-id>

Deferred private Paykit endpoint publish / Private Paykit is not available
```

First failures appear immediately after profile re-create (~12:23 iOS, ~12:26 Android on contact re-add).

#### Contact payments fall back to public

Payments use a **public BIP21 unified invoice** from the contact’s published profile — not an encrypted private payment list:

- Shared public address in logs: `bcrt1q2h4c7ghs2lj3glrm77mxdae3w2r5h6f3ph258l?lightning=lnbcrt1…`
- **Android** (`AppViewModel`): `Handling decoded scan data: OnChain(… params={lightning=lnbcrt1…})` → `PaymentSuccessful`
- **iOS** (`LightningService` / `SendConfirmationView`): `Paying bolt11: lnbcrt1…` → `Lightning payment successful`

#### Second profile delete blocked

Profile delete runs **private Paykit endpoint cleanup first**. With private Paykit already broken, cleanup throws `PrivateUnavailable` and delete aborts before homeserver sign-out.

**Android** (`EditProfileViewModel` → `PrivatePaykitRepo.removePublishedEndpointsForCleanup`):

```
Failed to remove private Paykit endpoints during 'EditProfileViewModel'
  [PrivateUnavailable='Private Paykit is not available']
```

**iOS** (`PubkyProfileManager.deleteProfile` → `removePrivatePaykitEndpoints`):

```
Failed to remove private Paykit endpoints before clearing session: privateUnavailable
ERROR Failed to delete profile: privateUnavailable - EditProfileView
```

#### Profile reset sequence (both sides)

1. Profile delete → contacts removed, `PAYKIT_SESSION` / `PAYKIT_SDK_STATE` cleared  
2. Re-create → homeserver returns **409 User already exists** → app signs in with existing key (same pubky identity)  
3. Public Paykit endpoints sync; **no successful private encrypted-link handshake** in logs  
4. After re-adding contact, `RestoreReplayError` persists through contact payments  
5. Second delete attempt fails — user stuck unless disconnect/retry workaround is used  

Likely cause: local Paykit SDK state is wiped on delete/re-create, but encrypted-link handshake state is inconsistent across peers. SDK reports recovery is required; the app logs warnings, skips private publish, and resolves contact payments via public endpoints (intentional fallback).

**Useful grep patterns:** `RestoreReplayError`, `Encrypted Link recovery`, `PrivateUnavailable`, `Failed to delete profile`, `Handling decoded scan data: OnChain`

---

### Verdict

| Scope | Result |
|-------|--------|
| Session 1 — fresh profiles: contacts, on-chain + LN, private receive (“Received from contact”) | ✅ Pass (smoke) |
| Session 2 — profile reset: contact payments work (public fallback) | ✅ By design |
| Session 2 — private Paykit restored; “Received from contact” on receive | ❌ **Regression** |
| Session 2 — second profile delete blocked (`PrivateUnavailable`) | ❌ **Regression** |

**Not approving** on “private contact payments survive profile delete/re-add.” Session 1 private Paykit looks fine; session 2 regresses on private Paykit recovery and blocks a second profile delete.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6`. The SDK state store now uses a dedicated synchronous Keychain access path for SDK sync callbacks instead of calling the suspend Keychain APIs through the service coroutine context inside the store callback. The state blob transaction is still serialized because the SDK storage callback is synchronous and does the revision check/write as one boundary.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6`. `clearSessionAccess()` now uses the same synchronous Keychain access path instead of calling suspend Keychain deletes from the SDK session-provider callback.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6` and mirrored on iOS. If the SDK returns a Paykit/contact profile with a blank name, Bitkit now falls back to the saved contact label instead of showing an empty display name. Added repo/model coverage for it.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6` and mirrored on iOS. If SDK auth completion already activated a session but the app flow was canceled or superseded, Bitkit clears that completed session before restoring UI state. Added `PubkyRepoTest` coverage on Android and `PubkyProfileManagerTests` coverage on iOS.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Rechecked this one and no extra code change is needed. `createIdentity` already calls `notifyBackupStateChanged()` after sign-up/sign-in and profile state are persisted, so the metadata Pubky session snapshot is queued separately from the SDK wallet backup-state version. Restore paths notify too.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6` and mirrored on iOS. Auth approval now parses the incoming auth URL through the SDK and passes that requested capabilities string into approval. Android also starts the sheet in `Loading` and reparses in `confirmAuthorize` if the user confirms before load finishes.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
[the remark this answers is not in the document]
ben-kaufman: Fixed in `82bb55cf6` and mirrored on iOS. Pending private drain retry keys are merged now, and a generation guard prevents an older retry task from clearing a newer batch.

ben-kaufman:
Fixed in `82bb55cf6` on Android and `51b7c2ce` on iOS.

Main thing is we now use Paykit `v0.1.0-rc23`, which includes the SDK fix for the stale recovery-required encrypted-link state after deleting/recreating a profile. It also fixes the Ring `startSignInAuth` Tokio runtime crash, so Android is pinned to rc23 now too.

I also fixed the related app-side edges:
- sign out/delete no longer get blocked if private cleanup is temporarily unavailable
- pending private drain retries now keep all queued peers instead of replacing older ones
- auth approval uses the capabilities from the actual auth URL
- if Ring auth completes but the app flow is canceled/superseded, we clear that session
- blank SDK profile names fall back to the saved contact label

Public fallback while private recovery/link work is unavailable is still intentional so contact payments can still complete. Ring is still public-only for now; this fixes the crash path, not full Ring private payments support.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -482,1239 +387,474 @@ class PrivatePaykitRepo @Inject constructor(
...
-                    generation = generation,
-                    forceRefreshLightning = forceRefreshLightning,
-                ).onFailure {
-                    if (scheduleRetries) schedulePendingPublicationRetry(normalizedKey)
+
+                val report = paykitSdkService.syncPrivatePaymentListsWithReservations(
+                    updates = preparation.updates,
+                    clearUnlistedLinkedPeers = false,
+                )
+                val firstError = preparation.firstError ?: applyPrivatePaymentListDeliveryReport(report, reason)
```
Jasonvdb: `applyPrivatePaymentListDeliveryReport(report, reason)` is skipped for the whole batch whenever any single contact fails to prepare endpoints.

`val firstError = preparation.firstError ?: applyPrivatePaymentListDeliveryReport(report, reason)` short-circuits: when `preparation.firstError != null`, the report is never applied. The early return at line 489 only fires when `preparation.updates.isEmpty()`, so a partial-failure batch (one contact with empty endpoints -> `firstError = PrivateUnavailable`, another contact succeeds -> `updates` non-empty) still runs the sync at line 496 and delivers the successful contact's endpoints.

Trigger: `refreshKnownSavedContactEndpoints` publishes for all known contacts at once, so one un-publishable contact (e.g. Lightning can't receive and on-chain disabled) suppresses report application for every contact in the batch. The successful contacts then never get `hasPublishedPrivatePaymentList=true`, `report.cleared` cache cleanup, `updateDeletedContactCleanupPending(pk, false)`, the failedToQueue/failedToDeliver warnings, or `persistState(markWalletBackup=true)`, so publish state isn't persisted and the wallet backup version isn't bumped.

Fix: evaluate the report unconditionally, e.g. `val deliveryError = applyPrivatePaymentListDeliveryReport(report, reason)` then `val firstError = preparation.firstError ?: deliveryError`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -2086,13 +1051,28 @@ class PrivatePaykitRepo @Inject constructor(
...
+        return PrivatePaykitState(cacheStore.data.first()).also { state = it }
+    }
 
     private suspend fun persistState(
         markWalletBackup: Boolean = false,
         preserveCleanupMarkers: Boolean = true,
     ) {
-        stateStore.persistState(markWalletBackup, ::notifyBackupStateChanged, preserveCleanupMarkers)
+        val currentState = state ?: PrivatePaykitState()
+        val stored = cacheStore.data.first()
```
Jasonvdb: `persistState` reads the cleanup markers from a snapshot taken outside the `cacheStore.update {}` transform, dropping the atomicity the previous `PrivatePaykitStateStore.persistState` had and allowing concurrently-set markers to be lost.

Line 1064 reads `val stored = cacheStore.data.first()`, then the `cacheStore.update { ... }` at 1065 ignores its transform input and writes `cleanupPending`/`deletedContactCleanupPendingPublicKeys` computed from that pre-read `stored`. All repo methods run on `serializedDispatcher = ioDispatcher.limitedParallelism(1)`, which serializes CPU work but still lets coroutines interleave at suspension points.

Trigger: coroutine A calls `updateContactSharingCleanupPending(true)` (which correctly mutates via `it.copy(...)`) while coroutine B is inside `persistState` between its `data.first()` (1064) and its `update {}` (1065). B writes from the stale `stored` and overwrites A's marker back to false, silently dropping it (same for `deletedContactCleanupPendingPublicKeys`). Those markers drive `removePublishedEndpoints` for removed/disabled contacts, so losing them can leave private payment endpoints published for contacts the user removed.

Fix: read the marker values from the transform's own input instead of the pre-read `stored`, e.g. `cacheStore.update { current -> currentState.cacheState(cleanupPending = if (preserveCleanupMarkers) current.cleanupPending else false, ...) }`, and drop the separate `data.first()` read.

app/src/main/java/to/bitkit/ui/screens/profile/ProfileViewModel.kt
```diff
@@ -77,20 +77,20 @@ class ProfileViewModel @Inject constructor(
...
-                    ToastEventBus.send(
-                        type = Toast.ToastType.ERROR,
-                        title = context.getString(R.string.profile__sign_out_title),
-                        description = it.message,
-                    )
-                }
+            privatePaykitRepo.removePublishedEndpointsForCleanup(TAG)
+            val result = pubkyRepo.signOut()
+            if (result.isSuccess) {
+                privatePaykitRepo.closeAndClear()
```
Jasonvdb: `privatePaykitRepo.closeAndClear()` runs only in the `result.isSuccess` branch, but `pubkyRepo.signOut()` clears the pubky session state unconditionally (it force-signs-out and clears the `_publicKey`/`_profile` StateFlows even on failure, returning `Result.failure` only when both `signOut()` and `forceSignOut()` throw).

Trigger: the rare double-failure path where both server and local sign-out throw. The pubky session is already cleared, but `closeAndClear()` is skipped and no `SignedOut` effect is emitted, so the private-Paykit local state (cached contacts, address reservations, and the on-disk `PAYKIT_SDK_STATE` blob via `paykitSdkService.clearState()`) is left intact and inconsistent with the now-empty session. This is a regression from the pre-PR code, which called `closeAndClear()` unconditionally before `signOut()`. `EditProfileViewModel.disconnectProfile` (EditProfileViewModel.kt:233) has the identical pattern with the same `signOut()` call.

Fix: clear the local Paykit state whenever `pubkyRepo.signOut()` has cleared the session (move `closeAndClear()` out of the `isSuccess` guard) while still surfacing the error toast on failure. Apply the same to `disconnectProfile`.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -686,37 +671,20 @@ class PubkyRepo @Inject constructor(
...
                             async {
-                                runCatching {
-                                    val uri = "$PUBKY_SCHEME$strippedOwnerKey${Env.contactsBasePath}$contactKey"
-                                    val json = pubkyService.fetchFileString(uri)
-                                    PubkyProfileData.decode(json).toPubkyProfile(contactKey)
+                                runSuspendCatching {
+                                    contactProfile(record.publicKey, record.label, record.profile, overrides)
                                 }.onFailure {
-                                    Logger.warn("Failed to load contact '$contactKey'", it, context = TAG)
+                                    Logger.warn("Failed to load contact '${record.publicKey}'", it, context = TAG)
```
Jasonvdb: Full unredacted pubky public keys (the user's and their contacts') are written to the app log here and at several sibling sites, while the rest of the codebase (including `PrivatePaykitRepo` touched in this same PR) routes keys through `PubkyPublicKeyFormat.redacted()`.

Line 685 logs `Failed to load contact '${record.publicKey}'`. Related sites: 615 (`deleteAllContacts`), 718, 741, 772, 783, 799 (fallback/add/update/remove/import), 1048 (`resolveContactProfile` retry), plus the info-level `'$pk'` auth log at 318. The Paykit SDK deliberately renders keys/labels as `<redacted>` in its own Debug impls, and app logs are routinely adb-exported per the debugging workflow, so this discloses the user's identity and their full contact social graph.

Fix: wrap these values with `PubkyPublicKeyFormat.redacted(...)` as `PrivatePaykitRepo` already does, e.g. `Logger.warn("Failed to load contact '${PubkyPublicKeyFormat.redacted(record.publicKey)}'", it, context = TAG)`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `14dd28d1a`. The delivery report is applied before choosing the first error now, so partial publishes still update cache/cleanup state and bump backup state. Added `PrivatePaykitRepoTest` coverage for a mixed batch.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `14dd28d1a`. `persistState` now preserves cleanup markers from the `cacheStore.update {}` input instead of a pre-read snapshot, so concurrent marker updates cannot be overwritten. Added coverage for preserving cleanup markers while saving publication state.

app/src/main/java/to/bitkit/ui/screens/profile/ProfileViewModel.kt
ben-kaufman: Fixed in `14dd28d1a` in both sign-out paths. `closeAndClear()` now runs after `pubkyRepo.signOut()` regardless of success, while still showing the error toast if sign-out reports a failure. Added tests for `ProfileViewModel` and `EditProfileViewModel`.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in `14dd28d1a`. `PubkyRepo` now redacts pubky keys in auth/contact/profile logs using `PubkyPublicKeyFormat.redacted(...)`. I also covered the sibling key-bearing log sites, not only this one.

piotr-iohk:
@ben-kaufman is pubky-ring option disabled?
Gating_no_profile_pubky_profile_1_-_Contactsprofile_entry_points_lead_to_choice_screen-2026-07-02T10-13-40-607Z

ben-kaufman:
@piotr-iohk Added it back for now, but we will likely remove it, still waiting for final decision on that...

piotr-iohk:
> @piotr-iohk Added it back for now, but we will likely remove it, still waiting for final decision on that...

OK, atm clicking at Import with Pubky ring results in error toast. Not sure then if we want to resolve that or just leave for now? that is on both iOS and Android

https://github.com/user-attachments/assets/61034e6f-e747-4589-9ff9-d9d67c63e581

piotr-iohk:
## Manual regression retest (Jul 3, post rc23)

**Environment:** regtest, staging  
**PRs:** [bitkit-android #1040](https://github.com/synonymdev/bitkit-android/pull/1040) · [bitkit-ios #606](https://github.com/synonymdev/bitkit-ios/pull/606)  
**Build:** `codex/paykit-sdk-native-integration`, Paykit **v0.1.0-rc23**

**Logs:**
- Android: [logs.zip](https://github.com/user-attachments/files/29633131/logs.zip)
- iOS: [bitkit_logs_2026-07-03_10-54-56.zip](https://github.com/user-attachments/files/29633088/bitkit_logs_2026-07-03_10-54-56.zip)


Same flow as Jul 1: create profiles → add contacts → LN + on-chain (verify private) → delete → re-create (same pubky) → re-add → LN + on-chain → delete again.

---

### Results

| Step | Android | iOS |
|------|:-------:|:---:|
| Session 1 — profiles, contacts, LN + on-chain | ✅ | ✅ |
| Session 1 — private receive (“Received from [contact]”) | ✅ | ✅ |
| Session 1 — `RestoreReplayError` in logs | Not seen | Not seen |
| Delete → re-create → re-add contact | ✅ | ✅ |
| Session 2 — LN + on-chain (payments complete) | ✅ | ✅ |
| Session 2 — private Paykit / “Received from [contact]” | ❌ | ❌ |
| Session 2 — `RestoreReplayError` after re-add | ❌ | ❌ |
| Second profile delete (while private Paykit broken) | ✅ | ✅ |

---

### Session 2 — private Paykit still broken after profile reset

After delete/re-create/re-add, private link fails again on **both** platforms:

```
RestoreReplayError: failed to restore Encrypted Link handshake
Encrypted Link recovery is required for counterparty …
Private Paykit is not available (deferred publish)
```

Contact payments still **complete via public fallback** (by design). On Android, post-reset sends resolve to public BIP21 `bcrt1qd8yaa9mwfcr5wwqyd999wmuj2vpyfs4s5emuy4?lightning=…` after `RestoreReplayError` on the contact payment path — same pattern as Jul 1. UI: no “Received from [contact]” on incoming activity.

First failures after re-add: ~10:52 Android, ~10:52 UTC iOS.

---

### Fixed since Jul 1 — profile delete no longer blocked

Second delete **succeeds** even when private cleanup fails. Logs show `PrivateUnavailable` warnings during cleanup, but **no** `Failed to delete profile: privateUnavailable` (iOS) and profile/session clears (`Deleted all contacts`, `PAYKIT_SESSION` removed). Jul 1 blocker is resolved.

---

### Verdict

| Scope | Result |
|-------|--------|
| Session 1 smoke (private contact payments) | ✅ Pass |
| Public fallback when private unavailable | ✅ By design |
| Private Paykit recovery after profile delete/re-add | ❌ **Still failing** (rc23 did not fix this in manual test) |
| Profile delete when private cleanup fails | ✅ **Fixed** |

**Not approving** on “private contact payments survive profile delete/re-add.” Happy to re-test after another SDK/app fix; delete trap fix looks good.

**Useful grep patterns:** `RestoreReplayError`, `Encrypted Link recovery`, `PrivateUnavailable`, `Handling decoded scan data: OnChain`, `Deleted all contacts`

ben-kaufman:
Fixed and tested now. I reran the Android rc26 E2E with two fresh dev installs: Bitkit profiles on both sides, Pay Contacts enabled, contacts added/resolved both ways, Alice paid Bob from Send -> Contact, and Bob's received activity was assigned to Alice with the contact chip + Detach action. I also checked the app logs/DB for the run: no no-endpoint/public-fallback/private-unavailable/send-failure markers, and both latest activity rows have the expected contact keys.

piotr-iohk:
## Manual regression retest (Jul 7)

**Environment:** regtest, staging  
**PRs:** [bitkit-android #1040](https://github.com/synonymdev/bitkit-android/pull/1040) · [bitkit-ios #606](https://github.com/synonymdev/bitkit-ios/pull/606)  
**Build:** `codex/paykit-sdk-native-integration`, Paykit **v0.1.0-rc23**

**Logs:**
- Android: 
[logs.zip](https://github.com/user-attachments/files/29743910/logs.zip)

- iOS: 
[bitkit_logs_2026-07-07_11-47-27.zip](https://github.com/user-attachments/files/29743912/bitkit_logs_2026-07-07_11-47-27.zip)


Cross-platform pair: Android ↔ iOS sim. Same flow as prior retests (Jul 1 / Jul 3) plus PR QA checklist from #1040.

---

### PR QA checklist

| # | Test | Android | iOS |
|---|------|:-------:|:---:|
| 1 | Create/edit profile → add contact → contact survives restart | ✅ | ✅ |
| 2 | Send → Contact → pay (private first, public fallback ok) | ✅ | ✅ |
| 3 | Backup/restore wallet with Pubky → pay contact | ✅ | ✅ |
| 4 | Settings → Payment Preference → toggle public/private | ✅ | ✅ |
| 5 | Sign out / delete / disconnect — cleanup then local state cleared | ✅ | ✅ |

---

### Session flow (regression focus)

| Step | Android | iOS |
|------|:-------:|:---:|
| Session 1 — fresh profiles, contacts, LN + on-chain | ✅ | ✅ |
| Session 1 — private contact payments | ✅ | ✅ |
| Delete → re-create (same pubky, 409 → sign-in) → re-add contact | ✅ | ✅ |
| Session 2 — LN + on-chain after reset | ✅ | ✅ |
| Session 2 — private contact payments (incl. “Received from [contact]”) | ✅ | ✅ |
| Second profile delete in same session | ✅ | ✅ |

**Jul 3 blockers — status in this run:**
- `RestoreReplayError` / encrypted-link recovery after profile reset → **not seen** (fixed)
- Profile delete blocked by `PrivateUnavailable` → **not seen** (still fixed)

Log support: multiple `PaymentSuccessful` / `Lightning payment successful` on both sides; iOS `setContact` after incoming payments in session 1 and session 2; `Deleted all contacts` on both platforms without `Failed to delete profile`.

---

### Known issue — deferred (Android only)

**Pubky Ring profile import on Android** fails after Ring returns auth success:

```
Received Pubky Ring auth success callback
Auth approval failed: code=identity_error, context=complete Pubky auth flow
```
Screenshot 2026-07-07 at 13 56 09


UI: “Authorization Failed” toast on Join the Pubky Web screen (Import with Pubky Ring).

**iOS:** Ring import works (`Pubky auth completed for pubkyc97…`).

Agreed with @ben-kaufman on Slack to **merge without blocking on this** — Android Ring import tracked as follow-up, not a Paykit SDK regression.

---

### Verdict

| Scope | Result |
|-------|--------|
| Paykit SDK integration — contact payments, profile lifecycle, backup/restore | ✅ Pass |
| Private Paykit recovery after profile delete/re-add (Jul 3 regression) | ✅ Pass |
| Profile delete when private cleanup flaky | ✅ Pass |
| Android Pubky Ring import | ❌ Deferred (Android-only, post-merge) |

**LGTM** on #1040 / #606 for merge, modulo deferred Android Ring import.

piotr-iohk (dismissed):
tACK

app/src/main/java/to/bitkit/data/keychain/Keychain.kt
```diff
@@ -66,6 +66,54 @@ class Keychain @Inject constructor(
...
+                emitLoadDiagnosticsOnce(key, t)
+                throw KeychainError.FailedToLoad(key, cause = t)
+            }
+        }
+
+        @Suppress("TooGenericExceptionCaught")
+        fun upsert(key: String, value: ByteArray) {
+            try {
+                val encryptedValue = keyStore.encrypt(value)
+                kotlinx.coroutines.runBlocking {
```
jvsena42: nit: import `runBlocking` instead of using full qualifier

app/src/main/java/to/bitkit/data/keychain/Keychain.kt
```diff
@@ -66,6 +66,54 @@ class Keychain @Inject constructor(
...
+            } catch (t: Throwable) {
+                throw KeychainError.FailedToSave(key, cause = t)
+            }
+            Logger.info("Upserted value for key '$key'", context = TAG)
+        }
+
+        @Suppress("TooGenericExceptionCaught")
+        fun delete(key: String) {
+            try {
+                kotlinx.coroutines.runBlocking {
```
jvsena42: same https://github.com/synonymdev/bitkit-android/pull/1040/changes#r3543842494

app/src/main/java/to/bitkit/data/keychain/Keychain.kt
```diff
@@ -128,6 +175,17 @@ class Keychain @Inject constructor(
             return "${this}_$walletIndex".let(::stringPreferencesKey)
         }
 
+    private val blockingSnapshot: Preferences
+        get() = kotlinx.coroutines.runBlocking { keychain.data.first() }
```
jvsena42: same https://github.com/synonymdev/bitkit-android/pull/1040/changes#r3543842494

app/src/main/java/to/bitkit/data/keychain/Keychain.kt
```diff
@@ -128,6 +175,17 @@ class Keychain @Inject constructor(
...
+
+    private val String.indexedBlocking: Preferences.Key<String>
+        get() {
+            val walletIndex = kotlinx.coroutines.runBlocking {
```
jvsena42: same https://github.com/synonymdev/bitkit-android/pull/1040/changes#r3543842494

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -261,55 +266,102 @@ class PubkyRepo @Inject constructor(
...
         val attemptId = _activeAuthAttemptId.value ?: return Result.failure(PubkyAuthAttemptInactive())
-        return runCatching {
-            withContext(ioDispatcher) {
-                val sessionSecret = pubkyService.completeAuth()
-                ensureAuthAttemptActive(attemptId)
-                val pk = pubkyService.importSession(sessionSecret).ensurePubkyPrefix()
-                ensureAuthAttemptActive(attemptId)
+        var didCompleteAuth = false
+        return try {
+            waitForAuthApproval(attemptId)
```
jvsena42: this method call `ensureAuthAttemptActive`, that  may may throw `PubkyAuthAttemptInactive()` and is not catch by completeAuthentication()

jvsena42 (changes requested):
Approved except for one comment that worth addressing

ben-kaufman:
@jvsena42  Fixed in 0c0dd99e1. Ring auth completion now returns a failed Result if the auth attempt is canceled/superseded while waiting for approval, instead of throwing or waiting forever. Also cleaned up the Keychain runBlocking nits from the review.

jvsena42 (approved):

Issue #998, opened 2026-06-08
epic: Trezor Hardware Wallet UI


ovitrif:
**Design:** [Figma — Trezor Screens](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151646)

**Related (iOS):** synonymdev/bitkit-ios#589

## Context

Bitkit adds a **Hardware Wallet** balance type alongside Savings (on-chain) and Spending (Lightning), backed by Trezor. The hardware balance is a watch-only view of funds held on a paired Trezor: Bitkit displays the amount and the device's on-chain activity and never holds the keys. Removing a device stops Bitkit from displaying that balance.

Paired Trezor identities appear on Home, Hardware Wallet detail, Settings, and in the Connect Hardware flow. Transfer to Spending from a hardware wallet is a send signed on the device.

Trezor v2 (milestone 2.5.0) extends that surface:

- Send on-chain from a paired Trezor through the normal Send flow, signing on the device.
- Receive on-chain to a paired Trezor, including on-device address verification.
- Pair passphrase (hidden) wallets as separate watch-only identities.
- Keep hardware activity, tags, and custom names across backup and restore.

### Device naming & connection indicator

- **Display name** = `label ?: model ?: "Trezor"` — `label` is the device name set on the Trezor (editable on-device / in Trezor Suite), `model` is the hardware model (e.g. "Safe 3" / "Safe 7").
- The Figma **"LABEL FUNDS"** field is a Bitkit-side display name for the balance, defaulting to the resolved device name; persisted per device so the user can rename within Bitkit without touching the device.
- **Connection indicator** = connection type (Bluetooth **or** USB), tinted green while the device is currently connected. Figma examples: [bluetooth](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151650) · [usb](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=44997-151655).

ovitrif:
## Android implementation notes

Current follow-up split:
- #1030 tracks runtime connection reliability and UX polish.
- #1038 tracks emulator/E2E coverage and automation.

The notes below are intentionally stale and kept as historical implementation guidance for comparing the original AI-generated spec against the later discovered work.

Order mirrors the original subtasks. All patterns confirmed in the codebase — reuse, don't reinvent.

[...]

Issue #1030, opened 2026-06-17
Hardware wallet connection reliability and UX polish


ovitrif:
Part of #998.

Automated emulator/E2E coverage has been split to #1038. This issue tracks runtime reliability and UX polish for the hardware-wallet flows.

## Scope

Polish and harden hardware-wallet connection UX after the initial flows:

- Improve reconnection reliability.
- Optimise how often Bitkit pings the Trezor device; repeated pings can currently keep the device busy and prevent the user from unlocking it.
- Improve pairing-code and PIN retry UX so wrong-code / device-confirmation states do not read as indefinite loading.
- Handle passphrase-enabled Trezor wallets explicitly and safely.

## Passphrase-enabled wallets

ℹ️ this part moved to https://github.com/synonymdev/bitkit-android/issues/1060

Passphrase-enabled Trezors need a deliberate UX and persistence model rather than accidentally inheriting state from another Trezor session.

- Decide the user-facing model for pairing a passphrase-enabled device:
  - default to the standard wallet unless the user explicitly chooses a hidden/passphrase wallet;
  - if hidden wallets are supported in this flow, provide an explicit prompt/step for host-entered passphrase and/or on-device passphrase entry instead of silently reusing prior `TrezorUiHandler` state.
- Treat the standard wallet and each hidden/passphrase wallet as separate hardware-wallet identities when their watched account xpubs/descriptors differ.
- Ensure the paired record, label, `walletId`, activity metadata, and removal behavior all stay scoped to the selected wallet identity, not just the physical Trezor device.
- Reconnect and Transfer to Spending must reconnect/sign with the same wallet identity the user paired; they must not fall back to another hidden wallet or accidentally sign from the standard wallet.
- Never persist or log raw passphrases. Clear any cached host passphrase / wallet selection on cancel, disconnect, removal, and sheet dismissal unless the current session still intentionally needs it.

## Pairing-code / PIN retry UX

Improve user-facing states while the device is waiting for code entry, device confirmation, or retry after a wrong code.

- For THP pairing-code retry: when the user enters a wrong 6-digit pairing code and the Trezor asks for device confirmation / shows a new code, Bitkit should stop looking like a generic spinner-only loading state. Show explicit copy such as "Confirm on your Trezor and enter the new code" and reset the input when a new code is requested.
[...]

PR #1044, opened 2026-06-24
feat: persist hardware activities

### LINKED ISSUE #1029: Bitkit Core support for hardware wallet activities
Part of #998.

Blocked by synonymdev/bitkit-core#112 and synonymdev/bitkit-core#113.

## Scope

Use Bitkit Core wallet-scoped activity support for hardware-wallet watcher activity, so hardware transactions can reuse normal activity capabilities where appropriate:

- Stable detail loading.
- Inputs/outputs via Bitkit Core transaction details fetched from the configured Electrum/electrs backend.
- Tag add/remove support.
- Unified activity listing so hardware-wallet activity and normal Bitkit wallet activity can be returned in the same sorted list.
- Activity Detail should show tags and inputs/outputs for hardware-wallet activity instead of the current watch-only fallback.

## App integration

After the Core work lands:

- Use the current Core activity model shape from synonymdev/bitkit-core#112.
- Use the shared wallet id scheme for hardware-wallet activity. Do not generate ad hoc app-only wallet ids in this repo.
- Keep wallet id handling behind the app's typed wallet-id model/helpers, not raw string literals passed around the codebase.
- Persist hardware watcher activities and transaction details through Core with the derived hardware wallet id.
- Read Activity Detail data through Core using the activity's wallet id.
- Add/remove tags through Core using the activity's wallet id.
- Use Core global list/search only where the UI intentionally merges the normal Bitkit wallet with paired hardware wallets.
- Use wallet-scoped Core calls for detail, txid lookup, tags, seen state, and deletion.

## Device removal

Hardware activity visibility must stay scoped to paired hardware-wallet identities:

- Removing a paired device deletes that wallet's hardware activity/details/tag metadata from Bitkit Core.
- Re-adding the same wallet should rebuild visible history from the watcher and fetch transaction details again as needed.
- Activity id and txid collisions across wallet ids must stay scoped.

## Out of scope

Backup/VSS migration for Core-owned activity data is tracked in #1046 and blocked by synonymdev/bitkit-core#113.

ovitrif:
Closes #1029
Refs #998

### Description

This PR:
1. Persists hardware-wallet watcher activities and Electrum transaction details in `bitkit-core` under the wallet ID already derived and stored by `master`.
2. Uses the current `bitkit-core` activity model and the existing `HwWalletId`/`KnownDevice` identity path across watcher events, navigation, detail loading, tags, seen state, boost relationships, and deletion.
3. Returns app and paired hardware-wallet activity through one sorted Core list while retaining wallet-scoped row identity, detail lookup, transaction lookup, tags, and UI keys.
4. Reconciles each hardware watcher snapshot with Core, keeps live watcher state visible during persistence recovery, and emits received-transaction notifications after successful persistence.
5. Preserves Transfer To Spending metadata on the hardware wallet activity so one blue `Transfer / From Savings` row represents the funding transaction.
6. Deletes the removed device's Core activity scope and rebuilds that wallet's visible history from the watcher after re-pairing.

### Preview

No visual redesign. Home, All Activity, Activity Detail, tag filtering, and Explore now render the wallet-scoped hardware activities supplied by `bitkit-core`.

#### Manual Tests
- [ ] **1.** Settings → Advanced → Dev Settings → Trezor → pair the Bridge emulator: Home shows the connected hardware wallet tile, balance, and blue hardware activities.
- [ ] **2.** Fund two hardware addresses with distinct amounts → mine a block → Home → Show All: both blue received rows appear once in the unified activity list.
- [ ] **3.** Blue hardware receive → Activity Detail → add a tag → back → reopen → Explore: the tag persists and Inputs plus Outputs contain the Electrum transaction details.
- [ ] **4.** All Activity → tag filter → select the hardware tag: the tagged blue row remains and untagged rows are absent.
- [ ] **5.** Hardware wallet → Transfer To Spending → 25% → Continue → Open Trezor Connect → approve Recipient, Amount, Locktime, and Summary → Continue Using Bitkit: Home shows exactly one new blue `Transfer / From Savings` row and its detail shows `TO SPENDING`.
- [ ] **6.** All Activity → switch Sent, Received, Other, and All: received hardware rows appear under Received, the hardware Transfer appears under Other, and All restores the combined list without a default-wallet duplicate.
- [ ] **7.** Settings → Hardware Wallets → remove the paired device → re-pair the same wallet: its scoped activity disappears on removal and rebuilds without duplicate rows after pairing.
- [ ] **8.** `regression:` app on-chain and LN activity → Activity Detail, tags, Explore, and wallet-specific lists: behavior matches `master`.

#### Automated Checks
- Unit tests updated:
	- coverage active watcher ownership, Core snapshot persistence, persistence recovery, wallet removal, transfer metadata, and activity/transaction ID collisions across wallet scopes: `HwWalletRepoTest.kt`, `ActivityRepoTest.kt`, and `TransferRepoTest.kt` 
	- coverage for scoped detail/tag calls, global unified listing, scoped navigation, hardware notification routing, and Transfer To Spending ownership: `ActivityDetailViewModelTest.kt`, `ActivityListViewModelTest.kt`, `AppViewModelSendFlowTest.kt`, and `TransferViewModelTest.kt`
	- retain default-wallet behavior with the scoped repository API: `NotifyPaymentReceivedHandlerTest.kt` and `BoostTransactionViewModelTest.kt` 
- Hardware-wallet journeys updated:
	- `activity-blue-icons.xml` and `transfer-to-spending.xml`; `activity-detail-hw-tags.xml` adds persisted tag and inputs/outputs coverage.
- Functional verification:
	- Android 17 emulator and deterministic T2T1 Bridge emulator completed pairing, two confirmed hardware receives, persisted tags, tag filtering, Core inputs/outputs, all activity-tab assertions, four-step device signing, and one wallet-scoped pending transfer row.
- Local checks:
	- `just compile`, `just test`, and `just lint` passed; all hardware-wallet journey XML files passed `xmllint`.


app/src/main/java/to/bitkit/ui/screens/trezor/TrezorViewModel.kt
```diff
@@ -714,6 +712,7 @@ class TrezorViewModel @Inject constructor(
             }
             val result = trezorRepo.startWatcher(
                 watcherId = watcherId,
+                walletId = watcherId,
                 extendedKey = key,
```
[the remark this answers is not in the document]
ovitrif: I changed the setup watcher to use the typed Trezor wallet scope instead of the raw UUID, and  now ignores watcher events unless they belong to an active wallet-scoped watcher it started. That prevents setup/dev watcher events from persisting orphan Core activities.

Resolved in a896d9f05

ovitrif:
Handoff note while I'm on PTO:

The core walletId/current activity work has landed. This PR is now waiting for the published Android package from the `bitkit-core` release path.

Current core state:

- synonymdev/bitkit-core#117 is merged and contains the walletId/current activity payload changes this PR needs.
- synonymdev/bitkit-core#118 is merged and contains the binding update for that work.
- `bitkit-core` `v0.3.8` is published from the synonymdev/bitkit-core#118 binding update head.
- The `v0.3.8` Gradle Package run is still in progress: https://github.com/synonymdev/bitkit-core/actions/runs/28467224244. It needs to finish successfully so `com.synonym:bitkit-core-android` is published and native debug symbols are uploaded.

When core unblocks this:

1. Fix any merge conflicts on this PR branch.
2. Update `gradle/libs.versions.toml` to consume `com.synonym:bitkit-core-android` `0.3.8`, or the next release if `0.3.8` publish fails and is replaced.
3. Resolve any generated API changes in this PR branch.
4. Run the normal PR checks: `just compile`, `just test`, `just lint`.
5. Update this PR description/status with the consumed `bitkit-core` release.

ovitrif:
Status update:

- This PR now consumes `bitkit-core` `0.3.9`.
- Local verification passed on `c0274e311`: `just compile`, `just test`, and `just lint`.
- There are no unresolved review threads currently open.
- Coordination note: #1062 touches much of the same Core/HW wallet surface. I’d recommend keeping this PR as the base for the Core hardware-activity refactor, then rebasing or splitting #1062 afterward so the hardware reliability fixes stay separately reviewable.

ovitrif:
Status update after HW journey validation:

- The core unblock is resolved for this branch: it consumes `bitkit-core` `0.3.9` and the Android artifact from the publish task.
- I re-ran the Bridge Trezor emulator flows on `8372a730f`: `connect-home-tile.xml`, `activity-blue-icons.xml`, `activity-detail-hw-tags.xml`, and `detail-overview.xml`.
- Validation passed for the important HW activity/wallet paths: deterministic Trezor pair/fund, blue Core-backed activity rows, All Activity tabs, persisted `hwtest` tag, tag filter, Explore txid/input/output data, hardware wallet detail, transfer amount screen after the first-run intro, remove cancel, final remove, watcher stop, and hardware activity purge.
- The app log confirms cleanup after removal: `Deleted '1' activities for hardware wallet ...`.
- I also updated the HW transfer journey specs so fresh emulator runs handle the first-run Transfer To Spending intro before asserting `HardwareTransferAmount`; otherwise the journey can false-fail before reaching the amount screen.
- No fatal/crash signatures showed up in logcat. Existing local checks remain: `just compile`, `just test`, and `just lint` passed on `c0274e311`; `xmllint` passed for the updated HW journey XMLs on `8372a730f`.

Remaining handoff: wait for checks/reviews on `8372a730f`. If those settle cleanly, this should be ready to move out of draft/request review. #1062 overlaps with this work and should likely be rebased or split after this lands, but treat that as coordination guidance rather than a blocker for this branch.

piotr-iohk:
# Trezor Android: #1044 vs #1062 — QA findings

**Date:** 2026-07-06  
**Tested:** `refactor/core-hww-activities-1029` (#1044) on physical device (`to.bitkit.dev`)  
**Comparison:** `feat/wallet-scoped-core-0.3.9` (#1062) shows correct Trezor balance on same device

## Scope mismatch

- **#1062** — 2.4.0 / #1030: minimal Trezor v1 (pair, balance, transfer-to-spending, BLE reconnect). HW balance/activities from watcher **in memory**. No tags, Explore I/O, Boost, unified Activity tab integration.
- **#1044** — 2.5.0 / #1029: full Core-backed HW activities (tags, Explore inputs/outputs, Boost, wallet-scoped persist, unified activity list). QA checklist explicitly covers tags + Explore.

#1044 may be a stretch for **2.4.0**: iOS Trezor is still behind, and for cross-platform consistency we should not ship HW tags or Explore inputs/outputs in 2.4.0. That fits **2.5.0 / #1029** instead.

## Finding 1: Trezor shows 0 balance + empty activity list (#1044)

**Symptom:** Paired Trezor with known funds/history → app shows **0 sats** and **no activities**.

**Not a Trezor/connect issue.** Watcher starts; Core emits data; UI never updates.

**Cause:** In `HwWalletRepo.observeWatcherEvents()`, `persistHardwareActivities()` runs **before** `_watcherData` is updated. On failure it `return@collect` — balance and activities never reach UI.

**Log evidence** (`bitkit_2026-07-06_08-34-36.log`):
[bitkit_logs_2026-07-06_08-50-23.zip](https://github.com/user-attachments/files/29695285/bitkit_logs_2026-07-06_08-50-23.zip)


```
Started watcher 'ble:5B:B1:C8:E9:4C:BE|nativeSegwit'
ERROR Failed to persist hardware activities
  AppError='Failed to upsert onchain_activity: UNIQUE constraint failed: onchain_activity.id'
hardwareWallets=[{"id":"ble:5B:B1:C8:E9:4C:BE","sats":0}]
```

**Likely persist root cause:** HW activity ids = **txids**. Error on `onchain_activity.id` suggests stale pre-0.3.9 schema (id globally unique) or failed migration — collides when same tx exists on Bitkit wallet and Trezor (e.g. transfer between wallets). Fresh install / clear app data should confirm.

**#1062:** Updates `_watcherData` from watcher first; does not gate UI on persist. Works on same device.

## Finding 2: Transfer shows duplicate rows — Transfer + Sent (#1044)

On transfer to spending, activity list shows **two rows** for one flow:

- **Transfer** (Bitkit wallet)
- **Sent** (Trezor / HW wallet)

Both appear in unified list; **both are taggable**. Not a quick polish issue — needs product rules for cross-wallet dedup/linking (`transfer_tx_id`, filtering, etc.). #1044 merges dupes across **address-type watchers**, not Bitkit↔Trezor transfer pairs.

## Finding 3: Scope ahead of 2.4.0 (#1044)

#1044 adds:

- Tag support on HW activities
- Explore inputs/outputs
- Boost with wallet-scoped Core lookup
- HW activities in global Activity tab (filters, tags, blue icons)

Reasonable for **2.5.0 / #1029**, but not aligned with a minimal, iOS-consistent **2.4.0** Trezor release.

## Prod / migration concern

**Not a prod issue today** — no Trezor in prod.

**Before HW launch (2.5.0):** migration must work on upgrade; UI must not depend on persist succeeding; transfer dedup needs design.

Clear app data is a **dev QA workaround**, not a ship strategy.

## Recommendation

| Release | PR | Action |
|---------|-----|--------|
| **2.4.0 Trezor v1** | **#1062** | Ship — minimal scope, works on device, #1030 reliability |
| **2.5.0** | **#1044** | Keep draft — fix persist gate, migration/collision, transfer dedup, then tags/Explore/Boost QA |

Later: cherry-pick #1062 reliability commits onto #1044 for 2.5.0; don’t merge as competing 0.3.9 paths.

## #1044 fixes needed (2.5.0, not blocking v1)

1. **Decouple UI from persist** — update watcher state even if Core persist fails; log error separately.
2. **Fix persist failure** — ensure 0.3.9 `(wallet_id, id)` migration on upgrade; handle shared txids across wallets.
3. **Transfer dedup UX** — one logical row per transfer, not Transfer + Sent both taggable.
4. Continue #1029 scope (tags, Explore, Boost) only after above is stable.

## Links

- https://github.com/synonymdev/bitkit-android/pull/1062
- https://github.com/synonymdev/bitkit-android/pull/1044

ovitrif:
> Decouple UI from persist — update watcher state even if Core persist fails; log error separately.

Addressed this Android-side gating issue: watcher state now updates before Core persistence, so the Trezor balance/activity UI stays live even if a Core upsert fails. Persistence failures are still logged with the watcher id, and received-transaction emission remains behind successful persistence.

Validation: `just compile`, `just test`, `just lint`.

The migration/collision and transfer dedup points remain broader draft follow-ups outside this scoped fix.

Resolved in f546f58

ovitrif:
Set milestone to `2.5.0` for clarity 👍🏻 , thanks for review & checks @piotr-iohk 🙏🏻

ovitrif:
> Trezor shows 0 balance + empty activity list
>
> Transfer shows duplicate rows - Transfer + Sent
>
> Scope ahead of 2.4.0

The branch is now rebased on the merged #1062 and latest `master`, and consumes `bitkit-core` `0.4.0`. The old pre-release database state from this report is no longer the branch/base under review. Wallet-scoped persistence is the source of truth, and watcher state is published only after the matching activities and transaction details are available for list/detail navigation.

I reran the full Bridge flow on the current head with two independently funded and mined hardware receives. Both appeared once with the correct wallet scope, and list tabs, tags, tag filtering, Activity Detail, and Explore all used the persisted entries.

Transfer-to-spending now reconciles the watcher-created sent activity with the transfer metadata. The Bridge run showed one pending `Transfer / From Savings` row, then the same row confirmed after mining while Spending increased by the transferred amount. Device removal purged that wallet's activities, and re-pairing rebuilt its confirmed history.

This work remains scoped to `2.5.0` / #1029. `just compile`, `just test`, and `just lint` all pass on the current head.

Resolved in 5ae816a98

app/src/main/java/to/bitkit/models/ActivityWalletType.kt
```diff
@@ -0,0 +1,15 @@
+package to.bitkit.models
+
+import java.util.Locale
+
+enum class ActivityWalletType {
+    BITKIT, TREZOR;
```
jvsena42: available in bitkit-core

ovitrif:
AFAIU this needs a new agent run to make sure it builds correctly on top of latest merges into master and doesn't try to reinvent wheels that have been added in the meantime.
I can give it a go once master and Trezor v1 PRs settle

jvsena42:
> AFAIU this needs a new agent run to make sure it builds correctly on top of latest merges into master and doesn't try to reinvent wheels that have been added in the meantime. I can give it a go once master and Trezor v1 PRs settle

Indeed, most of these changes were already implemented

ovitrif:
> Indeed, most of these changes were already implemented

I rebuilt the branch from current `master` and retained the missing `bitkit-core` persistence and unified activity integration. The Trezor lifecycle, wallet identity, watcher APIs, and hardware-wallet foundation now come directly from `master`; the branch-local type and identity implementations are removed. The full repository checks and Bridge-backed hardware journeys pass on the reconstructed head.

Resolved in b4bbc26e9

piotr-iohk:
There are some conflicts that would be good to resolve before testing.

ovitrif:
> There are some conflicts that would be good to resolve before testing.

I rebased this branch onto current `origin/master` (`fe0ae42af`), resolved the notification-handler test against the new presentation lifecycle, and reran `just compile`, `just test`, and `just lint`.

Resolved in 64cf108e5

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -375,20 +399,34 @@ class HwWalletRepo @Inject constructor(
...
+                    _watcherData.update { it + (watcherId to watcher) }
+                    val previousIds = persistedActivityIds.getOrPut(watcherId) {
+                        activities.map { it.scopedId() }.toSet()
+                    }
+
+                    val persistedActivities = activityRepo.persistHwSnapshot(
+                        walletId = walletId,
+                        activities = activities,
+                        transactionDetails = transactionDetails,
+                    ).getOrElse { return@withLock }
```
jvsena42: This method is called on every `TransactionsChanged` event, and `ActivityService.replaceHwSnapshot` unconditionally does getActivities → delete-orphans → upsertActivities → getActivities again, then fires notifyActivitiesChanged(). 

iOS deliberately short-circuits this:
```
// Skip the core upsert + activity-list reload when nothing changed for this group.
guard lastPersisted[group.walletId] != merged else { return }
```

Android has no equivalent guard, so an unchanged Electrum re-sync re-writes the whole wallet history to Core and forces a full activity-list reload. Recommend caching the last-persisted merged set per walletId and bailing when equal.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -286,22 +304,28 @@ class HwWalletRepo @Inject constructor(
...
-            failures.firstOrNull()?.let { throw it }
-            check(ids.none { it in remaining }) { "Hardware wallet '$deviceId' still present after removal" }
+                activeWatchers.toList()
+                    .filter { it.toDeviceId() in ids }
+                    .forEach {
+                        if (!stopActiveWatcherLocked(it)) {
+                            throw AppError("Failed to stop hardware wallet watcher '$it'")
+                        }
+                    }
+                walletId?.let { activityRepo.deleteForWallet(it).getOrThrow() }
```
jvsena42: This deletion only happens inside `removeDevice`. iOS additionally reconciles in 

```
updateDevices: for walletId in previousWalletIds.subtracting(hwWalletIds) { deleteActivities(walletId) }
```

Android's reconcileWatchers stops stale watchers but never purges their Core scope, so any removal that bypasses `removeDevice` (external forget, store rewrite) leaves orphaned hardware rows.

jvsena42 (changes requested):

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: I now cache each wallet's last successfully persisted watcher snapshot and reuse the Core-enriched activities when it is unchanged. Repeated Electrum syncs can still update the balance without rewriting Core or reloading the activity list.

Resolved in 0182de38a

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: I now track watcher wallet scopes across device-store snapshots. When a wallet disappears, reconciliation stops its watcher, deletes the wallet's Core activity scope, and clears the snapshot cache.

Resolved in 0182de38a

app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -281,9 +290,62 @@ class ActivityService(
...
+            Activity.Onchain(
+                onchain.v1.copy(
+                    isTransfer = onchain.v1.isTransfer || existing.isTransfer,
+                    channelId = onchain.v1.channelId ?: existing.channelId,
+                    transferTxId = onchain.v1.transferTxId ?: existing.transferTxId,
+                )
+            )
+        }
+        if (mergedActivities.isNotEmpty()) upsertActivities(mergedActivities)
+        if (transactionDetails.isNotEmpty()) upsertTransactionDetails(transactionDetails)
```
jvsena42, line 348: nit: Extracting the pure merge (given existing + incoming → (toDelete, toUpsert)) into a testable function would protect the "exactly one Transfer / From Savings row" guarantee against regressions

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
jvsena42: could remove this unused injection

jvsena42: or check if it is needed somewhere

app/src/main/java/to/bitkit/services/CoreService.kt
ovitrif: Extracted the snapshot reconciliation into the pure `mergeHwSnapshot` function and added direct tests covering preservation of one transfer row plus deletion of stale non-transfer rows.

Resolved in 482b94e38

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: Confirmed the `Clock` dependency was unused and removed it together with the obsolete test fixture setup and experimental-time opt-in.

Resolved in 482b94e38

jvsena42 (commented):
## Review — wallet-scoped hardware activity

Reviewed the diff and ran the feature end-to-end on a Pixel 9 / Android 17 AVD with the deterministic T2T1 Bridge emulator (`TREZOR_BRIDGE=true TREZOR_BRIDGE_URL=http://10.0.2.2:21325`), against the staging regtest backend.

**The design holds up.** Verified working on device:

| Check | Result |
|---|---|
| Pairing → home tile, balance, blue rows | ✅ |
| Two receives with distinct amounts (12 345 / 5 712) | ✅ each rendered once |
| Same txid in both scopes (`trezor:…` Received + `bitkit` Sent) | ✅ both rows, no collision |
| Tag on a hardware activity, persists across re-navigation | ✅ `activity_tags` row written with `wallet_id = trezor:8df6…` |
| Explore → INPUT + OUTPUTS(2) | ✅ 46 `transaction_details` rows persisted in the HW scope |
| Tag filter in All Activity | ✅ only the tagged blue row survives |
| Sent / Received / Other / All tabs | ✅ |
| Transfer To Spending (25% → 4 device approvals) | ✅ exactly one blue `Transfer / From Savings`, detail shows `TO SPENDING` |
| Restart with device disconnected | ✅ history + balance restored from Core, **no duplicates** (`GROUP BY wallet_id, tx_id HAVING count>1` → empty) |

Not exercised: device removal → re-pair (covered by unit tests), CJIT/boost paths.

**One blocker**, in `TransferRepo.markActivityAsTransfer` — the new `search = txid` lookup can never match, which silently kills transfer-marking for the **default wallet** as well as hardware. Details and evidence in the inline thread. Two 🟠 items (backup tag asymmetry, `emit` under mutex) are worth resolving before merge; the rest are follow-ups or nits.

Conventions are otherwise clean: the `runSuspendCatching` migration, the `System.currentTimeMillis()` → `nowTimestamp()` swap, quoted log parameters, `@Stable` on `ActivityDetailUiState`, and extracting `reconcileWatchers`/`toWatcherSpecs` out of the nested `collect` are all improvements. Test coverage is genuinely good — the only visible gap maps exactly to the blocker.

app/src/main/java/to/bitkit/repositories/TransferRepo.kt
```diff
@@ -260,7 +265,21 @@ class TransferRepo @Inject constructor(
...
+            sortDirection = SortDirection.DESC,
+        ).filterIsInstance<Activity.Onchain>()
+            .filter { it.v1.txId == txid }
+            .let { matches ->
+                matches.firstOrNull { it.v1.isTransfer }
+                    ?: matches.firstOrNull {
+                        it.isFromHardwareWallet() && it.v1.txType == PaymentType.SENT
+                    }
+                    ?: matches.firstOrNull()
+            }?.v1 ?: return
```
jvsena42: **🔴 This lookup can never match — `markActivityAsTransfer` is now dead code (regression on the default wallet too).**

`search` in bitkit-core only `LIKE`s four columns. From the shipped `bitkit-core-android-0.4.1.aar` (`strings jni/arm64-v8a/*.so`):

```
o.address LIKE '  /  a.contact LIKE '  /  l.invoice LIKE '  /  l.message LIKE '
```

No `tx_id`, no `a.id`. Verified against a live device DB (`files/regtest/wallet0/core/activity.db`) for a real funding txid:

```
simulate core search=txid → 0 rows
direct tx_id lookup       → 1 row
```

So this always falls through to `?: return`. Both callers break silently:

* `persistResolvedChannel` (L264) — `channel_id` is never written to the funding activity.
* the force-close batched-sweep path (L247) — sweeps stop being marked as transfers and render as plain Sent/Received.

Reproduced on device: after a hardware Transfer To Spending, the funding row is `is_transfer=1, channel_id=NULL` (the `1` comes from `createPendingToSpendingActivity`, not from here), while the pre-branch `bitkit` transfer row has `channel_id` populated.

Suggested shape — keep the direct lookup and iterate candidate scopes instead of searching:

```kotlin
private suspend fun findOnchainActivity(txid: String): OnchainActivity? =
    (listOf(WalletScope.default) + hwWalletRepo.knownWalletIds())
        .firstNotNullOfOrNull { coreService.activity.getOnchainActivityByTxId(txid, it) }
```

…or add a cross-scope `getActivityByTxId(walletId = null)` in core. Either way this needs a `TransferRepoTest` case — `createPendingToSpendingActivity uses the funding wallet scope` covers the sibling path, but nothing covers `markActivityAsTransfer`, which is the one thing here that regresses.

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
```diff
@@ -330,6 +361,7 @@ class ActivityRepo @Inject constructor(
     }
 
     suspend fun getActivities(
+        walletId: String? = WalletScope.default,
```
jvsena42, line 365: **🟠 Backup/restore is now asymmetric for hardware tags.**

`BackupRepo.kt:510` calls `activityRepo.getActivities()` with no argument, so it resolves to `WalletScope.default` and hardware activities are excluded from the backup payload — reasonable, they're watch-only and rebuildable from the watcher.

But `getAllActivitiesTags()` (`ActivityRepo.kt:826` → `CoreService.kt:479` → `com.synonym.bitkitcore.getAllActivitiesTags()`) is **unscoped**, so hardware tag rows *do* go into the payload.

On restore, `upsertTags` then inserts tags whose parent row doesn't exist in that wallet scope:

```sql
CREATE TABLE activity_tags (
  …, FOREIGN KEY (wallet_id, activity_id) REFERENCES activities(wallet_id, id) ON DELETE CASCADE);
```

Core uses `INSERT OR IGNORE INTO activity_tags`, and SQLite's `ON CONFLICT` clause does **not** apply to FOREIGN KEY constraints. Best case the tags are orphaned and silently lost; worst case the whole `ACTIVITY` restore throws.

This PR is what makes hardware tags possible in the first place, so it should decide explicitly: either scope the tag backup to the default wallet, or include hardware activities in the payload.

jvsena42, line 365: Separate consequence of the same default: `HomeViewModel.kt:127` still calls `activityRepo.getActivities(limit = 1u)` without a scope, so the "has any activity" flag only counts the default wallet.

A user whose entire history is a paired hardware wallet now sees blue rows in the list but still gets the empty state driven by `hasActivityFlow`. Probably wants `walletId = null` here, same as `ActivityListViewModel.refreshActivityState`.

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
```diff
@@ -593,15 +638,15 @@ class ActivityRepo @Inject constructor(
...
-                return@withContext Result.failure(Exception("Activity not deleted"))
-            }
+    suspend fun deleteActivity(
+        id: String,
+        walletId: String = WalletScope.default,
+    ): Result<Unit> = withContext(bgDispatcher) {
+        runSuspendCatching {
+            val deleted = coreService.activity.delete(id, walletId)
+            check(deleted) { "Activity not deleted" }
+            cacheStore.addActivityToDeletedList(id)
```
jvsena42: The `deletedActivities` cache is still keyed by **raw** id, not `scopedId()` — here and at the three read sites (`:548`, `:657`, `:671`), plus `CoreService.kt:1069`.

`deleteActivity` has no production caller today so it's latent, but cross-scope id collisions are now an *expected* state (I confirmed the same txid living in both `bitkit` and `trezor:8df6…` on device). Deleting one would suppress the other on re-sync. Worth scoping while everything else is being scoped.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -398,22 +462,25 @@ class HwWalletRepo @Inject constructor(
...
+                if (scopedId in previousActivityIds || !emittedReceivedTxIds.add(scopedId)) return@forEach
+                val sats = mergedActivities.findOnchain(onchain.v1.txId, onchain.v1.walletId)?.v1?.value
+                    ?: onchain.v1.value
+                _receivedTxs.emit(
+                    HwWalletReceivedTx(
+                        txid = onchain.v1.txId,
+                        sats = sats,
+                        walletId = onchain.v1.walletId,
+                    )
+                )
```
jvsena42: **🟠 This suspending `emit` now runs while `watcherMutex` is held.**

`emitReceivedTxs` is called from inside `watcherMutex.withLock` in the `TransactionsChanged` handler. `_receivedTxs` is `MutableSharedFlow(extraBufferCapacity = 8)` with the default `BufferOverflow.SUSPEND`, and the collector (`AppViewModel.kt:343`) shows a transaction sheet per emission.

A slow or backed-up collector suspends this `emit` **while the mutex is held**, which now also blocks `reconcileWatchers`, `removeDevice` and `resetState` — none of which were lock-coupled before this PR.

Cheap fix: build the list inside the lock and emit after releasing it, or use `tryEmit` (the buffer already exists for exactly this).

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -434,68 +501,85 @@ class HwWalletRepo @Inject constructor(
...
+        (activeWatchers - desiredIds).forEach { stopActiveWatcherLocked(it) }
+
+        removedWalletIds
+            .filterNot { it in activeWatcherWalletIds.values }
+            .forEach { walletId ->
+                activityRepo.deleteForWallet(walletId).onSuccess {
+                    trackedWalletIds -= walletId
+                    lastPersistedHwSnapshots -= walletId
+                }
+            }
```
jvsena42: **🟡 This cleanup can't see scopes orphaned before the process started.**

`trackedWalletIds` is in-memory and starts empty (L101, cleared in `resetState`), so `removedWalletIds` at L516 can only ever contain wallets this process already reconciled. If a device is forgotten by a path that bypasses `removeDevice`, or the removal fails and the process dies before the retry, that wallet's rows stay in Core indefinitely.

They also stay *visible*: `ActivityListViewModel.refreshActivityState` reads `walletId = null`, and `_persistedHardwareIds` marks every non-default scope as hardware — so they keep rendering blue with no watcher behind them.

Consider a startup reconcile: list the distinct wallet ids in Core, diff against `hwWalletStore.loadKnownDevices()`, and `deleteForWallet` anything unknown.

app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -281,9 +316,48 @@ class ActivityService(
...
+            minDate = null,
+            maxDate = null,
+            limit = null,
+            sortDirection = null,
+        ).filterIsInstance<Activity.Onchain>()
+        val merge = mergeHwSnapshot(existing = existingActivities, incoming = activities)
+        merge.toDelete.forEach {
+            deleteActivityById(walletId = walletId, activityId = it.v1.id)
+            deleteTransactionDetails(walletId = walletId, txId = it.v1.txId)
+        }
```
jvsena42, line 345: **🟡 Snapshot replace is wallet-scoped, but it's driven per *watcher*.**

`mergeHwSnapshot` marks every existing non-transfer row in the scope that isn't in `incoming` for deletion. `HwWalletRepo.toWatcherSpecs` does `distinctBy { it.addressType to it.xpub }`, so two watchers with different address types can legitimately share one `walletId` — and they'd delete each other's rows on every event, thrashing.

Latent today because `SUPPORTED_WATCHER_ADDRESS_TYPES` holds a single entry (`HwWalletRepo.kt:89`), but the surrounding code is written for multiple types and the xpubs for all types are already captured on connect.

Either key the snapshot by `watcherId` instead of `walletId`, or drop a comment here pinning the single-address-type assumption so the next person adding Taproot doesn't hit it.

app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -281,9 +316,48 @@ class ActivityService(
...
+            maxDate = null,
+            limit = null,
+            sortDirection = null,
+        )
+    }
+
     private fun mapToCoreTransactionDetails(
         txid: String,
         details: TransactionDetails,
+        walletId: String = defaultWalletId,
```
jvsena42: This parameter has no caller that passes a non-default value — both call sites (`:1239`, `:1270`) use `mapToCoreTransactionDetails(txid, details)`.

`CLAUDE.md` says *ALWAYS apply the YAGNI principle for new code*. Either wire it up to the hardware path or drop it.

app/src/main/java/to/bitkit/viewmodels/ActivityDetailViewModel.kt
```diff
@@ -197,34 +205,39 @@ class ActivityDetailViewModel @Inject constructor(
...
-                reloadActivity(id)
+                reloadActivity(id, currentActivity.walletId())
             }.onFailure {
                 Logger.error("Failed to detach contact for activity '$id'", it, context = TAG)
             }
         }
     }
 
     fun fetchTransactionDetails(txid: String) {
+        val walletId = activity?.walletId() ?: return
```
jvsena42: This early-return sits **outside** the coroutine, so when `activity` is null `_isTxDetailsLoading` never flips to `true`.

Combined with the new Explore condition (`ActivityExploreScreen.kt:333`, `isTxDetailsLoading && !onchain.v1.isTransfer`, replacing `!isHardware && …`), a skipped or failed fetch now renders as *nothing at all* — no spinner, no inputs/outputs, no error. The previous code at least showed a spinner.

I couldn't trigger it on device (Explore rendered INPUT + OUTPUTS(2) correctly for a persisted hardware receive), but the blank-on-failure path is worth an explicit empty/error state.

app/src/main/java/to/bitkit/viewmodels/ActivityDetailViewModel.kt
```diff
@@ -236,10 +249,10 @@ class ActivityDetailViewModel @Inject constructor(
     }
 
     suspend fun getBoostTxDoesExist(boostTxIds: List<String>): ImmutableMap<String, Boolean> =
-        activityRepo.getBoostTxDoesExist(boostTxIds).toImmutableMap()
+        activityRepo.getBoostTxDoesExist(boostTxIds, activity?.walletId() ?: WalletScope.default).toImmutableMap()
 
     suspend fun isCpfpChildTransaction(txId: String): Boolean {
-        return activityRepo.isCpfpChildTransaction(txId)
+        return activityRepo.isCpfpChildTransaction(txId, activity?.walletId() ?: WalletScope.default)
```
jvsena42: Nit: `activity?.walletId() ?: WalletScope.default` appears three times in this class (here, L255, and the `loadActivity` resolution). A single accessor would DRY it and keep the fallback in one place:

```kotlin
private val currentWalletId: String get() = activity?.walletId() ?: WalletScope.default
```

app/src/main/java/to/bitkit/repositories/TransferRepo.kt
ovitrif: Replaced the Core text search with direct `tx_id` lookup across the default and persisted hardware wallet scopes. Added regressions for the hardware funding activity and default-wallet batched sweep paths.

Resolved in cdc27da

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
ovitrif: Scoped activity-tag backup export to the default wallet, matching the activities included in the existing backup payload. Hardware/Trezor backup expansion remains tracked in #1046 under the #998 epic.

Resolved in cdc27da

ovitrif: Updated the home activity check to query all wallet scopes with `walletId = null`, so persisted hardware activity prevents the empty state.

Resolved in cdc27da

app/src/main/java/to/bitkit/repositories/ActivityRepo.kt
ovitrif: Scoped deleted-activity cache entries by wallet and activity ID at every write/read site. Legacy raw IDs remain default-wallet-only, and collision regressions cover the hardware scope.

Resolved in cdc27da

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: Received transactions are now built while holding `watcherMutex` and emitted only after the lock is released, preserving delivery without lock-coupling watcher lifecycle operations.

Resolved in cdc27da

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: Startup reconciliation now loads persisted non-default activity scopes, compares them with resolved known-device wallet IDs, and deletes unknown scopes. Added a regression for an orphan present before process start.

Resolved in cdc27da

app/src/main/java/to/bitkit/services/CoreService.kt
ovitrif: Documented the complete-wallet snapshot contract and pinned the current one-supported-address-type invariant at `replaceHwSnapshot`, so adding another watcher type requires merging its wallet snapshot first.

Resolved in cdc27da

app/src/main/java/to/bitkit/services/CoreService.kt
ovitrif: Removed the unused wallet parameter from `mapToCoreTransactionDetails`; the default-wallet mapping is now explicit inside the helper.

Resolved in cdc27da

app/src/main/java/to/bitkit/viewmodels/ActivityDetailViewModel.kt
ovitrif: Added an explicit terminal unavailable state for skipped, null, and failed transaction-detail loads, plus visible Explore feedback and regression coverage.

Resolved in cdc27da

app/src/main/java/to/bitkit/viewmodels/ActivityDetailViewModel.kt
ovitrif: Centralized wallet resolution in `currentWalletId`, including the requested route scope during load, and reused it for transaction details and boost/CPFP queries.

Resolved in cdc27da

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif: I extended the snapshot cache identity to ignore the volatile watcher timestamp while an activity is pending, while confirmation still invalidates the cache. The regression now covers two pending snapshots with different timestamps followed by confirmation.

Resolved in b8ef428

jvsena42 (approved):
tAck

PR #1053, opened 2026-06-29
feat: make foreground service opt-in

### LINKED ISSUE #1007: Polish Notifications Permissions Request
Some UX improvements could be addressed in follow-up PRs:

- Request permission once even without LN balance because of CJIT and channel opening, or at least in these specific flows
- Make the FG service optional, since it became less vital after ldk-node caching improvements
  - Could be an option in [background payments settings](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=40364-134589&m=dev). Would need some design work

_Originally posted by @jvsena42 in https://github.com/synonymdev/bitkit-android/issues/1004#issuecomment-4680332867_

jvsena42:
Fixes #1007

This PR:
1. Adds a "Keep Bitkit active in background" toggle to the Background Payments screen that makes the persistent foreground service opt-in and disabled by default. The Lightning node still runs while the app is open, and background payments still arrive via push notifications; enabling the toggle keeps Bitkit active in the background for more reliable payments, at the cost of a persistent notification.
2. Updates the Background Payments screen and the notification preview to match the latest design.
3. Always shows the amount in payment notifications and removes the "Include amount in notifications" option.
4. Fixes a Lightning event-handler leak so toggling the service on and off no longer accumulates stale handlers.

### Description

- New persisted setting, off by default. The foreground service start gate now also requires it, and turning it off stops a running service immediately so the persistent notification disappears at once.
- Stopping the foreground service no longer stops the in-app node when an activity is active; in the foreground the node lifecycle is owned by the wallet view model, so toggling the service off leaves the node running. When the app is backgrounded, the node still stops as before.
- The Background Payments screen is reordered to match the design: "Get paid when Bitkit is closed", "Keep Bitkit active in background", a Notifications section with the Customize button, and a Preview of the payment notification. The "Keep Bitkit active in background" switch is disabled while notifications are denied.
- The notification preview is redesigned to the Android notification card style, with the app icon, title, time, amount, and an expand chevron.
- Removed the show-notification-details setting and its privacy toggle. Payment notifications now always include the amount across the foreground service, in-app handler, and the background push path.
- Added a remove-event-handler API and de-register the foreground service handler when the service is destroyed, preventing stale handlers from accumulating across service restarts.

### Preview

[disable-fg-switch.webm](https://github.com/user-attachments/assets/e6c6f016-50fe-44e6-aa60-7e144018ecc8)

[disable-notifications.webm](https://github.com/user-attachments/assets/abaec2c7-47da-4601-8ef8-da151c36848f)

[running-from-master.webm](https://github.com/user-attachments/assets/a8c1d999-603f-4b4f-9c44-337e13f59a55)

[wake-to-pay.webm](https://github.com/user-attachments/assets/96616815-3b17-4761-94dd-f58e47eb6228)

[enable-fg-service.webm](https://github.com/user-attachments/assets/5bb03978-9d9a-4ca5-87ce-69c7a9f37131)

### QA Notes

[FIGMA](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=40364-135743&m=dev)

#### Manual Tests

- [x] **1.** Settings → Background Payments: "Keep Bitkit active in background" is off by default and no persistent Bitkit notification is shown in the status bar.
- [x] **2.** Background Payments → toggle Keep active on: persistent Bitkit notification appears → background the app: node stays alive.
- [x] **3.** Background Payments → toggle Keep active off: persistent notification disappears immediately → with the app still open, receive a payment: still works (node keeps running).
- [x] **4.** Revoke notification permission in Android settings → Background Payments: "Get paid when Bitkit is closed" reads off and "Keep Bitkit active in background" is disabled.
- [x] **5.** `regression:` Keep active off, app closed → receive a payment: push notification arrives and shows the amount.
- [x] **6.** Background Payments: layout matches the design (two switches with explanations, Notifications/Customize, Preview card).

#### Automated Checks

- Unit tests added: default value and setter for the new toggle in `SettingsViewModelTest.kt`.
- Test coverage removed: hidden-amount assertions deleted from `ReceivedNotificationContentTest.kt` and `NotifyChannelReadyHandlerTest.kt` because the show-notification-details option no longer exists; `SettingsData` stubs updated in `NotifyPaymentReceivedHandlerTest.kt`.
- Ran locally: `just compile`, the affected unit tests, and `just lint`.

### DIFF
--- a/app/src/main/java/to/bitkit/ui/ContentView.kt
+++ b/app/src/main/java/to/bitkit/ui/ContentView.kt
@@ -246,6 +246,7 @@ fun ContentView(
 
     val isRecoveryMode by walletViewModel.isRecoveryMode.collectAsStateWithLifecycle()
     val notificationsGranted by settingsViewModel.notificationsGranted.collectAsStateWithLifecycle()
+    val keepActiveInBackground by settingsViewModel.keepBitkitActiveInBackground.collectAsStateWithLifecycle()
     val walletExists = walletUiState.walletExists
 
     val requestNotificationPermission = rememberRequestNotificationPermission(
@@ -273,11 +274,10 @@ fun ContentView(
                 }
 
                 Lifecycle.Event.ON_STOP -> {
-                    if (walletExists && !isRecoveryMode && !notificationsGranted) {
-                        // App backgrounded without notification permission - stop node
+                    val keptAliveByService = notificationsGranted && keepActiveInBackground
+                    if (walletExists && !isRecoveryMode && !keptAliveByService) {
                         walletViewModel.stop()
                     }
-                    // If notificationsGranted=true, service keeps node running
                 }
 
                 else -> Unit
--- a/app/src/main/java/to/bitkit/ui/ContentView.kt
+++ b/app/src/main/java/to/bitkit/ui/ContentView.kt
@@ -274,7 +274,9 @@ fun ContentView(
                 }
 
                 Lifecycle.Event.ON_STOP -> {
-                    val keptAliveByService = notificationsGranted && keepActiveInBackground
+                    val keptAliveByService = notificationsGranted &&
+                        keepActiveInBackground &&
+                        appViewModel.isForegroundServiceRunning()
                     if (walletExists && !isRecoveryMode && !keptAliveByService) {
                         walletViewModel.stop()
                     }
--- a/changelog.d/next/fg-service-optional.added.md
+++ b/changelog.d/next/fg-service-optional.added.md
@@ -1,0 +1,1 @@
+Background payments now include a "Keep Bitkit active in background" option for more reliable payments.
--- a/changelog.d/next/fg-service-optional.added.md
+++ b/changelog.d/next/fg-service-optional.added.md
@@ -1,1 +1,0 @@
-Background payments now include a "Keep Bitkit active in background" option for more reliable payments.
--- a/app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
+++ b/app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
@@ -35,6 +35,7 @@ import to.bitkit.models.NewTransactionSheetDetails
 import to.bitkit.models.NotificationDetails
 import to.bitkit.repositories.LightningRepo
 import to.bitkit.repositories.WalletRepo
+import to.bitkit.services.NodeEventHandler
 import to.bitkit.services.NodeServiceFgState
 import to.bitkit.ui.ID_NOTIFICATION_NODE
 import to.bitkit.ui.MainActivity
@@ -80,18 +81,20 @@ class LightningNodeService : Service() {
 
     private var hasStartedNode = false
 
+    private val nodeEventHandler: NodeEventHandler = { event ->
+        Logger.debug("LDK-node event received in $TAG: ${jsonLogOf(event)}", context = TAG)
+        handlePaymentReceived(event)
+        if (event is Event.ChannelReady) handleChannelReady(event)
+        handlePendingPaymentResolved(event)
+    }
+
     private fun setupService() {
         if (hasStartedNode) return
         hasStartedNode = true
 
         serviceScope.launch {
             lightningRepo.start(
-                eventHandler = { event ->
-                    Logger.debug("LDK-node event received in $TAG: ${jsonLogOf(event)}", context = TAG)
-                    handlePaymentReceived(event)
-                    if (event is Event.ChannelReady) handleChannelReady(event)
-                    handlePendingPaymentResolved(event)
-                }
+                eventHandler = nodeEventHandler,
             ).onSuccess {
                 walletRepo.setWalletExistsState()
                 walletRepo.refreshBip21()
@@ -238,6 +241,8 @@ class LightningNodeService : Service() {
     override fun onDestroy() {
         Logger.debug("onDestroy", context = TAG)
         nodeServiceFgState.setForegroundServiceRunning(false)
+        // Drop our event handler so it isn't retained by the repo singleton across service restarts.
+        lightningRepo.removeEventHandler(nodeEventHandler)
         // Only stop the node when no activity is active; in the foreground WalletViewModel owns the
         // node lifecycle, so toggling off the foreground service must leave the node running.
         // Safe to call even if already stopped — guarded by lifecycleMutex + isStoppedOrStopping()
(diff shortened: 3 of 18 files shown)


ovitrif (commented):
Reviewed locally. I left two comments: one lifecycle issue around the background transition, and one changelog cleanup. Leaving this as a neutral review.

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -273,11 +274,10 @@ fun ContentView(
                 }
 
                 Lifecycle.Event.ON_STOP -> {
-                    if (walletExists && !isRecoveryMode && !notificationsGranted) {
-                        // App backgrounded without notification permission - stop node
+                    val keptAliveByService = notificationsGranted && keepActiveInBackground
```
ovitrif: I think this observer can use stale `notificationsGranted` and `keepActiveInBackground` values because the effect is keyed only by `lifecycle`. If the user toggles Keep Bitkit active during the same activity session, the next `ON_STOP` can still follow the old setting and either stop the node despite the service being enabled or leave it running after the service was disabled. Could we keep these values current inside the observer, for example with `rememberUpdatedState` or by recreating the effect when they change?

changelog.d/next/fg-service-optional.added.md
```diff
@@ -0,0 +1 @@
+Background payments now include a "Keep Bitkit active in background" option for more reliable payments.
```
ovitrif: Could we rename this fragment to `1053.added.md` and update the text to also mention that payment notifications now always include the amount? The collection script only appends the PR reference when the fragment ref is numeric, and this PR removes the previous hidden-amount behavior.

ovitrif (dismissed):
tAck. The two comments from my previous review are more like nits, the first one didn't really repro in my tests.

piotr-iohk (dismissed):
tACK

jvsena42:
draft for check comments

app/src/main/java/to/bitkit/ui/ContentView.kt
jvsena42: both `notificationsGranted` and `keepActiveInBackground` are delegated Compose State, not plain captured values.  
`keptAliveByService` is only evaluated when `ON_STOP` is triggered, getting a fresh read of State.value for each variable

jvsena42:
No logic has been changed, only the changelog

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -273,11 +274,10 @@ fun ContentView(
...
+                    if (walletExists && !isRecoveryMode && !keptAliveByService) {
                         walletViewModel.stop()
```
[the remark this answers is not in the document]
jvsena42: false positive, see https://github.com/synonymdev/bitkit-android/pull/1053#discussion_r3506555024

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -273,11 +274,10 @@ fun ContentView(
...
                     }
```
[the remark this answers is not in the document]
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1053/commits/0466cee1576da7ab0f0c32010dbc3d4f127ee20e

piotr-iohk:
Tested again on device (Samsung S22, Android 16).

**Observations**

Testing "Get paid when Bitkit is closed" ON:

- Keep Bitkit active OFF + Battery savings OFF: background payments work. Push wake path runs end-to-end (FCM → `WakeNodeWorker` → payment notification). Matches QA case **5**.
- Keep Bitkit active OFF + Battery savings ON: background payments don't work for me. No reliable push wake in logs; payments only land when the app is open or woken manually.
- Keep Bitkit active ON + Battery savings ON: works. Node stays up via `LightningNodeService` and the persistent notification — expected workaround on aggressive OEM power saving.

Attaching session logs. 
[bitkit_logs_2026-07-02_09-06-43.zip](https://github.com/user-attachments/files/29591433/bitkit_logs_2026-07-02_09-06-43.zip)

Implementation looks correct to me — **approve**.

**Non-blocking:** the text under "Get paid when Bitkit is closed" (`settings__bg__enabled`) over-promises a bit. It says you can get paid when the app is closed if you're online, but battery saver can block that path even with notifications on. Something like:

> Background payments are enabled. You can receive funds when the app is closed, if your device is online. On some phones, turning off battery saver improves reliability; if payments are still missed, enable "Keep Bitkit active in background" below.

Worth a QA note too: case **5** with battery saver off; Samsung + battery saver on = known limitation for the default (no persistent notification) path.

piotr-iohk:
Disabled auto-merge so there’s room to update `settings__bg__enabled` if you want

piotr-iohk (approved):

jvsena42:
> Tested again on device (Samsung S22, Android 16).
> 
> **Observations**
> 
> Testing "Get paid when Bitkit is closed" ON:
> 
> * Keep Bitkit active OFF + Battery savings OFF: background payments work. Push wake path runs end-to-end (FCM → `WakeNodeWorker` → payment notification). Matches QA case **5**.
> * Keep Bitkit active OFF + Battery savings ON: background payments don't work for me. No reliable push wake in logs; payments only land when the app is open or woken manually.
> * Keep Bitkit active ON + Battery savings ON: works. Node stays up via `LightningNodeService` and the persistent notification — expected workaround on aggressive OEM power saving.
> 
> Attaching session logs. [bitkit_logs_2026-07-02_09-06-43.zip](https://github.com/user-attachments/files/29591433/bitkit_logs_2026-07-02_09-06-43.zip)
> 
> Implementation looks correct to me — **approve**.
> 
> **Non-blocking:** the text under "Get paid when Bitkit is closed" (`settings__bg__enabled`) over-promises a bit. It says you can get paid when the app is closed if you're online, but battery saver can block that path even with notifications on. Something like:
> 
> > Background payments are enabled. You can receive funds when the app is closed, if your device is online. On some phones, turning off battery saver improves reliability; if payments are still missed, enable "Keep Bitkit active in background" below.
> 
> Worth a QA note too: case **5** with battery saver off; Samsung + battery saver on = known limitation for the default (no persistent notification) path.

I worth a specific investigation. It is possible to display some warning when the app is in battery save mode and also enable  specific permissions to don't limit the app on battery safe mode, but first we need to improve the node consumption.
Also we have we need to be careful not use too long texts

jvsena42:
I left comment for future iterations

https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-135743&m=dev#1827426454

PR #1062, opened 2026-07-03
feat: bump core to 0.3.9 and hw reliability

piotr-iohk:
Part of #1030

This PR:
1. Bumps bitkit-core from 0.1.75 to 0.3.9 and migrates CoreService, hardware wallet, activity, and migration paths to wallet-scoped APIs.
2. Improves BLE Trezor reconnect reliability for Transfer to Spending after app restart or device sleep.
3. Suppresses empty error toasts when the user cancels signing on the Trezor.
4. Serializes transfer reconnect with in-flight auto-reconnect and scopes sign-screen warm-up to the selected hardware device.

### Description

The bitkit-core upgrade brings wallet-scoped storage and Trezor stack improvements from the 0.3.x line (device-busy handling, typed session errors, transport callback updates). Android wiring moves default wallet identity through `WalletScope`, routes CoreService calls through scoped wallet ids, and updates hardware wallet, activity, migration, and dev Trezor tooling to match.

On top of the core bump, this branch hardens hardware transfer UX:

- Transfer to Spending retries BLE reconnect with backoff instead of failing immediately when the Trezor is asleep after restart.
- Opening Trezor Connect on the sign screen warms up that specific BLE device instead of triggering a global foreground reconnect for any known wallet.
- Connect paths wait for an in-flight auto-reconnect before starting another attempt, reducing "connection already in progress" races.
- User cancellation during signing (`UserCancelled`, `PinCancelled`, `PassphraseCancelled`) is treated as a silent no-op: no error toast, no stale-session disconnect churn.

Out of scope for this PR (still tracked elsewhere):

- Passphrase / hidden wallet pairing (#1060).
- Trezor logging consolidation into the main app log file.
- E2E emulator coverage (#1038).

### Preview

| Before (master / 0.1.75) | After (this branch) |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/6a86256a-6399-44d9-be70-f066a5eb940a" width="320" controls></video> | <video src="https://github.com/user-attachments/assets/1e90c095-c51a-486b-8077-86831a82ade6" width="320" controls></video> |

Upload your local recordings into the table and below when opening the draft PR on GitHub.

#### Manual Tests

- [ ] **1.** Pair BLE Trezor → force-stop app → reopen → Savings with HW balance → Transfer to Spending → Open Trezor Connect: reconnect succeeds, PIN/sign flow completes (compare with before video on master).
[...]

PR #1064, opened 2026-07-08
fix: migrate legacy backups missing wallet id on restore

coreyphillips:
## Summary

Stacked on #1062. Fixes the P1 raised by the Codex review there: legacy backups that predate wallet-scoped activity data fail to decode on restore because their activity, tag and pre-activity-metadata records have no wallet id, and with bitkit-core 0.4.0 those records require one. The decode throws before any repository migration can fill a default.

## Approach

Before decoding a restored envelope, each Core-owned array (`activities`, `activityTags`, `tagMetadata`) is handed to the matching bitkit-core migration helper as raw JSON:

- `migrateBackupActivitiesJson`
- `migrateBackupActivityTagsJson`
- `migrateBackupPreActivityMetadataJson`

Core fills in the default wallet id for records that lack one (or have it empty) and preserves any existing non-default id such as `trezor:{hash}`. The app only extracts and reinserts the sub-arrays, so it never edits Core model JSON itself. The helpers are idempotent, so current backups that already carry a wallet id are unaffected.

This is the app half of synonymdev/bitkit-core#113; the migration APIs it calls shipped in bitkit-core 0.4.0.

## Changes

- `BackupRepo.kt`: add `migrateCoreOwnedBackupFields` and apply it at the `METADATA` and `ACTIVITY` restore sites.
- Bump bitkit-core to 0.4.0.

## Notes

- Base is `feat/wallet-scoped-core-0.3.9` (#1062), which introduces the wallet-scoped restore path this builds on. Retarget to `master` once #1062 merges.


piotr-iohk:
Build failure looks like a missing Maven artifact, not a code issue: `Could not find com.synonym:bitkit-core-android:0.4.0`.

[bitkit-core Gradle Package for v0.4.0](https://github.com/synonymdev/bitkit-core/actions/runs/28949827135/job/85893273254) is still running

piotr-iohk (approved):
tACK

 - In-place upgrade 2.3.1 → head : wallet state intact, Trezor pair + TO_SPENDING worked
 - 2.3.1 VSS backup → wipe → restore on head: legacy wallet restored cleanly; logs show Full restore success, Restored 3 activities, 1 activity tags; balances/channel/order came back (on-chain + spending), Trezor pair + TO_SPENDING worked
 - no issues in logs

piotr-iohk:
Merging, my latest push only resolves the `libs.versions.toml` conflict with master (paykit bump), with no further code changes.

PR #1066, opened 2026-07-10
feat: support paykit receiver paths

ben-kaufman:
This PR:
1. Updates Paykit to v0.1.0-rc33 and configures the mobile app as the `bitkit/wallet` receiver.
2. Discovers and saves supported `bitkit/wallet` and `bitkit/server` receivers while keeping one visible contact per Pubky identity.
3. Scopes private links, payment details, reservations, retries, and cleanup to the exact receiver path.
4. Publishes the wallet receiver marker with the app's current public/private capabilities and removes it when Paykit sharing is disabled.
5. Refreshes receiver discovery when an existing contact is scanned or pasted, while normal send-to-contact continues to target `bitkit/wallet` explicitly.

### Description

Paykit now supports multiple apps under one Pubky identity. This change adopts that model without changing Bitkit's contact UI: a person remains one contact, while their saved SDK contact record can track multiple receiver paths.

Private links are maintained for supported receivers that advertise private capabilities. Receiving details are published independently per receiver and only when that receiver can send payments, so a `bitkit/server` receiver can be discovered and linked without receiving wallet invoices or addresses. Public contact payments and fallback remain scoped to `bitkit/wallet`.

Receiver-record failures are isolated per contact, and rescanning an existing contact refreshes only that contact while showing the validation result immediately.

No migration is included because the receiver-path Paykit state has not shipped and existing development state can be discarded.

Companion iOS PR: https://github.com/synonymdev/bitkit-ios/pull/620

### Preview

N/A

#### Manual Tests

- [ ] **1.** Contacts → add a Paykit-enabled Bitkit user: one contact is saved and its `bitkit/wallet` receiver is available for payment.
- [ ] **2.** Send → Contact → select the saved contact: private payment resolves through `bitkit/wallet`; public fallback still works when private payment is unavailable.
- [ ] **3.** Existing contact → scan or paste the same Pubky key after a `bitkit/server` receiver is published: the contact remains one row and the new receiver is saved in the background.
- [ ] **4.** Contact with a `bitkit/server` marker that cannot send payments: Bitkit may maintain its private link but does not publish wallet invoices or addresses to that receiver.
- [ ] **5.** Settings → Payment Preferences → disable contact payment sharing: receiver-scoped payment details and the local receiver marker are cleaned up.

#### Automated Checks

- `PrivatePaykitRepoTest.kt` and `PublicPaykitRepoTest.kt` cover receiver-scoped resolution, publication, cleanup, reservations, and fallback behavior.
- `RefreshContactPaykitReceiversUseCaseTest.kt`, `PrivatePaykitRepoTest.kt`, and contact view-model tests cover targeted receiver refresh and per-contact failure isolation.
- Full `testDevDebugUnitTest`, `compileDevDebugKotlin`, and `detekt` passed locally.
- `git diff --check` passed.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
@@ -34,6 +34,20 @@ sealed class PrivatePaykitAddressReservationError(message: String) : AppError(me
     )
 }
 
+internal data class ContactAssignmentKey(
+    val publicKey: String,
+    val receiverPath: String,
+) {
+    fun encoded(): String =
+        if (receiverPath == PaykitReceiverPaths.WALLET) publicKey else "$publicKey$SEPARATOR$receiverPath"
+
+    companion object {
+        private const val SEPARATOR = '#'
+
+        fun publicKeyOf(encoded: String): String = encoded.substringBefore(SEPARATOR)
+    }
+}
+
 @Singleton
 @Suppress("TooManyFunctions")
 class PrivatePaykitAddressReservationRepo @Inject constructor(
@@ -378,12 +392,10 @@ class PrivatePaykitAddressReservationRepo @Inject constructor(
     private fun normalizedPublicKeyOrNull(publicKey: String): String? =
         PubkyPublicKeyFormat.normalized(publicKey)
 
-    private fun contactAssignmentKey(publicKey: String, receiverPath: String): String {
-        val normalizedKey = normalizedPublicKey(publicKey)
-        return if (receiverPath == PaykitReceiverPaths.WALLET) normalizedKey else "$normalizedKey#$receiverPath"
-    }
+    private fun contactAssignmentKey(publicKey: String, receiverPath: String): String =
+        ContactAssignmentKey(normalizedPublicKey(publicKey), receiverPath).encoded()
 
-    private fun String.publicKeyFromAssignmentKey(): String = substringBefore("#")
+    private fun String.publicKeyFromAssignmentKey(): String = ContactAssignmentKey.publicKeyOf(this)
 
     private fun redacted(publicKey: String): String =
         PubkyPublicKeyFormat.redacted(publicKey)
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
@@ -15,12 +15,14 @@ import to.bitkit.data.PrivatePaykitReservationStore
 import to.bitkit.data.PrivatePaykitStoredAssignmentData
 import to.bitkit.data.SettingsStore
 import to.bitkit.di.IoDispatcher
+import to.bitkit.ext.runSuspendCatching
 import to.bitkit.models.DEFAULT_ADDRESS_TYPE
 import to.bitkit.models.PubkyPublicKeyFormat
 import to.bitkit.models.addressTypeFromAddress
 import to.bitkit.models.toAddressType
 import to.bitkit.models.toSettingsString
 import to.bitkit.services.CoreService
+import to.bitkit.services.PaykitReceiverPaths
 import to.bitkit.utils.AppError
 import to.bitkit.utils.Logger
 import javax.inject.Inject
@@ -80,33 +82,24 @@ class PrivatePaykitAddressReservationRepo @Inject constructor(
             }
         }
 
-    suspend fun currentOrRotatedAddress(publicKey: String): Result<String> = withContext(ioDispatcher) {
-        runCatching {
-            val normalizedKey = normalizedPublicKey(publicKey)
-            val current = locked { it.contactAssignments[normalizedKey] }
+    suspend fun currentOrRotatedAddress(
+        publicKey: String,
+        receiverPath: String,
+    ): Result<String> = withContext(ioDispatcher) {
+        runSuspendCatching {
+            val assignmentKey = contactAssignmentKey(publicKey, receiverPath)
+            val current = locked { it.contactAssignments[assignmentKey] }
             if (current != null && isAddressTypeMonitored(current.addressType)) {
                 val address = resolvedAddress(current).getOrThrow()
-                if (!isReservedAddressUsed(address)) return@runCatching address
+                if (!isReservedAddressUsed(address)) return@runSuspendCatching address
             } else if (current != null) {
-                clearCurrentAssignment(normalizedKey)
+                clearCurrentAssignment(assignmentKey)
             }
 
-            allocateAddress(normalizedKey).getOrThrow()
+            allocateAddress(assignmentKey).getOrThrow()
         }.onFailure {
             Logger.warn(
                 "Failed to get private Paykit address for '${redacted(publicKey)}'",
-                it,
-                context = TAG,
-            )
-        }
-    }
-
-    suspend fun rotateAddress(publicKey: String): Result<String> = withContext(ioDispatcher) {
-        runCatching {
-            allocateAddress(normalizedPublicKey(publicKey)).getOrThrow()
-        }.onFailure {
-            Logger.warn(
-                "Failed to rotate private Paykit address for '${redacted(publicKey)}'",
                 it,
                 context = TAG,
             )
@@ -163,7 +156,7 @@ class PrivatePaykitAddressReservationRepo @Inject constructor(
         assignments.firstOrNull { (_, assignment) ->
             assignment.addressType == addressType &&
                 (assignment.address == address || resolvedAddress(assignment).getOrNull() == address)
-        }?.first
+        }?.first?.publicKeyFromAssignmentKey()
     }
 
     suspend fun currentContactPublicKeyForReservedAddress(address: String): String? = withContext(ioDispatcher) {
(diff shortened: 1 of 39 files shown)


app/src/main/java/to/bitkit/viewmodels/SettingsViewModel.kt
```diff
@@ -219,6 +221,12 @@ class SettingsViewModel @Inject constructor(
                     settingsStore.update { it.copy(publicPaykitCleanupPending = true) }
                     Logger.warn("Failed to remove public Paykit endpoints after disabling Paykit UI", it, context = TAG)
                 }
+        } else if (hadPrivatePaykitState) {
+            publicPaykitRepo.syncLocalReceiverMarker(publicSharingEnabled = false, privateSharingEnabled = false)
```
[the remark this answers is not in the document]
ben-kaufman: This call only runs when the entire Paykit feature is disabled, so `false/false` and marker removal are intentional here. Turning off public sharing alone goes through `syncPublishedEndpoints(false)`, which leaves the private setting in place, so `syncLocalReceiverMarker` keeps the marker while private sharing is enabled.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -145,11 +153,14 @@ class PrivatePaykitRepo @Inject constructor(
         }
     }
 
-    suspend fun refreshSavedContactEndpoints(publicKeys: Collection<String>): Result<Unit> =
+    suspend fun refreshSavedContactEndpoints(publicKey: String): Result<Unit> =
         withContext(serializedDispatcher) {
             runSuspendCatching {
-                val keys = rememberSavedContacts(publicKeys, replacing = true)
-                if (!canPublishPrivateEndpoints()) return@runSuspendCatching
+                val keys = rememberSavedContacts(listOf(publicKey), replacing = false)
```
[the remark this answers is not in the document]
ben-kaufman: Good catch. Fixed in `48df7a477`: targeted refresh now registers the full current saved-contact set while still republishing only the selected contact. The use-case test verifies that the complete set is passed.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -896,6 +1083,57 @@ class PrivatePaykitRepo @Inject constructor(
...
+    private suspend fun receiverPathsForCleanup(publicKey: String): List<String> {
+        val paths = (
+            linkedReceiverPaths(publicKey) +
+                publishedPrivatePaymentReceiverPaths(publicKey)
+            )
+            .filter { it in PaykitReceiverPaths.supported }
+            .distinct()
+            .sorted()
+        return paths
+    }
```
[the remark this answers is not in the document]
ben-kaufman: This is intentional. A saved receiver path only means the receiver was discovered, not that private payment details were shared. We clear every path that is linked or recorded as published. Any path that received a private list has linked state, while queued publication is also retained in local publication state. Clearing every saved path could instead fail cleanup for receivers that never linked or received a list.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -700,22 +837,27 @@ class PrivatePaykitRepo @Inject constructor(
                 return retryKeys
             }
             .mapNotNull { peer ->
-                normalizedPublicKey(peer.counterparty)?.let { publicKey -> publicKey to peer.state }
+                normalizedPublicKey(peer.counterparty)?.let { publicKey ->
+                    PrivateMessageDrainRetryKey(publicKey, peer.counterpartyReceiverPath) to peer.state
+                }
             }
             .toMap()
```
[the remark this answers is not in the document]
ben-kaufman: The SDK cannot return duplicate records for this key. rc33 stores `linked_peers` in a `HashMap<(PubkyPublicKey, PaykitReceiverPath), LinkedPeerRecord>` and `linkedPeers()` returns its values, so each `(publicKey, receiverPath)` pair is unique. Converting the result to a map is safe here.

jvsena42:
@ben-kaufman I'll review today, please solve the conflict

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -79,6 +82,21 @@ data class PaykitResolvedPaymentEndpoint(
...
+    val cleanupProtectedReceiverPaths: List<String>,
+    val error: Throwable?,
+)
+
+internal object PaykitReceiverPaths {
+    const val WALLET = "bitkit/wallet"
+    const val SERVER = "bitkit/server"
+
+    /** Current Bitkit flows only route its own receivers; cross-wallet routing can broaden this allowlist. */
+    val supported = listOf(WALLET, SERVER)
```
jvsena42: nit: A Set would be a marginally better fit for repeated contains

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -302,11 +324,24 @@ class PaykitSdkService @Inject constructor(
...
+    suspend fun contactRecord(publicKey: String): ContactRecord? {
         isSetup.await()
         return operationMutex.withLock {
-            handle().saveContact(ContactUpdate(publicKey, label)).also {
-                notifyBackupStateChanged()
+            handle().contactRecord(publicKey)
+        }
+    }
+
+    suspend fun saveContact(
```
jvsena42, line 334: nit: could also implement a pruning logic for dead paths in other PR

app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
```diff
@@ -377,6 +378,13 @@ class PrivatePaykitAddressReservationRepo @Inject constructor(
     private fun normalizedPublicKeyOrNull(publicKey: String): String? =
         PubkyPublicKeyFormat.normalized(publicKey)
 
+    private fun contactAssignmentKey(publicKey: String, receiverPath: String): String {
+        val normalizedKey = normalizedPublicKey(publicKey)
+        return if (receiverPath == PaykitReceiverPaths.WALLET) normalizedKey else "$normalizedKey#$receiverPath"
+    }
+
+    private fun String.publicKeyFromAssignmentKey(): String = substringBefore("#")
```
jvsena42: nit: could move the encode/decode logic to a dedicated class to make it more testable and better document the separator.

```
internal data class ContactAssignmentKey(
    val publicKey: String,
    val receiverPath: String,
) {
    /** WALLET encodes to the bare public key, so pre-existing wallet-scoped assignments keep their key. */
    fun encoded(): String =
        if (receiverPath == PaykitReceiverPaths.WALLET) publicKey else "$publicKey$SEPARATOR$receiverPath"

    companion object {
        private const val SEPARATOR = '#'

        /** Receiver paths and normalized pubkeys never contain [SEPARATOR], so the prefix is the public key. */
        fun publicKeyOf(encoded: String): String = encoded.substringBefore(SEPARATOR)
    }
}


private fun contactAssignmentKey(publicKey: String, receiverPath: String): String =
    ContactAssignmentKey(normalizedPublicKey(publicKey), receiverPath).encoded()

private fun String.publicKeyFromAssignmentKey(): String =
    ContactAssignmentKey.publicKeyOf(this)

```

jvsena42 (dismissed):

jvsena42:
Approved with some suggestions

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Updated in 32b8c1d11: PaykitReceiverPaths.supported is now a Set, matching its membership-only use.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Agreed; dead-path pruning belongs in a follow-up. I left behavior unchanged here because this PR does not yet define an authoritative liveness policy or pruning trigger.

app/src/main/java/to/bitkit/repositories/PrivatePaykitAddressReservationRepo.kt
ben-kaufman: Updated in 32b8c1d11: extracted ContactAssignmentKey to centralize separator encoding and decoding, with direct wallet and server assignment-key tests.

jvsena42 (changes requested):
> [bitkit_e2e.apk Android #0-0] add contact pubky9tm89bfrr3qxic5nbf9oyxfym4cd4jagx68ukogw1j1pz66ksq6y: expected AddContactPay visible (fixture ableToPay=true)

Please, check if this an issue from the PR or the E2E setup

Image

ben-kaufman:
Already asked @piotr-iohk to help with the e2e tests. Since the way we save data changed we need to updated the existing tests to use this new structure with the pre made users

piotr-iohk (approved):
tACK

Manual regression (regtest / staging), Android ↔ iOS, on `codex/paykit-rc33-receivers`:

- Native profile create + mutual contacts ✅
- Cross-platform contact LN payments (both directions) ✅
- Contact attribution after payments (`setContact` / no public fallback noise) ✅
- Profile delete → recreate (same pubky) → re-add → pay again ✅
- Staging paykit fixtures republished under `/pub/paykit/v0/bitkit/wallet/`; e2e `@paykit` / `@pubky` green ✅

Known deferred (unchanged vs [#1040](https://github.com/synonymdev/bitkit-android/pull/1040#issuecomment-4903550817)): Android Pubky Ring import still fails after Ring auth success with `code=identity_error, context=complete Pubky auth flow`. iOS Ring import succeeded in the same session. Not blocking this PR.

jvsena42 (approved):
found same points of https://github.com/synonymdev/bitkit-ios/pull/620#pullrequestreview-4746775893. It is safe to address on other PR if necessary

Issue #1060, opened 2026-07-03
Hardware wallet: passphrase (hidden) wallet pairing


piotr-iohk:
Part of #998.

**Related (iOS epic):** synonymdev/bitkit-ios#589

**Related:** #1030 — reconnect/signing safety for passphrase identities (runtime). This issue owns the **pairing UX** and **watch-only identity model**.

## Design

| Screen | Figma |
|---|---|
| Enter Passphrase | [46312-107873](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46312-107873&t=Y2WiEWMOBV2Mh0wk-0) |
| Passphrase funds found | [46312-107925](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46312-107925&t=Y2WiEWMOBV2Mh0wk-0) |

Extends the existing Connect Hardware **Paired** step (standard wallet): secondary **Passphrase** action → enter passphrase → confirm label/balance → **Passphrase** again or **Finish**.

## Goal

Pair and watch **multiple Trezor wallet identities** from one physical device:

- **Standard wallet** (empty passphrase) — already shipped.
- **Hidden wallet(s)** — user enters a passphrase; Bitkit persists xpubs and shows a separate hardware balance tile.

Trezor does not store passphrases or list hidden wallets. Bitkit must never persist raw passphrases — watch-only via xpubs; re-enter passphrase when signing.

## Scope

- **Connect flow:** Passphrase button on Paired → Enter Passphrase → Passphrase Paired (label + balance) → loop or Finish.
- **Passphrase entry:** host (phone) and/or on-device on capable models (Safe 5/7), same as dev Trezor screen.
- **Identity model:** each passphrase-derived xpub set = separate `walletKey` / home tile / settings row / remove action.
- **Settings follow-up:** Add passphrase wallet on an already-paired device (not only during initial connect loop).
- **Zero balance:** allow add; non-blocking warning if balance is 0 (*"No balance found — double-check passphrase if you expected funds"*).
- **Duplicate:** same xpub already watched → "already added", no duplicate tile.
- **Signing:** Transfer to Spending from a hidden wallet reconnects/signs with that identity (coordinate with #1030).

## Out of scope

- Enabling Trezor "passphrase protection" device setting (Trezor Suite).
- Discovering unknown passphrases / storing passphrase for convenience.
[...]

Issue #1038, opened 2026-06-22
Hardware wallet emulator E2E coverage


ovitrif:
Part of #998.

Split from #1030 so connection reliability and UX polish can stay focused on runtime behavior while this issue owns automated coverage.

## Scope

Add hardware-wallet automated test support so the connect and transfer flows can be run by AI agents and in CI without a physical device.

- Audit and add UI test hooks (`testTag`s / accessibility IDs) on the hardware-wallet screens where needed.
- Keep Android journey docs aligned with the current hardware-wallet screen tags and flow names.
- Add a hardware-wallet E2E shard that can run the Connect Hardware flow against a Trezor emulator.
- Drive the E2E path through a Trezor emulator, not a physical device.
- Cover the Transfer to Spending signing path once the emulator/device bridge can exercise it reliably.
- Include coverage for USB and BLE-equivalent route handling where the test environment can model both paths.
- Include regression coverage for locked-device, pairing-code retry, PIN retry, reconnect, and device-busy/backoff behavior once the supporting bitkit-core releases are consumed.
- Include passphrase-enabled wallet scenarios if that UX is supported by the app flow.

## Related work

- #1008 tracks aligning Android Trezor dev tooling with `bitkit-docker`.
- #1030 tracks the runtime connection reliability and UX polishing that this automation should eventually prove.
- synonymdev/bitkit-core#104 adds lock-state and busy-state support.
- synonymdev/bitkit-core#105 tracks typed Trezor PIN failure mapping for mobile bindings.

## Acceptance criteria

- Hardware-wallet journeys can be run locally by an agent without a physical Trezor.
- CI has a dedicated hardware-wallet E2E path or shard once the emulator setup is reliable enough for hosted runners.
- Connect Hardware flow coverage includes the happy path and at least one retry/error path.
- Transfer to Spending coverage proves the app reaches and handles the signing path with the emulator.
- Test hooks and journey docs stay aligned with the implemented Compose `testTag`s.

piotr-iohk:
Done.

Issue #123, opened 2025-05-02
EditInvoiceScreen is not keeping balance state


jvsena42:
The input is cleared when the user navigates to other screen

PR #1067, opened 2026-07-10
fix: polish trezor reliability and ux

### LINKED ISSUE #1030: Hardware wallet connection reliability and UX polish

piotr-iohk:
Fixes #1030

This PR:
1. Resets the six-digit Trezor pairing flow when the device requests a new code.
2. Fails pending BLE operations promptly when Bluetooth turns off.
3. Improves locked-device, reconnect, and connection guidance during hardware-wallet operations.
4. Separates hardware signing from broadcasting so a signed funding transaction can be retried without reopening Trezor.
5. Cancels abandoned hardware transfers and keeps Trezor diagnostics in the main app log with sensitive values redacted.

### Description

This completes the app-level v1 reliability and UX polish for Trezor hardware wallets.

Pairing prompts now carry a request identity, so a wrong code followed by a new device code resets the input instead of leaving a spinner-only state. Device unlock remains on the Trezor itself.

Transfer to Spending now distinguishes reconnect, signing, and broadcast failures. Locked or busy devices show unlock guidance, network failures retain the signed transaction for an in-flow retry, and leaving the sign screen cancels the active operation so a later device approval cannot broadcast after the user exits.

The hardware wallet count is hidden when empty, connection failures include practical BLE guidance, and Trezor diagnostics are forwarded to the normal app log with pairing codes, PINs, passphrases, xpubs, PSBTs, and raw transactions redacted.

Passphrase/hidden-wallet pairing remains separately tracked in #1060. Emulator E2E coverage was handled in #1038. Durable signed-transaction recovery across process termination and deeper upstream transport retry changes are deliberate non-goals for this v1 polish.

#### Manual Tests
- [x] **1.** Connect Hardware → enter an incorrect six-digit pairing code → confirm the new code on Trezor: input resets and accepts the replacement code.
- [x] **2.** `regression:` Connect Hardware → turn Bluetooth off during connect → turn it back on: operation exits without an indefinite wait and reconnect can proceed.
- [x] **3.** Locked Trezor → Hardware Transfer Sign → Open Trezor Connect: informational unlock guidance appears instead of a raw firmware/timeout error.
- [x] **4.** Hardware Transfer Sign → approve on Trezor while network broadcast fails → Retry: “Transaction signed” remains visible and broadcast retries without signing again.
- [x] **5.** Hardware Transfer Sign → start signing → back: later approval on Trezor does not broadcast or advance the order.
- [x] **6.** Settings → General with no paired hardware wallet: Hardware Wallets does not show a `0` value.
- [ ] **7.** Known BLE Trezor unavailable → Open Trezor Connect: informational Connect Device toast explains unlock, proximity, and other-device connection checks.

#### Automated Checks
- `TrezorTransportTest.kt` covers unique pairing requests and releasing pending BLE operation latches.
- `TrezorRepoTest.kt` covers locked-state handling, reconnect retries, and current-device reuse.
- `HwConnectViewModelTest.kt` and `AppViewModelSendFlowTest.kt` cover repeated pairing prompts and sheet routing.
- `HwWalletRepoTest.kt` covers separate signing/broadcasting and hardware-wallet business behavior.
- `TransferViewModelTest.kt` covers locked-device guidance, signed-broadcast retry, broadcast timeout, reconnect guidance, successful funding, and flow-exit cancellation.
- `TrezorDebugLogTest.kt`, `TrezorExceptionExtTest.kt`, and `TrezorErrorPresenterTest.kt` cover diagnostic redaction and user-facing error mapping.
- Full compile, unit test, and detekt checks passed locally; targeted suites were rerun after the final cancellation change.


app/src/main/java/to/bitkit/services/TrezorDebugLog.kt
```diff
@@ -29,6 +36,7 @@ object TrezorDebugLog {
                 updated.toImmutableList()
             }
         }
+        Logger.debug("Recorded Trezor diagnostic '$tag': '$sanitizedMessage'", context = TAG)
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. The sanitizer now accepts quoted JSON-style keys as well as key=value/key:value diagnostics, and delimiters stop unquoted values at JSON braces. Added coverage for structured xpub, PSBT, raw transaction, PIN, and passphrase fields.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3103,28 +3103,28 @@ class AppViewModel @Inject constructor(
...
-    private fun showPairingCodeSheet() {
+    private fun showPairingCodeSheet(requestId: Long) {
         if (isHighPrioritySheet(_currentSheet.value)) {
-            isPairingCodeSheetQueued = true
+            queuedPairingCodeRequestId = requestId
             return
         }
 
         // The Connect Hardware flow is itself a Hardware sheet and drives the pair-code step
         // inline within its own NavHost; replacing it here would tear down that back stack.
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. App-wide PairCode sheets now replace their start route when a newer request id arrives, while an active Connect Hardware flow still preserves its NavHost and handles the request through HwConnectViewModel. Added regression coverage for replacing request 1 with request 2.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -488,12 +497,18 @@ class TransferViewModel @Inject constructor(
...
+        cancelHardwareTransfer()
         _spendingUiState.update { TransferToSpendingUiState() }
         _transferValues.update { TransferValues() }
     }
 
+    fun cancelHardwareTransfer() {
+        hwTransferSignJob?.cancel()
+        hwTransferSignJob = null
+        pendingHwFundingBroadcast = null
+        _spendingUiState.update { it.copy(isSigning = false, hasPendingHwBroadcast = false) }
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. Cancellation is now phase-aware: leaving cancels reconnect/compose/sign work before a signed transaction exists, but once signing completes the in-flight broadcast and bookkeeping are allowed to finish. Pending retry state is cleared only after fundPaidOrder succeeds, and tests cover both pre-sign cancellation and leaving during broadcast.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -237,16 +241,31 @@ class HwWalletRepo @Inject constructor(
                     trezorRepo.disconnectStaleSession(deviceId)
                 }
             }
-            val txId = trezorRepo.broadcastRawTx(serializedTx = signed.getOrThrow().serializedTx).getOrThrow()
-            HwFundingBroadcastResult(
-                txId = txId,
+            val signedTx = signed.getOrThrow()
+            HwFundingSignedTx(
+                serializedTx = signedTx.serializedTx,
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. HwFundingSignedTx now retains the signer-provided txid. Duplicate/already-known Electrum responses recover that txid so transfer bookkeeping can complete without another signature. Added duplicate-broadcast coverage.

app/src/main/java/to/bitkit/repositories/TrezorRepo.kt
```diff
@@ -716,13 +720,27 @@ class TrezorRepo @Inject constructor(
...
-            return@withContext reconnectKnownBluetoothDevice(deviceId)
+        val result = awaitConnectedOrNull(deviceId)?.let { Result.success(it) } ?: run {
+            if (isKnownBluetoothDevice(deviceId)) {
+                reconnectKnownBluetoothDevice(deviceId)
+            } else {
+                connectKnownDevice(deviceId, forceSession = true)
+            }
         }
-        connectKnownDevice(deviceId, forceSession = true)
+        result.requireUnlocked()
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. A cached connected session with locked features now calls the typed trezorRefreshFeatures API before requireUnlocked, updates the cached connected state, and succeeds after the device reports unlocked. Added locked-still-busy and locked-then-unlocked regression coverage.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -638,21 +671,37 @@ class TransferViewModel @Inject constructor(
...
+    }
+
     private suspend fun handleHardwareTransferFailure(e: Throwable, deviceId: String) {
         if (e.isTrezorUserCancellation()) {
             Logger.info("Hardware transfer cancelled on device for '$deviceId'", context = TAG)
             return
         }
-        if (e.isTrezorDeviceBusy()) {
-            Logger.warn("Hardware transfer blocked by busy Trezor for '$deviceId'", e, context = TAG)
+        if (e.isTrezorLockedOrBusy() || e.isHardwareInteractionTimeout()) {
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in d132d02a. Compose timeouts no longer enter the locked-device branch. They now show the connectivity warning, while only typed/recognized Trezor busy errors show unlock guidance. Updated the regression test accordingly.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -542,16 +559,24 @@ class TransferViewModel @Inject constructor(
...
-                        fundPaidOrder(
-                            order = order,
-                            txId = result.txId,
-                            createTransferActivity = true,
-                            fee = result.miningFeeSats,
-                            feeRate = result.feeRate,
-                        )
-                        setTransferEffect(TransferEffect.OnHwTxSigned)
+                        runCatching {
+                            fundPaidOrder(
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in 88fe961c. The post-broadcast bookkeeping path now uses runSuspendCatching so CancellationException remains structured cancellation instead of being logged/handled as a recoverable hardware failure.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -672,11 +739,37 @@ class TransferViewModel @Inject constructor(
         }
     }
 
+    private suspend fun showHardwareBroadcastError(error: HardwareBroadcastError) {
+        if (!error.isHardwareBroadcastConnectivityFailure()) {
+            ToastEventBus.send(error.cause ?: error)
+            return
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in 88fe961c. Non-connectivity/permanent broadcast failures now clear the pending signed transaction and reset the signed retry state so the user can recompose/sign instead of repeatedly rebroadcasting the same rejected transaction. Added regression coverage.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3103,28 +3103,35 @@ class AppViewModel @Inject constructor(
...
-        // The Connect Hardware flow is itself a Hardware sheet and drives the pair-code step
-        // inline within its own NavHost; replacing it here would tear down that back stack.
-        if (_currentSheet.value is Sheet.Hardware) return
+        // The Connect Hardware flow drives pair-code requests through its own NavHost. An app-wide
+        // PairCode sheet has no connect back stack to preserve, so replace its start route when the
+        // device requests a new code and force the input state to reset.
+        val currentSheet = _currentSheet.value
+        if (currentSheet is Sheet.Hardware) {
+            if (currentSheet.route is HardwareRoute.PairCode && currentSheet.route.requestId != requestId) {
+                showSheet(Sheet.Hardware(route = HardwareRoute.PairCode(requestId)))
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in 88fe961c. New app-wide PairCode requests update the active Hardware sheet route in place instead of calling showSheet/nulling the sheet, and HardwareSheet navigates to the new PairCode route on route changes. This resets the input without disposing the sheet and cancelling the fresh request.

app/src/main/java/to/bitkit/services/TrezorDebugLog.kt
```diff
@@ -7,20 +7,28 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
 import java.util.Locale
 
 object TrezorDebugLog {
+    private const val TAG = "TrezorDebugLog"
     private const val MAX_LINES = 300
+    private val secretValuePattern = Regex(
+        """(?i)["']?\b(mnemonic|seed|passphrase|pin|pairing[ _-]?code|credential|xpub|""" +
+            """extended[ _-]?key|psbt|raw[ _-]?tx|serialized[ _-]?tx)\b["']?\s*[:=]\s*""" +
+            """("[^"]*"|'[^']*'|[^\s,;}]+)""",
+    )
```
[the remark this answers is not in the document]
piotr-iohk: Fixed in 78eef902. Redaction now treats mnemonic/seed/passphrase as multi-word unquoted values and consumes through a safe delimiter, so diagnostics like `mnemonic=abandon ability able ...` or `passphrase=my hidden wallet` no longer leak the trailing words. Added regression coverage for unquoted multi-word secrets.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -672,11 +750,39 @@ class TransferViewModel @Inject constructor(
         }
     }
 
+    private suspend fun showHardwareBroadcastError(error: HardwareBroadcastError) {
+        if (!error.isHardwareBroadcastConnectivityFailure()) {
+            pendingHwFundingBroadcast = null
+            _spendingUiState.update { it.copy(hasPendingHwBroadcast = false) }
+            ToastEventBus.send(error.cause ?: error)
+            return
+        }
```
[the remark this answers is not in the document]
piotr-iohk: Addressed in 92ac225e. With bitkit-core 0.4.1, an already-known/already-in-mempool rebroadcast is converted by Core into a successful result carrying the canonical txid, so this path does not enter `showHardwareBroadcastError` or clear the retained signed transaction. Added end-to-end ViewModel regression coverage: the first broadcast succeeds but bookkeeping fails, Retry reuses the retained signed transaction without signing again, Core returns the txid on rebroadcast, and bookkeeping completes while clearing pending state.

jvsena42:
E2E Migration tests passed
https://github.com/synonymdev/bitkit-android/actions/runs/29242118065

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
```diff
@@ -444,6 +553,273 @@
...
+        )
+        whenever(hwWalletRepo.ensureConnected(DEVICE_ID))
+            .thenReturn(Result.success(mock<TrezorFeatures>()))
+        whenever(lightningRepo.getFeeRateForSpeed(any(), anyOrNull())).thenReturn(Result.success(FEE_RATE))
+        whenever(hwWalletRepo.composeFundingTransaction(any(), any(), any(), any())).thenReturn(Result.success(funding))
+        whenever(hwWalletRepo.signFunding(DEVICE_ID, funding)).thenReturn(Result.success(signed))
+        whenever(hwWalletRepo.broadcastFunding(signed)).thenReturn(Result.success(broadcast))
+        var bookkeepingAttempts = 0
+        whenever(cacheStore.addPaidOrder(order.id, TXID)).thenAnswer {
+            if (bookkeepingAttempts++ == 0) throw RuntimeException("cache failed")
```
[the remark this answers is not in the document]
jvsena42: no problem, it is just a test

app/src/main/java/to/bitkit/ext/TrezorExceptionExt.kt
```diff
@@ -11,3 +11,10 @@ fun Throwable.isTrezorUserCancellation(): Boolean =
 
 fun Throwable.isTrezorDeviceBusy(): Boolean =
     generateSequence(this) { it.cause }.any { it is TrezorException.DeviceBusy }
+
+fun Throwable.isTrezorLockedOrBusy(): Boolean =
+    isTrezorDeviceBusy() ||
+        generateSequence(this) { it.cause }.any {
+            val message = it.message.orEmpty()
+            "Device error (code 99)" in message && "Firmware error" in message
```
jvsena42: - Extract this error code to a private constant

jvsena42: could open an issue to bitkit-core to check the feasibility to create a typed error DeviceLocked

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -237,16 +241,31 @@ class HwWalletRepo @Inject constructor(
                     trezorRepo.disconnectStaleSession(deviceId)
                 }
             }
-            val txId = trezorRepo.broadcastRawTx(serializedTx = signed.getOrThrow().serializedTx).getOrThrow()
-            HwFundingBroadcastResult(
-                txId = txId,
+            val signedTx = signed.getOrThrow()
```
jvsena42: `signTxFromPsbt` already wraps on runCatching. could simplify this code to something like:

```
            val signed = trezorRepo.signTxFromPsbt(
                    trezorRepo.disconnectStaleSession(deviceId)
                }.getOrThrow()
```

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -965,6 +1089,8 @@ data class TransferToSpendingUiState(
     val quarterAmount: Long = 0,
     val isLoading: Boolean = false,
     val isSigning: Boolean = false,
+    val hasPendingHwBroadcast: Boolean = false,
+    val hwMiningFeeSats: ULong = 0u,
```
jvsena42: looks like dead code, it is not being consumed anywhere except from tests

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -1037,7 +1070,12 @@ class TrezorTransport @Inject constructor(
             TrezorTransportReadResult(success = true, data = data, error = "", errorCode = null)
         } catch (e: Exception) {
             Logger.error("BLE read failed", e, context = TAG)
-            TrezorTransportReadResult(success = false, data = byteArrayOf(), error = e.message ?: "Read failed", errorCode = null)
+            TrezorTransportReadResult(
+                success = false,
+                data = byteArrayOf(),
+                error = e.message ?: "Read failed",
+                errorCode = null
```
jvsena42: add trailing comma

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -875,14 +897,22 @@ class TrezorTransport @Inject constructor(
...
             var bondAttempts = 0
             while (device.bondState != BluetoothDevice.BOND_BONDED && bondAttempts < MAX_BOND_POLL_ATTEMPTS) {
                 Thread.sleep(BOND_POLL_INTERVAL_MS)
                 bondAttempts++
                 if (device.bondState == BluetoothDevice.BOND_NONE) {
-                    return TrezorTransportWriteResult(success = false, error = "Bonding failed or rejected", errorCode = null)
+                    return TrezorTransportWriteResult(
+                        success = false,
+                        error = "Bonding failed or rejected",
+                        errorCode = null
```
jvsena42: add trailing comma

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -875,14 +897,22 @@ class TrezorTransport @Inject constructor(
         if (device.bondState == BluetoothDevice.BOND_NONE) {
             Logger.info("Device not bonded, initiating bonding: '$address'", context = TAG)
             if (!device.createBond()) {
-                return TrezorTransportWriteResult(success = false, error = "Failed to initiate bonding", errorCode = null)
+                return TrezorTransportWriteResult(
+                    success = false,
+                    error = "Failed to initiate bonding",
+                    errorCode = null
```
jvsena42: add trailing comma

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -785,7 +802,12 @@ class TrezorTransport @Inject constructor(
             TrezorTransportReadResult(success = true, data = buffer.copyOf(bytesRead), error = "", errorCode = null)
         } catch (e: Exception) {
             Logger.error("USB read failed", e, context = TAG)
-            TrezorTransportReadResult(success = false, data = byteArrayOf(), error = e.message ?: "Unknown error", errorCode = null)
+            TrezorTransportReadResult(
+                success = false,
+                data = byteArrayOf(),
+                error = e.message ?: "Unknown error",
+                errorCode = null
```
jvsena42: add trailing comma

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -706,7 +719,11 @@ class TrezorTransport @Inject constructor(
             val usbInterface = device.getInterface(0)
             if (!connection.claimInterface(usbInterface, true)) {
                 connection.close()
-                return TrezorTransportWriteResult(success = false, error = "Failed to claim interface", errorCode = null)
+                return TrezorTransportWriteResult(
+                    success = false,
+                    error = "Failed to claim interface",
+                    errorCode = null
```
jvsena42: add trailing comma

jvsena42 (changes requested):
No big issues so far, haven't tested yet

app/src/main/java/to/bitkit/services/TrezorTransport.kt
jvsena42, line 1: multiple `detekt - MaxLineLength` errors, mostly already existent, but could use the opportunity to fix them

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed in 0a7129b0.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Removed in 0a7129b0. `hwMiningFeeSats` was left behind after restoring the original hardware transfer fee display and had no production consumer.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed in 0a7129b0.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed the `TrezorTransport` max-line-length findings in 0a7129b0. The remaining detekt findings are pre-existing and outside this transport change; `just lint` completes successfully with no changed-file max-line violations.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed in 0a7129b0.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed in 0a7129b0.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
piotr-iohk: Simplified in 0a7129b0. `signTxFromPsbt` now uses its existing `Result` directly with `getOrElse`, disconnecting stale sessions for non-user-cancellation failures before rethrowing.

app/src/main/java/to/bitkit/ext/TrezorExceptionExt.kt
piotr-iohk: Addressed in d2e2cbdb after further research. Trezor protocol code 99 is the generic `Failure_FirmwareError`, not a device-locked code. It previously appeared while testing a locked device and was incorrectly interpreted as lock-specific. The code is now extracted as `FIRMWARE_ERROR_CODE`, classified separately from typed `DeviceBusy`, and routes to the Reconnect Hardware Device toast. Typed `TrezorException.DeviceBusy` remains the only unlock-guidance path. I did not open a `DeviceLocked` Core issue from this response because code 99 itself is generic rather than evidence of a distinct lock error.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Fixed in 0a7129b0.

coreyphillips:
## Review notes

Verified locally: `compileDevDebugKotlin` and `testDevDebugUnitTest` both pass, and bitkit-core `0.4.1` resolves. The overall shape is good: separating sign from broadcast, gating retry on an order/device/address identity match, and giving pairing requests an identity are all the right calls. A few things below, one of which I think should block.

### Blocker: Trezor diagnostics now land in the persisted, user-shareable log

`TrezorDebugLog.log()` gained a `Logger.debug(...)` forward (`TrezorDebugLog.kt:40`). `Logger.debug` calls `saver.save(message)` unconditionally (`Logger.kt:146-153`), so these lines now get written to `files/logs/bitkit_*.log` in every flavor including mainnet. That is the file users export and attach to support tickets. Before this PR they only lived in a 300 line in-memory buffer behind the dev screen.

The app's own call sites are fine, they log lengths rather than values. The risk is `TrezorTransport.kt:556`: `logDebug()` is the callback bitkit-core invokes to stream its own Rust debug output, and it is forwarded verbatim as `RUST:$tag`. We do not control that content and it changes with every core release. The new regex only redacts labeled values from a fixed key list, so it misses most of what a transport level log actually emits. Running the exact patterns from this PR against realistic lines:

| Line | Result |
|---|---|
| `psbt=cHNidP8BAHECAAAAAf` | redacted |
| `passphrase=hunter2` | redacted |
| `thp_credential=8f3ac91be77d0011` | **leaked**, the `_` kills the `\b` before `credential` |
| `xpub6CUGRUonZSQ4TWtTMmz...` (bare value, no key) | **leaked** |
| `Sending frame: 3f23230012a4b8cc...` | **leaked** |
| `master_key: 0011223344556677` | **leaked** |
| `{"credential": ["aGVsbG8=", ...]}` | **leaked**, array value is not matched |
| `PIN entered: 1234` | **leaked** |

So "sensitive values redacted" only holds for labeled values whose key names we anticipated. Consumer side regex scrubbing of a stream we do not own is not a security boundary. Suggest either not forwarding `RUST:*` to `Logger` at all (keep it in the in-memory ring buffer), or gating the forward behind a debug/dev flag. Secondary point: transport level Rust logging fires per BLE chunk, so this will also flood the app log.

### Stale session cleanup silently does not run on the paths that need it most

`HwWalletRepo.signFunding` (`HwWalletRepo.kt:233-238`) calls `trezorRepo.disconnectStaleSession(deviceId)` when signing fails. But `disconnectStaleSession` is a `withContext(ioDispatcher)` call (`TrezorRepo.kt:1220`), and on both the `withTimeout(HW_SIGN_TIMEOUT)` path and the new back out cancellation path it is invoked from an already cancelled coroutine, so `withContext` throws immediately without running the body. The disconnect never happens.

This is pre-existing (master has the same hole), but the new `cancelHardwareTransfer()` makes it reachable through a deliberate and common user action, and a dirty session is exactly what produces the "device busy" state this PR is trying to fix. `withContext(NonCancellable) { ... }` would fix it.

Related: `TrezorRepo.signTxFromPsbt` (`:509`) and `broadcastRawTx` (`:523`) use `runCatching` rather than `runSuspendCatching` around suspend calls. So every user back out now logs the `CancellationException` as `Logger.error("Trezor signTxFromPsbt failed")`, which is both log noise and against the repo rules.

### Backing out during an in-flight broadcast completes the transfer with no confirmation

`cancelHardwareTransfer()` deliberately early returns when a signed tx is pending (`TransferViewModel.kt:515`), so the job keeps running. If the broadcast then succeeds, `fundPaidOrder` runs and the order is funded, but `setTransferEffect(TransferEffect.OnHwTxSigned)` is emitted into a `MutableSharedFlow()` with replay 0 (`:98`) whose only collector was the now disposed sign screen. The event is dropped and the user is never navigated. The funds move and the UI never says so.

### "Fails pending BLE operations promptly when Bluetooth turns off" is only partly true

`releasePendingBleOperations` covers the connection, write, and disconnect latches, and the wake up handling is correct (the connect path re-checks `isConnected` at `:994`, the write path re-checks `writeStatus` at `:1159`). But a pending read blocks on `readQueue.poll(BLE_READ_TIMEOUT_MS)` (`:1073`) and is not released, so it still waits out the full 5s.

### Minor

- `DisposableEffect(viewModel)` in `SpendingHwSignScreen.kt:63` also fires on activity recreation (locale, dark mode, font scale; rotation is safe since the activity is portrait locked), cancelling an in-flight signing job. Low probability, but worth knowing.
- `composeHardwareFundingTransaction` picked up a pointless `val funding = ...; return funding` local.
- The `.cursor/rules/rules.main.mdc` strings policy relaxation is fine on the merits, but it is an unrelated repo-wide policy change riding along in a Trezor fix PR. Worth splitting out.

### What should move to bitkit-core

The best commit here is `ccb3c1b`, which deleted `ALREADY_BROADCAST_MARKERS` and moved already-broadcast txid recovery into core `0.4.1`. That is exactly the right instinct, and three more instances of the same pattern are still in the app:

1. **`isTrezorFirmwareError()`** (`TrezorExceptionExt.kt:19-23`) string sniffs `"Device error (code 99)"` and `"Firmware error"` out of a message, with a hardcoded protocol constant. Core already exports a typed `TrezorException` hierarchy (`UserCancelled`, `DeviceBusy`, `Timeout`, `DeviceDisconnected`, `SessionException`, and others), so this should be a `TrezorException.FirmwareError` variant.
2. **`"Failed to connect to Electrum" in it.message`** (`TransferViewModel.kt:775`) is the load bearing one. It decides whether a signed transaction is retained for retry or thrown away. If core ever rewords that string, the app silently drops a signed tx and shows a generic error toast instead of offering Retry. That is a signed transaction loss bug waiting on a copy edit in another repo. Core should return a typed network/broadcast failure.
3. **Locked device detection** (`TrezorRepo.kt:743-752`) reimplements a device state rule app side (`pinProtection == true && unlocked == false`) and then synthesizes a semantically wrong `TrezorException.DeviceBusy()` to carry it. Core should expose `TrezorException.DeviceLocked`. Today `hardware__device_busy` ("Your Trezor is busy. Unlock it on the device...") is doing double duty for two distinct states.

The redaction itself also belongs in core: core should not emit secrets into its debug stream at all, rather than the app trying to scrub them after the fact.

Deferring durable signed tx persistence across process death seems like a reasonable v1 non-goal, though core, which owns the Electrum client, is arguably its natural home.

piotr-iohk:
@coreyphillips Addressed in 2ab41b00e.

**Blocker — Trezor diagnostics in persisted log:** Removed the `Logger.debug` forward from `TrezorDebugLog`. Rust/Kotlin transport diagnostics stay in the in-memory dev buffer only. Verified on device export — no `TrezorDebugLog` / `RUST:*` lines in `bitkit_*.log`.

**Stale session cleanup:** `disconnectStaleSession` now runs in `NonCancellable`. `cancelHardwareTransfer()` also triggers disconnect when backing out during signing.

**Cancellation logging:** `signTxFromPsbt` / `broadcastRawTx` use `runSuspendCatching`.

**Back-out during in-flight broadcast:** Added `hwFundingComplete` state; `RootNavHost` navigates to the signed screen even if the sign composable was disposed. Device test confirmed retry-after-DNS-failure → paid order → SettingUp.

**BLE read on Bluetooth off:** `readBleChunk` polls in 100ms intervals and fails fast when `isConnected` is false, since `releasePendingBleOperations` does not unblock queue-based reads.

**Electrum connectivity classification:** Replaced `"Failed to connect to Electrum"` string match with `BroadcastException.ElectrumException` (core 0.4.1 / #123).

Agree on core follow-ups for typed `FirmwareError` / `DeviceLocked` and source-level log redaction — will track separately.

jvsena42 (approved):
Approved with some OBS:

> 3. Locked Trezor → Hardware Transfer Sign → Open Trezor Connect: informational unlock guidance appears instead of a raw firmware/timeout error.

When signing with trezor locked, the app woke up the trezor instead of displaying the info toast (ok for me)

https://github.com/user-attachments/assets/579567af-3b42-4341-87ae-177368e8c5c5

When setting the wrong PIN, there was no error message, the trezor just requested the pin again ( can be improved in the next PR)

https://github.com/user-attachments/assets/2993125c-dfec-4a84-ab52-86f13787dc5c

coreyphillips:
## Follow up on 2ab41b00e

Thanks, this covers most of it. The blocker is properly resolved: the `Logger.debug` forward is gone, so transport diagnostics stay in the in-memory dev buffer and never reach `bitkit_*.log`. The sanitize regex remaining is fine as defense in depth now that it is no longer the only thing between core output and a user shareable file. Replacing the Electrum string match with the typed `BroadcastException.ElectrumException` is exactly right, and that was the load bearing one. `runSuspendCatching`, the `NonCancellable` nesting, and the BLE read poll all look correct. CI is green.

One item is not actually fixed, and the fixes introduced three new things.

### The sign timeout path still leaves the session dirty

`NonCancellable` does not rescue the 120s `HW_SIGN_TIMEOUT` path.

`HwWalletRepo.signFunding` only calls `disconnectStaleSession` inside `.getOrElse { ... }` on the result of `signTxFromPsbt`. But `signTxFromPsbt` now uses `runSuspendCatching`, which rethrows `CancellationException` rather than capturing it into a `Result.failure`. `TimeoutCancellationException` is a `CancellationException`, so a timeout from the outer `withTimeout(HW_SIGN_TIMEOUT)` propagates straight past `getOrElse` and the disconnect branch never executes. The `HardwareSigningTimeoutError` branch in `handleHardwareTransferFailure` only calls `showHardwareTimeoutError()`, with no disconnect either.

Slightly ironic: the `runSuspendCatching` change, which is correct on its own terms, removed the only path by which cancellation used to reach that `getOrElse`. Previously it reached it but the disconnect was a silent no-op on the cancelled coroutine. Now it does not reach it at all. So `NonCancellable` only ends up mattering for the new explicit `cancelHardwareTransfer()` call, which runs in a fresh uncancelled `viewModelScope` coroutine and did not need it.

Net: backing out is fixed, timing out after 120s is not. Adding `disconnectStaleSession(deviceId)` to the `HardwareSigningTimeoutError` branch would close it.

### RootNavHost now recomposes on every spending state change

`ContentView.kt:649` collects the whole `spendingUiState` at the app root just to watch one boolean, so every `isLoading`, `feeEstimate`, `order`, and amount entry update now recomposes the root NavHost.

Collecting a derived `map { it.hwFundingComplete }.distinctUntilChanged()` would fix it, but the cleaner option is to collect the existing `transferEffects` SharedFlow in `RootNavHost` instead of in the screen. That also solves the next point.

### TransferEffect.OnHwTxSigned is now dead

Still emitted at `TransferViewModel:603`, but the only remaining reference is a no-op `-> Unit` branch in `SpendingAdvancedScreen.kt:96`. Nothing acts on it since the sign screen's collector was removed.

### System.currentTimeMillis() in the new BLE read loop

Two calls in `readBleChunk`. Repo rules require the `ext/DateTime.kt` helpers (`nowMillis()` / `Clock.nowMs()`) so the code stays unit testable. Detekt did not catch it.

### Worth a conscious call

The new `LaunchedEffect` navigates to `Routes.SpendingHwSigned` on `hwFundingComplete` regardless of the current screen. That is the intended fix for the silent completion case and is defensible since the user initiated the transfer, but if they backed out to Home or moved into another flow, a background broadcast completing will now pull them out of it. Fine if that is deliberate, just flagging it.

### Minor leftovers

- `isHardwareBroadcastConnectivityFailure()` is now a redundant private wrapper that just delegates to `isBroadcastConnectivityFailure()`.
- The `DisposableEffect` in `SpendingHwSignScreen` still fires on activity recreation (locale, dark mode, font scale), and now also disconnects the device session rather than only cancelling the job. Low probability, slightly higher impact than before.

Agreed on tracking typed `FirmwareError` / `DeviceLocked` and source level redaction as core follow ups.

jvsena42:
> NonCancellable does not rescue the 120s HW_SIGN_TIMEOUT path.

120s is already a reasonable timing for `HW_SIGN_TIMEOUT`

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -672,14 +770,44 @@ class TransferViewModel @Inject constructor(
...
+
+    private suspend fun showHardwareConnectivityError() {
+        ToastEventBus.send(
+            type = Toast.ToastType.WARNING,
+            title = context.getString(R.string.other__connection_issue),
+            description = context.getString(R.string.other__connection_issues_explain),
+        )
+    }
+
+    private fun Throwable.isHardwareBroadcastConnectivityFailure(): Boolean =
```
jvsena42: just use isBroadcastConnectivityFailure

jvsena42:
> Stale session cleanup silently does not run on the paths that need it most
> 
> HwWalletRepo.signFunding (HwWalletRepo.kt:233-238) calls trezorRepo.disconnectStaleSession(deviceId) when signing fails. But disconnectStaleSession is a withContext(ioDispatcher) call (TrezorRepo.kt:1220), and on both the withTimeout(HW_SIGN_TIMEOUT) path and the new back out cancellation path it is invoked from an already cancelled coroutine, so withContext throws immediately without running the body. The disconnect never happens.

The `withTimeout(HW_SIGN_TIMEOUT)` lives in `TransferViewModel.signHardwareFundin`  and the wrapper there already disconnects:

```
        }.getOrElse {
            if (it is CancellationException && it !is TimeoutCancellationException) throw it
            if (it is TimeoutCancellationException) {
                hwWalletRepo.disconnectStaleSession(deviceId)
                throw HardwareSigningTimeoutError(it)
            }
            throw it
        }
```

`withTimeout` cancels only its child scope, so runCatching in the uncancelled caller catches the TimeoutCancellationException and the disconnect runs

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -646,6 +646,13 @@ private fun RootNavHost(
     onHomeCalculatorInputActiveChanged: (Boolean) -> Unit,
 ) {
     val scope = rememberCoroutineScope()
+    val spendingState by transferViewModel.spendingUiState.collectAsStateWithLifecycle()
+    LaunchedEffect(spendingState.hwFundingComplete) {
+        if (spendingState.hwFundingComplete) {
+            transferViewModel.consumeHwFundingComplete()
+            navController.navigateTo(Routes.SpendingHwSigned)
+        }
+    }
```
jvsena42: replace with

```
 LaunchedEffect(Unit) {                                                                                                                                                                                       
      transferViewModel.transferEffects.collect { effect ->                                                                                                                                                    
          if (effect is TransferEffect.OnHwTxSigned) {                                                                                                                                                         
              navController.navigateTo(Routes.SpendingHwSigned)                                                                                                                                                
          }                                                                                                                                                                                                    
      }                                                                                                                                                                                                        
  }  
```

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -965,6 +1111,8 @@ data class TransferToSpendingUiState(
     val quarterAmount: Long = 0,
     val isLoading: Boolean = false,
     val isSigning: Boolean = false,
+    val hasPendingHwBroadcast: Boolean = false,
+    val hwFundingComplete: Boolean = false,
```
jvsena42: can be deleted after https://github.com/synonymdev/bitkit-android/pull/1067/changes#r3572928500

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -484,16 +497,41 @@ class TransferViewModel @Inject constructor(
...
+        hwTransferSignJob = null
+        _spendingUiState.update { it.copy(isSigning = false) }
+        if (deviceId != null) {
+            viewModelScope.launch {
+                hwWalletRepo.disconnectStaleSession(deviceId)
+            }
+        }
+    }
+
+    fun consumeHwFundingComplete() {
```
jvsena42: can be deleted after delete https://github.com/synonymdev/bitkit-android/pull/1067/changes#r3572937980

app/src/main/java/to/bitkit/services/TrezorTransport.kt
```diff
@@ -1025,19 +1071,39 @@ class TrezorTransport @Inject constructor(
...
-                ?: return TrezorTransportReadResult(
-                    success = false,
-                    data = byteArrayOf(),
-                    error = "Read timeout",
-                    errorCode = null,
-                )
-
-            Logger.debug("BLE read ${data.size} bytes from '$path'", context = TAG)
-            TrezorTransportReadResult(success = true, data = data, error = "", errorCode = null)
+            val deadlineMs = System.currentTimeMillis() + BLE_READ_TIMEOUT_MS
```
jvsena42: inject `private val clock: Clock,` and use `clock.nowMs()` instead

jvsena42 (changes requested):
findings after double-check https://github.com/synonymdev/bitkit-android/pull/1067#issuecomment-4960692642 review

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `288147c0f` — removed the redundant `isHardwareBroadcastConnectivityFailure()` wrapper.

app/src/main/java/to/bitkit/ui/ContentView.kt
piotr-iohk: Done in `288147c0f` — `RootNavHost` now collects `transferEffects` exactly as suggested. `OnHwTxSigned` is emitted after `fundPaidOrder` succeeds.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `288147c0f` — `hwFundingComplete` removed from `TransferToSpendingUiState`.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `288147c0f` — `consumeHwFundingComplete()` removed.

app/src/main/java/to/bitkit/services/TrezorTransport.kt
piotr-iohk: Done in `288147c0f` — `TrezorTransport` ctor takes `Clock`; `readBleChunk` uses `clock.nowMs()` for the poll deadline. `TrezorTransportTest.kt` updated.

piotr-iohk:
@coreyphillips Addressed in `288147c0f` and `9e2ce151c`.

### Sign timeout path / stale session

Agree with @jvsena42 ([comment](https://github.com/synonymdev/bitkit-android/pull/1067#issuecomment-4960929769)): the disconnect is not in `HwWalletRepo.signFunding().getOrElse` — it lives in `TransferViewModel.signHardwareFunding()`. `withTimeout(HW_SIGN_TIMEOUT)` cancels only its child; the caller's `runCatching { ... }.getOrElse` runs in the uncancelled scope and handles `TimeoutCancellationException` before rethrowing `HardwareSigningTimeoutError`:

```kotlin
if (it is TimeoutCancellationException) {
    hwWalletRepo.disconnectStaleSession(deviceId)
    throw HardwareSigningTimeoutError(it)
}
```

`runSuspendCatching` inside `signTxFromPsbt` rethrowing `CancellationException` does not bypass this — timeout is handled one level up.

`9e2ce151c` adds `onTransferToSpendingHwConfirm disconnects when HW_SIGN_TIMEOUT elapses` (virtual time via `StandardTestDispatcher`) to lock that behavior.

### RootNavHost recompose + dead `OnHwTxSigned`

Fixed in `288147c0f` per @jvsena42's suggestion ([#r3572928500](https://github.com/synonymdev/bitkit-android/pull/1067#discussion_r3572928500)):

- Removed `hwFundingComplete` from `spendingUiState` and `consumeHwFundingComplete()` ([#r3572937980](https://github.com/synonymdev/bitkit-android/pull/1067#discussion_r3572937980), [#r3572940732](https://github.com/synonymdev/bitkit-android/pull/1067#discussion_r3572940732)).
- `RootNavHost` collects `transferViewModel.transferEffects` and navigates on `TransferEffect.OnHwTxSigned` only — no whole-`spendingUiState` collection at root.
- `OnHwTxSigned` is emitted after `fundPaidOrder` succeeds (broadcast + bookkeeping), not on sign alone. Removed the no-op branch from `SpendingAdvancedScreen`.

### `System.currentTimeMillis()` in BLE read loop

Fixed in `288147c0f` ([#r3572952417](https://github.com/synonymdev/bitkit-android/pull/1067#discussion_r3572952417)): `TrezorTransport` takes injected `Clock`; `readBleChunk` uses `clock.nowMs()`.

### Background navigation (`hwFundingComplete` / silent completion)

Clarifying observed behaviour vs the edge case this fixes:

**Normal paths (device-tested):**

- **Back / navigate away during signing** — `SpendingHwSignScreen` `DisposableEffect` calls `cancelHardwareTransfer()`, which cancels the job and disconnects. Signing is lost.
- **Phone locks while user is signing on Trezor** — after unlocking the phone, user stays on `SpendingHwSignScreen`. If signing completed while locked, `hasPendingHwBroadcast` flips the same screen to signed state + Retry — no navigation away.

**What `OnHwTxSigned` + root collector is for:**

After sign completes, `pendingHwFundingBroadcast` is set and `cancelHardwareTransfer()` no longer cancels the job. If the user leaves while **broadcast/bookkeeping** is still in flight, the coroutine can finish, call `fundPaidOrder`, and emit `OnHwTxSigned`. Root `NavHost` collects that so we navigate to `SpendingHwSigned` even if the sign composable was disposed — fixing the silent-completion bug from your first review.

This is not "navigate when signing completes"; signing completion stays in-place on the sign screen via `hasPendingHwBroadcast`. Root navigation fires only after the order is funded in-app. Conscious v1 trade-off for the narrow post-sign broadcast edge case.

### Minor leftovers

- **`isHardwareBroadcastConnectivityFailure()`** — removed in `288147c0f`; call sites use `isBroadcastConnectivityFailure()` directly ([#r3572840857](https://github.com/synonymdev/bitkit-android/pull/1067#discussion_r3572840857)).
- **`DisposableEffect` on activity recreation** — acknowledged; low probability. Not changing in this PR.

Core follow-ups (`FirmwareError` / `DeviceLocked`, source-level redaction) still tracked separately.

jvsena42 (approved):
Approved. tested:

 1. Connect Hardware → enter an incorrect six-digit pairing code → confirm the new code on Trezor: input resets and accepts the replacement code.
 2. regression: Connect Hardware → turn Bluetooth off during connect → turn it back on: operation exits without an indefinite wait and reconnect can proceed.
 3. Locked Trezor → Hardware Transfer Sign → Open Trezor Connect: informational unlock guidance appears instead of a raw firmware/timeout error.
 4. Hardware Transfer Sign → approve on Trezor while network broadcast fails → Retry: “Transaction signed” remains visible and broadcast retries without signing again.
 5. Hardware Transfer Sign → start signing → back: later approval on Trezor does not broadcast or advance the order.
 6. Settings → General with no paired hardware wallet: Hardware Wallets does not show a 0 value.

Could not find how to reproduce test 7, maybe some something in my phone

Found some possible refactors in TransferViewmodelTest, I can do it on the next PR

PR #1075, opened 2026-07-14
fix: match usb icon with design

jvsena42:
This PR redraws the USB connection icon so the hardware wallet connection badge matches the design.

### Description

The USB badge shown next to a paired hardware wallet was still using an oversized 32dp glyph built from a different icon set: a plug-and-sparks mark that read as a broken link rather than the USB symbol in the design, and that sat visually heavier than the Bluetooth badge it appears alongside.

### Preview

[Screen_recording_20260714_115704.webm](https://github.com/user-attachments/assets/9a9fc331-2c6e-48cf-aebd-12e993f2e7bd)

#### Manual Tests

- [x] **1a.** Paired USB hardware wallet, connected → wallet home hardware cell: badge shows the USB glyph from the design in green, at the same visual weight as the Bluetooth badge.
  - [x] **1b.** Same wallet, disconnected: the same glyph shows in gray.
- [x] **2.** Settings → General → Hardware Wallets → USB device row: badge shows the same glyph and connection colour as on the wallet home.
- [x] **3.** `regression:` Bluetooth-paired hardware wallet → wallet home and Hardware Wallets settings: Bluetooth badge is unchanged in shape, size, and connection colour.

#### Automated Checks

- No automated coverage changed: the change is a single vector drawable, and no unit or E2E test asserts on this icon.
- `just compile` and `just lint` passed locally.


piotr-iohk (dismissed):

ovitrif:
A note on the long reasoning fail of AI: actually the glyphs should use the "main component" from the design frames, which IIRC if followed through by clicking figma's "go to main component" option in the context menu on that component instance, will lead to the grid of 32x32 (or 16x16) glyphs from Synonym Branding Figma file. Doesn't mean changes are needed to this PR though, although ideally all android icons would be relying of that grid which is the actual source icons, that is also the only reliable way to end up with uniform icons in the app's resources.
The accurate reasoning is much simpler: initial Figma designs matched by 1st implementation used other icon, it had to be updated to match the new design.

jvsena42:
> A note on the long reasoning fail of AI: actually the glyphs should use the "main component" from the design frames, which IIRC if followed through by clicking figma's "go to main component" option in the context menu on that component instance, will lead to the grid of 32x32 (or 16x16) glyphs from Synonym Branding Figma file. Doesn't mean changes are needed to this PR though, although ideally all android icons would be relying of that grid which is the actual source icons, that is also the only reliable way to end up with uniform icons in the app's resources. The accurate reasoning is much simpler: initial Figma designs matched by 1st implementation used other icon, it had to be updated to match the new design.

Re-imported for using root component. video updated

piotr-iohk (approved):

PR #1076, opened 2026-07-14
Release 2.3.2

pwltr:
The purpose of this PR is only to run the e2e suite.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.iml
 .gradle
 .idea
+.vscode
 !.idea/codeStyles/
 .kotlin
 .DS_Store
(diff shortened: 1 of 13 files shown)


piotr-iohk:
Re-Running from dispatch against release-2.3.2 branch: https://github.com/synonymdev/bitkit-android/actions/runs/29349079016/job/87140403130

The failures here are mostly caused by the fact that CI uses the workflow from the merge preview (release-2.3.2 into master), not from release-2.3.2 alone and there were changes to the e2e workflow in the meantime (e.g. hardware shard added) which spills into here as well.

piotr-iohk:
Dispatch passed: https://github.com/synonymdev/bitkit-android/actions/runs/29349079016/job/87142127256
(the single `@paykit` failure may be ignored, it is due to old fixture pubky used in e2e staging that was fixed and changed in paykit refactor PR that is in 2.4.0)

PR #1081, opened 2026-07-15
feat: add boltz savings swap

coreyphillips:
This PR adds a swap-based path for transferring spending balance to savings, so Lightning funds can move on-chain through a Boltz reverse swap instead of only by closing a channel.

### Description

- Adds a Boltz reverse swap flow to the transfer-to-savings journey. The confirm screen now shows a live quote (amount to savings, network fee, service fee, and the amount you will receive) and lets the user proceed with a swap or fall back to closing the channel.
- The whole swap flow is off by default and switched on under Dev Settings -> Swaps -> Enable Savings Swap, since the confirm screen changes are not in design yet. With the flag off the journey is byte for byte the pre-PR one.
- Swaps are offered on mainnet only. Boltz's testnet deployment is deprecated and regtest expects a local backend, so on every other network the journey behaves exactly as it did before this PR: the swipe closes the channel, with no quote, fees, slider, or extra close action.
- Closing a channel is the default and a priced quote is the only thing that upgrades a transfer to a swap. Without one, for any reason (unsupported network, Boltz unreachable, or an amount below the swap minimum), the swipe closes the channel rather than doing nothing.
- Waits for the on-chain claim during the progress step and resumes any pending swaps when the wallet starts, refreshing balances automatically once a swap lands on-chain. The updates stream now also stops with the node when the app goes to background, and only starts where swaps are supported.
- Adds a Swaps section under Dev Settings to list swaps and claim a reverse swap, backed by a new service that wraps the Boltz module. The manual claim button only shows while the swap is actually claimable (lockup on-chain and no claim broadcast yet).
- Adds a reusable slider component used by the swap amount controls.
- Bumps bitkit-core to 0.5.2, which introduces the Boltz module this flow depends on.

#### Manual Tests (swap path is mainnet only and needs Dev Settings -> Swaps -> Enable Savings Swap)
- [ ] **1.** With a spending balance, Transfer to Savings → Savings Confirm: quote shows amount to savings, network fee, service fee, and you'll receive.
- [ ] **2.** Savings Confirm → change amount: quote recomputes for the new amount.
- [ ] **3.** Savings Confirm → set an amount below the swap minimum: no quote or fees show and the swipe closes the channel instead.
- [ ] **4a.** Savings Confirm → confirm swap → Savings Progress: completes after the on-chain claim.
  - [ ] **4b.** After claim: savings balance reflects the swapped funds without a manual sync.
- [ ] **5.** `regression:` Savings Confirm → Close channel instead: falls back to the channel-close path.
- [ ] **6.** `regression:` Start the app with a pending swap: swap resumes and balance updates once it lands on-chain.
- [ ] **7a.** Dev Settings → Swaps → tap a swap → Swap Detail: Claim now shows only for an unclaimed reverse swap with its lockup in mempool or confirmed.
  - [ ] **7b.** Swap Detail of a created, expired, refunded, or claimed swap: no Claim button.
- [ ] **8.** `regression:` On a regtest or testnet build, Transfer to Savings: confirm screen shows no quote, fees, slider, or "close channel instead", and the swipe closes the channel as before.
- [ ] **8b.** `regression:` With Enable Savings Swap off on a mainnet build, Transfer to Savings: same pre-PR confirm screen and the swipe closes the channel.
- [ ] **9.** `regression:` After a swap fails, start a second transfer to savings in the same session: the second attempt runs normally instead of bouncing back to the transfer root.
#### Automated Checks
- Unit tests added: savings swap quote limits, slider clamping, swap execution outcomes (claimed, error, timeout, payment failure), that a second start is ignored while a swap is in flight, that the quote is skipped when swaps are unsupported or disabled in dev settings, that a failed limits fetch resolves to the close path, and that confirming clears a previous swap outcome, in `TransferViewModelTest.kt`.
- Unit tests added: claimable status gating, swap list sorting, and manual claim in `SwapsViewModelTest.kt`.
- Unit tests added: the swap updates stream starts only where swaps are supported and enabled, in `WalletViewModelTest.kt`.
- Ran locally against bitkit-core 0.5.2: `compileDevDebugKotlin`, `testDevDebugUnitTest`, and `detekt` all pass.


app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -921,6 +926,150 @@ class TransferViewModel @Inject constructor(
...
+            Logger.info("Created savings transfer swap ${swap.id}", context = TAG)
+
+            coroutineScope {
+                // Subscribe before paying so a claim settling faster than the payment
+                // call returns cannot be missed (the events flow has no replay).
+                val claim = async { awaitSwapClaim(swap.id) }
+
+                // Pay the hold invoice (amount is encoded). It stays pending until Boltz
+                // locks funds on-chain and we claim them, which is the expected happy path.
+                lightningRepo.payInvoice(bolt11 = swap.invoice).getOrThrow()
```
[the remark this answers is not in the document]
coreyphillips: Good catch, this was a real race. The comment claimed we subscribed before paying, but `async` with the default start only schedules the collector, so `payInvoice()` could run first and a fast `Claimed`/`Error` would be dropped by the non-replaying flow.

Fixed in 41c6c85 by starting the collector with `CoroutineStart.UNDISPATCHED`, which runs the body inline until it suspends inside the `SharedFlow` subscribe, so the subscription is live before the payment call.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -319,6 +325,36 @@ class WalletViewModel @Inject constructor(
...
+     */
+    private suspend fun startSwapUpdates() {
+        runSuspendCatching {
+            val speed = settingsStore.data.first().defaultTransactionSpeed
+            val feeRate = lightningRepo.getFeeRateForSpeed(speed).getOrNull()?.toDouble()
+            boltzService.startUpdates(feeRateSatPerVb = feeRate)
+        }.onFailure {
+            Logger.error("Failed to start swap updates", it, context = TAG)
+        }
+    }
```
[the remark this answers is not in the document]
coreyphillips: Agreed. Fixed in 41c6c85: `startSwapUpdates()` now retries up to 3 times with a linear backoff (5s base) before giving up and logging an error. Without the stream a paid swap has nothing to broadcast its claim, so a single failed attempt was too fragile.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -921,6 +926,150 @@ class TransferViewModel @Inject constructor(
...
+                reverseLimits = null
+                _savingsSwapState.update { it.copy(isLoading = false, quote = null, error = e.message) }
+                return@launch
+            }
+            reverseLimits = limits
+
+            // Reserve headroom for Lightning routing fees. Paying an invoice for 100% of
+            // outbound capacity leaves nothing for fees and fails with RouteNotFound, so cap
+            // the swap at outbound minus ~1% (with a small floor).
+            val spendable = walletRepo.balanceState.value.maxSendLightningSats.toLong()
```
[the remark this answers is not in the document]
coreyphillips: Agreed. Fixed in 41c6c85 by tracking the quote job and cancelling any in-flight request at the start of `loadSavingsSwapQuote()`, so a slower earlier request can no longer overwrite `pendingSwapAmountSat`, the limits, or the displayed quote.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -319,6 +325,36 @@ class WalletViewModel @Inject constructor(
...
+    private fun collectSwapEventsOnce() {
+        if (swapEventsCollected) return
+        swapEventsCollected = true
+        viewModelScope.launch {
+            boltzService.events.collect { event ->
+                if (event is BoltzSwapEvent.Claimed) {
+                    Logger.info("Savings swap claimed: ${event.swapId}", context = TAG)
+                    walletRepo.syncBalances()
+                }
+            }
```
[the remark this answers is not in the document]
coreyphillips: Same class of bug as the `TransferViewModel` one. Fixed in 41c6c85: `collectSwapEventsOnce()` now launches with `CoroutineStart.UNDISPATCHED` so it subscribes before `startSwapUpdates()` runs, and a swap claimed early can no longer be dropped before `syncBalances()`.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsProgressScreen.kt
```diff
@@ -55,43 +57,39 @@ fun SavingsProgressScreen(
     val context = LocalContext.current
     var progressState by remember { mutableStateOf(SavingsProgressState.PROGRESS) }
 
-    // Effect to close channels & update UI
+    // Effect to execute the transfer & update UI
     // TODO move this logic to viewmodel so it can outlive the screen lifecycle
     LaunchedEffect(Unit) {
-        val channelsFailedToCoopClose = transfer.closeSelectedChannels()
+        when (transfer.savingsTransferMode.value) {
```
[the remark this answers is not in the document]
coreyphillips: Agreed, and the TODO here said as much. Fixed in 41c6c85: the create-pay-wait operation moved into `TransferViewModel.startSavingsSwap()`, which runs in `viewModelScope` so it survives leaving the progress screen. The outcome is published on a new `savingsSwapResult` flow the screen collects. `startSavingsSwap()` is also a no-op while a swap is in flight, so re-entering the journey cannot create and pay a second swap. Covered by a new test in `TransferViewModelTest.kt`.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsProgressScreen.kt
```diff
@@ -102,6 +100,59 @@ fun SavingsProgressScreen(
...
+    onSuccess: () -> Unit,
+    onFailure: (String) -> Unit,
+) {
+    when (val result = transfer.executeSavingsSwap()) {
+        is SavingsSwapResult.Success, SavingsSwapResult.Pending -> {
+            wallet.refreshState()
+            onSuccess()
+        }
+
+        is SavingsSwapResult.Failure -> onFailure(result.reason)
```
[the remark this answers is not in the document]
coreyphillips: Leaving this as-is intentionally. By the time we reach the timeout the hold invoice is paid and the swap is committed, and the updates stream auto-broadcasts the claim once the lockup confirms, so `Pending` means "settling", not "may not happen". Treating it as a failure would be the more misleading outcome since the funds have already left the spending balance.

The related risks you raised are what actually needed fixing, and both are addressed in 41c6c85: the subscription race that made `Pending` far more likely than it should be, and the un-retried updates stream that could leave the claim unmanaged. I have also expanded the comment on the screen to explain why pending counts as success.

app/src/main/java/to/bitkit/services/BoltzService.kt
```diff
@@ -0,0 +1,201 @@
...
+ * Mirrors the existing service pattern (e.g. [TrezorService]): a Hilt singleton
+ * that wraps the FFI and bridges the [BoltzEventListener] foreign callback to a
+ * [SharedFlow]. bitkit-core persists only a derivation index, never key material;
+ * swap keys are re-derived on demand from the wallet mnemonic.
+ *
+ * The Lightning side (issuing/paying invoices, fresh onchain addresses) is owned
+ * by [LightningService]; this service only talks to Boltz + the chain.
+ */
+@Singleton
+class BoltzService @Inject constructor(
```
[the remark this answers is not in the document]
coreyphillips: This was already addressed before this comment landed: `BoltzService` carries a `@Suppress("TooManyFunctions")` annotation with a class doc explaining why the FFI wrapper legitimately mirrors the surface of the underlying Boltz module.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsProgressScreen.kt
```diff
@@ -102,6 +100,59 @@
...
+/** Legacy path: cooperatively close the selected channel(s), retrying on failure. */
+@Suppress("MagicNumber")
+private suspend fun runChannelClose(
+    transfer: TransferViewModel,
+    wallet: WalletViewModel,
+    onSuccess: () -> Unit,
+    onInterrupted: () -> Unit,
+    onGiveUp: () -> Unit,
+    onUnavailable: () -> Unit,
+) {
```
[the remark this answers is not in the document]
coreyphillips: Already addressed: `runChannelClose` carries `@Suppress("MagicNumber", "LongParameterList")`. The parameters are the distinct terminal outcomes of the channel-close path, and collapsing them into a wrapper type would not make the call site clearer.

jvsena42 (changes requested):
1. The E2E is failing because of this error. Maybe the CI need to be updated

Image

2. Is this UI change in design? otherwise it should be gated under dev settings flag

coreyphillips:
### Fixed the failing E2E shards

Two shards were failing deterministically on every run of this branch, and both were caused by this PR:

- `e2e-tests-local - lnurl_transfer` → `@transfer_2 - Can open a channel to external node` (`transfer.e2e.ts:358`)
- `e2e-tests-staging - multi_address_2_regtest` → inside `transferSpendingToSavings` (`actions.ts:1036`)

Both perform the same steps and then time out waiting for `TransferSuccess` / `TransferSuccess-button`. Both shards build `assembleDevDebug`, so both run `BoltzNetwork.REGTEST`.

**Root cause.** bitkit-core resolves the Boltz regtest API to `http://localhost:9001`, which inside the emulator is the emulator's own loopback. The device logs from the failing run show it directly:

```
ERROR [TransferViewModel.kt:968] Failed to load reverse swap limits
[ApiException='errorDetails=Failed to fetch reverse pairs:
error sending request for url (http://localhost:9001/v2/swap/reverse)']
```

That left `savingsSwapState` as `{quote = null, error = <msg>, amountTooLow = false}`, and the confirm swipe guard was:

```kotlin
if (!amountTooLow && quote == null) return@SwipeToConfirm
```

so the swipe became a silent no-op. The failure screenshots confirm it: the GRAB circle is latched at the far right, the "close channel instead" button is visible, and there is no coop-close attempt anywhere in the log after the swipe. On `master` the swipe always navigated, so this was a regression from this branch, not a flake. Both shards are green on every other branch in the sampled window.

**Why not fix it in the E2E repo.** These specs assert post-transfer state only a channel close can produce: `expectSpendingBalance(0)` (`actions.ts:1045`) and `expectText('Connection 1', {visible: false})` (`transfer.e2e.ts:365`). A reverse swap keeps the channel open and always leaves the routing-fee reserve behind, so it cannot satisfy them on any network. `bitkit-e2e-tests` is untouched.

**Fix**, in 4fea1ad:

1. `Env.isSwapSupported` gates swaps to mainnet. Boltz's testnet deployment is deprecated (`api.testnet.boltz.exchange` currently returns 502) and regtest expects a local backend, so mainnet is the only network with a reachable API. Off mainnet the quote is never requested and the flow behaves exactly as it did before this PR.
2. The swipe can no longer be inert. `CLOSE` is now the default mode and a quote is the single thing that upgrades it to `SWAP`, so a missing quote for any reason (unsupported network, Boltz down, amount below the swap minimum) closes the channel instead of doing nothing. This was a real dead-end on mainnet too, not just in CI.
3. The limits fetch is bounded by a 15s timeout, so a hanging request cannot leave the swipe stuck in its loading state.
4. `ensureSwapUpdatesRunning()` is gated the same way, so the updates stream no longer retries against an unreachable backend off mainnet.
5. `onTransferToSavingsConfirm` now clears `savingsSwapResult`. Without this, a stale `Failure` from an earlier attempt would fire `onTransferUnavailable()` on the next transfer and bounce the user out mid-close. This became reachable once a failed swap can fall back to close in the same session.

The raw `error` and `amountTooLow` fields are gone from `SavingsSwapUiState`; `quote != null` is now the only predicate, and the unlocalized FFI error string is no longer rendered on screen (it is still logged).

Tests added in `TransferViewModelTest.kt`: the quote is skipped entirely when swaps are unsupported, a failed limits fetch resolves to `CLOSE`, mode resolution for both swipe and "close instead", and the stale-result reset. Added in `WalletViewModelTest.kt`: the updates stream starts only when swaps are supported. `compileDevDebugKotlin`, `testDevDebugUnitTest`, and `detekt` all pass.

### Corrections to earlier replies in this thread

- I said `startSwapUpdates()` "retries up to 3 times with a linear backoff (5s base) before giving up". That was true when written, but a later commit changed it to an unbounded `while (true)` loop capped at 60s. It stays unbounded; the network gate is what stops it from running where Boltz is unreachable.
- The description says this bumps bitkit-core to 0.5.1. It actually bumps 0.4.1 → 0.5.2.

### Known limitation, not addressed here

The swap is sized from `walletRepo.balanceState.maxSendLightningSats` (whole-wallet outbound), while the confirm screen is a per-channel selection flow. On the swap path the selected channels only feed `channelsToClose`, which is unused, so selecting one connection can move funds routed over another. The close path honours the selection exactly, as before. Happy to fix here or track it separately.

jvsena42:
We could run a Boltz instance in dev env in the future for better testing

coreyphillips:
> We could run a Boltz instance in dev env in the future for better testing

Agreed. For now I've been using mainnet for testing which will not be ideal long term.

coreyphillips:
> 2. Is this UI change in design? otherwise it should be gated under dev settings flag

Correct, the confirm screen changes are not in design yet, so they are now behind a dev flag in 146af8f.

- New `isSavingsSwapEnabled` setting, default `false`, toggled at Dev Settings -> Swaps -> Enable Savings Swap. The existing Swaps list row moved into that same section.
- `BoltzService.isSwapEnabled()` is the single gate: the network must have a reachable Boltz backend (mainnet) *and* the dev flag must be on. Both consumers use it, so the flag cannot be honoured in one place and missed in the other.
- `TransferViewModel.loadSavingsSwapQuote()` returns before touching state when the gate is closed, so no quote is requested and `SavingsSwapUiState` stays at its defaults. Since the confirm screen keys every new element off `quote != null`, that means no quote, fees, slider, or "close channel instead", and the swipe closes the channel exactly as it did before this PR.
- `WalletViewModel.ensureSwapUpdatesRunning()` is gated the same way, so with the flag off nothing subscribes to swap events and the updates stream never opens.

Tests added: the quote is skipped when the flag is off in `TransferViewModelTest.kt`, and the updates stream does not start when the flag is off in `WalletViewModelTest.kt`. The existing swap tests now enable the flag explicitly.

I also dropped the changelog fragment, since a dev-flagged flow is not user-facing yet. It can come back in the PR that turns the flag on once the design lands.

`compileDevDebugKotlin`, `testDevDebugUnitTest`, and `detekt` pass.

jvsena42:
Please, solve the conflict

app/src/main/java/to/bitkit/services/BoltzService.kt
```diff
@@ -0,0 +1,215 @@
...
+ * between onchain Bitcoin and Lightning).
+ *
+ * Mirrors the existing service pattern (e.g. [TrezorService]): a Hilt singleton
+ * that wraps the FFI and bridges the [BoltzEventListener] foreign callback to a
+ * [SharedFlow]. bitkit-core persists only a derivation index, never key material;
+ * swap keys are re-derived on demand from the wallet mnemonic.
+ *
+ * The Lightning side (issuing/paying invoices, fresh onchain addresses) is owned
+ * by [LightningService]; this service only talks to Boltz + the chain.
+ */
```
jvsena42, line 38: nit: this comment could be simplified

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -319,6 +331,72 @@ class WalletViewModel @Inject constructor(
...
+
+    /**
+     * Open the swap updates stream so any pending LN -> onchain swaps resume and auto-claim.
+     * Uses the wallet's current fee rate for the claim tx. Retries until started: without
+     * the stream a paid swap has nothing to broadcast its claim. Once started, bitkit-core
+     * keeps the WebSocket alive with its own reconnect loop.
+     */
+    private suspend fun startSwapUpdates() {
+        var attempt = 0
+        while (true) {
```
jvsena42: could add a ceiling

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -319,6 +331,72 @@ class WalletViewModel @Inject constructor(
...
+            val started = runSuspendCatching {
+                val speed = settingsStore.data.first().defaultTransactionSpeed
+                val feeRate = lightningRepo.getFeeRateForSpeed(speed).getOrNull()?.toDouble()
+                boltzService.startUpdates(feeRateSatPerVb = feeRate, acceptZeroConf = true)
+            }.onFailure {
+                Logger.warn("Failed to start swap updates, attempt '${attempt + 1}'", context = TAG)
```
jvsena42: could set the throwable to the warn parameter so we know the failure reason

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -1100,15 +1106,206 @@ class TransferViewModel @Inject constructor(
...
+    }
+
+    /**
+     * Run the swap in [viewModelScope] so it outlives the progress screen: navigating away
+     * mid-flight must not cancel a swap whose hold invoice is already paid. Idempotent while
+     * a swap is in flight, so re-entering the journey cannot create and pay a second swap.
+     * The outcome lands in [savingsSwapResult].
+     */
+    fun startSavingsSwap() {
+        if (savingsSwapJob?.isActive == true) return
```
jvsena42: add also `|| _savingsSwapResult.value != nul`, because this method can be called multiple times from `SavingsProgressScreen` in case of configuration changes

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -1100,15 +1106,206 @@ class TransferViewModel @Inject constructor(
...
+        }
+    }
+
+    /**
+     * Execute the LN -> onchain swap: derive a fresh claim address, create the swap,
+     * pay the returned hold invoice over Lightning, then wait for the on-chain claim.
+     * The claim is auto-broadcast by the updates stream as soon as the lockup hits the
+     * mempool, so a timeout here is not a failure; the swap completes in the background.
+     */
+    private suspend fun executeSavingsSwap(): SavingsSwapResult {
```
jvsena42, line 1250: `payInvoice` -> `LightningService.send(bolt11)` returns a PaymentId as soon as the HTLC is dispatched; it does not await settlement (and for a hold invoice it can't). Nothing listens for LDK's PaymentFailed. So an unroutable payment just falls through the 30s awaitSwapClaim timeout into Pending -> SETTLING, which tells the user "Your transfer has been initiated and is settling on-chain. Your savings balance will update automatically once it completes." for a transfer that will never complete.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
jvsena42, line 1: could have a specific `isSwapEnabled` flag

app/src/main/java/to/bitkit/viewmodels/SwapsViewModel.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    private val _isLoading = MutableStateFlow(false)
+    val isLoading = _isLoading.asStateFlow()
+
+    private val _error = MutableStateFlow<String?>(null)
+    val error = _error.asStateFlow()
+
+    init {
+        refresh()
+    }
```
jvsena42, line 38: this, along with `LaunchedEffect(Unit) { viewModel.refresh() `} fire refresh wise on ever open

jvsena42 (changes requested):

app/src/main/java/to/bitkit/services/BoltzService.kt
coreyphillips: Done. Trimmed the class KDoc down to the essentials: what it wraps, that it bridges the callback to a SharedFlow, and that keys are re-derived from the mnemonic rather than persisted.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
coreyphillips: Done. The loop now stops after SWAP_UPDATES_MAX_ATTEMPTS (20, roughly 14 min of backoff) and logs that it gave up. Giving up is safe here because ensureSwapUpdatesRunning is re-triggered on the next node start and when entering a swap flow, so a paid swap is not left stranded.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
coreyphillips: Done. The per-attempt warn now passes the throwable so the failure reason is logged.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
coreyphillips: Done. The guard is now `if (savingsSwapJob?.isActive == true || _savingsSwapResult.value != null) return`. A fresh journey still clears the outcome in onTransferToSavingsConfirm, so the redundant reset that used to sit here was removed.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
coreyphillips: Good catch. executeSavingsSwap now races the on-chain claim against a new awaitPaymentFailure, which watches lightningRepo.nodeEvents for LDK PaymentFailed matching the paid invoice and maps it to SavingsSwapResult.Failure. An unroutable payment now surfaces as an error instead of falling through the claim timeout into Pending then SETTLING. Covered by a new test: startSavingsSwap fails when the paid invoice reports a lightning routing failure.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
coreyphillips: Done. Added an explicit isSwapEnabled field to SavingsSwapUiState, populated from boltzService.isSwapEnabled() in loadSavingsSwapQuote, and the confirm screen now branches the swap-pricing UI on that flag instead of inferring it from quote presence.

app/src/main/java/to/bitkit/viewmodels/SwapsViewModel.kt
coreyphillips: Done. Removed the redundant LaunchedEffect(Unit) { refresh() } from SwapsScreen. Kept init { refresh() } because SwapDetailScreen resolves its own separate SwapsViewModel instance that relies on it, so both screens now load with a single refresh each.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
jvsena42: haven't find  `SWAP_UPDATES_MAX_ATTEMPTS` in the code, probably not pushed

app/src/main/java/to/bitkit/viewmodels/SwapsViewModel.kt
```diff
@@ -0,0 +1,101 @@
...
+ */
+@HiltViewModel
+class SwapsViewModel @Inject constructor(
+    private val boltzService: BoltzService,
+) : ViewModel() {
+    private val _swaps = MutableStateFlow<ImmutableList<BoltzSwap>>(persistentListOf())
+    val swaps = _swaps.asStateFlow()
+
+    private val _isLoading = MutableStateFlow(false)
+    val isLoading = _isLoading.asStateFlow()
```
jvsena42: value never collected

app/src/main/java/to/bitkit/services/BoltzService.kt
```diff
@@ -0,0 +1,215 @@
...
+        // Boltz does not operate on signet; fall back to testnet for development.
+        else -> BoltzNetwork.TESTNET
+    }
+
+    /** Current Electrum URL, used by Boltz for claim/refund broadcasting. */
+    suspend fun electrumUrl(): String = settingsStore.data.first().electrumServer
+
+    private suspend fun credentials(): Pair<String, String?> {
+        val mnemonic = keychain.loadString(Keychain.Key.BIP39_MNEMONIC.name)
+            ?: error("Mnemonic not found")
```
jvsena42: nit app conventions use `AppError/requireNotNull`

app/src/main/java/to/bitkit/ui/settings/SwapsScreen.kt
```diff
@@ -0,0 +1,358 @@
...
+        )
+    }
+}
+
+private val epochFormatter: DateTimeFormatter =
+    DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss").withZone(ZoneId.systemDefault())
+
+private fun formatEpochSeconds(seconds: ULong): String =
+    runCatching { epochFormatter.format(Instant.ofEpochSecond(seconds.toLong())) }
+        .getOrDefault(seconds.toString())
```
jvsena42: there is existing utils for this in  `ext/DateTime`

app/src/main/java/to/bitkit/ui/settings/SwapsScreen.kt
jvsena42, line 1: `cardColors`, `InfoCard`, `InfoCell`, `DetailRow` could be extracted and shared under `ui/components/settings/` The same components exist in `ChannelOrdersScreen.kt`

jvsena42 (changes requested):
Some style findings, no major issues

app/src/main/java/to/bitkit/viewmodels/SwapsViewModel.kt
coreyphillips: Done. Removed `_isLoading`/`isLoading` in 5618b2fe2, nothing collected them so the flow was dead weight.

app/src/main/java/to/bitkit/services/BoltzService.kt
coreyphillips: Done. `credentials()` now throws `ServiceError.MnemonicNotFound()` in 5618b2fe2, matching how `LightningRepo` loads the same keychain entry.

app/src/main/java/to/bitkit/ui/settings/SwapsScreen.kt
coreyphillips: Done. Dropped the local `epochFormatter`/`formatEpochSeconds` and switched both call sites to `ULong.formatToString()` from `ext/DateTime` in 5618b2fe2.

app/src/main/java/to/bitkit/ui/settings/SwapsScreen.kt
coreyphillips: Done. Moved `cardColors`, `InfoCard`, `InfoCell` and `DetailRow` into `ui/components/settings/InfoCard.kt` in 5618b2fe2, and pointed both `SwapsScreen` and `ChannelOrdersScreen` at them. The shared `DetailRow` keeps `ChannelOrdersScreen`s `isError` flag plus `SwapsScreen`s `weight(1f, fill = false)` on the value, so long values ellipsize instead of squeezing the label.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
coreyphillips: Good catch, it was on the wrong branch. That whole round of fixes got committed to `feat/send-swap` (#1101) instead of here, so this PR never carried it. Moved onto this branch in 6b0abdf5f, and #1101 was rebuilt on top of it so it now only holds its own commits.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
```diff
@@ -135,7 +179,48 @@ private fun SavingsConfirmContent(
 
             Caption13Up(text = stringResource(R.string.lightning__savings_confirm__label), color = Colors.White64)
             Spacer(modifier = Modifier.height(8.dp))
-            MoneyDisplay(sats = amount.toLong(), onClick = onAmountClick)
+            MoneyDisplay(sats = headlineAmount.toLong(), onClick = onAmountClick)
+
+            if (quote != null) {
+                Spacer(modifier = Modifier.height(24.dp))
```
jvsena42: Should use `VerticalSpacer`

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
```diff
@@ -135,7 +179,48 @@ private fun SavingsConfirmContent(
...
+                    )
+                    FeeInfo(
+                        label = stringResource(R.string.lightning__savings_confirm__receive),
+                        amount = quote.receiveSat.toLong(),
+                    )
+                }
+
+                // Adjust how much to move to savings, bounded to a payable range.
+                if (maxSat > minSat) {
+                    Spacer(modifier = Modifier.height(28.dp))
```
jvsena42: same https://github.com/synonymdev/bitkit-android/pull/1081/changes#r3656320289

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
```diff
@@ -156,32 +241,55 @@ private fun SavingsConfirmContent(
...
             )
+            if (quote != null) {
+                Spacer(modifier = Modifier.height(12.dp))
+                // Fallback: drain a whole channel on-chain by closing it instead of swapping.
+                TertiaryButton(
+                    text = stringResource(R.string.lightning__savings_confirm__close_instead),
+                    onClick = onCloseConfirm,
+                )
+            }
             Spacer(modifier = Modifier.height(16.dp))
```
jvsena42: The AI probably used this pattern to build the other spacers for this screen instead of using `VerticalSpacer`

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsProgressScreen.kt
```diff
@@ -203,7 +241,30 @@ private fun Content(
...
+            R.string.lightning__savings_settling__title to R.string.lightning__savings_settling__text
+
+        SavingsProgressState.SUCCESS ->
+            R.string.lightning__transfer_success__title_savings to R.string.lightning__transfer_success__text_savings
+
+        SavingsProgressState.INTERRUPTED ->
+            R.string.lightning__savings_interrupted__title to R.string.lightning__savings_interrupted__text
+    }
+    Display(text = stringResource(titleRes).withAccent())
+    Spacer(modifier = Modifier.height(8.dp))
```
jvsena42: Please, replace all these occurrences in this screen with `VerticalSpacer`

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -80,6 +90,7 @@ class TransferViewModel @Inject constructor(
     private val settingsStore: SettingsStore,
     private val cacheStore: CacheStore,
     private val transferRepo: TransferRepo,
+    private val boltzService: BoltzService,
```
jvsena42, line 93: nit: `BoltzService` could be injected into `TransferRepo`. It will facilitate the process of making swaps transfer activities

jvsena42: Can be done in other PR

jvsena42: Also injected on LightingRepo for payment related features

jvsena42 (changes requested):
The reverse swaps are not marked as transfer to savings. Acceptable for now, giving this  PR is a dev-only feature and we will probably have a more [structured one for the final user](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ?node-id=40364-128564&m=dev#1855941137)

The dependency injection refactor can also be done in the next PR

Please, apply the Spacer refactor for prevent the AIs using it as reference

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
coreyphillips: Done in 8163828f7. Every raw `Spacer(modifier = Modifier.height(...))` in this screen is now `VerticalSpacer(...)`.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
coreyphillips: Done in 8163828f7. Every raw `Spacer(modifier = Modifier.height(...))` in this screen is now `VerticalSpacer(...)`.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsConfirmScreen.kt
coreyphillips: Done in 8163828f7. Every raw `Spacer(modifier = Modifier.height(...))` in this screen is now `VerticalSpacer(...)`.

app/src/main/java/to/bitkit/ui/screens/transfer/SavingsProgressScreen.kt
coreyphillips: Done in 8163828f7. Replaced all of them: the fixed heights use `VerticalSpacer(...)` and the two `Modifier.weight(1f)` spacers use `FillHeight()`. Dropped the now unused `Spacer`/`height` imports from both screens.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
coreyphillips: Agreed, leaving the `BoltzService` move into `TransferRepo`/`LightningRepo` for the follow-up PR along with marking reverse swaps as transfers to savings.

coreyphillips:
Pushed 8163828f7:

- `SavingsConfirmScreen` and `SavingsProgressScreen` now use `VerticalSpacer`/`FillHeight` instead of raw `Spacer(modifier = Modifier.height(...))`, and the unused `Spacer`/`height` imports are gone. Checked the rest of the diff too, no other raw spacers left in the touched UI files.
- Marking reverse swaps as transfers to savings and moving `BoltzService` into `TransferRepo`/`LightningRepo` are deferred to the follow-up PR, as agreed.

`compileDevDebugKotlin`, `testDevDebugUnitTest` and `detekt` all pass.

jvsena42 (approved):
Safe to merge, all swaps flows are gated under the dev flag

Production design will come later in [V62](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-146397&m=dev)

PR #1086, opened 2026-07-16
fix: refresh receive after payment

ben-kaufman:
### Description

- Invalidate a settled wallet Lightning invoice before publishing its payment event so stale invoices are not restored.
- Carry settlement metadata with the exact node event instead of shared mutable state, preventing adjacent events from taking another receive request context.
- Persist the cached BOLT11 payment hash with the invoice so settlement correlation does not depend on decoding during event handling; migrated cache entries fall back to direct BOLT11 parsing.
- Correlate on-chain transaction outputs with the exact cached receive address, then invalidate and rotate only that address and its BIP21 request.
- Track the exact invoice and on-chain address shown by each receive sheet so a paid sheet closes before received-payment presentation without dismissing a replacement sheet.
- Process a received Lightning payment once using its payment ID, while still notifying activity observers after the core update.
- Coordinate notification and in-app presentation through one synchronous in-process claim and a shared presentation helper, then mark the activity seen only after presentation is handed off.

### Preview

Not applicable; this fixes receive-state lifecycle behavior without changing the UI layout. The existing Android sheet animation remains unchanged, as discussed in review.

### QA Notes

- ./gradlew compileDevDebugKotlin
- ./gradlew testDevDebugUnitTest
- ./gradlew detekt --rerun-tasks (passes with the repository baseline findings)
- git diff --check
- Regression coverage in AppCacheDataTest.kt, CacheStoreTest.kt, LightningRepoTest.kt, WalletRepoTest.kt, AppViewModelSendFlowTest.kt, LightningNodeServiceTest.kt, and NotifyPaymentReceivedHandlerTest.kt

### DIFF
--- a/app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
+++ b/app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
@@ -15,6 +15,7 @@ import androidx.core.content.ContextCompat
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.lightningdevkit.ldknode.Event
 import to.bitkit.App
 import to.bitkit.R
@@ -106,10 +107,17 @@ class LightningNodeService : Service() {
         if (event !is Event.PaymentReceived && event !is Event.OnchainTransactionReceived) return
         val command = NotifyPaymentReceived.Command.from(event, includeNotification = true) ?: return
 
-        notifyPaymentReceivedHandler(command).onSuccess {
-            Logger.debug("Payment notification result: $it", context = TAG)
-            if (it !is NotifyPaymentReceived.Result.ShowNotification) return
-            showPaymentNotification(it.sheet, it.notification)
+        notifyPaymentReceivedHandler(command).onSuccess { result ->
+            Logger.debug("Handled payment notification with result '$result'", context = TAG)
+            if (result !is NotifyPaymentReceived.Result.ShowNotification) return
+            withContext(uiDispatcher) {
+                notifyPaymentReceivedHandler.present(
+                    command = command,
+                    canPresent = { App.currentActivity?.value == null },
+                ) {
+                    showPaymentNotification(result.sheet, result.notification)
+                }
+            }
         }
     }
 
@@ -132,10 +140,6 @@ class LightningNodeService : Service() {
         sheet: NewTransactionSheetDetails,
         notification: NotificationDetails,
     ) {
-        if (App.currentActivity?.value != null) {
-            Logger.debug("Skipping payment notification: activity is active", context = TAG)
-            return
-        }
         Logger.debug("Showing payment notification: ${notification.title}", context = TAG)
         serviceScope.launch { cacheStore.setBackgroundReceive(sheet) }
         pushNotification(notification.title, notification.body)
--- a/app/src/main/java/to/bitkit/data/CacheStore.kt
+++ b/app/src/main/java/to/bitkit/data/CacheStore.kt
@@ -55,12 +55,38 @@ class CacheStore @Inject constructor(
         store.updateData { it.copy(onchainAddress = address) }
     }
 
-    suspend fun saveBolt11(bolt11: String) {
-        store.updateData { it.copy(bolt11 = bolt11) }
+    suspend fun saveBolt11(bolt11: String, paymentHash: String) {
+        store.updateData { it.copy(bolt11 = bolt11, bolt11PaymentHash = paymentHash) }
     }
 
     suspend fun setBip21(bip21: String) {
         store.updateData { it.copy(bip21 = bip21) }
+    }
+
+    suspend fun invalidateReceiveLightningInvoice(expectedBolt11: String): Boolean {
+        var invalidated = false
+        store.updateData {
+            if (it.bolt11 == expectedBolt11) {
+                invalidated = true
+                it.invalidateReceiveLightningInvoice()
+            } else {
+                it
+            }
+        }
+        return invalidated
+    }
+
+    suspend fun invalidateReceiveOnchainAddress(expectedAddress: String): Boolean {
+        var invalidated = false
+        store.updateData {
+            if (it.onchainAddress == expectedAddress) {
+                invalidated = true
+                it.invalidateReceiveOnchainAddress()
+            } else {
+                it
+            }
+        }
+        return invalidated
     }
 
     suspend fun cacheBalance(balanceState: BalanceState) {
@@ -123,6 +149,7 @@ data class AppCacheData(
     val paidOrders: Map<String, String> = mapOf(),
     val onchainAddress: String = "",
     val bolt11: String = "",
+    val bolt11PaymentHash: String = "",
     val bip21: String = "",
     val balance: BalanceState? = null,
     val backupStatuses: Map<BackupCategory, BackupItemStatus> = mapOf(),
@@ -132,5 +159,9 @@ data class AppCacheData(
     val addressSearchLastUsedReceiveIndexes: Map<String, Int> = mapOf(),
     val addressSearchLastUsedChangeIndexes: Map<String, Int> = mapOf(),
 ) {
-    fun resetBip21() = copy(bip21 = "", bolt11 = "", onchainAddress = "")
+    fun resetBip21() = copy(bip21 = "", bolt11 = "", bolt11PaymentHash = "", onchainAddress = "")
+
+    fun invalidateReceiveLightningInvoice() = copy(bip21 = "", bolt11 = "", bolt11PaymentHash = "")
+
+    fun invalidateReceiveOnchainAddress() = copy(bip21 = "", onchainAddress = "")
 }
--- a/app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
+++ b/app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
@@ -4,6 +4,7 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.withContext
 import to.bitkit.di.IoDispatcher
+import to.bitkit.ext.runSuspendCatching
 import to.bitkit.models.NewTransactionSheetDetails
 import to.bitkit.models.NewTransactionSheetDirection
 import to.bitkit.models.NewTransactionSheetType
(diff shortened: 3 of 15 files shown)


app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
```diff
@@ -107,10 +107,13 @@ class LightningNodeService : Service() {
...
         val command = NotifyPaymentReceived.Command.from(event, includeNotification = true) ?: return
 
-        notifyPaymentReceivedHandler(command).onSuccess {
-            Logger.debug("Payment notification result: $it", context = TAG)
-            if (it !is NotifyPaymentReceived.Result.ShowNotification) return
-            showPaymentNotification(it.sheet, it.notification)
+        notifyPaymentReceivedHandler(command).onSuccess { result ->
+            Logger.debug("Payment notification result: $result", context = TAG)
+            if (result !is NotifyPaymentReceived.Result.ShowNotification) return
+            if (App.currentActivity?.value != null) return
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in `4061e04`. Received-payment presentation now uses a process-wide atomic claim shared by the service and UI. The service claims only when it owns the background path; `AppViewModel` always attempts the fallback claim after the service handler and no longer rechecks activity state, so a pause/rotation cannot leave both paths declining the event. The losing path is deduplicated by the claim. Covered by the foreground/background handoff and service-ownership tests in `AppViewModelSendFlowTest.kt` and `LightningNodeServiceTest.kt`.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1014,7 +1053,10 @@ class AppViewModel @Inject constructor(
         val command = NotifyPaymentReceived.Command.from(event) ?: return
         val result = notifyPaymentReceivedHandler(command).getOrNull()
         if (result !is NotifyPaymentReceived.Result.ShowSheet) return
-        showTransactionSheet(result.sheet)
+        if (App.currentActivity?.value == null) return
+        notifyPaymentReceivedHandler.markPresented(command).onSuccess {
+            showTransactionSheet(result.sheet)
+        }
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in `4061e04`. `claimPresentation` records the in-process claim before best-effort persistence. If marking the activity as seen fails, the failure is logged but the claimant still returns true and presents the sheet/notification; the retained claim prevents the other in-process path from duplicating it. Covered by `claimPresentation keeps the claim when marking as seen fails` in `NotifyPaymentReceivedHandlerTest.kt`.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -439,10 +442,55 @@ class LightningRepo @Inject constructor(
...
+            Logger.error("Failed to read cached receive invoice", it, context = TAG)
+        }.getOrNull() ?: return null
+        if (cachedBolt11.isEmpty()) return null
+
+        val cachedPaymentHash = runSuspendCatching {
+            (coreService.decode(cachedBolt11) as? Scanner.Lightning)?.invoice?.paymentHash?.toHex()
+        }.onFailure {
+            Logger.error("Failed to decode cached receive invoice", it, context = TAG)
+        }.getOrNull()
+        if (cachedPaymentHash != event.paymentHash) return null
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in `4061e04`. The cache now stores the BOLT11 payment hash atomically with the invoice, and settlement correlation uses that stored identity without calling the core decoder. Existing cache entries without the new field fall back to direct `Bolt11Invoice` parsing, and invalidation clears both invoice and hash. `LightningRepoTest.kt` now verifies invalidation succeeds even when the core decoder is unavailable.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -439,10 +442,55 @@ class LightningRepo @Inject constructor(
     private suspend fun onEvent(event: Event) {
         handleLdkEvent(event)
         recordProbeOutcome(event)
+        val settledReceiveInvoice = if (event is Event.PaymentReceived) {
+            val settledInvoice = invalidateSettledReceiveInvoice(event)
+            val paymentId = event.paymentId ?: event.paymentHash
+            runSuspendCatching { coreService.activity.handlePaymentEvent(paymentId) }
```
jvsena42: AppViewModel.handlePaymentReceived still calls `activityRepo.handlePaymentEvent(paymentHash)`. The same payment is processed twice with different keys. Is this intentional?

app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
```diff
@@ -43,12 +49,39 @@ class NotifyPaymentReceivedHandler @Inject constructor(
...
+    ): Boolean {
+        val key = presentationKey(command) ?: return false
+        return synchronized(presentationClaimsLock) {
+            canPresent() && presentationClaims.add(key)
+        }
+    }
+
+    suspend fun recordPresentation(command: NotifyPaymentReceived.Command) {
+        withContext(ioDispatcher) {
+            runSuspendCatching { markAsSeen(command) }
```
jvsena42, line 77: nit: could remove the key from `presentationClaims`

app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
```diff
@@ -107,10 +108,17 @@ class LightningNodeService : Service() {
...
+            Logger.debug("Handled payment notification with result '$result'", context = TAG)
+            if (result !is NotifyPaymentReceived.Result.ShowNotification) return
+            val presented = withContext(uiDispatcher) {
+                if (!notifyPaymentReceivedHandler.claimPresentation(command) { App.currentActivity?.value == null }) {
+                    return@withContext false
+                }
+                showPaymentNotification(result.sheet, result.notification)
+                true
+            }
+            if (presented) notifyPaymentReceivedHandler.recordPresentation(command)
```
jvsena42: could extract this duplicated logic to a `present` method like: 

```
suspend fun present(
    command: NotifyPaymentReceived.Command,
    canPresent: () -> Boolean = { true },
    block: suspend () -> Unit,
): Boolean {
    if (!claimPresentation(command, canPresent)) return false
    block()
    recordPresentation(command)
    return true
}
```

jvsena42 (commented):
Nothing blocking so far, still haven't tested

piotr-iohk:
Nice fix overall — the LN receive settlement / refresh path brings Android closer to iOS and feels much more consistent.

Tested LN + on-chain receive on this branch (foreground, receive sheet open). Compared behavior with iOS. 

### LN receive (foreground)
Works as expected: paid receive sheet closes and the received-payment UI shows. Replacement / refresh path looks good for Lightning.

### On-chain receive (foreground) — gap vs LN / iOS
With the receive sheet open, after an on-chain payment:

1. The receive sheet **does not dismiss** (LN does; iOS dismisses/replaces for both).
2. The on-chain address / BIP21 request **stays the same** on the still-open receive sheet.

On iOS, showing `receivedTx` replaces the single sheet stack, so receive goes away for both LN and on-chain. On Android, `hideSheet()` is only wired for `PaymentReceived` + matching settled invoice — `OnchainTransactionReceived` never clears `Sheet.Receive`.

Would be good also to close/refresh the receive sheet for on-chain the same way as LN (or otherwise match iOS), and confirm the reusable address rotates after the current one is used.

https://github.com/user-attachments/assets/e71c7f84-e731-4c19-a4bb-fd6cee2bacc3

### UX nit (non-blocking)
iOS slides receive away before showing received; Android drops the receive sheet abruptly. Nice-to-have parity, not blocking if the on-chain dismiss/refresh above is fixed.

https://github.com/user-attachments/assets/12aa4ea5-25b5-4681-b796-5214fd04c70c

jvsena42:
> iOS slides receive away before showing received; Android drops the receive sheet abruptly. Nice-to-have parity, not blocking if the on-chain dismiss/refresh above is fixed.

This is Android native behavior. Better wait for a improvement on sheet library

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
ben-kaufman: Fixed in ff8ac68. LightningRepo is now the single owner of core received-payment processing and uses paymentId with paymentHash only as its fallback. AppViewModel no longer processes the payment again by paymentHash; it only signals that the already-updated activity changed.

app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
ben-kaufman: I kept the claim intentionally. Removing it immediately after persistence can let the other service/UI path reuse a stale Show result and present the same payment twice. The set is process-local, and the regression test now documents that a successful presentation retains its claim.

app/src/main/java/to/bitkit/androidServices/LightningNodeService.kt
ben-kaufman: Implemented in ff8ac68. Both the service notification path and the in-app sheet path now use NotifyPaymentReceivedHandler.present, which owns the claim, presentation handoff, and seen-state recording.

ben-kaufman:
@piotr-iohk Fixed in ff8ac68. On-chain receive events now match the exact cached receive address against transaction outputs, invalidate and rotate only that address/BIP21 request, and dismiss the matching receive sheet before the received-payment handoff. Unrelated or newly opened replacement sheets are preserved. Regression coverage was added across CacheStoreTest, LightningRepoTest, WalletRepoTest, and AppViewModelSendFlowTest.

The optional animation change is intentionally not included, following the note that the current transition is native behavior of the Android sheet library.

piotr-iohk (approved):

app/src/main/java/to/bitkit/data/CacheStore.kt
```diff
@@ -55,14 +55,40 @@ class CacheStore @Inject constructor(
         store.updateData { it.copy(onchainAddress = address) }
     }
 
-    suspend fun saveBolt11(bolt11: String) {
-        store.updateData { it.copy(bolt11 = bolt11) }
+    suspend fun saveBolt11(bolt11: String, paymentHash: String) {
```
jvsena42, line 58: if this cached hash ends up being bringing too much issues to manage, we can also extract it from Bolt11 invoice instead of caching

jvsena42 (approved):
tAck

Issue #1034, opened 2026-06-19
[Bug]: Transfer to spending should use 'fast' transaction speed


pwltr:
### What happened?

For regular channel orders, the order payment transaction should use 'fast' transaction speed, so it matches the copy "~10 minutes" and improved UX

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/edbd9762-414d-4925-a7f0-18eb5e9fad8c" />

PR #1089, opened 2026-07-16
fix: use fast fee for hw transfer

### LINKED ISSUE #1088: [Bug]: Trezor transfer to spending uses low on-chain fee rate
## What happened?

During 2.4.0 testing, **Transfer to Spending from a paired Trezor** stayed pending for a long time (>1h).

On-chain inspection of a funding tx from the Android test showed the tx was still **unconfirmed** at about **1 sat/vB**, not stuck only in LSP/channel setup.

Example (Android test funding tx):
https://mempool.space/tx/9a776705f6fb8f470c2a29b950fdb4499016c030c564f2550cc34d8b874654d9
- fee ≈ 209 sats / ~208.5 vB ≈ **1.0 sat/vB**
- unconfirmed for a long time while next-block market was ~2–3 sat/vB

Absolute fee looked “higher” than iOS only because the Android tx had more inputs (larger vsize); the **rate** matched iOS at ~1 sat/vB.

## Expected behavior

Trezor → spending funding should confirm promptly enough for LSP channel open, ideally matching (or beating) the fee urgency of savings → spending transfers.

At minimum: do not default HW funding to economy/minimum (~1 sat/vB) when estimates are available or when fallback is used.

## Steps to Reproduce

1. Pair a Trezor and fund it with spendable UTXOs.
2. Start **Transfer → Spending** from the hardware wallet.
3. Sign and broadcast the funding payment on the device.
4. Observe the transfer remaining pending for a long time.
5. Open the funding txid in a block explorer and check **sat/vB** + confirmations.

- Mempool evidence: https://mempool.space/tx/9a776705f6fb8f470c2a29b950fdb4499016c030c564f2550cc34d8b874654d9
- Related iOS child/sibling tx from same session: https://mempool.space/tx/c5fa3d989dab47ced6864f0ac18a251def6cb28a02a501c33f4169972d6f6ce1 (spends Android change output; also ~1 sat/vB)

## Bitkit Version

2.4.0 (release candidate / testing build)

## Device / OS

Android (QA device used in Trezor transfer testing session)

## Reproducibility

Often (>50%)

## Additional context

**Epic:** synonymdev/bitkit-android#998  
**Related (iOS):** synonymdev/bitkit-ios#633

### Root cause

HW funding fee rate comes from settings default speed via `TransferViewModel.fetchHwFundingSatsPerVByte()` → `defaultTransactionSpeed` (default **Medium** → Blocktank `mid`). Fallback when estimates fail: `HW_FUNDING_FALLBACK_SATS_PER_VBYTE = 1`.

When `mid` is 1, funding lands at **1 sat/vB**. iOS savings already hardcodes **fast**, which is why software transfers feel faster than Trezor there.

### Proposed change

1. Use **fast** for Trezor → spending funding (`TransactionSpeed.Fast` instead of `defaultTransactionSpeed`).
2. Raise HW fallback floor from `1` to a safer value (e.g. `3`, or `max(fastEstimate, 3)`).
3. Optional: also use **fast** for savings → spending so all LSP funding is urgent/consistent with iOS.
4. Update HW fee-related unit tests + changelog fragment if user-facing.

### Environment

- Mainnet
- Blocktank LSP order funding from Trezor
- Same Trezor account also used on iOS in the same session (chained unconfirmed spend)

piotr-iohk:
Fixes #1088

This PR makes Trezor → spending funding use a faster on-chain fee rate so LSP channel funding confirms more reliably.

### Description

- Uses the fast fee tier for hardware-wallet transfer funding instead of the settings default (medium), which can resolve to 1 sat/vB in a quiet fee market.
- Raises the estimate-unavailable fallback from 1 to 3 sat/vB so compose still picks a usable rate when Blocktank fee estimates fail.
- Updates unit coverage for the fast-rate path and the new fallback floor.

Related iOS: synonymdev/bitkit-ios#633

### Preview

N/A

#### Manual Tests
- [ ] **1.** Pair Trezor with spendable UTXOs → Transfer → Spending → Sign: funding tx uses a fast-tier sat/vB (check explorer), not ~1 sat/vB when the market’s mid is 1.
- [ ] **2.** Complete HW transfer → Spending: funding confirms and the transfer progresses toward channel open without lingering unconfirmed for hours under a normal fee market.
- [ ] **3.** `regression:` Savings → Transfer → Spending: still works with the existing savings fee path.

#### Automated Checks
- Unit tests modified: assert HW compose requests `TransactionSpeed.Fast` and uses a 3 sat/vB fallback when fee lookup fails in `TransferViewModelTest.kt`.
- Local: `just test file TransferViewModelTest`, `just lint`.


app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -892,10 +892,8 @@ class TransferViewModel @Inject constructor(
     private suspend fun hwFundingSatsPerVByte(): ULong =
         fetchHwFundingSatsPerVByte().getOrDefault(HW_FUNDING_FALLBACK_SATS_PER_VBYTE)
 
-    private suspend fun fetchHwFundingSatsPerVByte(): Result<ULong> {
-        val speed = settingsStore.data.first().defaultTransactionSpeed
-        return lightningRepo.getFeeRateForSpeed(speed)
-    }
+    private suspend fun fetchHwFundingSatsPerVByte(): Result<ULong> =
+        lightningRepo.getFeeRateForSpeed(TransactionSpeed.Fast)
```
jvsena42, line 896: the regular savings -> spending  `onTransferToSpendingConfirm` still uses `defaultTransactionSpeed` instead of hardcode fast. we should either update there too in this PR or at least open an issue

piotr-iohk: Good catch — left it out of the first commit to keep the change scoped to the HW path from #1088, but you’re right that savings should match iOS (which already hardcodes fast).

Pushed a follow-up that defaults savings → spending funding to `TransactionSpeed.Fast` as well, so both funding sources are consistent with iOS.

piotr-iohk:
Tested end-to-end on mainnet (Trezor → spending / Blocktank channel open).

- Funding tx used ~3 sat/vB (not the previous ~1 sat/vB mid/default path).
- Tx confirmed within minutes and the channel opened successfully.
- Previously the same flow could sit unconfirmed for hours at ~1 sat/vB.

Fix confirmed end-to-end on mainnet.

pwltr:
Failing e2e test seems valid on first glance, also note related issue https://github.com/synonymdev/bitkit-android/issues/1034

piotr-iohk:
Will address savings → spending (onTransferToSpendingConfirm / #1034) separately — that path needs more than a fee-tier swap (fee headroom / send-all edge cases under fast rates). This PR stays a targeted HW-wallet funding fee fix for #1088.

jvsena42 (approved):
I'll handle #1034

PR #1091, opened 2026-07-17
fix: use fast fee for savings transfer

### LINKED ISSUE #1034: [Bug]: Transfer to spending should use 'fast' transaction speed

piotr-iohk:
Fixes #1034

This PR makes savings → spending channel funding use a faster on-chain fee rate, matching the “~10 minutes” copy and iOS behavior.

### Description

- Defaults savings → spending funding to the fast fee tier instead of the settings default.
- Reserves a fast send-all mining fee when sizing max transfer, so “transfer all” leaves room to pay the order on-chain (fixes the tight-balance failure mode seen in `multi_address_2`).
- Aligns send-all with iOS: drain only for real dust change; fixed-send failures toast without surprise-draining the wallet.
- Prepares and shows the real mining fee on confirm before swipe, and pays funding from `viewModelScope` so broadcast + paid-order caching are not cancelled by Compose scope teardown.

### Preview

N/A

#### Manual Tests
- [x] **1.** Funded savings wallet → Transfer → Spending → confirm max amount: funding tx uses a fast-tier sat/vB (check explorer), not the settings mid/default when that mid is ~1.
- [x] **2.** Transfer → Spending → use max → confirm: order is sized with mining-fee headroom and funding succeeds (no coin-selection / insufficient-funds toast on a normal fee market).
- [x] **3.** Tight multi-UTXO balance (transfer all) → Transfer → Spending → confirm: channel funding progresses; leftover is not left as an unaffordable fixed send.
- [x] **4.** `regression:` Transfer → Spending with a partial amount well below balance: still funds with a normal (non-drain) send when change is above dust.

#### Automated Checks
- Unit tests added: fast mining-fee reserve on max sizing; dust send-all; no drain when normal fee leaves non-dust change; fixed-send failure without drain, in `TransferViewModelTest.kt`.
- Local: `just test file TransferViewModelTest`.
- CI: standard compile, unit test, and detekt checks run by the PR bot. E2E `multi_address_2` should pass once this lands on the resolved e2e branch.


app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -212,7 +212,8 @@ class TransferViewModel @Inject constructor(
     }
 
     /** Pays for the order and start watching it for state updates */
-    fun onTransferToSpendingConfirm(order: IBtOrder, speed: TransactionSpeed? = null) {
+    @Suppress("LongMethod")
+    fun onTransferToSpendingConfirm(order: IBtOrder, speed: TransactionSpeed = TransactionSpeed.Fast) {
```
jvsena42:
```suggestion
    fun onTransferToSpendingConfirm(order: IBtOrder) {
```

A custom speed is never set anywhere, could remove this parameter

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -230,53 +231,70 @@ class TransferViewModel @Inject constructor(
 
             val expectedChange =
                 spendableBalance.toLong() - order.feeSat.toLong() - sendAllFee.toLong()
+            // Match iOS: proactive drain only for dust change. Negative change means the drain
+            // output would underpay the order — never send-all in that case.
             val shouldUseSendAll =
                 expectedChange >= 0 && expectedChange < TRANSFER_SEND_ALL_THRESHOLD_SATS
```
jvsena42: nit: could apply a small refactor converting this to a range check

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -230,53 +231,70 @@ class TransferViewModel @Inject constructor(
...
-                    0uL
-                }
-            }
-            val txTotalSats = if (shouldUseSendAll) {
-                spendableBalance
-            } else {
-                order.feeSat.safe() + miningFee.safe()
-            }
+            val maxSendable =
+                if (sendAllFee >= spendableBalance) 0uL else spendableBalance - sendAllFee
```
jvsena42: use `.safe()` here. It has this clamp-to-zero internal logic

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -415,10 +433,21 @@ class TransferViewModel @Inject constructor(
...
+            val spendable = lightningRepo.getBalancesAsync().getOrNull()?.spendableOnchainBalanceSats
+                ?: 0uL
+            val miningFee = lightningRepo.estimateSendAllFee(
+                speed = TransactionSpeed.Fast,
+            ).getOrElse {
+                Logger.warn("Failed to estimate transfer mining fee reserve", it, context = TAG)
+                (spendable.toDouble() * Defaults.fallbackFeePercent).toULong()
+            }
+            val availableAmount =
+                if (miningFee >= spendable) 0uL else spendable - miningFee
```
jvsena42: use .safe() here too

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -230,53 +231,70 @@ class TransferViewModel @Inject constructor(
...
+                Logger.warn("Failed to estimate transfer funding fee", it, context = TAG)
+                0uL
+            }
+            fund(
+                isMaxAmount = false,
+                txTotalSats = order.feeSat.safe() + miningFee.safe(),
+            ).onFailure { error ->
+                // Match iOS: if fixed send fails (e.g. coin selection), drain only when the
+                // drain output still covers order.feeSat — never underpay by emptying the wallet.
+                if (maxSendable >= order.feeSat) {
```
jvsena42: I'm still checking this logic

jvsena42 (changes requested):

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `79eec29d2` — dropped the param; confirm always uses `TransactionSpeed.Fast` internally.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `79eec29d2` — now `expectedChange in 0 until …` (later replaced by the real dust helper on selected UTXOs).

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `79eec29d2` — `spendableBalance.safe() - sendAllFee.safe()`.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `79eec29d2` — `spendable.safe() - miningFee.safe()`.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: That block evolved after your review:
- drain only on real dust from selected UTXOs + normal fee (aligned with iOS `DustChangeHelper`)
- mining fee prepared on confirm appear and shown up-front
- no surprise drain retry after a failed fixed send

app/src/main/java/to/bitkit/ui/screens/transfer/SpendingConfirmScreen.kt
```diff
@@ -268,16 +279,22 @@ private fun Content(
...
+                    loading = isLoading || !isConfirmFeeReady,
                     color = Colors.Purple,
                     onConfirm = {
+                        if (!canConfirm) return@SwipeToConfirm
                         scope.launch {
                             isLoading = true
-                            delay(300)
-                            onTransferToSpendingConfirm(order)
-                            onConfirm()
+                            val paid = onTransferToSpendingConfirm(order)
```
jvsena42: It is not recommended expose suspend functions to the view, because te scope may be cancelled mid processing.

In this case there is the chance of the method be cancelled between sendOnChain broadcasting the funding tx and fundPaidOrder writing cacheStore.addPaidOrder

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42, line 1: nit: The description says "retry drain after a failed fixed send when the drain output still covers the order," but the implemented onFailure just toasts and returns false

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -211,73 +215,185 @@ class TransferViewModel @Inject constructor(
...
-                spendableBalance
-            } else {
-                order.feeSat.safe() + miningFee.safe()
+            .onFailure { ToastEventBus.send(it) }
+            .isSuccess
+    }
+
+    @Suppress("LongMethod")
+    private suspend fun buildSpendingConfirmFundingPlan(
+        order: IBtOrder,
```
jvsena42: nit: could separate normal selection and sendAll branches to remove `@Suppress("LongMethod")`

jvsena42 (changes requested):

app/src/main/java/to/bitkit/ui/screens/transfer/SpendingConfirmScreen.kt
piotr-iohk: Good catch — addressed in `4cf24ec29`.

Confirm pay no longer runs under Compose `rememberCoroutineScope`. `onTransferToSpendingConfirm` is non-suspend, launches on `viewModelScope`, and navigates via `TransferEffect.OnSpendingFundingPaid`. `fundPaidOrder` after broadcast also runs under `NonCancellable` so the paid-order cache write isn’t dropped if the job is cancelled mid-flight.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Agreed — that description was stale. PR body is updated: fixed-send failure only toasts (no surprise drain retry), matching the current implementation.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in `4cf24ec29` — split into `resolveNormalSpendingConfirmFunding` / `resolveSendAllSpendingConfirmFunding`, so the `@Suppress("LongMethod")` is gone.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -211,73 +218,236 @@
...
+                if (paid) {
+                    // Emit from this job (not a nested launch) so navigation is not raced/lost.
+                    transferEffects.emit(TransferEffect.OnSpendingFundingPaid)
+                } else {
+                    _spendingUiState.update { it.copy(isConfirmPaying = false) }
                 }
+            } catch (e: CancellationException) {
+                _spendingUiState.update { it.copy(isConfirmPaying = false) }
+                throw e
+            } catch (e: Throwable) {
```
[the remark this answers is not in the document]
piotr-iohk: [9a8e005](https://github.com/synonymdev/bitkit-android/pull/1091/commits/9a8e005ff9effbd657bfdd50f6ca037cf8feb473)

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -211,73 +218,238 @@ class TransferViewModel @Inject constructor(
...
+                    transferEffects.emit(TransferEffect.OnSpendingFundingPaid)
+                } else {
+                    _spendingUiState.update { it.copy(isConfirmPaying = false) }
                 }
+            } catch (e: CancellationException) {
+                _spendingUiState.update { it.copy(isConfirmPaying = false) }
+                throw e
+            } finally {
+                confirmPayJob = null
             }
```
jvsena42, line 282: it would be cleaner with `runSuspendCatching`, like:

```
    val paid = runSuspendCatching {
        paySpendingConfirmOrder(order)
    }.onFailure {
        Logger.error("Failed to pay spending confirm order", it, context = TAG)
        ToastEventBus.send(it)
    }.getOrDefault(false)

    if (paid) {
        // Emit from this job (not a nested launch) so navigation is not raced/lost.
        transferEffects.emit(TransferEffect.OnSpendingFundingPaid)
    } else {
        _spendingUiState.update { it.copy(isConfirmPaying = false) }
    }
```

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -649,8 +649,10 @@ private fun RootNavHost(
     val scope = rememberCoroutineScope()
     LaunchedEffect(Unit) {
         transferViewModel.transferEffects.collect { effect ->
-            if (effect is TransferEffect.OnHwTxSigned) {
-                navController.navigateTo(Routes.SpendingHwSigned)
+            when (effect) {
+                TransferEffect.OnHwTxSigned -> navController.navigateTo(Routes.SpendingHwSigned)
+                TransferEffect.OnSpendingFundingPaid -> navController.navigateTo(Routes.SettingUp)
```
jvsena42: Could add a test if possible, because the existing test still passes if this line is removed

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42, line 1: nit: could do a small refactor for this file extracting

`if (it is CancellationException && it !is TimeoutCancellationException) throw it`

to

```
  fun Throwable.rethrowIfCancellation() {                                                                                                                                                                      
      if (this is CancellationException && this !is TimeoutCancellationException) throw this                                                                                                                   
  }
```

or even

```
  /**                                                                                                                                                                                                          
   * Like [runSuspendCatching], but treats a [TimeoutCancellationException] from `withTimeout`                                                                                                                 
   * as a retriable failure instead of structured-concurrency cancellation.                                                                                                                                    
   */                                                                                                                                                                                                          
  @Suppress("TooGenericExceptionCaught")                                                                                                                                                                       
  suspend inline fun <T> runSuspendCatchingTimeout(crossinline block: suspend () -> T): Result<T> =                                                                                                            
      try {                                                                                                                                                                                                    
          Result.success(block())                                                                                                                                                                              
      } catch (t: TimeoutCancellationException) {                                                                                                                                                              
          Result.failure(t)                                                                                                                                                                                    
      } catch (c: CancellationException) {                                                                                                                                                                     
          throw c                                                                                                                                                                                              
      } catch (e: Throwable) {                                                                                                                                                                                 
          Result.failure(e)                                                                                                                                                                                    
      } 
```

Don't need to extend the refactor to other files, keep the PR scope

jvsena42 (changes requested):
Code is good, just need some style changes

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: Just a suggestion, I can do it later if you don't have time

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done — dropped the outer `CancellationException` catch; pay path is just `runSuspendCatching` + `finally` for `confirmPayJob`.

app/src/main/java/to/bitkit/ui/ContentView.kt
piotr-iohk: Good call. Extracted `transferEffectDestination()` and covered it in `ContentViewTest`; also assert `OnSpendingFundingPaid` emission in `TransferViewModelTest` so removing that wiring fails.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Done in this file only — private `Throwable.rethrowIfCancellation()` used by the HW timeout/`runCatching` paths. Left other modules alone to keep scope tight.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Pushed in `160d958a7`.

app/src/main/java/to/bitkit/ui/ContentView.kt
piotr-iohk: Pushed in `160d958a7`.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
piotr-iohk: Pushed in `160d958a7`.

jvsena42 (approved):

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: event the try/finally could be removed, as far I can see, but it is already good enough like this

PR #1092, opened 2026-07-17
fix: prevent crash during wallet backup

jvsena42:
This PR fixes a crash that could put the app into a restart loop shortly after launch.

### Description

The wallet backup payload was assembled outside of any error handling, so a failure while reading Paykit state escaped the backup coroutine and brought the whole app down. The automatic backup runs shortly after every launch, so an affected wallet crash-looped on startup with no way for the user to recover.

The payload is now built inside the existing failure handling. A failed read marks the backup as failed and retries later, the same way an upload failure already did. This also covers the retry action on the Backup settings screen, which previously crashed when tapped for a failed wallet backup.

The trigger seen in the wild was a stored Paykit session missing capabilities the SDK now requires, which is only reachable if Paykit was turned on from dev settings. The backup path itself is not gated, so any Paykit initialisation failure could take down the app the same way for any user.

#### Relationship to #1094

This branch was cut before #1094 merged, which is why the crash above was reachable at all: the backup scope had no exception handler, so the throw reached the system default handler and killed the process. #1094 has since been merged into this branch and now contains that throw.

The two fixes are complementary rather than overlapping. #1094 stops the process dying, but on its own it leaves the failure only logged, and the backup then retries every five seconds indefinitely because the category is never marked as failed. This PR is the concrete instance of the follow-up #1094 called out: guarding the specific background body so the failure is handled meaningfully instead of only logged.

The retry action on the Backup settings screen runs on the screen's view-model scope, which #1094 does not cover, so that path was fatal before #1094 and stays fatal without this PR.

Recovering a stale session and broader coroutine error handling are follow-ups (#1093). That includes the Paykit dependencies resolved when backup observation starts, which also run on a view-model scope and are still unguarded. This PR only stops the crash, so an affected wallet will report a failed wallet backup instead of crashing.

### Preview

N/A — no user-facing UI changes.

#### Manual Tests
- [x] **1.** `regression:` Fresh launch with wallet activity → wait for automatic backup: Settings → Backup shows wallet backup synced, no crash.
- [x] **2.** `regression:` Settings → Backup → tap retry on a failed wallet backup: retries without crashing.
- [x] **3.** Dev settings → Paykit enabled with a stale session → launch and wait for automatic backup: app stays up, wallet backup reports a failure once, and does not retry in a loop. With #1094 merged in, an unfixed build no longer crashes here, so check the session log for a backup attempt repeating every five seconds rather than for a process death.

#### Automated Checks
- Unit tests added: cover the wallet backup marking a failure instead of crashing when the Paykit snapshot fails, in `BackupRepoTest.kt`. Verified failing before the fix and passing after.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/BackupRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/BackupRepo.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import com.synonym.bitkitcore.migrateBackupActivitiesJson
 import com.synonym.bitkitcore.migrateBackupActivityTagsJson
 import com.synonym.bitkitcore.migrateBackupPreActivityMetadataJson
+import com.synonym.vssclient.VssItem
 import dagger.hilt.android.qualifiers.ApplicationContext
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.FlowPreview
@@ -425,7 +426,7 @@ class BackupRepo @Inject constructor(
         }
     }
 
-    suspend fun triggerBackup(category: BackupCategory) = withContext(ioDispatcher) {
+    suspend fun triggerBackup(category: BackupCategory): Result<VssItem> = withContext(ioDispatcher) {
         Logger.debug("Backup starting for: '$category'", context = TAG)
 
         val backupRequired = currentTimeMillis()
@@ -437,7 +438,7 @@ class BackupRepo @Inject constructor(
 
         val data = runSuspendCatching { getBackupDataBytes(category) }
             .onFailure { markBackupFailed(category, backupRequired, it) }
-            .getOrNull() ?: return@withContext
+            .getOrElse { return@withContext Result.failure(it) }
 
         vssBackupClient.putObject(key = category.name, data = data)
             .onSuccess {
--- a/app/src/main/java/to/bitkit/repositories/BackupRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/BackupRepo.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import com.synonym.bitkitcore.migrateBackupActivitiesJson
 import com.synonym.bitkitcore.migrateBackupActivityTagsJson
 import com.synonym.bitkitcore.migrateBackupPreActivityMetadataJson
-import com.synonym.vssclient.VssItem
 import dagger.hilt.android.qualifiers.ApplicationContext
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.FlowPreview
@@ -426,7 +425,7 @@ class BackupRepo @Inject constructor(
         }
     }
 
-    suspend fun triggerBackup(category: BackupCategory): Result<VssItem> = withContext(ioDispatcher) {
+    suspend fun triggerBackup(category: BackupCategory): Result<Unit> = withContext(ioDispatcher) {
         Logger.debug("Backup starting for: '$category'", context = TAG)
 
         val backupRequired = currentTimeMillis()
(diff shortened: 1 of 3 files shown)


jvsena42:
Waiting for upstream branch merge

app/src/main/java/to/bitkit/repositories/BackupRepo.kt
```diff
@@ -435,7 +436,11 @@ class BackupRepo @Inject constructor(
             it.copy(running = true, required = backupRequired)
         }
 
-        vssBackupClient.putObject(key = category.name, data = getBackupDataBytes(category))
+        val data = runSuspendCatching { getBackupDataBytes(category) }
+            .onFailure { markBackupFailed(category, backupRequired, it) }
+            .getOrNull() ?: return@withContext
```
ben-kaufman: `triggerBackup` now infers `Unit` because this path returns without a value, but it returned `Result<VssItem>` before this change. That makes a snapshot failure look like successful completion to callers and changes the method signature. Could we return the snapshot failure as a `Result` and declare the return type explicitly?

app/src/test/java/to/bitkit/repositories/BackupRepoTest.kt
```diff
@@ -111,6 +112,35 @@ class BackupRepoTest : BaseUnitTest() {
...
+                    required = 2_000,
+                ),
+            )
+        )
+        val allowWalletClear = CompletableDeferred<Unit>().apply { complete(Unit) }
+        stubBackupStatuses(backupStatuses, allowWalletClear) {}
+        stubBackupObservers()
+
+        try {
+            sut.startObservingBackups()
```
ben-kaufman, line 133: This test still passes if `startObservingBackups()` is removed, so it can pass without calling `backupSnapshot()`. The test clock is also `1_000`, which makes `triggerBackup()` set `required` equal to `synced` instead of leaving a failed backup pending. Could we set the clock after `synced`, verify `backupSnapshot()` is called once, assert the status is still required with `running = false`, and advance another debounce to confirm it is not retried?

jvsena42:
Tested from https://github.com/synonymdev/bitkit-android/actions/runs/29591563331 . My app stopped crashing

app/src/main/java/to/bitkit/repositories/BackupRepo.kt
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1092/commits/854544c2ea130f32dadaf3ab1704d89bd67850c0

app/src/test/java/to/bitkit/repositories/BackupRepoTest.kt
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1092/commits/854544c2ea130f32dadaf3ab1704d89bd67850c0

app/src/main/java/to/bitkit/repositories/BackupRepo.kt
```diff
@@ -424,7 +425,7 @@ class BackupRepo @Inject constructor(
         }
     }
 
-    suspend fun triggerBackup(category: BackupCategory) = withContext(ioDispatcher) {
+    suspend fun triggerBackup(category: BackupCategory): Result<Unit> = withContext(ioDispatcher) {
```
ben-kaufman, line 428: The cleanup commit changes `triggerBackup` back from the base `Result<VssItem>` contract to `Result<Unit>` and drops the successful VSS item with `.map {}`. This reintroduces the API change from the previous thread after `854544c2e` had fixed it. Could we keep `Result<VssItem>` here and return the `putObject` result directly while preserving the snapshot `Result.failure` path?

jvsena42: As far as I could see, the internal type wasn't used anywhere, that is why it was removed in the cleanup. I'll double check

ben-kaufman (approved):

app/src/main/java/to/bitkit/repositories/BackupRepo.kt
jvsena42: Confirmed, I is not used anywhere. `putObject` already logs the version ` Logger.verbose("VSS 'putObject' success for '$key' at version: '${it.version}'", context = TAG)`, so no log info is lost

PR #1084, opened 2026-07-15
feat: add Paykit server accounts

ben-kaufman:
### Description

This PR builds on #1066 to add Paykit Server watch-only accounts:

- Presents a dedicated watch-only consent step before the normal Pubky authorization screen.
- Requires `x-bitkit-claim=watch-only-account-v1` and the exact `/pub/paykit/v0/bitkit/server/:rw` capability, rejecting missing, unknown, duplicate, or mismatched companion parameters.
- Uses Paykit `approveAuthWithCompanionClaim`, failing closed if companion delivery or normal authorization fails.
- Allocates monotonic BIP84 accounts, reuses the same incomplete account on retry, and exports the exact versioned 84-byte claim payload.
- Automatically names new accounts from the requesting service; users can rename them later from Settings.
- Hides Settings → Advanced → Server Accounts until the experimental Paykit UI is enabled.
- Persists and backs up account metadata, serializes lifecycle and reconciliation changes, and supports disabling or restoring LDK tracking without deleting wallet state.
- Configures additional-wallet scanning and the rolling 1,000-address lookahead.
- Deterministically derives the receiver Noise key from the wallet's BIP39 seed using domain-separated HKDF-SHA256 with version, network, and receiver-path context; securely caches and verifies it so the wallet seed remains the only backup required.
- Reuses the derived Noise key across authentication and restarts, and publishes only its receiver marker when live session access becomes available.
- Updates to Paykit `0.1.0-rc37`, LDK Node `0.7.0-rc.57`, and Bitkit Core `0.4.2`; Bitkit Core now supplies trusted extended-public-key serialization, so BitcoinJ is no longer required.

The relay claim contains only the versioned account index, address type, and serialized xpub. Local names and tracking state are not disclosed. Ordinary Pubky Auth remains unchanged.

References:

- Base PR: https://github.com/synonymdev/bitkit-android/pull/1066
- Companion iOS PR: https://github.com/synonymdev/bitkit-ios/pull/630
- Server boundary ADR: https://github.com/pubky/paykit-server/blob/master/docs/ADRs/0001-minimal-paykit-server-boundaries.md#creator-setup-and-bitkit-watch-only-account-claim

### QA Notes

This slice implements ADR 0001's `watch-only-account-v1` claim, which is the currently agreed protocol. ADR 0002's proposed `watch-only-account-service-v1` delegation flow has not been agreed and is not part of the current implementation.

Reviewers can exercise the complete Bitkit-side setup and account lifecycle with this disposable test request:

```text
[...]

PR #1098, opened 2026-07-22
feat: support incoming paykit requests

ben-kaufman:
### Description

This PR builds on #1084 to support incoming Paykit payment requests:

- Updates Paykit to `0.1.0-rc39` and uses its separate public and private payment resolution APIs.
- Receives supported one-time Bitcoin payment requests while Bitkit is active, discarding expired or unsupported requests.
- Reuses the existing send confirmation and payment flow, with a Payment Request title, fixed requested amount, and requesting contact.
- After the user approves, validates that the request remains pending, consumes the private payment list, accepts the request, then submits payment.
- Uses public payment details only when no Noise channel is established and never falls back from private to public resolution.
- Persists consumed private payment-list versions before submission, treats submitted, pending, and uncertain payments as consumed, waits for newer details, and serializes private payments per contact and receiver path.
- Advertises payment-request support with the live Paykit session.

Payment proofs and receipts remain out of scope.

References:

- Base PR: #1084
- Companion iOS PR: https://github.com/synonymdev/bitkit-ios/pull/637

### Preview

N/A — the existing payment UI is reused, and no media is attached.

#### Manual Tests

- [ ] **1.** Receive a request from a linked Paykit contact while Bitkit is active: the Payment Request confirmation opens with the contact and requested amount.
- [ ] **2.** Swipe to pay: the request is accepted and payment continues through the existing send flow.
- [ ] **3.** Receive or retain a request past its expiration: it is discarded and not presented.
- [ ] **4.** Attempt another private payment before the contact publishes a newer payment list: Bitkit waits and does not reuse the consumed details or fall back to public details.

#### Automated Checks

- `PaykitPaymentRequestRepoTest.kt`: request mapping, expiry, lifecycle, and acceptance.
- `PrivatePaykitRepoTest.kt`: separate resolution, consumed-list persistence, and prevention of private payment-detail reuse.
- `AppViewModelSendFlowTest.kt`: presentation, polling, approval order, expiry, and duplicate or pending consumption.
- Gradle compilation, affected unit tests, and Detekt passed.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
@@ -1,0 +1,238 @@
+@file:OptIn(ExperimentalTime::class)
+
+package to.bitkit.repositories
+
+import com.synonym.paykit.OutboundPrivateCounterpartySendReport
+import com.synonym.paykit.PaymentRequestLifecycleState
+import com.synonym.paykit.PaymentRequestLocalRole
+import com.synonym.paykit.PaymentRequestRecord
+import com.synonym.paykit.PrivateStreamCounterpartyIntakeReport
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+import to.bitkit.di.IoDispatcher
+import to.bitkit.ext.runSuspendCatching
+import to.bitkit.models.PubkyPublicKeyFormat
+import to.bitkit.services.PaykitSdkService
+import to.bitkit.utils.AppError
+import to.bitkit.utils.Logger
+import java.math.BigDecimal
+import java.util.concurrent.atomic.AtomicLong
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.time.Clock
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
+
+data class PaykitPaymentRequestId(
+    val paymentRequestId: String,
+    val counterparty: String,
+    val counterpartyReceiverPath: String,
+)
+
+data class PaykitPaymentRequest(
+    val paymentRequestId: String,
+    val counterparty: String,
+    val counterpartyReceiverPath: String,
+    val amountValue: String,
+    val amountSats: ULong,
+    val paymentReference: String,
+    val expiresAt: Instant?,
+    val acceptedPaymentEndpointIdentifiers: List<String>,
+    val metadata: String,
+) {
+    val id: PaykitPaymentRequestId
+        get() = PaykitPaymentRequestId(paymentRequestId, counterparty, counterpartyReceiverPath)
+
+    fun isExpired(now: Instant): Boolean = expiresAt?.let { it <= now } == true
+}
+
+sealed class PaykitPaymentRequestError(message: String) : AppError(message) {
+    data object RequestUnavailable : PaykitPaymentRequestError("Payment request is unavailable")
+    data object RequestExpired : PaykitPaymentRequestError("Payment request has expired")
+}
+
+@Singleton
+class PaykitPaymentRequestRepo @Inject constructor(
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+    private val paykitSdkService: PaykitSdkService,
+    private val clock: Clock,
+) {
+    companion object {
+        private const val TAG = "PaykitPaymentRequestRepo"
+    }
+
+    private val operationMutex = Mutex()
+    private val stateGeneration = AtomicLong()
+    private val repoScope = CoroutineScope(SupervisorJob() + ioDispatcher)
+    private var expirationJob: Job? = null
+    private val _pendingRequests = MutableStateFlow<List<PaykitPaymentRequest>>(emptyList())
+    val pendingRequests: StateFlow<List<PaykitPaymentRequest>> = _pendingRequests.asStateFlow()
+
+    suspend fun refresh(): Result<Unit> {
+        val generation = stateGeneration.get()
+        return withContext(ioDispatcher) {
+            runSuspendCatching {
+                operationMutex.withLock {
+                    runSuspendCatching { synchronizeLocked(generation) }
+                        .onFailure { discardExpiredRequestsLocked() }
+                        .getOrThrow()
+                }
+            }.onFailure {
+                Logger.warn("Failed to refresh incoming Paykit payment requests", it, context = TAG)
+            }
+        }
+    }
+
+    suspend fun accept(request: PaykitPaymentRequest): Result<Unit> = updateRequest(request) {
+        paykitSdkService.acceptPaymentRequest(
+            counterparty = it.counterparty,
+            counterpartyReceiverPath = it.counterpartyReceiverPath,
+            paymentRequestId = it.paymentRequestId,
+        )
+    }.onFailure {
+        Logger.warn("Failed to accept incoming Paykit payment request", it, context = TAG)
+    }
+
+    suspend fun clear() {
+        stateGeneration.incrementAndGet()
+        withContext(ioDispatcher) {
+            operationMutex.withLock {
+                expirationJob?.cancel()
+                expirationJob = null
+                _pendingRequests.update { emptyList() }
+            }
+        }
+    }
+
+    private suspend fun synchronizeLocked(generation: Long) {
+        processPendingMessages()
+        paykitSdkService.receivePrivateMessagesFromLinkedPeers().also(::logIntakeFailures)
+        val now = clock.now()
+        val requests = paykitSdkService.actionableReceivedPaymentRequests().mapNotNull {
+            it.toPaykitPaymentRequest(now)
+        }
+        if (stateGeneration.get() != generation) return
+        _pendingRequests.update { requests }
+        scheduleExpirationLocked()
+    }
+
+    private suspend fun updateRequest(
+        request: PaykitPaymentRequest,
+        operation: suspend (PaykitPaymentRequest) -> Unit,
+    ): Result<Unit> = withContext(ioDispatcher) {
+        runSuspendCatching {
+            operationMutex.withLock {
+                if (request.isExpired(clock.now())) {
+                    discardExpiredRequestsLocked()
+                    throw PaykitPaymentRequestError.RequestExpired
+                }
+                val current = _pendingRequests.value.firstOrNull { it.id == request.id }
+                    ?: throw PaykitPaymentRequestError.RequestUnavailable
+
+                operation(current)
+                _pendingRequests.update { requests -> requests.filterNot { it.id == current.id } }
+                discardExpiredRequestsLocked()
+                processPendingMessages()
+            }
+        }
+    }
+
+    private suspend fun processPendingMessages() {
+        runSuspendCatching { paykitSdkService.processPendingPrivateMessages() }
+            .onSuccess(::logOutboundFailures)
+            .onFailure { Logger.warn("Failed to deliver pending Paykit private messages", it, context = TAG) }
+    }
+
+    private fun logOutboundFailures(reports: List<OutboundPrivateCounterpartySendReport>) {
+        reports.forEach {
+            val error = it.error ?: return@forEach
+            Logger.warn(
+                "Failed to deliver Paykit private messages to '${PubkyPublicKeyFormat.redacted(it.counterparty)}': " +
+                    "'${error.redactedContext()}'",
+                context = TAG,
+            )
+        }
+    }
+
+    private fun logIntakeFailures(reports: List<PrivateStreamCounterpartyIntakeReport>) {
+        reports.forEach {
+            val error = it.error ?: return@forEach
+            Logger.warn(
+                "Failed to receive Paykit private messages from '${PubkyPublicKeyFormat.redacted(it.counterparty)}': " +
+                    "'${error.redactedContext()}'",
+                context = TAG,
+            )
+        }
+    }
+
+    private fun discardExpiredRequestsLocked() {
+        val now = clock.now()
+        _pendingRequests.update { requests -> requests.filterNot { it.isExpired(now) } }
+        scheduleExpirationLocked()
+    }
+
+    private fun scheduleExpirationLocked() {
+        expirationJob?.cancel()
+        expirationJob = null
+
+        val nextExpiration = _pendingRequests.value.mapNotNull { it.expiresAt }.minOrNull() ?: return
+        val delayDuration = (nextExpiration - clock.now()).coerceAtLeast(Duration.ZERO)
+        expirationJob = repoScope.launch {
+            delay(delayDuration)
+            operationMutex.withLock {
+                expirationJob = null
+                discardExpiredRequestsLocked()
+            }
+        }
+    }
+}
+
+private val bitcoinAmountPattern = Regex("(?:[0-9]+(?:\\.[0-9]*)?|\\.[0-9]+)")
+
+@Suppress("ReturnCount")
+private fun PaymentRequestRecord.toPaykitPaymentRequest(now: Instant): PaykitPaymentRequest? {
+    if (localRole != PaymentRequestLocalRole.PAYER || state != PaymentRequestLifecycleState.PROPOSED) return null
+    val requestTerms = terms ?: return null
+    if (requestTerms.recurrence != null || requestTerms.amount.asset != "btc") return null
+    val amountSats = requestTerms.amount.value.toSats() ?: return null
+    val endpoints = requestTerms.acceptedPaymentEndpointIdentifiers
+        .filter { MethodId.fromRawValue(it) != null }
+        .distinct()
+    if (endpoints.isEmpty()) return null
+
+    val expiresAt = requestTerms.proposalExpiresAt?.let {
+        runCatching { Instant.parse(it) }.getOrNull() ?: return null
+    }
+    if (expiresAt != null && expiresAt <= now) return null
+
+    return PaykitPaymentRequest(
+        paymentRequestId = paymentRequestId,
+        counterparty = counterparty,
+        counterpartyReceiverPath = counterpartyReceiverPath,
+        amountValue = requestTerms.amount.value,
+        amountSats = amountSats,
+        paymentReference = requestTerms.paymentReference.exportText(),
+        expiresAt = expiresAt,
+        acceptedPaymentEndpointIdentifiers = endpoints,
+        metadata = requestTerms.metadata.exportText(),
+    )
+}
+
+private fun String.toSats(): ULong? {
+    if (!bitcoinAmountPattern.matches(this)) return null
+    return runCatching {
+        BigDecimal(this).movePointRight(8).toBigIntegerExact().toString().toULong()
+    }.getOrNull()?.takeIf { it > 0uL }
+}
--- a/app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
@@ -56,6 +56,14 @@ data class PaykitPaymentRequest(
         get() = PaykitPaymentRequestId(paymentRequestId, counterparty, counterpartyReceiverPath)
 
     fun isExpired(now: Instant): Boolean = expiresAt?.let { it <= now } == true
+
+    fun acceptsLightningInvoiceAmountMsats(amountMsats: ULong?): Boolean =
+        amountMsats == null || amountSats <= ULong.MAX_VALUE / 1000uL && amountMsats == amountSats * 1000uL
+
+    fun acceptsLightningInvoiceAmountSats(amountSats: ULong): Boolean =
+        amountSats == 0uL || acceptsPaymentAmount(amountSats)
+
+    fun acceptsPaymentAmount(amountSats: ULong): Boolean = amountSats == this.amountSats
 }
 
 sealed class PaykitPaymentRequestError(message: String) : AppError(message) {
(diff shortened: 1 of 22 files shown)


app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -428,73 +467,175 @@ class PrivatePaykitRepo @Inject constructor(
                 if (backup == null) {
                     paykitSdkService.clearState()
                 } else {
-                    paykitSdkService.restoreBackupState(backup)
+                    val decoded = json.decodeFromString<PrivatePaykitBackup>(backup)
```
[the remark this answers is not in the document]
ben-kaufman: No code change here: this raw backup format only exists in the unshipped parent Paykit work, so there are no production backups to migrate. Per the pre-release scope of this stack, we are intentionally not adding migration or backward-compatibility code.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -578,6 +596,84 @@ class AppViewModel @Inject constructor(
...
+                    currentSheet.value is Sheet.Send &&
+                    requests.none { it.id == activeRequest.id }
+                ) {
+                    hideSheet()
+                }
+            }
+        }
+    }
+
+    private suspend fun presentNextIncomingPaykitPaymentRequest() {
```
jvsena42, line 671: similar issue of https://github.com/synonymdev/bitkit-ios/pull/637#discussion_r3689850056

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -578,6 +596,84 @@ class AppViewModel @Inject constructor(
...
+                .collect { refreshIncomingPaykitPaymentRequests() }
+        }
+    }
+
+    private suspend fun refreshIncomingPaykitPaymentRequests() {
+        if (!isPaykitEnabled.value || pubkyRepo.publicKey.value == null || !walletRepo.walletExists()) return
+        paykitPaymentRequestRepo.refresh().onSuccess { presentNextIncomingPaykitPaymentRequest() }
+    }
+
+    fun startPaykitPaymentRequestPolling() {
```
jvsena42, line 627: Same https://github.com/synonymdev/bitkit-ios/pull/637#discussion_r3689912390

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2346,9 +2524,17 @@ class AppViewModel @Inject constructor(
     }
 
     @Suppress("LongMethod")
-    private suspend fun proceedWithPayment() {
+    private suspend fun proceedWithPayment(contactPaymentContext: ContactPaymentContext?) {
         delay(SCREEN_TRANSITION_DELAY) // wait for screen transitions when applicable
 
+        if (!validateAndAcceptIncomingPaymentRequest(contactPaymentContext)) return
+
+        consumePrivatePaymentListIfNeeded(contactPaymentContext).onFailure {
```
jvsena42, line 2592: Consuming the payment list before submission burns it on any send failure

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
jvsena42: this method has no callers. Dead code?

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -0,0 +1,251 @@
...
+    return PaykitPaymentRequest(
+        paymentRequestId = paymentRequestId,
+        counterparty = counterparty,
+        counterpartyReceiverPath = counterpartyReceiverPath,
+        amountValue = requestTerms.amount.value,
+        amountSats = amountSats,
+        paymentReference = requestTerms.paymentReference.exportText(),
+        expiresAt = expiresAt,
+        acceptedPaymentEndpointIdentifiers = endpoints,
+        metadata = requestTerms.metadata.exportText(),
```
jvsena42: `paymentReference` and `metadata` are not used. Should the be used in the branch?

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -427,72 +467,183 @@ class PrivatePaykitRepo @Inject constructor(
...
+                    receiverPath = receiverPath,
+                    consumedVersion = consumedVersion,
+                    amount = amount,
+                    allowPublicResolution = paymentRequest == null,
+                )
+                    ?: if (paymentRequest == null) {
+                        return@runSuspendCatching publicPaykitRepo.beginPayment(publicKey).getOrThrow()
+                    } else {
+                        throw PrivatePaykitError.PrivateUnavailable
                     }
```
jvsena42: this branch is unreachable

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -427,72 +467,183 @@ class PrivatePaykitRepo @Inject constructor(
...
-                } else {
-                    PublicPaykitPaymentResult.NotOpened
-                }
-            }
+        val privatePayable = privatePayableEndpoints(acceptedEndpoints, publicKey)
+        val paymentListVersion = resolution.privatePaymentListVersion
+        if (privatePayable.isNotEmpty() && paymentListVersion != null) {
+            return PublicPaykitPaymentResult.Opened(
+                paymentRequest = PublicPaykitRepo.paymentRequest(privatePayable),
+                privatePaymentContext = PrivatePaykitPaymentContext(receiverPath, paymentListVersion),
```
jvsena42, line 596: Should add at least a log here

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -0,0 +1,251 @@
...
+    private val paykitSdkService: PaykitSdkService,
+    private val clock: Clock,
+) {
+    companion object {
+        private const val TAG = "PaykitPaymentRequestRepo"
+    }
+
+    private val operationMutex = Mutex()
+    private val stateGeneration = AtomicLong()
+    private val repoScope = CoroutineScope(SupervisorJob() + ioDispatcher)
```
jvsena42: use `appScope(ioDispatcher, TAG)`

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2467,6 +2647,53 @@ class AppViewModel @Inject constructor(
...
+        val incomingPaymentRequest = contactPaymentContext?.incomingPaymentRequest ?: return false
+        if (!incomingPaymentRequest.acceptsPaymentAmount(_sendUiState.value.amount)) return true
+        if (_sendUiState.value.payMethod != SendMethod.LIGHTNING) return false
+
+        val lightningInvoice = _sendUiState.value.decodedInvoice ?: return false
+        return !incomingPaymentRequest.acceptsLightningInvoice(lightningInvoice)
+    }
+
+    private fun PaykitPaymentRequest.acceptsLightningInvoice(invoice: LightningInvoice): Boolean {
+        val amountMsats = runCatching { Bolt11Invoice.fromStr(invoice.bolt11).amountMilliSatoshis() }
```
jvsena42: `Bolt11Invoice.fromStr` is a native ffi call. could change the dispatcher

jvsena42 (changes requested):

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Addressed in cf14578b1. Unresolved requests now use per-request presentation retry delays of 30, 60, 120, then 300 seconds, and later payable requests are still considered in the same pass.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Addressed in cf14578b1. Bolt11Invoice.fromStr and amount extraction now run inside withContext(bgDispatcher), while the surrounding validation remains suspendable.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: Addressed in cf14578b1. PaykitPaymentRequestRepo now creates its expiration scope with appScope(ioDispatcher, TAG).

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Addressed in cf14578b1. Opening a payable private resolution now logs the redacted counterparty through PrivatePaykitRepo.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Addressed in cf14578b1. A null prepared result is only possible when public resolution is allowed, so the unreachable conditional is replaced by a direct public-resolution return.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Addressed in cf14578b1. Foreground polling now backs off from 30 to 60 to 120 seconds while the actionable request list is unchanged, and resets to 30 seconds when the list changes.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: They are not used by the current flow, so cf14578b1 removes both fields from the app model and avoids exporting them during mapping. They can be introduced with the later request-details design if needed.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Yes. Removed in cf14578b1 together with its now-dead on-chain filtering helper and test stubs. The lightning equivalent remains live.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: This is intentional and required by the rc39 private-payment contract. Using any private endpoint consumes the entire Private Payment List, including submitted, pending, or uncertain outcomes such as a send failure after invocation. Persisting the version before submission prevents reuse; the next private payment waits for a newer list and never falls back to public details.

ben-kaufman:
Restacked onto the current #1084 head after its force-push. The request/review-fix commits are signed; the retry fix previously referenced as `cf14578b1` is now `e01a3ce25`, with `44d58d2b8` preserving the Paykit rc39 integration across the rewritten base.

The current `build` failure is inherited unchanged from #1084: both jobs fail in `WatchOnlyAccountStore.kt` on unresolved `serializedExtendedPubkey` references at lines 6 and 29. I left #1084 untouched, as it is being handled separately.

jvsena42 (commented):
Re-reviewed the changes since my last pass. All 8 comments I raised are addressed — resolved those threads. Verified locally on `796bc2b1`: the Paykit unit tests (`PaykitPaymentRequestRepoTest`, `PrivatePaykitRepoTest`, `PublicPaykitRepoTest`, `AppViewModelSendFlowTest`, `PaykitSdkServiceTest`, `ContactPaymentSettingsRepoTest`) all pass and `detekt` is clean.

A few new findings below. Only the first one blocks.

One extra nit that has no diff line to attach to: `PrivatePaykitRepoTest.kt:32` — `import org.mockito.kotlin.anyOrNull` is now unused (base had 4 usages, this branch has 0). `detekt` reports it as `NoUnusedImports`, but `ignoreFailures = true` in `app/build.gradle.kts` keeps CI green.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1347,10 +1360,10 @@ class PrivatePaykitRepo @Inject constructor(
...
     }
 
-    private suspend fun hasPrivatePaymentAccessForCurrentProfile(): Boolean = runSuspendCatching {
-        pubkyService.currentPublicKey() ?: return@runSuspendCatching false
-        paykitSdkService.hasPrivatePaymentAccess()
-    }.getOrDefault(false)
+    private suspend fun hasPrivatePaymentAccessForCurrentProfile(): Boolean {
+        pubkyService.currentPublicKey() ?: return false
+        return paykitSdkService.hasPrivatePaymentAccess()
+    }
```
jvsena42: This dropped the `runSuspendCatching {}.getOrDefault(false)` that wrapped the helper on `master`, and that turns a settings toggle into a crash path.

`pubkyService.currentPublicKey()` reaches `PaykitSdkService.currentPublicKey()` → `handle()` + `handle.initialize()`, which throws `PaykitException`. The public `hasPrivatePaymentAccess()` (line 121) is called by `ContactPaymentSettingsRepo.enable()` **outside** its `runSuspendCatching`:

```kotlin
val canUsePrivateContactPayments = privatePaykitRepo.hasPrivatePaymentAccess()  // throws
return runSuspendCatching { ... }
```

So the throw escapes `setEnabled()`'s `Result<Unit>` contract and lands in `SettingsViewModel.setContactPaymentsEnabled`'s bare `viewModelScope.launch` — uncaught coroutine exception, and `_isUpdatingContactPayments` stays stuck at `true`. `PayContactsViewModel.continueToProfile` has a `finally` so it only loses the toast, but the exception still escapes.

The other callers (`canPublishPrivateEndpoints`, `prepareRelevantPrivateLinksIfAvailable`) are all inside a `runSuspendCatching`, so `hasPrivatePaymentAccess()` is the single exposure — but it's user-reachable. Please restore the guard:

```kotlin
private suspend fun hasPrivatePaymentAccessForCurrentProfile(): Boolean = runSuspendCatching {
    pubkyService.currentPublicKey() ?: return@runSuspendCatching false
    paykitSdkService.hasPrivatePaymentAccess()
}.getOrDefault(false)
```

app/src/main/java/to/bitkit/ui/screens/contacts/ContactDetailViewModel.kt
```diff
@@ -105,11 +106,19 @@ class ContactDetailViewModel @Inject constructor(
...
+                                    publicKey,
+                                    result.privatePaymentContext,
+                                )
+                            )
                         PublicPaykitPaymentResult.NoEndpoint ->
                             showPayError(R.string.slashtags__error_pay_empty_msg)
                         PublicPaykitPaymentResult.NotOpened ->
                             showPayError(R.string.slashtags__error_pay_not_opened_msg)
+                        PublicPaykitPaymentResult.WaitingForUpdatedPaymentList ->
+                            showPayError(R.string.slashtags__error_pay_empty_msg)
```
jvsena42: `slashtags__error_pay_empty_msg` means "no payment endpoint found", but the actual state here is "waiting for the contact to publish an updated payment list" — a transient, retryable condition the user could act on by waiting. Worth a dedicated string.

The same branch is duplicated verbatim in `SendContactSelectViewModel` and `AddContactViewModel`; folding the `PublicPaykitPaymentResult` → message mapping into one shared helper would keep the three in sync.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2352,9 +2579,17 @@ class AppViewModel @Inject constructor(
     }
 
     @Suppress("LongMethod")
-    private suspend fun proceedWithPayment() {
+    private suspend fun proceedWithPayment(contactPaymentContext: ContactPaymentContext?) {
         delay(SCREEN_TRANSITION_DELAY) // wait for screen transitions when applicable
 
+        if (!validateAndAcceptIncomingPaymentRequest(contactPaymentContext)) return
+
+        consumePrivatePaymentListIfNeeded(contactPaymentContext).onFailure {
```
jvsena42: Separate from my earlier consume-before-send comment (which you answered — the rc39 contract makes that unavoidable, understood): the concern here is the window *between* these two calls.

`validateAndAcceptIncomingPaymentRequest` accepts the request on the Paykit side and removes it from `pendingRequests`. If `consumePrivatePaymentListIfNeeded` then fails (e.g. `PaymentListAlreadyConsumed`), we've accepted a request, made no payment, and there's no path left to retry or decline it — it's simply gone.

Consuming first (or accepting last) would close that gap. `incoming payment request is accepted before its private list is consumed` pins the current order, so I assume this is deliberate — just want to confirm it's the order you want.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3416,6 +3682,13 @@ class AppViewModel @Inject constructor(
         private const val AUTH_CHECK_SPLASH_DELAY_MS = 500L
         private const val ADDRESS_VALIDATION_DEBOUNCE_MS = 1000L
         private const val PAYKIT_CHANNEL_USABILITY_REFRESH_DELAY_MS = 5_000L
+        private val PAYKIT_PAYMENT_REQUEST_REFRESH_INTERVALS = listOf(30.seconds, 60.seconds, 120.seconds)
+        private val PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS = listOf(
+            30.seconds,
+            60.seconds,
+            120.seconds,
+            300.seconds,
+        )
```
jvsena42, line 3707: The retry schedule saturates at 300s and never gives up. A request with `expiresAt == null` that never resolves will keep issuing `beginPaymentRequest` network calls every 5 minutes for as long as the app is foregrounded.

A max-attempt cap — after which the request falls back to the normal poll cycle — would bound it.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -580,6 +600,135 @@ class AppViewModel @Inject constructor(
...
+            isPresentingPaymentRequest = false
+        }
+    }
+
+    private suspend fun openIncomingPaymentRequestIfAvailable(request: PaykitPaymentRequest): Boolean {
+        val result = privatePaykitRepo.beginPaymentRequest(request).getOrNull()
+        if (currentSheet.value != null || hasActiveContactPaymentContext()) return true
+        if (result !is PublicPaykitPaymentResult.Opened || !paykitPaymentRequestRepo.isPending(request)) {
+            if (paykitPaymentRequestRepo.isPending(request)) deferPaymentRequestPresentation(request)
+            return false
```
jvsena42: Nit: `paykitPaymentRequestRepo.isPending(request)` is evaluated twice on adjacent lines — worth hoisting into a local.

Also, the `Boolean` return means two different things: `true` at line 689 is "abort, a sheet opened" while `true` at line 703 is "presented". The caller treats both as `return`, so behaviour is right, but the name `openIncomingPaymentRequestIfAvailable` reads like it returns "did I open it". A small sealed result, or a rename to something like `presentIncomingPaymentRequestOrStop`, would make the loop easier to follow.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -0,0 +1,246 @@
...
+    val expiresAt: Instant?,
+    val acceptedPaymentEndpointIdentifiers: List<String>,
+) {
+    val id: PaykitPaymentRequestId
+        get() = PaykitPaymentRequestId(paymentRequestId, counterparty, counterpartyReceiverPath)
+
+    fun isExpired(now: Instant): Boolean = expiresAt?.let { it <= now } == true
+
+    fun acceptsLightningInvoiceAmountMsats(amountMsats: ULong?): Boolean =
+        amountMsats == null || amountSats <= ULong.MAX_VALUE / 1000uL && amountMsats == amountSats * 1000uL
```
jvsena42: Nit: the `amountSats <= ULong.MAX_VALUE / 1000uL` guard is already enforced at construction by `toSats()` (`takeIf { it <= ULong.MAX_VALUE / 1000uL }`), so it's dead weight here. The unparenthesized `||`/`&&` mix also reads ambiguously even though the precedence is correct — parentheses around the `&&` clause would help.

app/src/main/java/to/bitkit/ui/screens/contacts/ContactDetailViewModel.kt
ben-kaufman: Addressed in `5866dc2ef`. `WaitingForUpdatedPaymentList` now uses a dedicated message explaining that the contact needs to publish updated payment details, and the same resource is used in all three flows. I kept the result handling local because the effects differ and extracting a helper for one shared resource choice would add more indirection than it removes.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: Addressed in `5866dc2ef`. The local overflow guard and mixed `||`/`&&` expression are gone; the comparison now uses the existing `satsToMsat` conversion helper, which keeps overflow-safe arithmetic explicit.

ben-kaufman:
All fresh review findings are addressed in `5866dc2ef`, including the no-diff-line `anyOrNull` import nit. The focused behavior changes have regression coverage, and detekt passes with only the repository's existing unrelated warnings. Local Gradle unit compilation was blocked at dependency resolution because this machine's GitHub Packages token lacks `read:packages`; the pushed CI run is the compile/test gate.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -580,6 +600,137 @@ class AppViewModel @Inject constructor(
...
+                val retryAttempts = paymentRequestPresentationRetryAttempts[request.id] ?: 0
+                request.id !in presentedPaymentRequestIds &&
+                    retryAttempts <= PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS.size &&
+                    paymentRequestPresentationRetryJobs[request.id]?.isActive != true
+            }) {
+                if (presentIncomingPaymentRequestOrStop(request)) return
+            }
+        } finally {
+            isPresentingPaymentRequest = false
+        }
```
jvsena42, line 687: could have a log on `retryAttempts > PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS.size ` give up

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2352,9 +2581,21 @@ class AppViewModel @Inject constructor(
...
+            handlePaymentPreparationFailure(contactPaymentContext, it)
+            return
+        }
+
+        acceptIncomingPaymentRequestIfNeeded(contactPaymentContext).onFailure {
+            handlePaymentPreparationFailure(contactPaymentContext, it)
+            return
+        }
+
         val amount = _sendUiState.value.amount
```
jvsena42, line 2602: nit: Could update the PR description  with the consumePrivatePaymentList → acceptIncomingPaymentRequestIfNeeded. 

The test was correctly renamed and updated

jvsena42 (changes requested):
Only one finding worth adressing from last changes

jvsena42 (approved):
utAck, implementation still not end-to-end testable

Issue #638, opened 2026-01-16
[Bug]:  Random toggle in screen after confirming PIN change


86667:
<p align="center">
<img src="https://github.com/user-attachments/assets/1543bf83-d97f-4339-8e90-3a8a346db967" width="25%">
<img src="https://github.com/user-attachments/assets/8d0cb725-3aae-4ea5-b77b-e7b7147e8c7f" width="25%">
</p>

PR #1099, opened 2026-07-22
refactor: batch paykit contact cleanup

ben-kaufman:
### Description

This PR builds on #1098 to batch and harden private Paykit contact cleanup:

- Reads one shared SDK linked-peer snapshot per multi-contact publication or cleanup pass, with one batch-level retry instead of per-contact fallback reads.
- Clears receiver-scoped payment lists, drains outbound private messages, and checks pending status once for the whole batch.
- Processes all pending deleted contacts through the collection cleanup path instead of repeating snapshots and drains per contact.
- Serializes publication and cleanup with the same mutex and preserves endpoint cache state that changes while remote cleanup is suspended.
- Keeps per-contact failure isolation for attributable clear failures: successful contacts are cleared while failed contacts remain cached for retry.
- Drops malformed persisted cleanup keys with a warning because they cannot address SDK state, allowing cleanup to converge.
- Batches deleted-contact marker updates into one DataStore write.

Error-path behavior is intentionally conservative:

- If linked-peer inspection fails twice, cleanup still clears locally tracked published paths, but retains all affected contacts for retry.
- If the shared post-drain inspection fails, the whole batch remains cached and pending because delivery status cannot be attributed safely per contact.

### Preview

N/A — repository-only refactor.

#### Manual Tests

N/A

#### Automated Checks

- `PrivatePaykitRepoTest.kt`: all 43 focused tests passed, including malformed-state recovery, shared snapshot retry, batched deleted-contact retry, drain-inspection failure, partial cleanup failure, and a concurrent cache update during remote cleanup.
- `./gradlew compileDevDebugKotlin` passed as part of the focused and full test runs.
- `./gradlew detekt` passed with only pre-existing findings outside this change.
- `git diff --check` passed.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
@@ -107,6 +107,12 @@ class PrivatePaykitRepo @Inject constructor(
         val firstError: Throwable?,
     )
 
+    private data class PrivateEndpointCleanupPreparation(
+        val clearedRetryKeys: List<PrivateMessageDrainRetryKey>,
+        val failedPublicKeys: Set<String>,
+        val firstError: Throwable?,
+    )
+
     private data class PrivateMessageDrainRetryKey(
         val publicKey: String,
         val receiverPath: String,
@@ -683,6 +689,7 @@ class PrivatePaykitRepo @Inject constructor(
         var firstError: Throwable? = null
         val updates = mutableListOf<PrivatePaymentListReservationUpdateInput>()
         val linkRetryKeys = mutableListOf<PrivateMessageDrainRetryKey>()
+        val linkedReceiverPaths = linkedReceiverPathsByPublicKey()
 
         for (publicKey in publicKeys) {
             val receiverPaths = runSuspendCatching { receiverPathsForSavedContact(publicKey) }
@@ -699,15 +706,12 @@ class PrivatePaykitRepo @Inject constructor(
             val publicationReceiverPaths = receiverPathSelection.publishableReceiverPaths
             receiverPathSelection.error?.let {
                 firstError = firstError ?: it
-                Logger.warn(
-                    "Failed to inspect private Paykit receiver markers for '${redacted(publicKey)}' during '$reason'",
-                    it,
-                    context = TAG,
-                )
+                logPrivateReceiverPathSelectionFailure(publicKey, reason, it)
             }
             val cleanupReceiverPaths = receiverPathsForPrivateEndpointCleanup(
                 publicKey = publicKey,
                 excludedReceiverPaths = publicationReceiverPaths + receiverPathSelection.cleanupProtectedReceiverPaths,
+                linkedReceiverPaths = linkedReceiverPaths[publicKey].orEmpty(),
             )
 
             (linkableReceiverPaths + cleanupReceiverPaths).distinct().forEach { receiverPath ->
@@ -815,6 +819,18 @@ class PrivatePaykitRepo @Inject constructor(
                 context = TAG,
             )
         }
+    }
+
+    private fun logPrivateReceiverPathSelectionFailure(
+        publicKey: String,
+        reason: String,
+        error: Throwable,
+    ) {
+        Logger.warn(
+            "Failed to inspect private Paykit receiver markers for '${redacted(publicKey)}' during '$reason'",
+            error,
+            context = TAG,
+        )
     }
 
     private suspend fun applyPrivatePaymentListDeliveryReport(
@@ -1170,41 +1186,77 @@ class PrivatePaykitRepo @Inject constructor(
     }
 
     private suspend fun removePublishedEndpoints(): Result<Unit> = withContext(serializedDispatcher) {
-        runSuspendCatching {
-            val keys = (knownSavedContactKeys + ensureState().contacts.keys + pendingDeletedContactCleanupPublicKeys())
-                .distinct()
-            val firstError = keys.mapNotNull { publicKey ->
-                removePublishedEndpoints(publicKey).exceptionOrNull()
-            }.firstOrNull()
-            if (firstError != null) throw firstError
-        }
+        val keys = (knownSavedContactKeys + ensureState().contacts.keys + pendingDeletedContactCleanupPublicKeys())
+            .distinct()
+        removePublishedEndpoints(keys)
     }
 
-    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> = withContext(serializedDispatcher) {
-        runSuspendCatching {
-            var firstError: Throwable? = null
-            receiverPathsForCleanup(publicKey).forEach { receiverPath ->
-                val result = runSuspendCatching {
-                    val report = paykitSdkService.clearPrivatePaymentList(
-                        counterparty = publicKey,
-                        receiverPath = receiverPath,
+    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> =
+        removePublishedEndpoints(listOf(publicKey))
+
+    private suspend fun removePublishedEndpoints(publicKeys: Collection<String>): Result<Unit> =
+        withContext(serializedDispatcher) {
+            runSuspendCatching {
+                val publicKeys = publicKeys.mapNotNull(::normalizedPublicKey).distinct()
+                if (publicKeys.isEmpty()) return@runSuspendCatching Unit
+
+                val linkedReceiverPaths = linkedReceiverPathsByPublicKey()
+                val preparation = clearPrivatePaymentLists(publicKeys, linkedReceiverPaths)
+                val failedPublicKeys = preparation.failedPublicKeys.toMutableSet()
+                var firstError = preparation.firstError
+
+                if (preparation.clearedRetryKeys.isNotEmpty()) {
+                    drainPendingPrivateMessages(
+                        reason = "private endpoint cleanup",
+                        advancingLinksFor = preparation.clearedRetryKeys,
                     )
+                    val pendingRetryKeys = pendingPrivateMessageDrainKeys(preparation.clearedRetryKeys)
+                    if (pendingRetryKeys.isNotEmpty()) {
+                        failedPublicKeys += pendingRetryKeys.map { it.publicKey }
+                        firstError = firstError ?: PrivatePaykitError.PrivateUnavailable
+                    }
+                }
+
+                val successfulPublicKeys = publicKeys.filterNot { it in failedPublicKeys }
+                clearPublishedEndpointCache(successfulPublicKeys)
+
+                firstError?.let { throw it }
+                Unit
+            }
+        }
+
+    private suspend fun clearPrivatePaymentLists(
+        publicKeys: Collection<String>,
+        linkedReceiverPaths: Map<String, Set<String>>,
+    ): PrivateEndpointCleanupPreparation {
+        val failedPublicKeys = mutableSetOf<String>()
+        val clearedRetryKeys = mutableListOf<PrivateMessageDrainRetryKey>()
+        var firstError: Throwable? = null
+
+        publicKeys.forEach { publicKey ->
+            receiverPathsForCleanup(
+                publicKey = publicKey,
+                linkedReceiverPaths = linkedReceiverPaths[publicKey].orEmpty(),
+            ).forEach { receiverPath ->
+                runSuspendCatching {
+                    val report = paykitSdkService.clearPrivatePaymentList(publicKey, receiverPath)
                     if (report.failedToQueue.isNotEmpty() || report.failedToDeliver.isNotEmpty()) {
                         throw PrivatePaykitError.PrivateUnavailable
                     }
-                    val retryKey = PrivateMessageDrainRetryKey(publicKey, receiverPath)
-                    drainPendingPrivateMessages("private endpoint cleanup", advancingLinksFor = listOf(retryKey))
-                    if (retryKey in pendingPrivateMessageDrainKeys(listOf(retryKey))) {
-                        throw PrivatePaykitError.PrivateUnavailable
-                    }
-                }
-                if (result.isFailure) {
-                    firstError = firstError ?: result.exceptionOrNull()
+                }.onSuccess {
+                    clearedRetryKeys += PrivateMessageDrainRetryKey(publicKey, receiverPath)
+                }.onFailure {
+                    failedPublicKeys += publicKey
+                    firstError = firstError ?: it
                 }
             }
+        }
 
-            firstError?.let { throw it }
+        return PrivateEndpointCleanupPreparation(clearedRetryKeys, failedPublicKeys, firstError)
+    }
 
+    private suspend fun clearPublishedEndpointCache(publicKeys: Collection<String>) {
+        publicKeys.forEach { publicKey ->
             state?.contacts?.get(publicKey)?.let { contactState ->
                 contactState.remoteEndpoints = emptyList()
                 contactState.localInvoicesByReceiverPath = emptyMap()
@@ -1214,6 +1266,9 @@ class PrivatePaykitRepo @Inject constructor(
                 }
             }
             updateDeletedContactCleanupPending(publicKey, isPending = false)
+        }
+
+        if (publicKeys.isNotEmpty()) {
             persistState(markWalletBackup = true)
         }
     }
@@ -1226,42 +1281,38 @@ class PrivatePaykitRepo @Inject constructor(
         return paths.ifEmpty { listOf(PaykitReceiverPaths.WALLET) }
     }
 
-    private suspend fun receiverPathsForPrivateEndpointCleanup(
+    private fun receiverPathsForPrivateEndpointCleanup(
         publicKey: String,
         excludedReceiverPaths: List<String>,
+        linkedReceiverPaths: Collection<String>,
     ): List<String> {
         val publishedPaths = publishedPrivatePaymentReceiverPaths(publicKey)
-        val linkedPaths = linkedReceiverPaths(publicKey)
-        return (publishedPaths + linkedPaths)
+        return (publishedPaths + linkedReceiverPaths)
             .filter { it in PaykitReceiverPaths.supported }
             .filterNot { it in excludedReceiverPaths }
             .distinct()
             .sorted()
     }
 
-    private suspend fun receiverPathsForCleanup(publicKey: String): List<String> {
-        val paths = (
-            linkedReceiverPaths(publicKey) +
-                publishedPrivatePaymentReceiverPaths(publicKey)
-            )
+    private fun receiverPathsForCleanup(
+        publicKey: String,
+        linkedReceiverPaths: Collection<String>,
+    ): List<String> {
+        return (linkedReceiverPaths + publishedPrivatePaymentReceiverPaths(publicKey))
             .filter { it in PaykitReceiverPaths.supported }
             .distinct()
             .sorted()
-        return paths
     }
 
-    private suspend fun linkedReceiverPaths(publicKey: String): List<String> {
-        val normalizedKey = normalizedPublicKey(publicKey) ?: return emptyList()
-        val paths = paykitSdkService.linkedPeers()
-            .mapNotNull { peer ->
-                val peerKey = normalizedPublicKey(peer.counterparty)
-                peer.counterpartyReceiverPath.takeIf {
-                    peerKey == normalizedKey && it in PaykitReceiverPaths.supported
-                }
+    private suspend fun linkedReceiverPathsByPublicKey(): Map<String, Set<String>> {
+        val linkedPaths = mutableMapOf<String, MutableSet<String>>()
+        paykitSdkService.linkedPeers().forEach { peer ->
+            val publicKey = normalizedPublicKey(peer.counterparty) ?: return@forEach
+            if (peer.counterpartyReceiverPath in PaykitReceiverPaths.supported) {
+                linkedPaths.getOrPut(publicKey, ::mutableSetOf) += peer.counterpartyReceiverPath
             }
-            .distinct()
-            .sorted()
-        return paths
+        }
+        return linkedPaths
     }
 
     private fun publishedPrivatePaymentReceiverPaths(publicKey: String): List<String> {
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
@@ -113,6 +113,16 @@ class PrivatePaykitRepo @Inject constructor(
         val firstError: Throwable?,
     )
 
+    private data class NormalizedPublicKeyBatch(
+        val publicKeys: List<String>,
+        val hadInvalidKey: Boolean,
+    )
+
+    private data class LinkedReceiverPathInspection(
+        val receiverPaths: Set<String>,
+        val error: Throwable?,
+    )
+
     private data class PrivateMessageDrainRetryKey(
         val publicKey: String,
         val receiverPath: String,
@@ -689,7 +699,7 @@ class PrivatePaykitRepo @Inject constructor(
         var firstError: Throwable? = null
         val updates = mutableListOf<PrivatePaymentListReservationUpdateInput>()
         val linkRetryKeys = mutableListOf<PrivateMessageDrainRetryKey>()
-        val linkedReceiverPaths = linkedReceiverPathsByPublicKey()
+        val linkedReceiverPathsSnapshot = linkedReceiverPathsSnapshotOrNull(reason)
 
         for (publicKey in publicKeys) {
             val receiverPaths = runSuspendCatching { receiverPathsForSavedContact(publicKey) }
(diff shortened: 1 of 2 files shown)


app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1181,41 +1197,77 @@ class PrivatePaykitRepo @Inject constructor(
...
-                        counterparty = publicKey,
-                        receiverPath = receiverPath,
+    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> =
+        removePublishedEndpoints(listOf(publicKey))
+
+    private suspend fun removePublishedEndpoints(publicKeys: Collection<String>): Result<Unit> =
+        withContext(serializedDispatcher) {
+            runSuspendCatching {
+                val publicKeys = publicKeys.mapNotNull(::normalizedPublicKey).distinct()
+                if (publicKeys.isEmpty()) return@runSuspendCatching Unit
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in e56f9dc51. Invalid keys now return `InvalidPublicKey` instead of being dropped from an empty batch, so the retry marker and cached state are preserved for another cleanup attempt. Added a regression test for this case.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1181,41 +1197,77 @@ class PrivatePaykitRepo @Inject constructor(
...
+
+                val linkedReceiverPaths = linkedReceiverPathsByPublicKey()
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in e56f9dc51. The shared snapshot remains the fast path, but if it fails we retry peer inspection per contact. One contact's failure no longer blocks later contacts, and its locally cached published paths are still cleared while its state remains pending. Added regression coverage for the partial-failure case.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1181,41 +1197,77 @@ class PrivatePaykitRepo @Inject constructor(
...
+                val preparation = clearPrivatePaymentLists(publicKeys, linkedReceiverPaths)
```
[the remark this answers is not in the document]
ben-kaufman: I don't think this can leave an untracked payment list. Cleanup unions the snapshot with `publishedPrivatePaymentReceiverPaths`; queued publications are recorded in that local cache before delivery, and publication/cleanup run on the serialized dispatcher. A path becoming linked by itself does not publish a payment list, so any newly linked path with a list to clear is already covered by the cached published path.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
-                        receiverPath = receiverPath,
+    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> =
+        removePublishedEndpoints(listOf(publicKey))
+
+    private suspend fun removePublishedEndpoints(publicKeys: Collection<String>): Result<Unit> =
+        withContext(serializedDispatcher) {
+            runSuspendCatching {
+                val normalizedBatch = normalizedPublicKeyBatch(publicKeys)
+                val publicKeys = normalizedBatch.publicKeys
+                var firstError: Throwable? = PrivatePaykitError.InvalidPublicKey.takeIf {
```
jvsena42: **Invalid pending key now poisons every cleanup pass, permanently.** (Android-only, new in `0b91ae2` — no iOS counterpart.)

Seeding `firstError` with `InvalidPublicKey` means it is rethrown at the end of the batch. If a non-normalizable key ever lands in `deletedContactCleanupPendingPublicKeys`, it can never be cleared, so:

- `retryPendingEndpointRemoval` fails on **every** call forever → `updateContactSharingCleanupPending(false)` (line 221) never runs;
- `removePublishedEndpointsForCleanup` sets `cleanupPending = true` forever — affects `ProfileViewModel:110`, `EditProfileViewModel:230`, `WipeWalletUseCase:57`;
- `disableSharingAndPruneUnsavedContactState` throws permanently.

Pre-PR, the un-normalized single-key path silently no-oped and then cleared the marker via `updateDeletedContactCleanupPending(key, false)` — it self-healed. Current writers all normalize, so this only bites on legacy/corrupt persisted data, but the failure is unrecoverable and takes every other contact's cleanup down with it.

An unnormalizable key cannot correspond to any SDK state (the SDK is keyed by normalized pubkeys), so I'd drop it from the pending set with a warning rather than retaining it forever. At minimum, don't let it feed the batch-wide `firstError`.

app/src/test/java/to/bitkit/repositories/PrivatePaykitRepoTest.kt
```diff
@@ -611,6 +632,51 @@ class PrivatePaykitRepoTest : BaseUnitTest(StandardTestDispatcher()) {
...
+    fun `invalid deleted contact key remains pending for cleanup`() = test {
+        val invalidPublicKey = "not-a-pubky"
+        cacheData.value = PrivatePaykitCacheData(
+            deletedContactCleanupPendingPublicKeys = setOf(invalidPublicKey),
+        )
+        sut = createSut()
+
+        val result = sut.retryPendingEndpointRemoval(emptyList())
+
+        assertTrue(result.isFailure)
```
jvsena42: This test locks in the non-converging behavior described on `PrivatePaykitRepo.kt:1244` — the key stays pending and the pass keeps failing, with nothing that can ever clear it. If the intent is instead "drop keys we can never act on," this should assert `deletedContactCleanupPendingPublicKeys` is empty and `result.isSuccess`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1246,42 +1342,85 @@ class PrivatePaykitRepo @Inject constructor(
...
+                )
+            }.getOrNull()
+
+    private suspend fun linkedReceiverPaths(
+        publicKey: String,
+        snapshot: Map<String, Set<String>>?,
+    ): Result<Set<String>> = if (snapshot != null) {
+        Result.success(snapshot[publicKey].orEmpty())
+    } else {
+        runSuspendCatching { linkedReceiverPathsByPublicKey()[publicKey].orEmpty() }
```
jvsena42: **This fallback re-issues the exact call that just failed** — same finding as iOS #638.

`linkedReceiverPathsByPublicKey()` is the same `paykitSdkService.linkedPeers()` that already threw when `linkedReceiverPathsSnapshotOrNull` took the snapshot. So on snapshot failure the pass makes **N+1** `linkedPeers()` calls instead of the pre-PR N — a regression in exactly the dimension this PR optimizes. The new `cleanup falls back per contact when shared linked receiver inspection fails` test encodes it (`linkedPeerReads <= 2` = snapshot + contact 1's re-fetch).

Suggest picking one: on snapshot failure either fail the pass once (set `firstError`, mark all keys failed, return) or continue with an empty snapshot — and drop the per-contact re-fetch entirely. The `"; retrying per contact"` wording in the warning above also oversells what the retry can accomplish. Applies equally to `preparePrivatePaymentListReservations` via `inspectLinkedReceiverPaths`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+                    firstError = firstError ?: it
                 }
             }
+        }
 
-            firstError?.let { throw it }
+        return PrivateEndpointCleanupPreparation(clearedRetryKeys, failedPublicKeys, firstError)
+    }
 
+    private suspend fun clearPublishedEndpointCache(publicKeys: Collection<String>) {
```
jvsena42, line 1309: **Reentrancy window for the cache wipe is now much wider** — Android analogue of the iOS actor-reentrancy finding.

`serializedDispatcher` is `ioDispatcher.limitedParallelism(1)`, which serializes *execution* but releases the slot at every suspension point — same semantics as the iOS actor. And `removePublishedEndpoints` is **not** run under `publicationMutex`, while `publishLocalEndpoints` is (line 674).

Before this PR each contact's wipe happened immediately after that contact's own clear+drain. Now every wipe is deferred to here, after the whole batch's clears *and* the full drain. Scenario: cleanup runs while a foreground `refreshKnownSavedContactEndpoints` is in flight; contact A is cleared early in the batch, and during the many subsequent awaits the concurrent publish calls `recordPublishedPrivatePaymentListCache` for A (or an incoming invoice writes `localInvoicesByReceiverPath`); this loop then wipes `remoteEndpoints` / `localInvoicesByReceiverPath` / `publishedPrivatePaymentReceiverPaths` for A, desyncing local cache from what the SDK actually holds.

The hazard pre-exists, but the batching widens it from one-contact latency to whole-batch latency. Cheapest mitigations: wrap the cleanup pass in `publicationMutex.withLock { }` too, or snapshot the relevant `contactState` fields before the clear and skip the wipe if they changed.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+                firstError = firstError ?: preparation.firstError
+
+                if (preparation.clearedRetryKeys.isNotEmpty()) {
+                    drainPendingPrivateMessages(
+                        reason = "private endpoint cleanup",
+                        advancingLinksFor = preparation.clearedRetryKeys,
                     )
-                    if (report.failedToQueue.isNotEmpty() || report.failedToDeliver.isNotEmpty()) {
-                        throw PrivatePaykitError.PrivateUnavailable
+                    val pendingRetryKeys = pendingPrivateMessageDrainKeys(preparation.clearedRetryKeys)
```
jvsena42: **Blast radius of a transient drain-check failure grew from one contact to the whole batch** — same as iOS #638.

`pendingPrivateMessageDrainKeys` returns *all* passed keys from both of its `getOrElse` branches (lines 1030 and 1042) when `linkedPeers()` or `pendingOutboundPrivateCounterparties()` throws. With the check now run once per batch, a single transient SDK hiccup marks every successfully-cleared contact as failed, so none of them get their cache cleared or their retry marker cleared.

The retry path covers it, so not a correctness bug — but it's a real reduction in transient-failure granularity that comes with the batching, and it's the kind of thing that surfaces as "cleanup never converges" if the SDK is flaky during a full pass. Worth noting in the description alongside the perf win, and no test covers it.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+            ).onFailure {
+                failedPublicKeys += publicKey
+                firstError = firstError ?: it
+                Logger.warn(
+                    "Failed to inspect private Paykit links for '${redacted(publicKey)}' during cleanup",
+                    it,
+                    context = TAG,
+                )
+            }.getOrDefault(emptySet())
+            receiverPathsForCleanup(
```
jvsena42, line 1288: **Behavior change worth calling out** — the description says "no user-facing behavior changes are intended." Same finding as iOS #638.

Previously `receiverPathsForCleanup` was `suspend` and called `linkedPeers()` directly, so a throw propagated out of `removePublishedEndpoints(publicKey)` and **zero** `clearPrivatePaymentList` calls were made for that contact. Now a link-inspection failure yields an empty `contactLinkedReceiverPaths` and execution falls through to here, so the contact still gets cleared for its locally-published paths (asserted by `cleanup keeps pending state when linked receiver inspection fails`).

This is arguably an improvement — partial progress, and the contact is marked failed either way so it retries — but it's a real change in what hits the SDK on the error path. Worth a line in the PR description.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+    private suspend fun clearPrivatePaymentLists(
+        publicKeys: Collection<String>,
+        linkedReceiverPathsSnapshot: Map<String, Set<String>>?,
+    ): PrivateEndpointCleanupPreparation {
+        val failedPublicKeys = mutableSetOf<String>()
+        val clearedRetryKeys = mutableListOf<PrivateMessageDrainRetryKey>()
+        var firstError: Throwable? = null
+
+        publicKeys.forEach { publicKey ->
+            val contactLinkedReceiverPaths = linkedReceiverPaths(
```
jvsena42: **Duplicated snapshot-or-fallback block** (mirrors the iOS #638 duplication finding, just in a different spot).

This hand-rolls exactly what `inspectLinkedReceiverPaths` (line 1398) already does — resolve, log on failure, default to empty — differing only in the log wording (`"during cleanup"` vs `"during '$reason'"`) and the extra `failedPublicKeys` bookkeeping. Per the repo's reuse convention, call the helper with `reason = "private endpoint cleanup"` and branch on `LinkedReceiverPathInspection.error` for the failure marking, so the two paths can't drift.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+                }
+                if (publicKeys.isEmpty()) {
+                    firstError?.let { throw it }
+                    return@runSuspendCatching Unit
+                }
+
+                val linkedReceiverPathsSnapshot = linkedReceiverPathsSnapshotOrNull("private endpoint cleanup")
+                val preparation = clearPrivatePaymentLists(publicKeys, linkedReceiverPathsSnapshot)
+                val failedPublicKeys = preparation.failedPublicKeys.toMutableSet()
+                firstError = firstError ?: preparation.firstError
```
jvsena42: `firstError` precedence hides the real failure: because `InvalidPublicKey` is seeded above, this `?:` discards the actual SDK error the batch hit. Callers log and surface only the first one, so diagnostics for a genuinely failing cleanup get replaced by "Contact public key is invalid." Prefer surfacing the SDK failure and tracking the invalid-key case separately.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1234,6 +1327,9 @@ class PrivatePaykitRepo @Inject constructor(
                 }
             }
             updateDeletedContactCleanupPending(publicKey, isPending = false)
```
jvsena42: Missed batching in a batching PR: `persistState` below is correctly hoisted out of the loop, but `updateDeletedContactCleanupPending` still runs per key — and each call is a `DataStore.updateData` (`PrivatePaykitStores.kt:35`), i.e. serialize + file write. That's N disk writes for N contacts, plausibly more expensive than the `linkedPeers()` call this PR batched away.

Collapse into a single update, e.g.:

```kotlin
cacheStore.update {
    it.copy(deletedContactCleanupPendingPublicKeys = it.deletedContactCleanupPendingPublicKeys - publicKeys.toSet())
}
```

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
-    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> = withContext(serializedDispatcher) {
-        runSuspendCatching {
-            var firstError: Throwable? = null
-            receiverPathsForCleanup(publicKey).forEach { receiverPath ->
-                val result = runSuspendCatching {
-                    val report = paykitSdkService.clearPrivatePaymentList(
-                        counterparty = publicKey,
-                        receiverPath = receiverPath,
+    private suspend fun removePublishedEndpoints(publicKey: String): Result<Unit> =
+        removePublishedEndpoints(listOf(publicKey))
```
jvsena42, line 1225: `retryPendingDeletedContactEndpointRemoval` (line 1563) still loops this single-key overload per pending contact, so each one gets its own `linkedPeers()` snapshot and its own drain — exactly the multi-contact pattern this PR set out to batch. Passing the whole pending set to the `Collection` overload would get the same win there.

Separately (pre-existing, but it interacts badly with the invalid-key issue): the `.getOrThrow()` inside that `forEach` aborts the remaining pending keys on the first failure, so one bad key blocks every key after it in iteration order.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1190,41 +1228,96 @@ class PrivatePaykitRepo @Inject constructor(
...
+
+    private suspend fun removePublishedEndpoints(publicKeys: Collection<String>): Result<Unit> =
+        withContext(serializedDispatcher) {
+            runSuspendCatching {
+                val normalizedBatch = normalizedPublicKeyBatch(publicKeys)
+                val publicKeys = normalizedBatch.publicKeys
```
jvsena42: Nit: this shadows the function parameter. It's consistent with the existing `val retryKeys = retryKeys.toSet()` in `pendingPrivateMessageDrainKeys`, so take it or leave it — but a distinct name (`normalizedKeys`) reads better and makes it obvious the raw input is intentionally no longer used.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -791,6 +802,21 @@ class PrivatePaykitRepo @Inject constructor(
         drainAndSchedulePrivateLinkRetries(reason, retryKeys.distinct())
     }
 
+    private suspend fun preparePrivateLinks(
+        publicKey: String,
+        receiverPaths: Collection<String>,
+        reason: String,
+    ): List<PrivateMessageDrainRetryKey> = receiverPaths.distinct().map { receiverPath ->
```
jvsena42: Nit: `.map { }` here is really a side-effecting loop — the lambda's actual work is `ensureLinkWithPeer`, and the returned value is unrelated to it. `onEach`/an explicit loop building the list would read more honestly about the intent. Behavior is preserved (the retry key is added regardless of whether the link succeeded), so this is purely stylistic.

jvsena42 (changes requested):
Reviewed the batching refactor against the parallel iOS findings on synonymdev/bitkit-ios#638.

The structure is a clear improvement over the base, and the per-contact failure isolation contract does hold — I traced clear-failure, drain-still-pending, and link-inspection-failure, and each ends with the contact in `failedPublicKeys`, cache retained, retry marker intact. Test coverage is also meaningfully ahead of the iOS PR, which ships none. `./gradlew testDevDebugUnitTest --tests to.bitkit.repositories.PrivatePaykitRepoTest` passes locally.

**iOS parity summary**

| iOS #638 finding | Android status |
|---|---|
| greptile **P1** — shared lookup failure drops cleanup + retries | ✅ already fixed in `0b91ae2` (ahead of the iOS fix) |
| N+1 `linkedPeers()` on snapshot failure | ⚠️ present |
| duplicated snapshot-or-fallback block | ⚠️ partly — helper exists but isn't reused in cleanup |
| clears now attempted after link-inspection failure | ⚠️ present, undocumented |
| drain-check failure blast radius = whole batch | ⚠️ present |
| widened reentrancy window on the deferred state wipe | ⚠️ present in analogous form |
| eager vs deferred failure markers | n/a — Android markers are caller-driven |

**Requesting changes on one item:** the invalid-key handling added in `0b91ae2` turns a previously self-healing no-op into a permanently non-converging failure that blocks profile-delete, disable-sharing, and wipe cleanup for *all* contacts. Details inline. Everything else is either an iOS-parity item worth resolving consistently across platforms, or a cleanup.

**Description nits:** "no user-facing behavior changes are intended" isn't quite accurate — the error path now issues `clearPrivatePaymentList` calls it previously skipped, and transient-failure granularity is coarser. Both worth a bullet.

**Changelog:** correctly omitted for a `refactor:` with no user-facing change.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Malformed persisted keys are now logged and removed from both cached and pending cleanup state without failing valid contacts. This is safe because a key that cannot be normalized cannot address SDK state.

app/src/test/java/to/bitkit/repositories/PrivatePaykitRepoTest.kt
ben-kaufman: Updated in `b6689414e`. The regression test now asserts that an invalid key is dropped, the retry succeeds, and both cached and pending cleanup state converge to empty without an SDK clear call.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Publication and cleanup now use one shared snapshot helper with at most one batch-level retry and no per-contact fallback reads. If both attempts fail, locally tracked paths are still cleared but all affected contacts remain pending.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Cleanup now runs under the same `publicationMutex` as publication, snapshots the three cleanup-owned cache fields before remote awaits, and skips the local wipe if any of them changed. Added a concurrent regression test that updates resolved endpoint cache while cleanup is suspended.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Kept conservative by design and documented in the PR description. The post-drain inspection cannot attribute an unknown delivery state safely, so the whole batch remains cached and pending for retry. Added explicit regression coverage for this failure path.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Documented in the updated PR description. After two link-inspection failures we still attempt locally tracked published paths, while retaining every affected contact and its pending state so cleanup retries safely.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. The duplicated per-contact inspection path is gone; publication and cleanup both consume the same `linkedReceiverPathsSnapshot` helper and shared result.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Invalid keys are handled separately and no longer seed `firstError`, so actual SDK or inspection failures retain diagnostic precedence.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Deleted-contact cleanup markers now use the collection overload and are removed with one DataStore update for the batch.

ben-kaufman: Fixed in `b6689414e`. Deleted-contact cleanup markers now use the collection overload and are removed with one DataStore update for the batch.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`. Pending deleted contacts now go through the collection cleanup overload once, sharing the snapshot and drain. Successful valid contacts are finalized even if another contact fails, while malformed keys self-heal instead of aborting the pass.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`; the normalized collection is now named `normalizedKeys`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `b6689414e`; `preparePrivateLinks` now uses an explicit loop that builds retry keys while performing link preparation.

ben-kaufman:
Restacked onto the current #1098 head after the base rewrite. The cleanup review-fix commit previously referenced as `b6689414e` is now signed commit `cd4cd1ec7`; the cleanup diff is unchanged.

The current `build` failure is the same inherited #1084 `WatchOnlyAccountStore.serializedExtendedPubkey` compile failure already present on #1084 and #1098.

ovitrif (approved):
Approved. The batched cleanup keeps per-contact clear isolation, drops malformed pending keys without failing the pass, and preserves cache state that changes while remote cleanup is suspended.

Issue #986, opened 2026-06-03
epic: Play store production crashes


piotr-iohk:
Elevated user-perceived crash rate on mainnet **2.2.0 (181)**. Tracks Play Vitals triage (Jun 2026).

**Vitals:** https://play.google.com/console/u/1/developers/5690237613698939049/app/4973032799696771282/vitals/crashes?days=28&isUserPerceived=true

**Bundles:** FGS (#980, #981, #443); native (#982 ± #984). `DidNotStartInTime` @ MainActivity (3/60d) → #981, no separate issue.

## Done
P0 closed; P2 closed/deferred; AAB + native symbols each release; 28-day crash rate improved vs Jun 2026.

ovitrif:
Milestones:
- `2.3.0` for high priority issues
- `2.4.0` for medium priority issues

ovitrif:
Update on milestones:
More certainty about fixing #982 requires update across all our Rust repos, won't land in `2.3.0`, now targeting `2.4.0`.

Work is ongoing but risk of delays for getting PRs in 4 repos merged is too impactful for the upcoming release.

ovitrif:
Update:
All rust repos now emit the artifact `native-debug-symbols.zip`, while the android release process includes scripts and instructions for bundling those and including them in release channels (store and GH).

Investigations should go better in future release crashes thanks to these symbols metadata.

piotr-iohk:
New Play Vitals issues from 2.3.1 (build 184) week of Jul 6–13:

- #1068 — Electrum config connect ANR in isValidURL regex on main thread (1 user, 1 ANR)
- #1069 — Backup failure toast crashes on formatPlural IllegalArgumentException (1 user, 2 events)

ovitrif:
Crash investigation closeout for the reports currently attached to this epic:

- #980 and #981 are Android foreground-service failures fixed by #987. #987 is the canonical fix for both.
- #982 is an ldk-node owned-Tokio-runtime self-drop. Exact rc.52 symbols confirm the electrum chain-sync task releases the last runtime owner from a runtime worker after timed-out shutdown aborts tasks without draining them. synonymdev/ldk-node#94 owns the fix.
- #983 has matching generated Kotlin/native checksums in the archived build 181 and 182 artifacts. The remaining cause is runtime-loaded binding/library skew; the issue now records the exact evidence required from a recurrence.
[...]

Issue #982, opened 2026-06-03
crash: Production SIGABRT in arm64 native code


piotr-iohk:
## What happened?

Google Play Console reports a top production crash cluster: SIGABRT in `split_config.arm64_v8a.apk` for package `to.bitkit` (mainnet).

- Native `abort()` (typical of Rust panic / fatal native assert in release).
- Stack frames #01–#29 show only raw program counters — not symbolicated to `libldk_node.so` / `libbitkit_core.so` in Play Vitals, despite `native-debug-symbols.zip` (25.7 MB) on bundle 181.aab (2.2.0).
- Native worker thread (`__pthread_start` / `__start_thread`), not main/UI thread.
- Play Vitals (60-day): ~38% of crash events in top cluster (~60 events, ~22 users); heavily version 181 (2.2.0); mostly foreground; Android 13–16, multiple OEMs.
- User-perceived crash rate elevated since native Bitkit 2.0 (Feb 2026).

## Expected behavior

App must not crash with SIGABRT during normal wallet/Lightning use on production mainnet. Play or local symbolication should resolve native PCs to function names for triage.

## Steps to Reproduce

1. Mainnet wallet, notifications enabled (`LightningNodeService` running).
2. Use app in foreground: sync, activity, send/receive Lightning, channels.
3. Optional: rapid background/foreground, large wallet, post-migration state.
4. Capture `adb logcat` around crash.

No deterministic repro confirmed yet.

## Logs / Screenshots / Recordings

```
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 10374 >>> to.bitkit <<<

backtrace:
  #00  pc 0x0000000000077980  /apex/com.android.runtime/lib64/bionic/libc.so (abort+160)
  #01  pc 0x0000000000ceaff0  /data/app/.../split_config.arm64_v8a.apk
  #02  pc 0x0000000000cf22cc  /data/app/.../split_config.arm64_v8a.apk
  #03  pc 0x0000000000cddf58  /data/app/.../split_config.arm64_v8a.apk
  #04  pc 0x00000000009ca2f4  /data/app/.../split_config.arm64_v8a.apk
  #05  pc 0x0000000000cddebc  /data/app/.../split_config.arm64_v8a.apk
  #06  pc 0x0000000000cf35a0  /data/app/.../split_config.arm64_v8a.apk
[...]

PR #1100, opened 2026-07-22
fix: harden ldk node teardown on stop

### LINKED ISSUE #1078: Validate Electrum server network before switching
## Summary

When changing the custom Electrum server in **Settings → Advanced → Electrum server**, the app does not validate that the server belongs to the wallet's network. Pointing a wallet at a server from a different network (e.g. a regtest wallet at a mainnet server) is accepted, after which the on-chain wallet sync fails indefinitely and the screen silently shows **Disconnected** with an endless retry loop.

## Steps to reproduce

1. Run a regtest (dev) wallet.
2. Open Settings → Advanced → Electrum server.
3. Enter a mainnet server, e.g. `electrum.blockstream.info` / `50002` (SSL), and tap Connect.
4. The change is accepted ("Successfully changed electrum server").
5. The Electrum server row shows **Disconnected** and the on-chain wallet sync keeps timing out.

## Root cause (from device logs)

The server switch succeeds, but the regtest node then syncs against a mainnet chain. The Lightning tx sync still finishes fast (only a few channel script-hashes), while the on-chain wallet full re-scan against the remote mainnet chain exceeds its deadline:

```
LightningRepo: Changing ldk-node electrum server to: 'ssl://electrum.blockstream.info:50002'
LightningRepo: Successfully changed electrum server
...
ldk_node::chain::electrum: Incremental sync of on-chain wallet timed out: deadline has elapsed
LightningRepo: Error executing 'sync' [AppError='A wallet operation timed out.']
LightningRepo: Sync retry failed, will retry in 15s
```

The failed sync sets `lastSyncError` (`LightningRepo.kt`), and the screen derives `isConnected = isNodeRunning && !hasSyncError` (`ElectrumConfigViewModel.observeState`), so it renders Disconnected. The status is technically accurate, but the real problem — a wrong-network server — is never surfaced, and the app retries forever.

## Expected behavior

Before switching, validate that the Electrum server's network matches the wallet's network. On mismatch, reject the change and show a clear error (e.g. "This server is on a different network") instead of accepting it and silently going Disconnected.

## Proposed approach

- Add a network check to the connect flow (`ElectrumConfigViewModel.connectToServer` / `LightningRepo.restartWithElectrumServer`), e.g. probe the server's genesis block hash (Electrum `server.features` / `blockchain.block.header` at height 0) and compare it against the wallet's expected network genesis before committing the switch.
- Surface a dedicated validation error string rather than relying on the generic sync-error → Disconnected state.
- Add unit coverage for the mismatch path in `ElectrumConfigViewModelTest.kt`.

## Notes

- Reproduced on a dev/regtest build; primarily affects users who manually enter a server, but the silent failure mode is poor UX on any network.
- Unrelated to the ANR/ReDoS fix in #1077 — this is a separate, pre-existing gap.

jvsena42:
Relates to #982 , #986
Fixes #1078

Upstream issue: synonymdev/ldk-node#94

This PR hardens LDK node teardown so it cannot orphan, deadlock, or overlap the native node:

1. Releases the LDK node handle deterministically on stop instead of leaving it to the garbage collector, with the wait bounded so a stuck `free_node` can never brick startup
2. Makes teardown non-cancellable, so a cancelled caller (including one cancelled during listener cleanup) can't abandon it halfway
3. Fixes a self-join deadlock when a stop or listener re-arm is requested from inside an LDK event handler
4. Gates rebuild and destructive storage work on the previous node's release so native lifetimes never overlap
5. Probes an Electrum server before the node is torn down for it, so a misconfigured server is rejected in seconds instead of failing a rebuild and leaving the release wedged
6. Serializes a whole server-change transaction against the background recovery a failed change launches, so a change can never report and persist a server that was not applied
7. Stops tearing the node down and rebuilding it when the app is only briefly backgrounded

### Description

Play Vitals reports a `SIGABRT` on mainnet as the top production crash cluster. Symbolicating it locally against the unstripped `libldk_node.so` for the shipped `ldk-node-android` version resolved every frame and gave a clear cause:

```
#19 ldk_node::runtime::spawn_background_task  (chain-sync task, tokio worker thread)
#14 drop_slow<ElectrumRuntimeClient>          ← last Arc released here
#13 drop_in_place<ldk_node::runtime::Runtime / RuntimeMode>
#12 drop_in_place<tokio::runtime::Runtime → BlockingPool>
#11 tokio blocking/shutdown.rs:51 → panic: "Cannot drop a runtime in a context
    where blocking is not allowed"
```

The defect itself is upstream: ldk-node's chain-sync task drops the last reference to the Electrum client, which owns the tokio runtime, while running on a worker thread of that same runtime. Release Rust aborts on panic, so the process dies with no catchable error. That is filed as synonymdev/ldk-node#94 and this PR does not fix it. What this PR fixes is the Android teardown that made the window wide, non-deterministic, and prone to overlapping native lifetimes.

#### Deterministic, bounded, non-cancellable release

Stopping the node cleared the reference but never destroyed the handle, so the native node stayed alive until the garbage collector freed it on the finalizer thread at an arbitrary later moment, potentially while a new node was already running. The handle is now released as part of the stop itself.

The release is bounded, not unconditionally synchronous. `free_node` returns in tens of milliseconds for a healthy node but blocks for tens of seconds when a failed start left the node's background tasks wedged (measured: ~38 ms healthy vs ~40 s wedged). So the release runs on the IO dispatcher and the stop waits on it only up to a short timeout — long enough that a healthy release completes inline, but bounded so the lifecycle state (`Stopped`) is published promptly rather than blocking behind a wedged drain.

Teardown was also abandonable. The stop ran through a cancellable context and was triggered from a ViewModel scope, so an activity finishing mid-stop left a live native node orphaned with the lifecycle state stranded. The whole teardown — including the listener cleanup join at the top — is now non-cancellable once committed. Reading ldk-node also showed the stop path could not fail the way the code assumed: it reports an error only when the node is already stopped and swallows every internal failure, so a failed stop no longer rethrows and skips the release.

#### Listener lifecycle safety

`WakeNodeWorker` calls `stop()` synchronously from its registered LDK event handler, which runs on the listener job. `listenerJob.cancelAndJoin()` then waited on the very job it was running on — a self-join. The same shape applied to `startEventListener()`'s re-arm. A stop from inside a handler now skips joining its own job (detected via `currentCoroutineContext()`, not the shadowing class-scope `coroutineContext`), and a re-arm from inside a handler is a no-op. The listener loop also keys on node identity, not just the shared `shouldListenForEvents` flag (now `@Volatile`), so a racing `start()` cannot make the old loop poll a node `destroy()` already freed. `startEventListener` uses `runSuspendCatching` so it no longer silently swallows the cancellation.

#### Overlap prevention: release gate + non-blocking recovery

Even with a bounded stop, the old node can still be draining after `Stopped` is published. The destructive storage operations — `wipeStorage()`, `resetNetworkGraph()`, and the pathfinding-scores VSS deletes — plus the normal rebuild path now wait on the previous node's release before touching storage, so a delete never races, and a new node never builds over, storage the old node still owns. That wait is itself bounded (~90 s, comfortably above the observed ~40 s wedge): a stuck `free_node` throws `NodeReleaseTimeout` rather than proceeding (which would overlap) or blocking forever (which would permanently brick every future rebuild).

Nothing opts out of that gate. An earlier revision of this PR let the Electrum server-change rebuild skip it (`awaitRelease = false`), because gating serialized each back-to-back change in `@settings_10` behind the previous — possibly wedged, ~40 s — node's drain. That bypass is gone: the pre-flight probe below removes the wedge it was working around, so the parameter itself has been deleted from `LightningService.setup`, `LightningRepo.setup`, and `LightningRepo.start`.

#### Electrum pre-flight probe

The ~40 s wedge was never intrinsic. It is produced by a **failed `node.start()`**, which leaves the node's Electrum background tasks stuck mid-handshake so `free_node` cannot drain. Every wedge measured in this PR came from tearing the node down and rebuilding it against a server that could never work.

So the Electrum change now probes the candidate server over its own socket *before* the node is touched, exactly as `restartWithRgsServer` already validates its URL before stopping anything. `ElectrumProbeService` connects, performs the TLS handshake when SSL is selected, and speaks enough of the protocol to classify four failure modes:

| Check | Rejects with |
| --- | --- |
| Socket connect | `Unreachable` — wrong host or port |
| TLS handshake | `ProtocolMismatch` — TLS against a plain-TCP server, the `@settings_10` condition |
| `server.version` | `NotElectrum` — reachable, but not an Electrum server |
| `server.features` → `genesis_hash` | `NetworkMismatch` — right protocol, wrong chain (regtest vs mainnet) |

The genesis-hash check closes #1078: a regtest wallet pointed at a mainnet server used to be accepted ("Successfully changed electrum server"), after which the on-chain sync timed out forever and the screen silently showed **Disconnected** with an endless retry loop. That server is now rejected before the switch, with the failure surfaced immediately instead of as a permanent Disconnected state. #1078 proposed exactly this mechanism — comparing the server's genesis hash from `server.features` against the wallet's network — and its reproduction case (`electrum.blockstream.info` on a regtest build) is one of the device journeys below.

Each reply is validated as a JSON-RPC envelope rather than merely parsed: the probe requires the `id` it asked for, no `error`, and a non-null `result`. Without that, a server erroring on version negotiation — one the real LDK client rejects at startup — would pass the version check, and a subsequent `server.features` error would then look like "no genesis hash" and probe clean, recreating the failed-start wedge. Version negotiation must succeed first; only then is `server.features` treated as optional.

`server.features` is optional in the protocol, so a server that omits `genesis_hash` is accepted rather than rejected — the network check degrades instead of false-rejecting working servers. Every step is bounded so the probe itself cannot become the thing that hangs.

Because the node is no longer torn down for a server that cannot work, this path stops manufacturing wedged releases — which is what makes the release gate affordable again on every rebuild, including recovery.

This is a mitigation, not a guarantee: a server can pass the probe and still fail at start (it goes down in between, or fails in a way the probe does not model). The gate therefore keeps its ~90 s bound and `NodeReleaseTimeout` for that residual case.

#### Server-change transaction serialization

A failed change still runs its recovery in the background, on the repository's process-lifetime scope, so the user sees the failure immediately rather than waiting on the restore. That detachment opened a race: `stop()` and `start()` take the lifecycle mutex separately, so a recovery could restart the previous config in the gap between a later change's stop and its start. The later change would then hit the already-running fast path in `start()`, report success, and persist a server that was never applied.

Two changes close it. A dedicated transaction lock is held across the whole stop → start → persist sequence, and is taken by the recovery too, so a change and a recovery can no longer interleave. And a `start()` carrying an explicit config is never satisfied by an already-running node — one that is already up demonstrably was not built with that config — so it fails with `NodeConfigNotApplied` instead of reporting a success the caller would persist. That second guard also covers racers the lock cannot see, such as a foreground start or `WakeNodeWorker`.

#### Brief-background debounce

Backgrounding the app stopped the node immediately, so a short task switch cost a full stop plus a full rebuild. The stop from backgrounding is now deferred by a few seconds and cancelled if the app returns. Only the backgrounding path is deferred; the notification stop action, service teardown, notification worker, and restore migration all still stop immediately. The deferred stop is owned by the repository's process-lifetime scope so it cannot be silently dropped when the activity goes away, and every `start()` cancels a pending stop up front so a start that short-circuits on its guards can't leave one to fire against a foregrounded node.

### Preview

[short-pause.webm](https://github.com/user-attachments/assets/57bc9efa-832b-416e-b9f5-c1d7fcfdaed9)

[long-pause.webm](https://github.com/user-attachments/assets/7fbb7942-bf36-404d-8739-8c3c030efd9b)

#### Manual Tests
- [x] **1a.** Home → background the app and return within ~2s: no node teardown or rebuild, wallet stays responsive.
  - [x] **1b.** Background and wait past the window, then return: node stops, then rebuilds and reaches started.
- [x] **2.** `regression:` Background Payments → enable Get paid when Bitkit is closed and Keep Bitkit active in background → background the app: node keeps running, no teardown.
- [x] **3.** `regression:` Node notification → tap Stop: node stops immediately and the app task is removed.
- [x] **4.** `regression:` Relaunch after a notification stop: node rebuilds and reaches started.
- [x] **5.** `regression:` Settings → Lightning → restart node, and Electrum/RGS server change to a valid server: node stops and restarts normally.
- [x] **6.** `regression:` Receive a Lightning payment in the background (app closed, Background Payments on): payment is received and the notification is delivered.
- [x] **7.** Settings → Advanced → Electrum → enter an unreachable server → Connect: the error surfaces promptly and the node is left running on the previous server.
- [x] **8.** Electrum → point TLS at a plain-TCP Electrum server → Connect: rejected in ~5 s with no node teardown, node stays on the previous server.
- [x] **9.** Electrum → point a regtest build at a mainnet Electrum server → Connect: rejected in ~1 s on the genesis-hash check, no node teardown.
- [x] **10.** Electrum → change to a valid server (local TCP and a remote TLS server): probe passes and the node stops, rebuilds and reaches started, with the setting persisted.
- [x] **11.** Electrum → a rejected change immediately followed by a valid one (the `@settings_10` shape): the valid change applies without waiting on anything.

#### Automated Checks
- Release + teardown coverage in `LightningServiceTest.kt`: deterministic handle release, release when the node is already stopped / when node stop throws, the no-node case, no double release, and teardown completing when the caller is cancelled (including cancellation during listener cleanup).
- Listener-safety coverage in `LightningServiceTest.kt`: a stop from inside an event handler completes without self-join deadlock, a re-arm from inside a handler keeps the listener running, an external stop cancels and joins the listener, and the loop stops polling a node once it is swapped out (no use-after-free).
- Release-gate coverage in `LightningServiceTest.kt`: rebuild (`setup`), `wipeStorage`, and `resetNetworkGraph` each wait for the previous release before proceeding; the gate adds no latency when no release is pending; the gate throws instead of blocking forever when the release never finishes; and the stop returns within its bound while the release is wedged. These use a real IO dispatcher plus a controllably delayed `destroy()`.
- Probe coverage in `ElectrumProbeServiceTest.kt`, driven by a fake Electrum server on a local socket: a server on the expected network is accepted, one on a different network is rejected on the genesis hash, a server that omits `server.features` is still accepted, a host that never answers the handshake and an unreachable host are both rejected, TLS against a plain-TCP server is rejected, and the error names the server that was probed. Envelope validation is covered by four more cases — a version reply carrying an `error`, one with no `result`, one answering a different `id`, and a server that errors on both version and features — each asserting rejection, and each verified load-bearing by reverting the check.
- Server-change coverage in `LightningRepoTest.kt`: a probe rejection returns the probe error without stopping the node or calling `setup`, a `start()` carrying a custom server URL fails instead of riding an already-running node, and a change issued while a recovery is in flight waits for it and then genuinely rebuilds with the requested server.
- Repo coverage in `LightningRepoTest.kt`: deferred stop timing and cancellation, a foreground/background cycle within the window never stopping the node, and `restartWithElectrumServer` surfacing failure before the background recovery completes. `WalletViewModelTest.kt`: a foreground start cancels a pending deferred stop even while startup is still active.
- Each new regression test was verified load-bearing by reverting its fix and confirming it fails (or, for the unbounded-gate case, hangs) before restoring.
- Device validation on the emulator, backgrounding for 1–8 seconds: under the window the node stays running and start is skipped; past it the log order is always `Stopping node…` → `Node stopped` → `Building node…` → `Node started`, confirming teardown completes before startup on the healthy path.
- Device validation of the foreground-service path: with background payments and keep-active enabled, backgrounding produced no stop and the node kept processing LDK events.
- Device validation of the notification stop action (immediate, no double-stop from service teardown) and of a background Lightning receive.
- Device validation of the Electrum pre-flight probe against the local regtest stack (electrs on plain TCP), with the release gate restored on every rebuild. Each rejection left the node completely untouched — no `Stopping node…`, no `Building node…`, and no `Node handle release still pending` anywhere in the session:

  | Journey | Before this PR | With the probe |
  | --- | --- | --- |
  | TLS pointed at the plain-TCP electrs (the `@settings_10` condition) | ~22 s, node torn down, release wedged | **5.0 s `ProtocolMismatch`, node untouched** |
  | Regtest build pointed at a mainnet Electrum server | started, then failed slowly | **1.3 s `NetworkMismatch`** (expected `0f9188f1…`, got `00000000…0a8ce26f`) |
  | Unreachable host | ~25 s | **46 ms `Unreachable`** |
  | Valid local TCP server | — | probe 8 ms, rebuild succeeds, gate costs **8 ms** |
  | Valid remote TLS server | — | probe **1.5 s**, accepted (confirms no false rejection of real TLS, and the genesis check passes on a legitimate server) |
  | Re-validated after adding envelope validation | — | mainnet server still rejected in **1.35 s**; valid local TCP accepted in **14 ms** and remote TLS in **1.5 s**, so the stricter checks do not false-reject real servers |
  | Rejected change immediately followed by a valid one | second change serialized behind the ~40 s drain | **second change completes in 8.3 s with zero gate wait** |

- Device validation that the restored gate costs nothing on the healthy path: no `Waiting for previous node release…` line appeared on any rebuild, because a healthy release finishes in ~38 ms.
- Device validation of the transaction lock on a real concurrent dispatcher: with a recovery in flight, a second Electrum change logged its entry and then blocked for ~9.5 s until the recovery finished, before running its own full stop → rebuild → start and only then persisting. Without the lock that window is where the recovery's start lands, leaving the change to report success without applying.
- No `Fatal signal`, `SIGABRT` or tombstone appeared in logcat across any of the runs above.

#### Notes for review
- **#1078's dedicated error string is not included.** The wrong-network server is now rejected rather than silently accepted, but the failure toast still uses the generic `settings__es__server_error_description`; the specific reason (`NetworkMismatch` / `ProtocolMismatch` / `NotElectrum` / `Unreachable`) is distinguished in the probe error and the logs only. Wiring those to distinct user-facing strings is a small, self-contained follow-up — say the word and I'll add it here instead. The mismatch path is covered by unit tests in `ElectrumProbeServiceTest.kt` and `LightningRepoTest.kt` rather than in `ElectrumConfigViewModelTest.kt` as #1078 suggested, since the check lives at the service/repo layer.
- **`@settings_10` timing changes.** Its Electrum error toasts now appear in ~5 s (wrong protocol) and ~50 ms (unreachable host) instead of ~22 s, since the probe rejects before any node work. Faster and more deterministic, but the E2E's timing assumptions are worth a look — CI is the real check.
- **The two-request barrier test pins the outcome, not the interleaving.** Under the single-threaded test dispatcher a parked recovery holds the lifecycle mutex for its whole rebuild, so a second change blocks there regardless; the window the transaction lock closes only opens on a genuinely concurrent dispatcher. That case is covered by the device validation above rather than by the unit test, and the test is commented to say so. The `start()` guard is the part pinned by a load-bearing unit test, and it is what prevents the user-visible symptom (persisting a server that was never applied).


app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -457,6 +461,20 @@ class LightningRepo @Inject constructor(
...
+    /**
+     * Defers [stop] so a brief background/foreground cycle does not tear the node down and rebuild it.
+     * Runs on the repo scope so a cancelled ViewModel cannot drop the pending stop.
+     */
+    fun stopDebounced() {
+        val job = scope.launch {
+            delay(BACKGROUND_STOP_DELAY)
+            stop()
+        }
+        pendingStopJob.getAndSet(job)?.cancel()
```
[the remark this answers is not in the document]
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1100/commits/1db7af253f53ec99000c556789dd600c72496dca

jvsena42:
checking CI

jvsena42:
### iOS port assessment: not needed

Checked this against the iOS teardown paths (`LightningService.stop`, `WalletViewModel.stopLightningNode`, `AppScene.handleScenePhaseChange`). This PR's three changes each target a mechanism that is Android-specific and structurally absent on iOS, so there is nothing to port.

| This PR's change | Android mechanism | iOS |
|---|---|---|
| Deterministic handle release (`node.destroy()`, bounded) | JVM cleared the reference but left the native node to the **GC finalizer**, freed at an arbitrary later moment, potentially racing a new node | **ARC**: dropping the last reference at `self.node = nil` frees the handle deterministically. No finalizer thread, no arbitrary-timing race, so no explicit release to add |
| Non-cancellable teardown (`withContext(NonCancellable)`) | `stop()` ran in a cancellable context from a `viewModelScope`, so an activity finishing mid-stop orphaned a live native node | Swift task cancellation is **cooperative** and does not interrupt the synchronous FFI `node.stop()`. `stop()` also unlocks via `defer`, and no Activity-like scope tears it down |
| Deferred stop on backgrounding (`stopDebounced`) | Backgrounding stopped the node immediately, so a short task switch cost a full stop + rebuild — this is what made the race window wide | **iOS never stops the node on backgrounding.** `handleScenePhaseChange` only handles PIN lock and foreground reconnect; the only `stop` callers are `resetNetworkGraph`, `wipe`, and internal restart/recovery. No churn, nothing to debounce |

The one genuinely shared item is the underlying defect, synonymdev/ldk-node#94 — which this PR explicitly does not fix, it only narrows the timing window. That's a Rust-side fix and applies to both platforms whenever it lands upstream; it isn't something to port at the app layer.

One minor, non-blocking note for the iOS side: `listenForEvents` holds a local strong `node` while parked at `await node.nextEventAsync()`, so after `self.node = nil` the actual free is deferred until that task unwinds rather than happening exactly at the assignment. It's far more bounded than a GC finalizer and there is no rebuild churn to race against, so it does not reproduce this crash — just the one place where iOS deallocation isn't strictly deterministic, if we ever want to tighten it.

app/src/main/java/to/bitkit/services/LightningService.kt
```diff
@@ -325,14 +334,39 @@ class LightningService @Inject constructor(
         }
 
         Logger.debug("Stopping node…", context = TAG)
-        ServiceQueue.LDK.background {
-            runCatching { node.stop() }
-                .onFailure { if (it !is NodeException.NotRunning) throw it }
-            this@LightningService.node = null
+        // Teardown must not be abandoned midway: a cancelled caller would leave the rust node alive
+        // and let the GC free it later on the finalizer thread, racing the next node.
+        withContext(NonCancellable) {
```
ovitrif: [P1] Cover listener shutdown with the non-cancellable section

This PR’s cancellation guarantee still has one uncovered entry path. `LightningService.stop()` currently:

1. Cancels and joins `listenerJob` in the caller’s cancellable context.
2. Enters `withContext(NonCancellable)` only after that join completes.
3. Performs `node.stop()`, clears `node`, and destroys the handle.

If the caller is cancelled while listener cleanup is still joining, step 1 exits `stop()` before native teardown begins. `LightningRepo.stop()` has an outer `NonCancellable`, but direct service-stop paths such as the stale-graph reset in `LightningRepo.start()` do not.

I reproduced this on head `cda00a419`: only `nextEventAsync()` ran; neither `node.stop()` nor `destroy()` was invoked. Making the entire `LightningService.stop()` body non-cancellable makes the same test pass.

Suggested regression test for `LightningServiceTest.kt`:

```kotlin
@Test
fun `stop completes teardown when cancelled during listener cleanup`() = test {
    val listenerEntered = CompletableDeferred<Unit>()
    val listenerCleanup = CompletableDeferred<Unit>()
    whenever(node.nextEventAsync()).doSuspendableAnswer {
        listenerEntered.complete(Unit)
        try {
            awaitCancellation()
        } finally {
            withContext(NonCancellable) { listenerCleanup.await() }
        }
    }
    sut.startEventListener()
    listenerEntered.await()

    val stopJob = launch { sut.stop() }
    stopJob.cancel()
    listenerCleanup.complete(Unit)
    stopJob.join()
    testScheduler.advanceUntilIdle()

    verify(node).stop()
    verify(node).destroy()
    assertNull(sut.node)
}
```

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -334,13 +334,7 @@ class WalletViewModel @Inject constructor(
...
         if (!walletExists) return
 
-        viewModelScope.launch(bgDispatcher) {
-            lightningRepo.stop()
-                .onFailure {
-                    Logger.error("Node stop error", it)
-                    ToastEventBus.send(it)
-                }
-        }
+        lightningRepo.stopDebounced()
```
ovitrif, line 432: [P1] Cancel the deferred stop on every foreground entry

The uncovered lifecycle sequence is:

1. `ON_START` begins wallet startup and sets `isStarting = true`.
2. `ON_STOP` calls this method and schedules the repo-scoped delayed stop.
3. The app returns before the delay expires, so `ON_START` calls `WalletViewModel.start()` again.
4. That call returns at the existing `isStarting` guard before reaching `LightningRepo.start()` and its `cancelPendingStop()`.
5. The delayed job fires after three seconds and stops the newly started node while the app is foregrounded.

The existing debounce test begins with startup already complete, so its foreground call reaches `LightningRepo.start()` and does not cover this sequence. I reproduced the gap on head `cda00a419`; moving `lightningRepo.cancelPendingStop()` to the top of `WalletViewModel.start()`, before the guard, makes the regression pass.

Suggested regression test for `WalletViewModelTest.kt`:

```kotlin
@Test
fun `foreground start cancels deferred stop while startup is active`() = test {
    walletState.value = walletState.value.copy(walletExists = true)
    advanceUntilIdle()
    val startEntered = CompletableDeferred<Unit>()
    val finishStart = CompletableDeferred<Unit>()
    whenever(
        lightningRepo.start(
            any(),
            anyOrNull(),
            any(),
            anyOrNull(),
            anyOrNull(),
            anyOrNull(),
            anyOrNull(),
            any(),
        ),
    ).doSuspendableAnswer {
        startEntered.complete(Unit)
        finishStart.await()
        Result.success(Unit)
    }

    sut.start()
    startEntered.await()
    sut.stop()
    sut.start()

    verify(lightningRepo).stopDebounced()
    verify(lightningRepo, times(2)).cancelPendingStop()

    finishStart.complete(Unit)
    advanceUntilIdle()
}
```

ovitrif (commented):
I re-audited the teardown and lifecycle paths against the current head and reproduced the two inline races with focused coroutine tests.

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1100/commits/332240cbaead6ce3c405f66b0d3302d698e0e1b2 fixed and added the regression test

[Screen_recording_20260724_074819.webm](https://github.com/user-attachments/assets/3545df6c-5157-4407-ae31-05ade0fe085c)

app/src/main/java/to/bitkit/services/LightningService.kt
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1100#discussion_r3644808887

app/src/main/java/to/bitkit/services/LightningService.kt
```diff
@@ -314,25 +323,49 @@ class LightningService @Inject constructor(
...
+     * when a failed start left them wedged, so the wait is bounded and the release is joined rather
+     * than run inline: cancelling a blocking FFI call does nothing, but abandoning the join lets the
+     * caller continue while the release finishes on its own.
+     */
+    private suspend fun releaseHandle(node: Node) {
+        val release = launch(ioDispatcher) {
+            runSuspendCatching { node.destroy() }
+                .onFailure { Logger.warn("Node handle release error", it, context = TAG) }
+        }
+        withTimeoutOrNull(NODE_RELEASE_TIMEOUT) { release.join() }
```
ovitrif, line 393: I’m treating native destruction as part of the lifecycle gate here. After this timeout, `stop()` reports success and `LightningRepo` publishes `Stopped`, even though the old Rust node can still be draining for tens of seconds. Immediate restart paths can then build a replacement over the same LDK state, while `wipeStorage()` and graph-reset paths can mutate that storage before the old owner releases it. That recreates the overlapping native lifetime this PR is intended to remove. Please retain an explicit release gate that prevents rebuild and destructive storage work until `destroy()` completes, or cross a safe process-recovery boundary, and cover it with distinct dispatchers plus a controllably delayed `destroy()`.

app/src/main/java/to/bitkit/services/LightningService.kt
```diff
@@ -314,25 +323,49 @@ class LightningService @Inject constructor(
         Logger.info("Node started", context = TAG)
     }
 
-    suspend fun stop() {
+    // Teardown must not be abandoned midway: a cancelled caller would leave the rust node alive and
+    // let the GC free it later on the finalizer thread, racing the next node. This covers the whole
+    // body, including the listener join, so cancellation during listener cleanup cannot skip it.
+    suspend fun stop() = withContext(NonCancellable) {
```
ovitrif, line 349: I need this stop path to handle calls made from the listener itself. `WakeNodeWorker.deliver()` calls `lightningRepo.stop()` synchronously from its registered LDK event handler while the app is backgrounded. That callback runs inside `listenerJob`, so `listenerJob.cancelAndJoin()` waits for the currently executing job to finish while that job is suspended inside `stop()`. Teardown never reaches `node.stop()` or `destroy()`, `deliverSignal` never completes, and the lifecycle mutex remains held. Please move handler-triggered teardown outside the listener job or avoid joining the current listener while preserving the external-caller cancellation guarantee, with a regression test whose event callback requests stop.

ovitrif (commented):
I re-audited exact head `2955bd0` after the two earlier P1 fixes. Those cancellation and deferred-stop findings are resolved, and all 18 hosted checks plus the focused `LightningServiceTest`, `LightningRepoTest`, and `WalletViewModelTest` suites pass.

Two distinct high-priority native-lifecycle findings remain inline, so I’m leaving this as a neutral review. I also installed this head on the Android 17 / 16 KB emulator and completed short and long background/foreground smoke cycles without a crash or ANR. The preserved dev wallet’s configured local Electrum endpoint (`10.0.2.2:60001`) was unavailable, so that device pass did not exercise teardown from a fully running node.

app/src/main/java/to/bitkit/services/LightningService.kt
jvsena42: Fixed in https://github.com/synonymdev/bitkit-android/pull/1100/commits/8ad0b71fae22e10233445ee0735ed64a640033da and https://github.com/synonymdev/bitkit-android/pull/1100/commits/b3930a0cae720ecea3672c4f68e2c4ec53d0da8a

app/src/main/java/to/bitkit/services/LightningService.kt
jvsena42: fixed in https://github.com/synonymdev/bitkit-android/pull/1100/commits/fc385e5abb9f9ee1921fff7f44eb5af6911b6035

app/src/main/java/to/bitkit/services/LightningService.kt
jvsena42: [receive-bg.webm](https://github.com/user-attachments/assets/76885d60-a4bf-4545-8345-c097117a9452)

app/src/main/java/to/bitkit/services/LightningService.kt
jvsena42: [wipe-and-restore.webm](https://github.com/user-attachments/assets/07fb8710-1771-4d3e-b065-010664dae211)

jvsena42:
checking the best approach for the electrum restart taking too long https://github.com/synonymdev/bitkit-android/actions/runs/30292125774/job/90105118405?pr=1100

jvsena42:
everything re-checked and description updated

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -756,9 +785,16 @@ class LightningRepo @Inject constructor(
...
+            // wedged, ~40s) node's drain. A rebuild-over-drain here is no worse than the pre-PR GC
+            // behaviour; the destructive storage deletes (wipe / graph-reset / scorer) keep the gate.
+            awaitRelease = false,
         ).onFailure {
-            Logger.warn("Failed ldk-node config change, attempting recovery…", context = TAG)
-            restartWithPreviousConfig()
+            // Recover in the background: a wedged node's release can gate the rebuild for tens of
+            // seconds, and the caller must surface this failure now rather than block on recovery.
+            Logger.warn("Failed ldk-node config change, recovering in background…", context = TAG)
+            scope.launch { restartWithPreviousConfig() }
```
ovitrif: I see this detached recovery—and the identical RGS branch—racing the next Connect attempt. Because recovery and the new request acquire the lifecycle mutex separately for `stop()` and `start()`, recovery can start the old config first; the new start then takes the `Running` fast path, reports success, and persists its URL without applying it.

Please serialize Electrum changes, RGS changes, and their recovery behind one dedicated configuration mutex, and assign each request a monotonically increasing generation before it waits for that mutex. Run the detached recovery under the same mutex and skip it when its captured generation is no longer current:

```kotlin
private val configChangeMutex = Mutex()
private val configGeneration = AtomicLong(0)

private fun recoverPreviousConfig(generation: Long) = scope.launch {
    configChangeMutex.withLock {
        if (generation != configGeneration.get()) return@withLock
        restartWithPreviousConfig()
    }
}
```

Keep the successful `settingsStore.update` inside that serialized transaction. Add a two-request barrier test that fails request A, pauses its recovery, starts request B, then releases A; assert A’s recovery is skipped, B’s server is the running configuration, and only B’s URL is persisted.

ovitrif (changes requested):
I re-audited exact head `140ba5e` after the rebase. Hosted build, lint, detekt, and E2E checks are green, and the earlier cancellation and self-join fixes remain resolved. I’m requesting changes for two native-lifecycle races: the Electrum/recovery path bypasses the pending-release gate, and detached recovery can race a later server change into reporting and persisting success for a configuration that never started.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
jvsena42: https://github.com/synonymdev/bitkit-android/pull/1100/commits/a42b95bdb1215c940cc09a87ae30b4ab526ad195 adds protection against stale config override on recoveries

jvsena42: I'm testing a better solution for the Electrum that could eliminate `awaitRelease` flag completely

jvsena42: https://github.com/synonymdev/bitkit-android/pull/1100/commits/ec66ea8e298f48f6d59f944020348b5a2e157867

app/src/main/java/to/bitkit/services/ElectrumProbeService.kt
```diff
@@ -0,0 +1,155 @@
...
+        runSuspendCatching {
+            openSocket(server).use { socket ->
+                socket.soTimeout = RESPONSE_TIMEOUT.inWholeMilliseconds.toInt()
+                val reader = socket.getInputStream().bufferedReader()
+                val writer = socket.getOutputStream().bufferedWriter()
+
+                // A host that accepts the connection but does not speak electrum drops or resets it
+                // mid-exchange, so report that as a probe verdict rather than a raw socket error.
+                runCatching {
+                    request(reader, writer, id = 0, method = "server.version", params = versionParams())
```
ovitrif: This probe needs to validate the JSON-RPC envelope before the working node can be torn down. `request()` currently accepts any parsed object, so a `server.version` response with a mismatched `id`, non-null `error`, or missing `result` passes. If `server.features` then returns an error, `verifyNetwork()` treats the absent genesis hash as optional and the probe succeeds. The real LDK client rejects invalid RPC responses during startup, leaving this path able to recreate the failed-start/release wedge.

Could you please parse a typed response and require the requested `id`, no `error`, and a valid non-null `result` for `server.version`; keep `server.features` optional only after successful version negotiation; and add fake-server cases for a version error, missing result, and mismatched `id`, each asserting rejection before `stop()`?

ovitrif (changes requested):
The serialization change closes the recovery race from my previous review. One release-blocking issue remains in the new Electrum preflight: `server.version` errors and mismatched responses are treated as success, so an incompatible endpoint can still take down the working node and recreate the failed-start teardown wedge. I’m limiting this request to that path.

app/src/main/java/to/bitkit/services/ElectrumProbeService.kt
jvsena42: Thanks! fixed in https://github.com/synonymdev/bitkit-android/pull/1100/commits/44911a0ef4049f874f52d8f5e9c6b5fba8f05c03 and refactored in https://github.com/synonymdev/bitkit-android/pull/1100/commits/918cde3196149c3f987a0902d6fdae96e37bc49c and https://github.com/synonymdev/bitkit-android/pull/1100/commits/4a6968b9cb06bdd5800721badcdca6d6f31bae22

ovitrif (approved):
I verified the contiguous `685d478..4a6968b` fix delta. The Electrum probe now rejects mismatched `id` values, non-null `error` responses, and missing `result` values before node teardown, with regression coverage for each path. My previous blocker is resolved, and all 18 hosted checks are green.

PR #1102, opened 2026-07-23
fix(lightning): tune scorer params

pwltr:
### Description

Improves Lightning route selection by tuning node-wide scorer fee params. These params have been used in nightly probing testing as the `next` variant. See comparison to current (stable) [here](https://legendary-disco-3qw4ekw.pages.github.io/).

iOS counterpart: https://github.com/synonymdev/bitkit-ios/pull/639


piotr-iohk (approved):
LGTM, given probing nightly results.

PR #1108, opened 2026-07-24
fix: allow terminal profile deletion

ben-kaufman:
Follow-up to #1097.

Companion iOS fix: https://github.com/synonymdev/bitkit-ios/pull/640

### Description

This PR allows terminal Pubky profile deletion to continue when private Paykit endpoint cleanup cannot finish, such as when a contact is offline or a private link is still being established.

- Still attempts private endpoint revocation before deleting the profile.
- Keeps the remote Pubky deletion authoritative, so a homeserver or connectivity failure still shows the existing retry/disconnect dialog.
- Clears local Paykit state after the Pubky profile is deleted.
- Warns that contacts may retain profile and payment information shared with them previously.

Paykit lifecycle/concurrency hardening and E2E repository changes remain outside this PR.

### Preview

N/A — behavior and confirmation-copy update only.

#### Manual Tests

- [ ] **1.** Profile → Edit Profile → Delete Profile with an offline or still-linking contact → confirm: Pubky deletion proceeds instead of being blocked by private cleanup.
- [ ] **2.** Profile → Edit Profile → Delete Profile while the Pubky homeserver is unavailable → confirm: Unable to Delete Profile still offers retry or disconnect.
- [ ] **3.** Profile → Edit Profile → Delete Profile: confirmation explains that contacts may retain previously shared profile and payment information.

#### Automated Checks

- `EditProfileViewModelTest.kt`: verifies private cleanup failure does not block a successful Pubky deletion and local Paykit cleanup.
- `./gradlew compileDevDebugKotlin` passed.
- `./gradlew detekt` passed with pre-existing findings outside this change.
- `git diff --check` passed.


piotr-iohk (approved):
tACK

ovitrif (approved):
utAck

Issue #1110, opened 2026-07-27
Channel details time uses different time format to system


guzino:
### What happened?

I use 12hr time but a connection created at 8:58 PM shows as 20:58.
"Created on" and "Opened on" both use a 24hr clock regardless of the device setting.

Settings > Advanced > Lightning Connections > (tap a connection)

Same as #677, which only covered the activity list and activity details.

<img width="640" height="824" alt="Image" src="https://github.com/user-attachments/assets/407eafcc-e6e2-46a6-8781-e6cb52c96d52" />

### Reproducibility

None

PR #1111, opened 2026-07-28
fix: respect system 24-hour time setting

### LINKED ISSUE #677: [Bug]: Activity time uses different time format to system
I use 24hr time but a payment received at 15:23 shows as 3:23 (without the pm).
iOS app correctly shows 15:23.

Image

guzino:
Fixes #677

This PR makes activity timestamps follow the device's 12/24-hour time setting.

### Description

`DatePattern.ACTIVITY_TIME` was `h:mm`, a 12-hour pattern with no meridiem, so a payment received at 15:23 rendered as `3:23`. That is wrong for anyone on a 24-hour clock and ambiguous even for 12-hour users. The two neighbouring row patterns (`ACTIVITY_ROW_DATE`, `ACTIVITY_ROW_DATE_YEAR`) were hardcoded the other way, to 24-hour, so the same activity row changed clock format depending on how old the item was. Nothing in the app consulted the system setting.

Rather than pass a 12/24-hour flag around the screens, this adds one entry point for user-facing timestamps:

- `uiDateText(timestamp, style)` in `ui/utils/DateText.kt`, backed by a `UiDateStyle` enum (`TIME`, `DATE`, `DATE_TIME`, `DATE_TIME_YEAR`) in `ext/DateTime.kt` that owns the patterns. The clock format comes from `LocalIs24HourFormat`, provided once in `ContentView` by `rememberIs24HourFormat()`, which registers for `ACTION_TIME_CHANGED` so timestamps follow a change to the setting.
- `ActivityRow` and `ActivityDetailScreen` pick a style and call it. Neither builds a pattern nor sees the flag. Picking the style by age is the pure `uiDateStyleFor(timestamp, today, zone)` in `ext/DateTime.kt`.
- `ext/Numbers.kt` (`toActivityItemDate`, `toActivityItemTime`) is removed as superseded, along with the `ACTIVITY_*` pattern constants and the unused `INVOICE_EXPIRY`, leaving `DatePattern` as plain constants.
- `Instant.formatted()` takes optional `locale` and `zone` so patterns can be asserted without depending on the CI machine's locale or timezone. Existing call sites are unchanged.

`ChannelDetailScreen` still hardcodes 24-hour via `DatePattern.CHANNEL_DETAILS` (#1110), and `toLocalizedTimestamp` defaults to `Locale.US` so App Status and Backup timestamps ignore the device locale (#1112). Both are separate screens and are left out here; with `uiDateText` in place each becomes a two-line adoption.

### Preview

before-activity-list
after-activity-list-24h
after-activity-list-12h

#### Manual Tests

- [ ] **1.** Set the device to 24-hour time (Settings > System > Date & time). Receive a payment, then check the activity row and activity detail: both show the time on a 24-hour clock (e.g. `14:38`), matching the status bar.
- [ ] **2.** Set the device to 12-hour time and reopen the activity list: the same entry reads `2:38 PM`.
- [ ] **3.** Check an activity item from earlier this year and one from a previous year: the date is unchanged and the time follows the same 12/24-hour choice as today's items.
- [ ] **4.** `regression:` Switch the device to a non-English locale and confirm activity dates still render with a localized month name.

#### Automated Checks

- Unit tests added in `DateTimeExtTest.kt`: each `UiDateStyle` in both clock formats, midnight and noon boundaries, month-name localization, and timezone handling. All use a fixed instant with explicit locale and zone so they are deterministic.
- Verified on an emulator in both 12- and 24-hour modes with a regtest transaction.


app/src/main/java/to/bitkit/ext/Context.kt
```diff
@@ -55,6 +56,11 @@ val Context.powerManager: PowerManager
 val Context.usageStatsManager: UsageStatsManager
     get() = getSystemService(Context.USAGE_STATS_SERVICE) as UsageStatsManager
 
+// Settings
+
+val Context.is24HourTimeFormat: Boolean
+    get() = DateFormat.is24HourFormat(this)
```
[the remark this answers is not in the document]
guzino: Fixed in 8ca5c38. `uiDateText` now resolves the clock format through `rememberIs24HourFormat()`, which registers for `ACTION_TIME_CHANGED` so the format updates without waiting for an unrelated recomposition.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
```diff
@@ -0,0 +1,52 @@
...
+            override fun onReceive(receiverContext: Context?, intent: Intent?) = run {
+                is24Hour = context.is24HourTimeFormat
+            }
+        }
+        ContextCompat.registerReceiver(
+            context,
+            receiver,
+            IntentFilter(Intent.ACTION_TIME_CHANGED),
+            ContextCompat.RECEIVER_NOT_EXPORTED,
+        )
```
jvsena42, line 54: every instance will registerReceiver/unregisterReceiver one receiver, causing a real junk risk for a big list

jvsena42: take a look at `ui/Locals.kt` pattern. Could create a LocalIs24HourFormat and provide a single `rememberIs24HourFormat` in ContentView/MainActivity with `uiDateText` reading `LocalIs24HourFormat.current`

app/src/main/java/to/bitkit/ui/screens/wallets/activity/components/ActivityRow.kt
```diff
@@ -363,18 +363,17 @@ private fun AmountViewContent(
...
-    val isToday = dateTime.toLocalDate() == now
-    val isThisYear = dateTime.year == now.year
-    return when {
-        isToday -> instant.formatted(DatePattern.ACTIVITY_TIME)
-        isThisYear -> instant.formatted(DatePattern.ACTIVITY_ROW_DATE)
-        else -> instant.formatted(DatePattern.ACTIVITY_ROW_DATE_YEAR)
+    val style = when {
+        dateTime.toLocalDate() == now -> UiDateStyle.TIME
+        dateTime.year == now.year -> UiDateStyle.DATE_TIME
+        else -> UiDateStyle.DATE_TIME_YEAR
```
jvsena42: Extracting a pure `fun uiDateStyleFor(timestamp: ULong, today: LocalDate, zone: ZoneId): UiDateStyle` into ext/DateTime.kt  would make it more testable

app/src/main/java/to/bitkit/ui/utils/DateText.kt
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Locale
+
+@Composable
+fun uiDateText(
+    timestamp: ULong,
+    style: UiDateStyle,
+    locale: Locale = Locale.getDefault(),
+    zone: ZoneId = ZoneId.systemDefault(),
+): String = Instant.ofEpochSecond(timestamp.toLong())
+    .formatted(style.pattern(rememberIs24HourFormat()), locale, zone)
```
jvsena42: add a remember for avoid rebuilding the pattern string and a DateTimeFormatter on every recomposition of every row

jvsena42 (changes requested):

app/src/main/java/to/bitkit/ui/utils/DateText.kt
guzino: Fixed in `3e90024f4`. `LocalIs24HourFormat` in `ui/Locals.kt`, provided once in `ContentView` — one receiver for the app instead of one per row. Left it nullable so previews and anything outside `ContentView` fall back to the device value rather than a hardcoded default.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
guzino: Fixed in `3e90024f4`. The `DateTimeFormatter` is remembered on `(style, is24Hour, locale, zone)` and the formatted string on `(formatter, timestamp)`.

app/src/main/java/to/bitkit/ui/screens/wallets/activity/components/ActivityRow.kt
guzino: Fixed in `0a102b0bc`. Pure `uiDateStyleFor(timestamp, today, zone)` in `ext/DateTime.kt`, with `today`/`zone` defaulted so the call site stays a one-liner. Four tests cover today / this year / previous year and the zone boundary.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
```diff
@@ -0,0 +1,61 @@
...
+
+@Composable
+fun uiDateText(
+    timestamp: ULong,
+    style: UiDateStyle,
+    locale: Locale = Locale.getDefault(),
+    zone: ZoneId = ZoneId.systemDefault(),
+): String {
+    val context = LocalContext.current
+    val is24Hour = LocalIs24HourFormat.current ?: context.is24HourTimeFormat
```
jvsena42: The fallback should also be remembered:  `remember(context) { context.is24HourTimeFormat }`

`context.is24HourTimeFormat` is a `Settings.System` ContentProvider query, and on the fallback path it runs on every recomposition

app/src/main/java/to/bitkit/ui/utils/DateText.kt
```diff
@@ -0,0 +1,61 @@
...
+import java.time.Instant
+import java.time.ZoneId
+import java.util.Locale
+
+@Composable
+fun uiDateText(
+    timestamp: ULong,
+    style: UiDateStyle,
+    locale: Locale = Locale.getDefault(),
+    zone: ZoneId = ZoneId.systemDefault(),
```
jvsena42, line 28: These parameters are not passed by any caller. could be removed

jvsena42 (changes requested):
The main points of last review were solved

app/src/main/java/to/bitkit/ui/utils/DateText.kt
ovitrif: I'm keeping `locale` and `zone` here. In this codebase, formatting APIs keep locale and time zone injectable so tests and previews can supply deterministic values, while their defaults keep production call sites concise.

@guzino, no change needed for this comment.

app/src/main/java/to/bitkit/ui/Locals.kt
```diff
@@ -21,6 +21,7 @@ val LocalCurrencies = compositionLocalOf { CurrencyState() }
 
 // Statics
 val LocalDrawerState = staticCompositionLocalOf<DrawerState?> { null }
+val LocalIs24HourFormat = staticCompositionLocalOf<Boolean?> { null }
```
ovitrif: Should not be a static local AFAIC (warning: I didn't consult AI), then we don't need to make it nullable… then we don't need a fallback, then we don't have to optimise at all anymore and the comment there becomes irrelevant.

https://github.com/synonymdev/bitkit-android/pull/1111/changes#diff-bac2243ecf17c09cab183ab8d66d7daf00a2a2981ee1644c4bdd0439cfe29265R31

Otherwise, even if an AI roasts my guess, you can alway just default it to `false` so you can avoid the nullable:
```kt
val LocalIs24HourFormat = staticCompositionLocalOf<Boolean> { false }
```

And still achieve what I suggest, freedom from the need to over optimise for an UI flag. There can't be many issues if default is false… we can't call this outside Compose UI context.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
ovitrif: nit: Should't even be falling back to null IMHO, see my other inline review comment on the local's declaration.

ovitrif (commented):
untested neutral review round 1 done, submitted some remarks

app/src/main/java/to/bitkit/ui/Locals.kt
guzino: Done in `3e0713edc`. Dropped the nullable and the fallback with it.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
guzino: Gone in `3e0713edc`, no fallback left.

guzino: Moot in `3e0713edc`: the fallback is gone, `uiDateText` reads `LocalIs24HourFormat.current` directly.

app/src/main/java/to/bitkit/ui/utils/DateText.kt
guzino: Keeping both then.

jvsena42 (approved):
tAck

ovitrif:
@greptile re-review latest tip

ovitrif (approved):
utAck, thanks @guzino for your first ready-to-merge PR 👏🏻 

To many more 🍻 !
This improvement might be released in the near future alongside a planned patch that addresses the most commonly occurring PROD crashes.

Issue #1118, opened 2026-07-30
refactor: migrate incrementally to navigation 3


ovitrif:
Parent: #1113
Depends on: #1114, #1115, and #1116

## Goal

Migrate Bitkit from Compose Navigation 2 to Navigation 3 through approachable, separate-concern pull requests, beginning with an isolated onboarding navigator and preserving a runnable, tested app after every merge.

## Preparation

- Update Android CLI and install or refresh the official `navigation-3` skill before refinement and implementation.
- Use the skill's current migration guidance and primary Android documentation as the implementation contract.
- Reinspect current `master` after the toolchain/dependency issues land.
- Use closed PR #554 as the coverage inventory for routes, navigator state, entry organization, sheets, transitions, deep links, timed overlays, drawer/tab behavior, QR flows, settings, transfers, widgets, startup/PIN handling, and receive/send in-sheet flows.
- Implement each slice from the settled current baseline with focused review and verification.

## Incremental boundary

Navigation 2 remains the root host during the first slice. Its onboarding destination hosts a Navigation 3 `NavDisplay` with a typed back stack. Completion and cancellation return through explicit callbacks to the legacy root. This provides a complete onboarding vertical slice without coupling it to the whole-app migration.

### PR 1: Navigation 3 foundation and onboarding

- Add mutually compatible stable Navigation 3 dependencies selected at implementation time.
- Add typed onboarding routes, `NavBackStack`, entry provider, state save/restore, and focused navigator tests.
- Host onboarding `NavDisplay` behind the existing root boundary.
- Migrate the complete onboarding route sequence and back behavior.
- Preserve completion, terms, wallet creation/import entry, process recreation, and return-to-root behavior.
- Add/update the onboarding journey and Compose/instrumentation coverage.

### PR 2: Shared route model and root screen navigation

- Establish the shared typed route hierarchy and repository `Navigator` contract.
- Migrate root wallet, drawer, tab, settings-entry, and non-sheet screen transitions.
- Preserve deep-link handoff boundaries and restored navigation state.
- Keep remaining legacy destinations reachable through explicit adapters.

### PR 3: Sheet scene strategy and in-sheet flows

- Introduce the Navigation 3 scene strategy for Bitkit bottom sheets.
- Migrate receive, send, selectors, suggestions, and nested in-sheet navigation.
- Preserve swipe, back, overlay-tap, transition, amount-retention, and nested-dismiss behavior.
- Keep sheet presentation aligned with the supported design.

[...]

PR #1119, opened 2026-07-30
feat: add deep links to every screen

### LINKED ISSUE #659: Add deep link linking to every screen in the app
Ideally with support to pass values for every input field if the destination screen has one via the link's query params.

It would help making the app more AI-self-check friendly.

guzino:
Fixes #659
Refs #1118
Refs #1126

This PR adds a `bitkit://screen/...` URI for every screen in the app, bottom sheets included, gated on dev mode.

### Description

A destination declares whether it may be entered directly, in the model that defines it. 88 root routes are `Routes.DeepLinkable` and 15 are `Routes.InternalOnly`, and `deepLinkableComposable` accepts only the former, so a new screen has to make the choice rather than inherit one. `ScreenDeepLinks` adapts the URI and nothing else: it derives the id by kebab-casing the class name, so `Routes.RgsServer` is reachable at `bitkit://screen/rgs-server`. Arguments without a default become path segments, arguments with a default become query parameters.

Bottom sheets needed a second mechanism. They are not in the root graph: they are `Sheet` values rendered by `SheetHost`, each carrying the start route of its own nested `NavHost`, so `NavController.handleDeepLink` cannot reach them. Each sealed route family declares `DeepLinkStart` or `InternalOnly` on its states and owns a `fromDeepLink` lookup, and `SheetDeepLinks` picks the family and wraps the result. Ids derive from class names throughout, so `bitkit://screen/widgets/price-edit` comes from `Sheet.Widgets` plus `WidgetsRoute.PriceEdit`.

- Registers 38 sheet paths across `send`, `receive`, `backup`, `widgets`, `hardware` and the three sheets with no nested graph.
- Gates screen links on `isDevModeEnabled` in `AppViewModel.handleDeeplinkIntent`, and holds one until the wallet is loaded; `ContentView` replays it.
- Detaches the screen URI from the Activity intent in `MainActivity` once the gated pipeline has read it, so graph creation cannot navigate outside the dev-mode gate.
- Classifies the recovery phrase, auth, transfer and LNURL destinations as internal, so they yield no link at all. Links navigate and prefill only, and never send, broadcast or change a setting without the usual confirmation.
- Leaves payment URIs (`bitcoin:`, `lightning:`, `lnurl*`) on the scanner decode path, untouched.
- Dismisses an open sheet only once a root-screen URI has matched, so a rejected link leaves the visible screen alone.
- Removes `docs/deeplinks.md`. The route markers, `deepLinkableComposable`, the sheet families' `fromDeepLink` and the unit tests are the source of truth, and a parallel doc drifted three times during review.
- Keeps the 25 flow-only sheet states internal. None is tap-reachable, so none is a user-facing defect. Direct support for the late transfer screens is tracked in #1126.

Not every nested state is a valid start destination. I fired all 51 against an emulator, then re-ran each rejected one in isolation against logcat. The states left internal fall into three groups:

- `SendRoute.FeeRate` and `FeeCustom` sit inside `navigationWithDefaultTransitions<SendRoute.FeeNav>`. A nested graph's child cannot be a `NavHost` start destination, so both throw `IllegalStateException: Cannot find startDestination ... from NavGraph`. `send/fee-nav` resolves but renders only the screen title.
- `send/quick-pay` throws on `requireNotNull(quickPayData)` (`SendSheet.kt:309`). `send/confirm` and the four receive confirm/liquidity routes do not crash but render empty or zero-amount screens. `send/confirm` offers "Swipe To Pay" over a payment that was never built.
- `backup/success` reports a backup that never ran, and its OK button persists `backupVerified = true` (`BackupNavSheetViewModel.onSuccessContinue`). `backup/warning` is one tap upstream of the same write. `hardware/searching` waits forever because discovery starts from the intro's continue action, and `hardware/paired` claims a paired device over default state.

Tests pin the classification, and a final sweep confirmed the rejected paths no-op with the wallet overview intact and zero fatal exceptions.

### Preview

N/A

### QA Notes

Dev mode is on by default on debug builds (Settings ▸ Advanced ▸ Dev Settings). The app must be past onboarding.

#### Manual Tests

- [x] **1.** `adb shell am start -a android.intent.action.VIEW -d "bitkit://screen/settings" to.bitkit.dev` → Settings opens. A mistyped id still resolves `MainActivity`, because the manifest accepts every `bitkit:` URI by scheme, so check logcat for `Unhandled screen deeplink` rather than relying on `-W`.
- [x] **2.** Deep-linked screen → Back: returns to the wallet overview, not the launcher.
- [x] **3.** `bitkit://screen/send` → Send sheet on the recipient picker. `bitkit://screen/widgets/price-edit` → Bitcoin Price editor.
- [x] **4.** `bitkit://screen/recovery-mnemonic` and `bitkit://screen/backup/show-mnemonic` → screen unchanged, recovery phrase never shown, logcat carries `Unhandled screen deeplink`.
- [x] **5.** `bitkit://screen/send/fee-rate` and `bitkit://screen/backup/success` → screen unchanged, no crash, and `backupVerified` is untouched.
- [x] **6.** PIN screen up → fire any screen link: navigates behind the PIN, which is still required before anything is visible.
- [x] **7.** Dev mode off → any `bitkit://screen/...` link is ignored, on warm start and on cold start. Settings ▸ Support, tap Version five times to toggle.
- [x] **8.** `regression:` scan a `bitcoin:` / `lightning:` / `lnurl` URI → still decodes through the scanner path.

#### Automated Checks

- Unit tests in `ScreenDeepLinksTest.kt`: id derivation, path vs query argument placement, and that every route declares its eligibility.
- Unit tests in `SheetDeepLinksTest.kt`: bare-id defaults, sub-route selection, case-insensitive lookup, ids derived from the `Sheet` class names, and that every state marked as a start is registered in its family lookup.
- Instrumented `ScreenDeepLinkDetachmentTest`: graph creation stays on Home after detachment, the captured URI reaches Settings only through the replay, and a denied route is not matched.
- Two journeys under `journeys/deeplinks/`, including cold-start cases for dev mode off and on.
- Fired all 38 sheet paths and the screen links against an API 33 emulator and confirmed each renders. The rejected paths no-op without crashing.


app/src/main/java/to/bitkit/ui/utils/SheetDeepLinks.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        Sheet.Receive(ReceiveRoute.QR),
+        Sheet.Receive(ReceiveRoute.Amount),
+        Sheet.Receive(ReceiveRoute.EditInvoice),
+        Sheet.Receive(ReceiveRoute.AddTag),
+        Sheet.Receive(ReceiveRoute.GeoBlock),
+
+        Sheet.Backup(BackupRoute.Intro),
+        Sheet.Backup(BackupRoute.Warning),
+        Sheet.Backup(BackupRoute.Success),
```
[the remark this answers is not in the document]
guzino: Correct, and the write is worse than the render. `onSuccessContinue` persists `backupVerified = true` (`BackupNavSheetViewModel.kt:139-141`), so the OK button on a deep-linked success screen would mark a wallet backed up without the recovery phrase ever being shown. That breaks this PR's own rule that links never change a setting without the usual confirmation.

Removed `backup/success`, and also `backup/warning`, which is one tap upstream of the same write. `backup` now exposes `intro`, `multiple-devices` and `metadata` only.

Resolved in `511250c90`.

app/src/main/java/to/bitkit/ui/utils/SheetDeepLinks.kt
```diff
@@ -0,0 +1,85 @@
...
+        Sheet.Widgets(WidgetsRoute.BlocksEdit),
+        Sheet.Widgets(WidgetsRoute.HeadlinesPreview),
+        Sheet.Widgets(WidgetsRoute.HeadlinesEdit),
+        Sheet.Widgets(WidgetsRoute.FactsPreview),
+        Sheet.Widgets(WidgetsRoute.CalculatorPreview),
+        Sheet.Widgets(WidgetsRoute.SuggestionsPreview),
+
+        Sheet.Hardware(HardwareRoute.Intro),
+        Sheet.Hardware(HardwareRoute.Searching),
+        Sheet.Hardware(HardwareRoute.Paired),
```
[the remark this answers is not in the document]
guzino: Confirmed both. Discovery is started by the intro's continue action rather than by the screen, so a deep-linked `hardware/searching` spins forever, and `hardware/paired` renders PAIRED TREZOR over default state with no device behind it.

Same class as `send/confirm`, which this PR already excluded for offering Swipe To Pay over a payment that was never built. My registry was inconsistent in keeping these two.

Removed both. `hardware` now exposes `intro` only.

Resolved in `511250c90`.

app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -234,6 +235,8 @@ class MainActivity : FragmentActivity() {
             return
         }
 
+        intent.data?.let { if (ScreenDeepLinks.isScreenDeepLink(it)) intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TASK) }
```
ovitrif: I found that cold launches bypass the dev-mode gate. AndroidX automatically handles the Activity's data intent when this Activity-backed `NavHost` sets its graph; this line retains the screen URI, while `AppViewModel` gates only the later manual replay. A valid root link can therefore navigate with dev mode off. Please clear the screen URI from the Activity intent before graph creation and replay it only through the gated pipeline, with cold-start coverage for dev mode off and on.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
```diff
@@ -0,0 +1,51 @@
...
+
+object ScreenDeepLinks {
+    const val SCHEME = "bitkit"
+    const val HOST = "screen"
+
+    private const val BASE_URI = "$SCHEME://$HOST"
+
+    private val CAMEL_HUMP = Regex("(?<=[a-z0-9])(?=[A-Z])")
+
+    private val DENIED: Set<KClass<out Routes>> = setOf(
```
ovitrif: I found that `bitkit://screen/external-confirm` creates a fresh `ExternalNav`-scoped view model with `peer = null`. Its enabled swipe reaches `onConfirm()`, which calls `requireNotNull(peer)` and crashes; `external-amount` can advance into the same state. Please deny these cold-start routes or restore and validate the required peer state before enabling confirmation, and cover both URIs from fresh state.

app/src/main/java/to/bitkit/ui/utils/Transitions.kt
```diff
@@ -100,7 +100,7 @@ inline fun <reified T : Any> NavGraphBuilder.navigationWithDefaultTransitions(
 @Suppress("LongParameterList", "MaxLineLength")
 inline fun <reified T : Any> NavGraphBuilder.composableWithDefaultTransitions(
     typeMap: Map<KType, NavType<*>> = emptyMap(),
-    deepLinks: List<NavDeepLink> = emptyList(),
+    deepLinks: List<NavDeepLink> = ScreenDeepLinks.linksFor(T::class),
```
ovitrif, line 128: I found that `bitkit://screen/home` is generated but never attached to the graph. `Routes.Home` still uses raw `composable<Routes.Home>` in `ContentView`, so this default is skipped and the URI is unhandled from another screen. Please attach the link explicitly and test matching against the assembled graph.

docs/deeplinks.md
```diff
@@ -0,0 +1,112 @@
...
+| `Contacts(showAddContactSheet = false)` | `bitkit://screen/contacts?showAddContactSheet={showAddContactSheet}` |
+
+### Firing a link
+
+```sh
+adb shell am start -W -a android.intent.action.VIEW \
+  -d "bitkit://screen/settings" to.bitkit.dev
+```
+
+Escape `&` as `\&` when passing more than one query parameter. `-W` reports whether the intent
```
ovitrif: I found that `-W` verifies only Activity resolution, not the screen ID. The manifest accepts every `bitkit:` URI by scheme, so `bitkit://screen/typo` still resolves `MainActivity`; the in-app handler rejects it later. Please document a destination or logcat assertion instead of treating `-W` as route validation.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3279,6 +3283,16 @@ class AppViewModel @Inject constructor(
             return@launch
         }
 
+        if (ScreenDeepLinks.isScreenDeepLink(uri)) {
```
ovitrif: I found that `bitkit://screen/recovery-mode` never reaches this deny gate. `isRecoveryModeDeeplink()` runs first and accepts any URI whose only path segment is `recovery-mode`, so this URL calls `setRecoveryMode(true)` and navigates before the dev-mode check. Please classify the `screen` namespace before the legacy recovery parser, or restrict that parser to the exact legacy URI, and add this denied URL to `AppViewModel` coverage.

ovitrif (changes requested):
Thanks for addressing the existing backup and hardware findings; I confirmed both fixes on 511250c.

I found three blocking issues: cold-start links bypass the dev-mode gate, `bitkit://screen/recovery-mode` is consumed by the legacy parser before the gate, and `external-confirm` can reach invalid fresh state and crash. I also left two non-blocking corrections for Home registration and the `adb -W` guidance.

app/src/main/java/to/bitkit/ui/MainActivity.kt
guzino: Good catch. Cold start went through AndroidX automatic intent handling rather than the gated replay, so links still navigated with dev mode off. The data is now cleared from the activity intent once the gate has read it, which also made the CLEAR_TASK flag unnecessary.

Resolved in 09239fbb0

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
guzino: Agreed. The legacy parser matched on the path segment alone, so it ran first and switched recovery mode on. The screen namespace is now classified ahead of it and that URL falls through to the deny list, with the legacy one unchanged.

Resolved in 09239fbb0

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
guzino: Denied both, and external-success too, since it reports a completed channel open over the same empty state. Restoring the peer instead would mean re-running the connect flow validation from a URI, which belongs with that flow.

Resolved in 09239fbb0

app/src/main/java/to/bitkit/ui/utils/Transitions.kt
guzino: Right, Home is the last destination still on raw composable, so it never picked up the default and nothing was attached to the graph. Passed the links explicitly at that call site and confirmed it resolves from another screen now.

Resolved in 09239fbb0

docs/deeplinks.md
guzino: You are right, -W only proves an activity resolved. The doc and the first QA note now assert on the Unhandled screen deeplink warning in logcat instead.

Resolved in 09239fbb0

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
```diff
@@ -0,0 +1,54 @@
...
```
ovitrif: Please deny `SavingsProgress`, `SettingUp`, `SpendingAdvanced`, `SpendingConfirm`, `SpendingHwSign`, and `SpendingHwSigned` in this PR. These destinations depend on transfer state created by earlier steps, but their generated links do not reconstruct that state: fresh links render empty state or return home, and `SavingsProgress` can execute using default or retained channel state. Please track direct support for these late-flow screens in a separate GitHub issue. That implementation will require the `lsp` skill and a flow-specific setup contract that prepares the state expected after the preceding user and app actions, performs the required server-side preparation, passes stable identifiers as route parameters, and reloads validated data before navigation.

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -301,6 +304,27 @@ fun ContentView(
...
+
+        navController.currentBackStackEntryFlow.first()
+        appViewModel.consumeScreenDeepLink()
+
+        SheetDeepLinks.sheetFor(uri)?.let {
+            appViewModel.showSheet(it)
+            return@LaunchedEffect
+        }
+
+        val request = Intent(Intent.ACTION_VIEW, uri)
```
ovitrif, line 320: Please dismiss the active sheet before forwarding a root-screen URI to `NavController`. This branch leaves `currentSheet` unchanged, so a link such as `bitkit://screen/settings` navigates behind an expanded send or receive sheet and does not expose the requested screen. Add coverage for sheet-to-root navigation while keeping sheet-to-sheet replacement working.

ovitrif (changes requested):
Thanks for addressing the earlier findings. Three blocking issues remain:

- Cold-start coverage is missing at the Activity/`NavHost` boundary.
- Late transfer destinations allow entry without the flow state they require.
- Root-screen links route behind an active sheet.

app/src/main/java/to/bitkit/ui/MainActivity.kt
ovitrif: The cold-start bypass is fixed in code. The added tests cover only `AppViewModel` queue/drop state, leaving the Activity/`NavHost` boundary that caused the bypass unprotected. Please extract the screen-intent detachment into an internal helper and add an instrumented regression test with a `ComponentActivity` and an Activity-backed `TestNavHostController` containing `Home` and `Settings`. The regression test should prove that graph creation stays on `Home` after detachment and that the captured URI reaches `Settings` only through the gated replay. Please also add force-stop cases for dev mode off and on to `screen-deeplink.xml`, asserting `Home` when off and `Settings` when on.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
guzino: Denied all six. They all read the activity-scoped transfer view model, so a fresh link gets default or retained state rather than the flow that built it. Direct support and the setup contract are tracked in #1126.

Resolved in ad4ec8049

app/src/main/java/to/bitkit/ui/ContentView.kt
guzino: Agreed. The root branch now hides an open sheet first, and a sheet URI still replaces the open sheet rather than dismissing it. Both paths are covered in ContentViewTest.

Resolved in ad4ec8049

app/src/main/java/to/bitkit/ui/MainActivity.kt
guzino: Extracted the detachment into detachScreenUri and covered it with the instrumented test: graph creation stays on Home once the URI is detached, and the captured URI reaches Settings only through the replay. The journey now has the cold-start cases too.

Resolved in 1ff7743d2

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -301,6 +305,31 @@ fun ContentView(
...
+
+        navController.currentBackStackEntryFlow.first()
+        appViewModel.consumeScreenDeepLink()
+
+        SheetDeepLinks.sheetFor(uri)?.let {
+            appViewModel.showSheet(it)
+            return@LaunchedEffect
+        }
+
+        if (shouldDismissSheetForScreenLink(uri, appViewModel.currentSheet.value)) {
```
ovitrif: `backup/show-mnemonic` is denied by `SheetDeepLinks`, but it reaches this branch and hides the active backup intro before `handleDeepLink` reports the URI as unhandled. The new sheet journey therefore returns to the wallet overview instead of leaving the visible screen unchanged; any rejected screen URI can similarly dismiss the current sheet. Could we confirm that the URI matches a root destination before hiding the sheet and add regression coverage for a denied URI while a sheet is open?

ovitrif (changes requested):
One blocking navigation regression remains: rejected screen links dismiss an active sheet before route handling determines that the URI is invalid.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
```diff
@@ -0,0 +1,69 @@
...
```
ovitrif: `ScreenDeepLinks` becomes a second owner of the route model because `DENIED` duplicates direct-entry eligibility outside the sealed `Routes` hierarchy. Every future destination change therefore needs to synchronize the route hierarchy, navigation graph, this registry, its tests, the documentation, and the journeys.

That duplication does not guard future changes:
- A new route becomes externally reachable automatically because `composableWithDefaultTransitions` calls `linksFor(T::class)` by default.
- A route that should remain internal can remain externally reachable because its author must also know to add it to `DENIED`.
- The every-route test does not enforce an explicit eligibility decision because every unlisted route is accepted as deep-linkable.

The navigation model should follow the same ownership principle we use from DDD: behavior and invariants live with the model that owns them. Within this navigation boundary, the sealed `Routes` hierarchy should own whether a destination may be entered directly, while `ScreenDeepLinks` only adapts external URIs.

The sealed hierarchy should make that choice explicit:

```kotlin
sealed interface Routes {
    sealed interface DeepLinkable : Routes
    sealed interface InternalOnly : Routes

    @Serializable
    data object Settings : DeepLinkable

    @Serializable
    data object SpendingConfirm : InternalOnly
}
```

The graph helper should then accept only `Routes.DeepLinkable`, making every new destination declare its policy in the same model that defines it. Could we move root-screen eligibility into the sealed `Routes` hierarchy and constrain the graph helper to `Routes.DeepLinkable`, leaving `ScreenDeepLinks` responsible only for URI adaptation?

app/src/main/java/to/bitkit/ui/utils/SheetDeepLinks.kt
```diff
@@ -0,0 +1,81 @@
...
+import android.net.Uri
+import to.bitkit.ui.components.Sheet
+import to.bitkit.ui.screens.wallets.receive.ReceiveRoute
+import to.bitkit.ui.sheets.BackupRoute
+import to.bitkit.ui.sheets.SendRoute
+import to.bitkit.ui.sheets.WidgetsRoute
+import to.bitkit.ui.sheets.hardware.HardwareRoute
+
+object SheetDeepLinks {
+    private val SHEETS: List<Sheet> = listOf(
```
ovitrif: `SheetDeepLinks` correctly fails closed because `SHEETS` is a positive allowlist. The maintenance defect is that this object becomes a second catalog of which nested states may safely start a flow, separate from the sealed `SendRoute`, `ReceiveRoute`, `BackupRoute`, `WidgetsRoute`, and `HardwareRoute` families that define those states.

Future flow changes therefore need synchronized maintenance:
- A startable state needs to be declared in its route family and registered in `SHEETS`.
- Reclassifying a flow-only state therefore changes this catalog and its parallel tests instead of one route contract.
- The route declaration does not communicate whether the state is a valid independent entry point.

The nested navigation models should follow the same DDD ownership principle: each sealed route family owns the invariant that determines whether a state can start independently, while `SheetDeepLinks` only adapts the external URI.

Each sealed route family should make that choice explicit and own its lookup:

```kotlin
sealed interface SendRoute {
    sealed interface DeepLinkStart : SendRoute
    sealed interface InternalOnly : SendRoute

    @Serializable
    data object Recipient : DeepLinkStart

    @Serializable
    data object Confirm : InternalOnly

    companion object {
        fun fromDeepLink(path: String): DeepLinkStart? = TODO()
    }
}
```

`SheetDeepLinks` should keep the standalone `Sheet` entries it directly owns. For nested flows, it should select the sheet family, delegate the child path to that family’s route-owned lookup, and wrap the eligible state in the corresponding `Sheet`. Could we move nested sheet-entry eligibility and lookup into the existing sealed route families, leaving `SheetDeepLinks` as the family and standalone URI adapter?

changelog.d/next/1119.added.md
```diff
@@ -0,0 +1 @@
+Every screen can now be opened directly by a `bitkit://screen/...` link while dev mode is on, so QA and bug reports can jump straight to a screen instead of navigating to it.
```
ovitrif: Changelog fragments should describe public-facing capabilities. This entry belongs here because people can build the app themselves and provide their own Firebase configuration when needed. Could we remove the internal `so QA and bug reports...` framing and use: “Developers running their own build can now open supported screens directly with `bitkit://screen/...` links while dev mode is enabled”?

ovitrif (changes requested):
The deep-link implementation currently splits navigation policy across the route models and parallel registries, so I came up with 4 required corrections to keep direct entry safe and make the route hierarchy the single source of truth, and fix the changelog:
- A denied root-screen link dismisses an active sheet before rejection, so rejected input changes the visible flow.
- Root-screen eligibility has a second owner in `ScreenDeepLinks.DENIED`, so new routes become reachable without an explicit decision and every route change must synchronize a parallel policy.
- `SheetDeepLinks.SHEETS` correctly fails closed while duplicating which nested states may start a flow outside the sealed route families that define them.
- The changelog is incorrect because it describes an internal QA workflow rather than the public developer capability available to people running their own builds.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
guzino: Moved eligibility into the sealed hierarchy, so DENIED is gone.

I could not constrain the shared helper: it also registers the sheet and startup families, so 73 of its 171 call sites are not Routes, and a same-name overload clashes. Root screens use a separate deepLinkableComposable.

Resolved in c4f4bcaf5

changelog.d/next/1119.added.md
guzino: Rewritten to the wording you suggested.

Resolved in 4362cc9

app/src/main/java/to/bitkit/ui/utils/SheetDeepLinks.kt
guzino: Moved this into the route families too. Each now declares DeepLinkStart or InternalOnly and owns its fromDeepLink lookup, so SheetDeepLinks only picks the family and wraps the result. A state marked as a start but missing from its lookup now fails a test.

Resolved in 0e0a16be3

guzino:
The red e2e-status is not from this branch. e2e-tests-staging - pubky_paykit fails the same four contact specs on ContactViewName here and on #1123, which touches only channel-details time formatting. My #1111 passed that job on 29 Jul and ContactDetailScreen has not changed since May, so the break is on master or in the staging env rather than in either PR.

The rest is green: build, build-local, build-staging, detekt, and both local e2e specs.

app/src/main/java/to/bitkit/ui/ContentView.kt
guzino: Reordered so the sheet closes only after the link has matched, which leaves a rejected URI with the visible screen untouched. ContentViewTest covers a rejected link arriving while a sheet is open.

Resolved in 4362cc9

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -1928,320 +1961,324 @@ fun NavController.navigateToLanguageSettings() = navigateTo(Routes.LanguageSetti
...
-    data object FundingAdvanced : Routes
+    data object FundingAdvanced : Routes.DeepLinkable
 
     @Serializable
-    data object ExternalNav : Routes
+    data object ExternalNav : Routes.DeepLinkable
 
     @Serializable
-    data class ExternalConnection(val scannedNodeUri: String? = null) : Routes
+    data class ExternalConnection(val scannedNodeUri: String? = null) : Routes.DeepLinkable
```
ovitrif, line 2157: `ExternalConnection` is marked `Routes.DeepLinkable`, but its NavHost registration still uses `composableWithDefaultTransitions`, which no longer attaches screen deep links by default. Before the ownership migration this URI was reachable; now `bitkit://screen/external-connection` is unhandled even though the route model says it may be entered directly. Could we register it with `deepLinkableComposable` like the other DeepLinkable leaves under nested graphs?

docs/deeplinks.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+bitkit://screen/<sheet-id>[/<route-id>]
+```
+
+The sheet id is the `Sheet` class name in kebab-case and the route id is the nested route's class
+name, so the same rule holds as for screens. The bare id opens the sheet at its first registered
+route (`bitkit://screen/send` is the recipient picker, `bitkit://screen/backup` the backup intro).
+
+Each sealed route family owns which of its states may start a flow, through a `DeepLinkStart` marker
```
ovitrif: This ownership note is right, but the rest of `docs/deeplinks.md` still contradicts the fail-closed model (auto-registration via `composableWithDefaultTransitions`, hand-maintained `SheetDeepLinks` entries). For this developer-oriented feature the durable source of truth is the route markers, `deepLinkableComposable`, sheet-family `DeepLinkStart` / `fromDeepLink`, and the unit tests, so a long parallel doc will keep drifting. Could we delete `docs/deeplinks.md` (or shrink it to a short pointer to those code/tests) instead of maintaining it as a second catalog?

ovitrif (changes requested):
Thanks for addressing the earlier findings. One blocking issue and one docs cleanup remain:
- `ExternalConnection` is still `Routes.DeepLinkable` but registered with `composableWithDefaultTransitions`, so `bitkit://screen/external-connection` lost its Nav deep link after that helper stopped auto-registering links.
- `docs/deeplinks.md` no longer matches the fail-closed route-owned model and should be removed rather than kept as a second source of truth for this developer-oriented feature.

PS. We can ignore the red e2e failures, it's caused by a premature merge of a PR in the e2e repo targeting paykit-related PRs that are not yet into `master`. Classic PR stacks issues 🙃 .

app/src/main/java/to/bitkit/ui/ContentView.kt
ovitrif: Confirmed fixed on c755819: the sheet now closes only after `handleDeepLink` succeeds, so a rejected URI such as `backup/show-mnemonic` leaves the open sheet unchanged.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
ovitrif: Confirmed fixed on c755819: `DENIED` is gone and root-screen eligibility now lives on `Routes.DeepLinkable` / `Routes.InternalOnly` with `deepLinkableComposable` registration, though `ExternalConnection` still needs that wiring.

app/src/main/java/to/bitkit/ui/utils/SheetDeepLinks.kt
ovitrif: Confirmed fixed on c755819: sheet families now own `DeepLinkStart` / `fromDeepLink`, and `SheetDeepLinks` only selects the family and wraps the result.

changelog.d/next/1119.added.md
ovitrif: Confirmed fixed on c755819: the changelog now describes the developer-facing capability.

app/src/main/java/to/bitkit/ui/ContentView.kt
guzino: Good catch, that one was imported bare as ExternalConnection so my sweep to the new helper missed it. Registered with deepLinkableComposable now and the URI opens Manual Setup again on device. The other 15 non-linking registrations are exactly the InternalOnly routes.

Resolved in ce7f8a847

docs/deeplinks.md
guzino: Deleted it. The markers, deepLinkableComposable, the family lookups and the unit tests are the source of truth, and that doc drifted three times during this review alone. The journeys stay since they are executable.

Resolved in ce7f8a847

ovitrif (approved):
The ExternalConnection deep link is restored and docs/deeplinks.md is removed; this looks good to merge.

app/src/main/java/to/bitkit/ui/ContentView.kt
ovitrif: Confirmed fixed on e420f42: `ExternalConnection` remains registered with `deepLinkableComposable` after the master merge, so `bitkit://screen/external-connection` stays reachable.

docs/deeplinks.md
ovitrif: Confirmed fixed on e420f42: `docs/deeplinks.md` remains deleted after the master merge.

PR #1122, opened 2026-07-31
fix: integrate ldk crash fixes

### LINKED ISSUE #1120: fix: integrate ldk crash fixes
## Goal

Upgrade `com.synonym:ldk-node-android` from `0.7.0-rc.58` to the first published release containing both native crash fixes so Bitkit Android ships them in `2.5.0`.

The runtime self-drop fix tracked in synonymdev/ldk-node#94 is published in `v0.7.0-rc.60`. The peer-store lock fix tracked in synonymdev/ldk-node#99 is under review in synonymdev/ldk-node#101, which prepares `v0.7.0-rc.61`. Adopt the combined release after its Android artifact and matching native debug symbols are published.

## Deliverables

- Update the LDK Node Android dependency to the published release containing both fixes.
- Adapt affected Android integration points to the released bindings.
- Preserve matching native build IDs and debug-symbol coverage for every packaged ABI.
- Add one user-facing fix fragment under `changelog.d/next/`.

## Acceptance criteria

- Bitkit Android resolves and packages the published LDK Node Android release containing both fixes.
- The packaged `libldk_node.so` files and the release's native debug-symbol archive have matching build IDs for every release ABI.
- `just compile`, `just test`, and `just lint` pass.
- A targeted device journey exercises node start, stop, restart, receive, and background/foreground transitions without a native abort, a `free_node` timeout, or an input-dispatch ANR.

ovitrif:
Fixes #1120

This PR updates the Android app to the published LDK node `v0.7.0-rc.62` release, integrating the native crash fixes from synonymdev/ldk-node#101.

### Description

- Consumes the release that removes the peer-store lock before persistence and avoids blocking the async runtime during Electrum client teardown.
- Includes the corrected peer-disconnect transaction: durable removal failures now propagate to Bitkit, keep the live peer connected, and remain retryable.
- Keeps the native debug-symbol artifact aligned with the packaged Android libraries for all release ABIs.

Release: https://github.com/synonymdev/ldk-node/releases/tag/v0.7.0-rc.62

### Preview

N/A — dependency integration only.

#### Manual Tests

- [x] **1.** Home → Receive: Savings and Spending tabs plus the receive QR render normally.
- [x] **2.** `regression:` Home → background the app → foreground Bitkit: Home remains responsive and the node stays usable.
- [x] **3.** Settings → Advanced → Dev Settings → LDK → Restart: Restart is disabled during stop/start, becomes enabled after completion, and Receive still works.

#### Automated Checks

- No test coverage added, modified, or removed; the change is dependency-only and the existing suite is the regression guard.
- Local verification: `just compile`, `just test`, `just lint`, and native debug-symbol synchronization pass using the authenticated published package.
- Published AAR SHA-256 is `d7e5f50a968fee1dfaec753fa8fa325e7b3c17b6b81105bd053a013104a6dde5`; packaged APK and release-symbol build IDs match for `armeabi-v7a`, `arm64-v8a`, and `x86_64`.
- A 24-action Pixel 6 emulator journey covers receive before and after restart, background/foreground handling, restart completion, and final Home responsiveness with no native crash, timeout, or ANR signature.


jvsena42 (approved):
Approving. Verified on a Pixel-class arm64 emulator (API 36) with a **fresh install** of this branch and a new regtest wallet, node connected to the staging LSP. Confirmed the build resolves `com.synonym:ldk-node-android:0.7.0-rc.62`.

### Test flows

All three manual tests from the QA Notes pass, driven as journeys (added under `journeys/ldk-crash-fixes/`):

| Flow | Result |
| --- | --- |
| **1.** Home → Receive: Savings QR + Spending tab | ✅ QR renders, Spending tab opens, sheet closes back to a responsive Home |
| **2.** `regression:` Home → background → foreground | ✅ node stops cleanly on background, restarts on foreground, no splash/onboarding re-entry, Receive still works |
| **3.** Dev Settings → LDK → Restart | ✅ Restart row disabled during stop/start, re-enabled on completion (~12s), Receive works after |

Extra stress on the same paths, beyond the listed tests:

- **5 background/foreground cycles** with mixed 1s and 8s backgrounds — every cycle stopped and restarted the node and reconnected the trusted LSP peer (`connectedPeers='1'`).
- **3 consecutive LDK restarts**, each with a second tap during the disabled window — 6 taps produced exactly 3 restarts, so the disabled state genuinely blocks re-entry.

No native crash, ANR, or tombstone across the whole session: `logcat -b crash` empty, no `SIGABRT`/`SIGSEGV`/`libldk_node` frames, app PID unchanged throughout. Only recurring log noise is environmental — stale staging RGS (`gossip data is more than two weeks old`) and `canReceive = false` on a channel-less fresh wallet.

Caveat on scope: this is positive evidence that the stop/start paths are clean, not a before/after demonstration of the original abort — I did not observe the native crash on the rc.58 baseline either.

### One pre-existing defect found (not from this PR, not blocking)

Backgrounding the app ~1s after tapping **Restart** makes the restart's `start()` stall inside the Electrum fee-rate estimation for its full ~15s deadline and then fail — 3/3 reproductions:

```
Starting up LDK Node …                                   ← restart's start()
   … 15.4s with no ldk_node progress …
ERROR [ldk_node::chain::electrum] Updating fee rate estimates timed out: deadline has elapsed
ERROR [LightningRepo.kt:1881] Failed to start node during restart
      [AppError='LDK Node error: Feerate estimation update timeout.']
Fee rate cache update finished in 682ms.                 ← next attempt, same work
```

The identical work completes in ~680ms on the following attempt, and unrelated Ktor calls (Blocktank FX, widgets) keep succeeding during the stall, so it is not loss of network. Because `restartNode()` calls `start(shouldRetry = false)`, the node is left in `ErrorStarting` behind a failure toast and recovers only via the start triggered by the next foreground resume.

**Verified this is not a regression from the bump:** I built and installed the `0.7.0-rc.58` baseline and reproduced it 2/2 with the same signature, then restored rc.62. Captured as a known-failing journey (`restart-while-backgrounded.xml`) — worth a separate issue.

ovitrif:
@jvsena42 Thanks for documenting the restart/backgrounding failure. I created #1125 as a Bug with your reproduction evidence and the `restart-while-backgrounded.xml` journey. It tracks bounded automatic recovery so a transient fee-rate timeout does not leave the node in `ErrorStarting`.

ovitrif:
### Merge rationale

I am treating this PR as ready to merge at the reviewed head c7340c4.

- The diff is limited to updating `com.synonym:ldk-node-android` from `0.7.0-rc.58` to `0.7.0-rc.62` and adding its changelog fragment.
- Joao approved this exact head after fresh-install testing, five background/foreground cycles, three consecutive LDK restarts, and checks for native crashes, ANRs, and tombstones.
- Build, lint, Detekt, both APK builds, and every other E2E shard passed.
- The remaining red `pubky_paykit` shard is attributable to shared E2E test sequencing and has no causal path through this PR's two changed files.

### Why `pubky_paykit` is failing across Android PRs

- #1097 implements the future save-contact behavior that opens Contact Detail.
- #1097 was a stacked PR: it merged into the feature branch owned by #1084, not into `master`. #1084 remains open, so Android `master` does not contain that navigation behavior.
- The companion test change in synonymdev/bitkit-e2e-tests#200 was then merged into the shared E2E `main` branch. Its `addContact()` helper now waits for `ContactViewName`, which exists on Contact Detail.
- Android `master` still completes the save and returns to the contacts list. The failing artifact confirms that the contact was added successfully, after which Appium timed out waiting for `ContactViewName` on a screen the current app never opens.
- Android PR workflows without a matching E2E feature branch consume the floating shared E2E `main`. That made the future-stack assertion run against every unrelated `master`-based PR.
- All five completed Android PR E2E runs today failed this same shard with the same missing `ContactViewName` assertion. The aggregate `e2e-status` check is red only because it propagates that shard result.

This is a shared test-branch sequencing failure rather than a regression signal from this dependency bump. The correct follow-up is to keep the future Contact Detail assertion on the matching feature branch until #1084 reaches `master`, or make the shared helper assert Android's current post-save destination. That correction belongs in `bitkit-e2e-tests`; changing this PR would neither address the cause nor add relevant safety.

The exact reviewed dependency integration therefore retains its approval and merge readiness despite this non-attributable check failure.

cc: @piotr-iohk

PR #1131, opened 2026-08-04
fix: require private Paykit auth scope

ben-kaufman:
### Description

- require Paykit Server watch-only authorization requests to include both the public and private server capabilities
- compare capability sets independently of ordering
- default local E2E Homegate traffic to the Android emulator host alias (`10.0.2.2`)
- update to Paykit `0.1.0-rc42` and set `localTestnetHost` to `10.0.2.2` for local E2E builds
- use the same Pubky client configuration for session bootstrap and the Paykit SDK
- keep production network defaults unchanged

Paykit `0.1.0-rc42` is published for Android and iOS. The local test environment fixes are in https://github.com/pubky/locks/pull/11.

### Preview

N/A — no user-facing UI changes.

### QA Notes

- `./gradlew detekt` passed; reported only existing unrelated warnings.
- The rc42 GitHub Packages publication succeeded and the package version is visible through the GitHub Packages API.
- Local compile/test dependency resolution remains blocked by this machine's credentials for unrelated private LDK/VSS packages; authenticated CI provides the full build verification.
- Build an E2E variant without overriding `E2E_HOMEGATE_URL` and confirm both Homegate and Paykit resolve through `10.0.2.2`.


app/src/test/java/to/bitkit/models/PubkyAuthRequestTest.kt
```diff
@@ -20,6 +20,18 @@ class PubkyAuthRequestTest {
         assertEquals(PubkyAuthClaim.WATCH_ONLY_ACCOUNT_V1, request.bitkitClaim)
     }
 
+    @Test
+    fun `parse recognizes watch-only account claim with reordered capabilities`() {
+        val capabilities = PubkyAuthClaim.WATCH_ONLY_ACCOUNT_CAPABILITIES.split(",").reversed().joinToString(",")
```
jvsena42, line 25: The order-insensitivity this test proves isn't carried through to the approval path, so the scenario can't actually complete end-to-end.

`PubkyAuthApprovalViewModel.approveRequest` passes `request.capabilities` (the actual, as-received string) for plain auth, but the watch-only branch goes through `PubkyRepo.approveAuthWithCompanionClaim`, which forwards the fixed-order constant:

```kotlin
expectedCapabilities = PubkyAuthClaim.WATCH_ONLY_ACCOUNT_CAPABILITIES,
```

So a reordered request now parses fine and reaches `WatchOnlyConsent`, but then hands the canonical-order string to the SDK. Unless the SDK also compares order-insensitively, approval fails at the last step.

Suggest threading `request.capabilities` through `approveAuthWithCompanionClaim` the same way the plain path does — otherwise this test asserts a path the app can't finish. (`PubkyRepoTest > approveAuthWithCompanionClaim forwards exact claim identifiers and capability` currently pins the fixed-order behavior, so it'd need updating alongside.)

app/src/main/java/to/bitkit/models/PubkyAuthRequest.kt
```diff
@@ -15,7 +15,13 @@ enum class PubkyAuthClaim(val wireValue: String) {
         const val QUERY_PARAMETER = "x-bitkit-claim"
 
         /** Capabilities required by the watch-only Paykit Server setup flow. */
-        const val WATCH_ONLY_ACCOUNT_CAPABILITIES = "/pub/paykit/v0/bitkit/server/:rw"
+        const val WATCH_ONLY_ACCOUNT_CAPABILITIES =
+            "/pub/paykit/v0/bitkit/server/:rw,/pub/paykit/v0/private/bitkit/server/:rw"
```
jvsena42, line 19: Nit: the KDoc no longer conveys the point of the change — that **both** the public and private scopes are now mandatory. Worth a word here, and `matchesWatchOnlyAccountCapabilities` below has no doc at all despite now being the security-relevant comparison.

app/src/main/java/to/bitkit/models/PubkyAuthRequest.kt
```diff
@@ -15,7 +15,13 @@ enum class PubkyAuthClaim(val wireValue: String) {
...
+
+        private val watchOnlyAccountCapabilitySet = WATCH_ONLY_ACCOUNT_CAPABILITIES.split(",").toSet()
+
+        fun matchesWatchOnlyAccountCapabilities(capabilities: String) =
+            capabilities.split(",").toSet() == watchOnlyAccountCapabilitySet
```
jvsena42: This doesn't normalize the incoming string, so `"a, b"` (space after the comma) or a trailing comma silently fails to match and the request falls through to the generic auth path. Since this is now the security-relevant comparison, worth being defensive:

```kotlin
private fun normalize(capabilities: String) =
    capabilities.split(",").map { it.trim() }.filter { it.isNotBlank() }.toSet()
```

Also note set semantics silently accept duplicated entries (`"a,a,b"` matches `"a,b"`) — probably harmless, just flagging it's now lenient in a way exact string comparison wasn't.

app/build.gradle.kts
```diff
@@ -76,7 +76,7 @@ val bcp47Locales = listOf(
 )
 val e2eEnv = envFlag("E2E", default = false)
 val e2eBackendEnv = providers.environmentVariable("E2E_BACKEND").orElse("local")
-val e2eHomegateUrlEnv = providers.environmentVariable("E2E_HOMEGATE_URL").orElse("http://127.0.0.1:6288")
+val e2eHomegateUrlEnv = providers.environmentVariable("E2E_HOMEGATE_URL").orElse("http://10.0.2.2:6288")
```
jvsena42: Verified this is needed for CI — `bitkit-e2e-tests/ci_run_android.sh` reverse-forwards 60001 / 9735 / 30001 but not 6288, so `127.0.0.1:6288` really was unreachable there. And `10.0.2.2` as a default is consistent with `trezorBridgeUrlEnv` on line 83, so the direction isn't unprecedented.

But it breaks the physical-device workflow. `Justfile:118` still does:

```sh
adb -s "$device_id" reverse tcp:6288 tcp:6288     # local homegate
```

which is now dead code, and README states *"`just run` prefers a physical device and falls back to an emulator"* — so `just run docker` on a real device can no longer reach Homegate. Two clean ways out:

- **(a)** Add `adb reverse tcp:6288` to `ci_run_android.sh` and keep `127.0.0.1` here — works on emulator *and* device.
- **(b)** Keep `10.0.2.2`, delete the now-dead reverse-forward from the Justfile, and document `E2E_HOMEGATE_URL=http://127.0.0.1:6288` as the physical-device override.

Either is fine, but leaving the Justfile line in place is misleading.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -845,6 +850,18 @@ internal fun paykitSdkConfig() = defaultConfig(PaykitReceiverPaths.WALLET).copy(
     publicContactSharing = BitkitPaykitSdkConfig.publicContactSharing,
 )
 
+internal fun paykitPubkyClientConfig(
+    isLocalE2eBackend: Boolean = Env.isLocalE2eBackend,
+    baseConfig: PubkyClientConfig = defaultPubkyClientConfig(),
+) =
+    if (isLocalE2eBackend) {
+        baseConfig.copy(
+            localTestnetHost = "10.0.2.2",
```
jvsena42: Unlike `E2E_HOMEGATE_URL`, this host has no override path, so local-E2E Paykit can't be made to work on a physical device at all — not even manually. It also duplicates the magic host across `build.gradle.kts` and this file, and now diverges from `Env.ElectrumServers.REGTEST.LOCAL = "tcp://127.0.0.1:60001"` within the same local stack.

A single `E2E_LOCAL_HOST` BuildConfig field feeding Homegate, Electrum and this would remove the duplication and the inconsistency in one go, and would keep the constant reuse rule in CLAUDE.md satisfied.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -786,14 +788,17 @@ class PaykitSdkService @Inject constructor(
...
+        PaykitSdk.withPaymentAdapterAndPubkyClientConfig(
             stateStore = stateStore,
             sessionProvider = sessionProvider,
             paymentAdapter = paymentAdapter,
             config = paykitSdkConfig(),
+            pubkyClient = paykitPubkyClientConfig(),
         ).also { sdk = it }
     }
 
+    private fun bootstrap() = PubkySessionBootstrap.withPubkyClientConfig(paykitPubkyClientConfig())
```
jvsena42: Nice consolidation — all five bootstrap call sites now funnel through one config.

Minor: `paykitPubkyClientConfig()`'s default arg evaluates `defaultPubkyClientConfig()`, which is a native FFI call, so it fires on every `bootstrap()` (5 call sites) plus each `handle()` miss. Cheap, but a `private val pubkyClientConfig by lazy { paykitPubkyClientConfig() }` would avoid the repeated hop and read a bit clearer.

jvsena42 (commented):
Reviewed and validated locally on this branch.

**Validation**

| Check | Result |
|---|---|
| `just lint` (detekt) | BUILD SUCCESSFUL — 10 warnings, all in files untouched by this PR (`WalletRepo`, `BackupRepo`, `TrezorPreviewData`, …). **Zero new warnings introduced.** |
| `just test` | BUILD SUCCESSFUL, 0 failures |
| Targeted `--rerun-tasks` on `PubkyAuthRequestTest`, `PaykitSdkServiceTest`, `PubkyRepoTest`, `PubkyAuthApprovalViewModelTest` | all PASSED, including both new tests |
| `E2E=true GEO=false ./gradlew assembleDevDebug` | built successfully (dependency resolution works here, so the blocker noted in the description looks machine-local) |
| Install + launch on an emulator | no crash, no `UnsatisfiedLinkError` — the rc42 native lib loads and the new FFI entry points resolve |

Runtime confirmation the new config actually takes effect (logcat):

```
I/paykit: pubky::client::core: Configuring testnet builders for host 10.0.2.2
```

I also confirmed the motivation for the host change is real: `bitkit-e2e-tests/ci_run_android.sh` reverse-forwards 60001 / 9735 / 30001 but **not** 6288, so `127.0.0.1:6288` was genuinely unreachable in CI.

Inline comments below — one correctness issue worth fixing before merge (the reorder test), plus some local-E2E ergonomics and nits.

app/src/test/java/to/bitkit/models/PubkyAuthRequestTest.kt
ben-kaufman: No app change is needed for this path. Paykit validates the approval URL by parsing both the request capabilities and expectedCapabilities through parse_capabilities(), which trims entries and calls Capabilities.normalize() before comparing them. Capability order is therefore ignored end-to-end, including approveAuthWithCompanionClaim; the fixed-order constant expresses the required semantic set rather than imposing URL order. I kept the reorder test because it covers that supported behavior.

app/src/main/java/to/bitkit/models/PubkyAuthRequest.kt
ben-kaufman: Fixed in e79199f72. The matcher now trims each capability and explicitly rejects empty entries before comparing the semantic sets. A request carrying the watch-only claim does not fall through on mismatch; it returns InvalidBitkitClaimCapabilities. Duplicate entries remain normalized as the same semantic set, matching Paykit Capabilities.normalize behavior.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Fixed in e79199f72. E2E_LOCAL_HOST is now a BuildConfig field shared by Homegate, local Electrum, and Paykit, with 10.0.2.2 as the emulator default and an explicit LAN-host override for physical devices. This removes the duplicated and divergent host values.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Fixed in e79199f72. PaykitSdkService now caches paykitPubkyClientConfig with a lazy property and reuses the same config for SDK handle creation and every session bootstrap.

app/src/main/java/to/bitkit/models/PubkyAuthRequest.kt
ben-kaufman: Fixed in e79199f72. The constant KDoc now explicitly says both public and private Paykit Server capabilities are required, and the matcher documents its exact security-relevant comparison behavior.

app/build.gradle.kts
ben-kaufman: Fixed in e79199f72. A shared E2E_LOCAL_HOST now defaults to 10.0.2.2 for emulators and feeds Homegate, Electrum, and Paykit. Physical-device runs can set it to the development machine LAN address. The obsolete Homegate and Electrum adb reverse lines were removed and the workflow is documented in README and .env.example.

ovitrif (approved):
Approved.

jvsena42 (approved):

PR #1133, opened 2026-08-05
fix: recover hw transfer after re-pairing

### LINKED ISSUE #1130: Hardware wallet: a settled transfer displays as a plain send after remove + re-pair
### Summary

After removing and re-pairing a hardware wallet, a past transfer-to-spending shows up as an ordinary send instead of a transfer.

Found and fixed on iOS (synonymdev/bitkit-ios#648, reproduced and verified there). This report is **derived from reading the Android code**, not reproduced on a device — the same two mechanisms are present, so the same journey should reproduce it. Worth confirming with a Bridge emulator before acting on it.

### Steps to reproduce

1. Pair a hardware wallet.
2. Transfer to spending from it, and wait for the channel to become usable (this is what makes the transfer settle — see below).
3. Settings → remove the paired device.
4. Re-pair the same device.
5. Open the activity list: the funding transaction is back, but rendered as a plain send. Expected: a transfer row.

### Root cause

Five steps, each a deliberate behaviour on its own; together they lose the flag.

1. `TransferRepo.syncTransferStates` re-marks the funding tx via `persistResolvedChannel` → `markActivityAsTransfer` — `TransferRepo.kt:263`.
2. That path only ever sees **unsettled** transfers: `syncTransferStates` reads `transferDao.getActiveTransfers()` (`TransferRepo.kt:144`), whose query is `WHERE isSettled = 0` (`TransferDao.kt:33`).
3. Once the channel is ready the transfer is settled (`markSettled`, `TransferRepo.kt:159`), so it permanently leaves that set and step 1 never runs for it again.
4. Removing the wallet deletes its activities: `HwWalletRepo.removeDevice` → `activityRepo.deleteForWallet(it)` (`HwWalletRepo.kt:326`).
5. On re-pair the watcher rediscovers the tx from Electrum, and `mergeHwSnapshot` carries `isTransfer` forward **only from a stored row** — `isTransfer = onchain.v1.isTransfer || stored.isTransfer` (`CoreService.kt:261`). Step 4 deleted that row, so there is nothing to carry, and it lands as a plain send.

This predicts the bug is specific to *settled* transfers. A transfer whose channel had not yet become usable is still in the active set, so step 1 would re-mark it on re-pair.

### Fixable with no new retention

The obvious fix — keep the hardware wallet's activity rows after removal — would be the wrong trade: removal should mean removal.

It is not needed. The transfer fact is already retained on the Bitkit side, legitimately: the Lightning channel is still open and the LSP order still exists, neither of which the user asked to delete. `TransferEntity` holds `fundingTxId` and `channelId`, and `removeDevice` never touches `TransferRepo`/`transferDao` — only `activityRepo`, `trackedWalletIds` and `lastPersistedHwSnapshots`.

So the match runs in the safe direction — **rediscovered tx id → existing local transfer record**, never remembered-hardware-data → display:

- Nothing hardware-specific is retained past removal; remove-and-never-re-pair leaves no trace.
- The tx id is not remembered by the app — the device's own watcher re-supplies it on re-pair.
- These records would be retained identically for a user who never owned a hardware wallet.

### Suggested fix (mirrors the iOS one)

Recover the flag inside `mergeHwSnapshot` rather than as a follow-up re-marking pass, so it lands in the same transaction as the snapshot write and does not run a lookup on every watcher poll.

1. Give `mergeHwSnapshot` a `transferChannelIdsByFundingTxId: Map<String, String>` parameter, defaulted to empty.
2. Apply it **after** the existing stored-row merge and guarded on `!isTransfer`, so stored metadata always wins and recovery only fills gaps:

   ```kotlin
   if (!merged.isTransfer) {
       transferChannelIdsByFundingTxId[merged.txId]?.let { channelId ->
           merged = merged.copy(isTransfer = true, channelId = merged.channelId ?: channelId)
       }
   }
   ```

3. Build the map where `replaceHwSnapshot` is called, from transfers that have both `fundingTxId` and `channelId`. Unsettled ones can be skipped — `syncTransferStates` still handles those, and a settled transfer always has a channel id since settling requires a ready channel.

Two Android-specific notes:

- The current structure early-returns when no stored row matches, which is exactly the re-pair case. On iOS that branch had to be restructured so recovery sits outside it; check whether `mergeHwSnapshot` needs the same.
- `transferDao.getByFundingTxId` already exists (`TransferDao.kt:39`), so a per-tx query is an option instead of building a map — probably preferable given the snapshot is usually small.

Match on **transaction id, not activity id**: a re-paired wallet may rebuild its rows under different activity ids.

### iOS reference

synonymdev/bitkit-ios#648 — `HwSnapshotMerge.plan`, `ActivityService.replaceHwSnapshot`, and `HwWalletManager.transferChannelIdsByFundingTxId()`, with unit tests in `BitkitTests/HwSnapshotMergeTests.swift` covering recovery with no stored row, stored metadata winning, tx-id matching, and the empty-map no-op.

One coverage gap worth repeating here: the merge rule is unit tested, but the wiring from transfer storage into the snapshot call is not, because the hardware wallet manager tests inject their own persistence seam and bypass it. The end-to-end journey needs the manual test in the Steps above.

jvsena42:
Fixes #1130

This PR restores the transfer label on a hardware wallet's funding transaction after the device is removed and re-paired.

### Description

A transfer to spending from a paired hardware wallet settles once its channel becomes usable. Removing the device deletes that wallet's activities, and on re-pair the watcher rediscovers the funding transaction from scratch with no stored row left to carry the transfer metadata forward, so it came back looking like an ordinary send. Only settled transfers were affected: one whose channel had not opened yet is still being tracked and gets re-marked on its own.

Bitkit's own record of the transfer survives removal untouched and still holds the funding transaction id and the channel id, so the flag is now recovered from there while the rediscovered snapshot is merged.

The match runs in the safe direction only, from a rediscovered transaction id to an existing local transfer record. Nothing hardware-specific is retained past removal: the transaction id is re-supplied by the device's own watcher on re-pair, and the transfer record belongs to the Lightning channel that is still open, which the user never asked to delete. Removing a device and never re-pairing leaves no trace.

Recovery only ever fills gaps. Anything the app already knows wins, an existing channel id is never overwritten, and transactions with no matching transfer record are left alone.

This mirrors the iOS fix in synonymdev/bitkit-ios#648, with two deliberate differences. Its partial-snapshot handling is not ported, because a wallet here reports through a single watcher and callers already merge everything before persisting. And the new value is required rather than defaulted, since a default is what allowed the equivalent iOS tests to pass while silently skipping the recovery path entirely.

One consequence worth flagging for future work: settled transfers are now load-bearing for this recovery. The note suggesting they could be deleted once settled has been rewritten to say so, and a test pins the behaviour so acting on it fails loudly instead of quietly reintroducing this bug.

### Preview

Before remove transfer row

1_before_remove_transfer_row

Bug - After repair, transfer row replaced with send row

2_bug_after_repair_sent_row

Fixed - After repair, transfer flow displayed

3_fixed_after_repair_transfer_row

### QA Notes

Requires the Trezor Bridge emulator from `bitkit-docker` and a Bridge-enabled build, see `journeys/hardware-wallet/README.md`.

#### Manual Tests

- [x] **1.** Pair a hardware wallet → Hardware Wallet → Transfer To Spending → sign, then wait for the channel to become usable: funding row reads Transfer with subtitle From Savings.
- [x] **2.** Settings → Hardware Wallets → remove the device → re-pair the same device → Hardware Wallet: the same funding row still reads Transfer, not Sent.
- [x] **3.** Tap the recovered row: Activity Detail shows FROM SAVINGS.
- [x] **4** `regression:` Transfer to spending while the channel is still opening → remove and re-pair before it settles: row still reads Transfer.

#### Automated Checks

- Unit tests added: cover the recovery rule in `CoreServiceTest.kt` — recovery when no stored row remains, stored metadata winning over a recovered value, matching on transaction id rather than activity id since re-paired rows are rebuilt under new ids, unrelated transactions left unmarked, and Lightning activities passing straight through.
- Unit tests added: cover the wiring in `ActivityRepoTest.kt` — known channel ids actually reach the merge, and a failed lookup still persists the snapshot instead of dropping a wallet's activity list. The equivalent iOS change has no coverage here.
- Unit tests added: cover the lookup in `TransferRepoTest.kt` — settled transfers are included, records missing a funding transaction or channel are skipped, duplicates keep the first entry, and a database failure surfaces.
- Regression pin added: `TransferRepoTest.kt` asserts that settling a transfer never deletes its record, so a future cleanup of settled transfers cannot silently reintroduce this bug.
- Verified on device: reproduced the plain send on `master` against a Trezor Bridge emulator, then confirmed the row returns as a transfer after a fresh remove and re-pair on this branch.


ovitrif (approved):
Approved.

PR #1121, opened 2026-07-31
chore: update kotlin and compose bom

guzino:
Closes #1115

This PR moves Kotlin to `2.3.21` and Compose to the `2026.06.01` BOM on the current AGP `8.13.2` baseline.

### Description

AGP `8.13.2` is the supported baseline for Kotlin 2.3, so the language and Compose set land here first and #1114 consumes them before touching AGP, Gradle, KSP and Hilt. `2.3.21` is the floor that satisfies the #1114 matrix: KSP `2.3.10` is published against Kotlin `2.3.20`, and AGP 9 built-in Kotlin cannot compile `2.2.21` at all, measured on #1114.

The stability analyzer plugin has to move with Kotlin, and its version is constrained from both sides. `0.6.6` throws `NoSuchMethodError` on a removed IR declaration origin under Kotlin 2.3. `0.8.0` and `0.12.0` pull AGP `9.3.1` onto the plugin classpath, which fails the AGP version check on Gradle `8.13` and would silently override a pinned AGP version later. `0.7.5` targets Kotlin `2.3.21` and leaves AGP alone.

Selected set, all stable releases:

| Component | From | To |
|---|---|---|
| Kotlin, Compose compiler and serialization plugins | `2.2.21` | `2.3.21` |
| Kotlin BOM | `2.2.21` | `2.3.21` |
| Compose BOM | `2025.12.01` | `2026.06.01` |
| Compose stability analyzer | `0.6.6` | `0.7.5` |
| AGP | `8.13.2` | unchanged |
| KSP | `2.3.2` | unchanged |

- Updates the shared Kotlin catalog version, which drives the Compose compiler plugin, the serialization plugin and the Kotlin BOM through `version.ref`.
- Updates the Compose BOM to the latest stable release set.
- Removes the empty Compose feature-flag set, which restated the plugin's own convention. Compose reports still generate without it.
- Leaves KSP and Hilt untouched. Both resolve and compile against Kotlin `2.3.21`, and #1114 owns moving them.
- Requires no source, compiler-option or stability-annotation changes. The upgrade introduces no deprecated compiler flags.
- Leaves the two follow-up issues the spec requires, `docs: align kotlin and compose agent rules` and `refactor: adopt current kotlin language features`, still to be filed and linked before #1115 closes.

### Preview

N/A

#### Manual Tests

- [ ] **1.** Android Studio → open project: Gradle sync succeeds and Compose previews render.
[...]

PR #1134, opened 2026-08-05
chore: prune fcm-tester deps to fix alerts

jvsena42:
This PR:
1. Replaces `googleapis` with `google-auth-library` in the FCM tester tool, clearing every open Dependabot alert
2. Adds a `dependabot.yml` for the three ecosystems this repo has — the `tools/fcm-tester` npm tree, workflow actions, and the Gradle catalog

### Description

All three open alerts, and both dismissed ones, came from `tools/fcm-tester/package-lock.json`. Nothing vulnerable reaches the app: the scope is development and there are no Gradle findings.

They were collateral from an over-broad dependency. The tool declared `googleapis`, a meta-SDK for every Google API, but uses one thing from it — `google.auth.JWT` — which lives in `google-auth-library` and is only re-exported. `qs` and one `uuid` came solely through `googleapis-common`; the second `uuid` came from the older `gaxios` that `google-auth-library` 9 depended on. Depending on the auth library directly resolves all of them with no pins, `overrides` or dismissals, and the regenerated lockfile also picks up the patched `jws`, clearing the dismissed high-severity alert. The tree goes from 42 packages to 23.

`google-auth-library` 11 requires Node 22, so `.nvmrc` moves off `v20.15.1`, end of life since April 2026. That forces a fix already overdue: the entry point imported the service account JSON with `assert { type: "json" }`, syntax removed in Node 22, so the tool would not start on any current runtime. The `JWT` constructor also became options-only, so the positional call was rewritten; `authorize` still accepts a callback, so the promise wrapper is unchanged.

#### Dependabot config

The repo has had alerts enabled for a long time but never a Dependabot pull request, because there was no config file. Hand-fixing advisories, as the first half of this PR does, is the work this is meant to stop from recurring.

Grouping is cost control: `ci.yml` and `lint.yml` run on every PR with no path filter, and `e2e.yml` fires on `gradle/**`, `*.gradle.kts` or its own workflow file. Ungrouped, six updates in a month is six full CI runs. npm and Gradle group minor and patch only, so majors fall through to their own PR. Actions need a second group because every one is pinned to a floating major tag, so a minor/patch group would never fire and each action would arrive separately.

Two sets of Gradle coordinates are ignored. The `com.synonym:*` artifacts resolve from credentialed GitHub Packages registries Dependabot cannot read, and move in lockstep with native code. The Kotlin toolchain — Kotlin, AGP, KSP, Compose compiler, stability analyzer, Compose BOM — is a compatibility matrix that took #1121 and #1127 to move deliberately; Dependabot resolves each coordinate independently with no view of it, so grouping would not help and only ignoring keeps it out. Everything else in the catalog updates normally. The rules use `org.jetbrains.kotlin:*` rather than a looser prefix so `org.jetbrains.kotlinx` packages keep updating.

#### Admin steps to finish enabling this

A config file only turns on *version* updates. Turning an advisory into a pull request is a separate repo setting. **Merging this alone will not make alerts arrive as pull requests.** Either tick Settings → Advanced Security → **Dependabot security updates**, or run the equivalent from the CLI. Both need admin and return `204 No Content`:

```bash
# already on here, but idempotent and safe to re-run
gh api --method PUT repos/synonymdev/bitkit-android/vulnerability-alerts

# this is the one that turns an alert into a pull request
gh api --method PUT repos/synonymdev/bitkit-android/automated-security-fixes
```

Verify afterwards:

```bash
# expect {"enabled": true, "paused": false}
gh api repos/synonymdev/bitkit-android/automated-security-fixes

# expect HTTP/2.0 204 — a 404 means either disabled or the caller is not an admin
gh api repos/synonymdev/bitkit-android/vulnerability-alerts --include | head -1
```

Both return 404 for this PR's author, who has `maintain` and not `admin`, so the current state cannot be read from here and someone with admin has to run them. Alerts themselves are on — this PR exists because of three.

### Preview

N/A — developer tooling and repository configuration only.

#### Manual Tests

> Dependabot reads `dependabot.yml` from the default branch only, so tests 3 onward cannot run until this merges. Each row in the Dependabot tab has a **Check for updates** button that forces a run.

- [ ] **1.** With `tools/fcm-tester/service-account.json` present → `nvm use && npm i` → `npm start`: prints the access token banner followed by a `ya29.…` token.
- [ ] **2.** `regression:` Paste that token into the Postman collection as `bearerToken` → send `messages/notification.json` to a device token: notification arrives and logcat shows the `FCM` tag entry.
- [ ] **3.** Insights → Dependency graph → Dependabot: three rows listed (npm, GitHub Actions, Gradle), each with a Last checked timestamp and no config error banner.
- [ ] **4a.** Dependabot tab → Gradle row → Check for updates → job log: androidx, Ktor, Room, Hilt and detekt resolve from `gradle/libs.versions.toml`.
  - [ ] **4b.** Same log: `com.synonym:*` and the Kotlin/AGP/KSP/Compose matrix are skipped as ignored, and `org.jetbrains.kotlinx` packages stay eligible.
- [ ] **5.** Dependabot tab → npm row → Check for updates: resolves `tools/fcm-tester/package-lock.json` and does not look at the repo root.
- [ ] **6.** Dependabot tab → GitHub Actions row → Check for updates: one grouped `github-actions-major` PR rather than one per action; `github-actions-minor` produces nothing.
- [ ] **7.** Repo admin → run the two `gh api --method PUT` commands above → `automated-security-fixes` returns `enabled: true`, `paused: false`.
- [ ] **8.** First grouped PR Dependabot opens → CI and Lint: both green. Dependabot PRs run with a read-only token and no secrets, so a failure on secret ownership rather than on the bump needs a Dependabot secret or an author guard in a follow-up.

#### Automated Checks

- No Kotlin or Gradle build code changed, so no test coverage was added, modified or removed.
- `npm audit` reports 0 vulnerabilities on the regenerated lockfile, down from three open advisories, and `npm ls qs uuid googleapis googleapis-common` returns empty for all four.
- `node index.js token` now fails only on the gitignored `service-account.json`, exercising the changed import, the JSON import attribute and the `JWT` construction.
- `npx -y js-yaml .github/dependabot.yml` parses cleanly with the intended ecosystems, groups and ignore rules. Schema validity is otherwise enforced by GitHub, which annotates an invalid config on the PR.
- Grouping and resolution behaviour can only be observed once the config is on the default branch (after merge); the manual steps cover that.
- CI: standard checks run by the PR bot.

### Linked Issues/Tasks

- Sibling precedent: synonymdev/bitkit-ios#655, the equivalent config for the iOS repo
- Toolchain matrix context for the Gradle ignore rules: #1121 and #1127


jvsena42:
@ovitrif I don't even remember using this tool, do you think we should keep it?

ovitrif (approved):
Approved: the fcm-tester google-auth-library migration and Dependabot config are consistent and ready to merge.

PR #1140, opened 2026-08-06
feat(send): improve Lightning send failure recovery

### LINKED ISSUE #829: Improve Send Error Sheet UX
> I fixed the toast message in this PR but the error sheet message is also misleading, I will remove in next PR, then on this error screen we'll have parity with iOS:

That screen needs work on both platforms. Design has some improvements to status quo but needs to be adjusted somewhat. For LN payments "Transaction" is not accurate, should be "Payment". And we should try to show a user actionable error description as well as we can.

image

_Originally posted by @pwltr in https://github.com/synonymdev/bitkit-android/issues/825#issuecomment-4014633605_

pwltr:
### Description

Ports the iOS Lightning send-failure retry/support behavior to Android.

- Shows the send failure screen for failed Lightning sends instead of relying only on toasts.
- Maps LDK payment failure reasons to localized, user-facing strings and avoids exposing internal error values like `Optional(...)`, `NodeError`, or `DuplicatePayment`.
- Uses the reusable generic route failure copy for `routeNotFound` and `retriesExhausted`.
- Preserves LDK failure reasons for payments that first enter `SendPendingScreen` and fail later.
- Shows the correct failure title for Lightning vs on-chain send errors.
- Shows Contact Support above Try Again on the Lightning send failure screen.
- Prefills the support report with failure type, payment method, routing cache reset status, and the available Lightning payment request.
- Resets routing caches only for routing-related failures, and only once per current send flow/payment attempt.
- Retries non-routing failures normally without resetting graph/scorer caches.
- Retry clears local/VSS network graph data and scorer/pathfinding score data before restarting ldk-node when a routing reset is needed.
- Retry waits for local routing data to be usable again without requiring a newer remote RGS snapshot timestamp or waiting only on delayed graph-cache persistence.
- Guards retry state in `WalletViewModel` so overlapping retry flows cannot surface stale timeout errors.
- Resets send navigation back to the retry route instead of stacking retry screens on top of the failure screen.

Closes https://github.com/synonymdev/bitkit-android/issues/829

### Preview

Normal payment:

https://github.com/user-attachments/assets/7980c47a-eb25-4557-85be-cabb5dd33c54

QuickPay:

https://github.com/user-attachments/assets/2ccb4ee8-6af1-4fef-b147-891f7d36c26d

### QA Notes

Tested on mainnet emulator:
- Confirmed `ROUTE_NOT_FOUND` surfaces user-facing payment failure copy.
- Confirmed pending `RETRIES_EXHAUSTED` failures surface the localized retries-exhausted copy instead of generic failure copy.
- Confirmed Try Again deletes local graph, VSS graph, scorer, and external scores cache for the first routing-related retry.
- Confirmed Try Again does not repeat routing cache reset after a reset has already been attempted in the current send flow.
- Confirmed retry restarts from an empty graph and accepts the same RGS snapshot timestamp.
- Confirmed retry waits through RGS/scorer refresh and returns to the send confirmation path without timing out when graph cache persistence lags behind the in-memory graph.
- Confirmed no stale retry timeout after repeated retry attempts.

Automated checks:
- `compileDevDebugKotlin`
- `PaymentFailureReasonExtTest.kt`
- `AppViewModelSendFlowTest.kt`
- `detekt`

### DIFF
--- a/app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
+++ b/app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
@@ -3,15 +3,125 @@ package to.bitkit.ext
 import android.content.Context
 import org.lightningdevkit.ldknode.PaymentFailureReason
 import to.bitkit.R
+import to.bitkit.models.SendFailureDetails
+import to.bitkit.utils.LdkError
 
 fun PaymentFailureReason?.toUserMessage(context: Context): String = when (this) {
     PaymentFailureReason.RECIPIENT_REJECTED ->
-        context.getString(R.string.wallet__toast_payment_failed_recipient_rejected)
+        context.getString(R.string.wallet__payment_recipient_rejected)
+    PaymentFailureReason.USER_ABANDONED ->
+        context.getString(R.string.wallet__payment_abandoned)
     PaymentFailureReason.RETRIES_EXHAUSTED ->
-        context.getString(R.string.wallet__toast_payment_failed_retries_exhausted)
+        context.getString(R.string.wallet__payment_retries_exhausted)
     PaymentFailureReason.ROUTE_NOT_FOUND ->
-        context.getString(R.string.wallet__toast_payment_failed_route_not_found)
+        context.getString(R.string.wallet__payment_route_not_found)
     PaymentFailureReason.PAYMENT_EXPIRED ->
-        context.getString(R.string.wallet__toast_payment_failed_timeout)
-    else -> context.getString(R.string.wallet__toast_payment_failed_description)
+        context.getString(R.string.wallet__payment_expired)
+    PaymentFailureReason.UNKNOWN_REQUIRED_FEATURES ->
+        context.getString(R.string.wallet__payment_unknown_required_features)
+    PaymentFailureReason.INVOICE_REQUEST_EXPIRED ->
+        context.getString(R.string.wallet__payment_invoice_request_expired)
+    PaymentFailureReason.INVOICE_REQUEST_REJECTED ->
+        context.getString(R.string.wallet__payment_invoice_request_rejected)
+    else -> context.getString(R.string.wallet__payment_failed_description)
 }
+
+fun PaymentFailureReason?.shouldResetRoutingCachesOnRetry(): Boolean =
+    this == PaymentFailureReason.ROUTE_NOT_FOUND || this == PaymentFailureReason.RETRIES_EXHAUSTED
+
+fun PaymentFailureReason?.toCompactFailureType(): String {
+    return this?.name?.snakeToLowerCamel() ?: UNKNOWN_FAILURE_TYPE
+}
+
+fun PaymentFailureReason?.toSendFailureDetails(
+    context: Context,
+    paymentRequest: String? = null,
+): SendFailureDetails {
+    return SendFailureDetails(
+        message = toUserMessage(context),
+        failureType = toCompactFailureType(),
+        resetRoutingCachesOnRetry = shouldResetRoutingCachesOnRetry(),
+        paymentRequest = paymentRequest,
+    )
+}
+
+fun Throwable.toSendFailureMessage(context: Context): String {
+    val fallbackMessage = context.getString(R.string.wallet__payment_failed_description)
+    val rawMessage = message?.trim().orEmpty()
+
+    if (this is LdkError || rawMessage.isBlank() || rawMessage.looksInternalPaymentError()) {
+        return fallbackMessage
+    }
+
+    return rawMessage
+}
+
+fun Throwable.toCompactFailureType(): String {
+    if (this is LdkError) return compactType ?: UNKNOWN_FAILURE_TYPE
+
+    val rawValue = message?.trim()?.takeIf { it.isNotEmpty() }
+        ?: this::class.simpleName
+        ?: UNKNOWN_FAILURE_TYPE
+
+    return rawValue.compactFailureType()
+}
+
+fun Throwable.toSendFailureDetails(
+    context: Context,
+    paymentRequest: String? = null,
+): SendFailureDetails {
+    return SendFailureDetails(
+        message = toSendFailureMessage(context),
+        failureType = toCompactFailureType(),
+        resetRoutingCachesOnRetry = false,
+        paymentRequest = paymentRequest,
+    )
+}
+
+private fun String.snakeToLowerCamel(): String {
+    return lowercase()
+        .split("_")
+        .filter { it.isNotBlank() }
+        .mapIndexed { index, segment ->
+            if (index == 0) segment else segment.replaceFirstChar { it.titlecase() }
+        }
+        .joinToString("")
+        .ifBlank { UNKNOWN_FAILURE_TYPE }
+}
+
+private fun String.compactFailureType(): String {
+    val strippedPrefixes = this
+        .removePrefix("LDK Node error:")
+        .removePrefix("LDK Build error:")
+        .trim()
+    val type = strippedPrefixes
+        .substringBefore("(")
+        .trim()
+        .trimEnd('.')
+        .trim()
+
+    return type.toCompactTypeName()
+}
+
+private fun String.toCompactTypeName(): String {
+    if (isBlank()) return UNKNOWN_FAILURE_TYPE
+    if (none { it.isWhitespace() || it == '-' || it == '_' }) return this
+
+    return split(Regex("[\\s_-]+"))
+        .filter { it.isNotBlank() }
+        .joinToString("") { it.replaceFirstChar(Char::titlecase) }
+        .ifBlank { UNKNOWN_FAILURE_TYPE }
+}
+
+private fun String.looksInternalPaymentError(): Boolean {
+    return INTERNAL_PAYMENT_ERROR_MARKERS.any { contains(it, ignoreCase = true) }
+}
+
+private val INTERNAL_PAYMENT_ERROR_MARKERS = listOf(
+    "DuplicatePayment",
+    "PaymentFailureReason",
+    "ldknode",
+    "LDK",
+)
+
+private const val UNKNOWN_FAILURE_TYPE = "Unknown"
--- a/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/LightningRepo.kt
@@ -48,6 +48,7 @@ import org.lightningdevkit.ldknode.ChannelDetails
 import org.lightningdevkit.ldknode.ClosureReason
 import org.lightningdevkit.ldknode.CoinSelectionAlgorithm
 import org.lightningdevkit.ldknode.Event
+import org.lightningdevkit.ldknode.Network
 import org.lightningdevkit.ldknode.NodeStatus
 import org.lightningdevkit.ldknode.PaymentDetails
 import org.lightningdevkit.ldknode.PaymentHash
@@ -92,6 +93,7 @@ import to.bitkit.services.LnurlChannelResponse
 import to.bitkit.services.LnurlService
 import to.bitkit.services.LnurlWithdrawResponse
 import to.bitkit.services.LspNotificationsService
+import to.bitkit.services.NetworkGraphInfo
 import to.bitkit.services.NodeEventHandler
 import to.bitkit.utils.AppError
 import to.bitkit.utils.Logger
@@ -648,8 +650,14 @@ class LightningRepo @Inject constructor(
     }
 
     private suspend fun clearNetworkGraph(walletIndex: Int): Result<Unit> {
-        lightningService.resetNetworkGraph(walletIndex)
-        return runCatching {
+        runSuspendCatching {
+            lightningService.resetNetworkGraph(walletIndex)
+        }.onFailure {
+            Logger.warn("Failed to clear local network graph", it, context = TAG)
+            return Result.failure(it)
+        }
+
+        return runSuspendCatching {
             vssBackupClientLdk.setup(walletIndex).getOrThrow()
             vssBackupClientLdk.deleteObject("network_graph").getOrThrow()
             Logger.info("Cleared network graph from VSS", context = TAG)
@@ -1870,7 +1878,7 @@ class LightningRepo @Inject constructor(
             vssBackupClientLdk.deleteObject(VSS_KEY_EXTERNAL_SCORES_CACHE).getOrThrow()
         }.onFailure {
             Logger.error("Failed to delete pathfinding scores from VSS", it, context = TAG)
-            start(walletIndex = walletIndex, shouldRetry = false).onFailure { startError ->
+            start(walletIndex = walletIndex, shouldRetry = false, shouldValidateGraph = false).onFailure { startError ->
                 Logger.error("Failed to restart node after pathfinding scores reset failure", startError, context = TAG)
             }
             return@withContext Result.failure(it)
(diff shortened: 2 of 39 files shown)


app/src/main/java/to/bitkit/ui/screens/wallets/send/SendQuickPayScreen.kt
```diff
@@ -53,17 +53,17 @@ fun SendQuickPayScreen(
...
-            is QuickPayResult.Pending -> onPaymentPending(result.paymentHash, result.amount)
-            is QuickPayResult.Error -> onShowError(result.message)
+            is QuickPayResult.Success -> {
+                app.resetQuickPay()
+                onPaymentComplete(result.paymentHash, result.amountWithFee)
+            }
+            is QuickPayResult.Pending -> {
+                app.resetQuickPay()
+                onPaymentPending(result.paymentHash, result.amount)
+            }
```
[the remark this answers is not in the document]
pwltr: Fixed [48391ce](https://github.com/synonymdev/bitkit-android/pull/1140/commits/48391cebe4b92f7984c3134aba9922114b1df962)

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
```diff
@@ -363,16 +402,44 @@ fun SendSheet(
...
+                            scope.launch {
+                                val shouldResetRoutingCaches = route.shouldResetRoutingCaches(
+                                    routingCacheResetAttempted = routingCacheResetAttempted
+                                )
+                                if (shouldResetRoutingCaches) routingCacheResetAttempted = true
+                                val resetResult = if (shouldResetRoutingCaches) {
+                                    walletViewModel.resetPaymentRoutingCachesAndWait()
+                                } else {
+                                    Result.success(Unit)
+                                }
```
jvsena42, line 425: dismissing the error screen mid-reset may cancel the coroutine between node stop() and start(), leaving the node stopped

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
```diff
@@ -461,10 +528,21 @@ sealed interface SendRoute {
...
 
     @Serializable
-    data class Error(val message: String? = null) : InternalOnly
+    data class Error(
+        val message: String? = null,
+        val retryRoute: SendRetryRoute = SendRetryRoute.Confirm,
+        val resetRoutingCachesOnRetry: Boolean = false,
+        val failureType: String = "Unknown",
+        val paymentRequest: String? = null,
+        val routingCacheResetAttempted: Boolean = false,
```
jvsena42: property never used

app/src/main/java/to/bitkit/domain/commands/NotifyPendingPaymentResolved.kt
```diff
@@ -9,12 +10,15 @@ sealed interface NotifyPendingPaymentResolved {
         val paymentHash: String
 
         data class Success(override val paymentHash: String) : Command
-        data class Failure(override val paymentHash: String) : Command
+        data class Failure(
+            override val paymentHash: String,
+            val reason: PaymentFailureReason? = null,
```
jvsena42: never called

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -1859,20 +1867,111 @@ class LightningRepo @Inject constructor(
...
+            !settingsStore.data.first().rgsServerUrl.isNullOrEmpty()
+        val requiresScorerRefresh = Env.ldkScorerUrl != null
+        val resetErrors = mutableListOf<Throwable>()
+
+        Logger.info(
+            "Started payment routing refresh rgs='$requiresRgsRefresh' scorer='$requiresScorerRefresh'",
+            context = TAG,
+        )
+
+        stop().onFailure {
```
jvsena42: the node stops here and I only see it starting again on the [VSS-delete failure path](https://github.com/synonymdev/bitkit-android/pull/1140/changes#diff-0185db3b5aed8e71acd13a790149b5b814a34d7b4c4314864087f8fecce5b412R1870) worth double checking if the is some path that Kees the node on stopped state

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2486,6 +2500,14 @@ class AppViewModel @Inject constructor(
                 }
             }
 
+            _sendUiState.update {
+                it.copy(
+                    amount = amountSats,
+                    decodedInvoice = invoice,
```
jvsena42: This will make it matches in `closeActiveSendForFailedPayment` and race with the similar logic of `watchUntil` in `QuickPayVIewModel`

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
jvsena42, line 2804: The `SendMethod.ONCHAIN` still do Toast + `hideSheet()`

app/src/main/java/to/bitkit/ui/settings/support/ReportIssueScreen.kt
```diff
@@ -32,9 +32,14 @@ import to.bitkit.ui.theme.Colors
 @Composable
 fun ReportIssueScreen(
     viewModel: ReportIssueViewModel = hiltViewModel(),
+    prefillMessage: String? = null,
     onBack: () -> Unit,
     navigateResultScreen: (Boolean) -> Unit,
 ) {
+    LaunchedEffect(prefillMessage) {
```
jvsena42, line 39: This re-runs on every composition recreation while the ViewModel survives

jvsena42 (dismissed):

ovitrif (dismissed):
Three defects remain because the earlier review findings are unresolved:
- Dismissing the send sheet during Try Again fails to restart Lightning because stop is NonCancellable and the reset job is cancelled.
- Writing QuickPay invoice state into sendUiState routes a failed QuickPay retry to Confirm.
- Report-issue prefill overwrites edits because it is reapplied on composition restart.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
pwltr: Seems more fitting for a follow-up as this PR is mainly scoped to Lightning send-failure recovery/retry

pwltr:
Addressed all comments, waiting for CI to pass.

app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
```diff
@@ -3,15 +3,111 @@ package to.bitkit.ext
...
+
+private fun String.compactFailureType(): String {
+    val unwrappedOptional = removeSurrounding("Optional(", ")")
+    val unwrappedNodeError = unwrappedOptional.removeSurrounding("NodeError(", ")")
+    return unwrappedNodeError
+        .substringBefore("(")
+        .substringAfterLast(".")
+        .trim()
+        .ifBlank { UNKNOWN_FAILURE_TYPE }
+}
```
jvsena42, line 114: Kotlin errors here comes in a different shape, like `"LDK Node error: Duplicate payment."`
using `.substringAfterLast(".")` makes the last part always 
`UNKNOWN_FAILURE_TYPE`.

app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
```diff
@@ -3,15 +3,111 @@ package to.bitkit.ext
...
+
+private fun String.looksInternalPaymentError(): Boolean {
+    return INTERNAL_PAYMENT_ERROR_MARKERS.any { contains(it, ignoreCase = true) }
+}
+
+private val INTERNAL_PAYMENT_ERROR_MARKERS = listOf(
+    "Optional(",
```
jvsena42: This is also iOS exclusive error shape

app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
```diff
@@ -3,15 +3,111 @@ package to.bitkit.ext
...
+
+    if (this is LdkError || rawMessage.isBlank() || rawMessage.looksInternalPaymentError()) {
+        return fallbackMessage
+    }
+
+    return rawMessage
+}
+
+fun Throwable.toCompactFailureType(): String {
+    val rawValue = message?.trim()?.takeIf { it.isNotEmpty() }
```
jvsena42, line 62: Android can derive the type from exception class `NodeException.DuplicatePayment -> DuplicatePayment`, could use it instead of depending on the message

jvsena42 (dismissed):

jvsena42 (commented):
Test-coverage note on `compactFailureType()`: the assertions only exercise iOS-shaped strings, so the shape Android actually produces is untested — and it currently resolves to `Unknown`. Details inline.

app/src/main/java/to/bitkit/ext/PaymentFailureReasonExt.kt
```diff
@@ -3,15 +3,111 @@ package to.bitkit.ext
...
+        .joinToString("")
+        .ifBlank { UNKNOWN_FAILURE_TYPE }
+}
+
+private fun String.compactFailureType(): String {
+    val unwrappedOptional = removeSurrounding("Optional(", ")")
+    val unwrappedNodeError = unwrappedOptional.removeSurrounding("NodeError(", ")")
+    return unwrappedNodeError
+        .substringBefore("(")
+        .substringAfterLast(".")
```
jvsena42: `substringAfterLast(".")` assumes the iOS input shape. This function is a port of `compactFailureType` in bitkit-ios (`Bitkit/Views/Wallets/Send/SendFailure.swift:41`), where the input is `String(describing: someType)` and the dot split strips a Swift module/type qualifier (`LDKNode.NodeError` -> `NodeError`), the same way the `Optional(` and `(` steps handle Swift's optional and associated-value rendering.

On Android the input is `Throwable.message`, which is a human-readable sentence: `Errors.kt:120` builds `"LDK Node error: $it"` over strings like `"Duplicate payment."`. So the dot split lands on sentence punctuation rather than a qualifier:

- `"LDK Node error: Duplicate payment."` -> `substringAfterLast(".")` == `""` -> `ifBlank { UNKNOWN_FAILURE_TYPE }` -> `"Unknown"`
- `"LDK Node error: Invalid custom TLVs"` (no trailing period) -> no `.` at all -> the whole `"LDK Node error: ..."` prefix leaks through as the failure type

Net effect: `Failure type: Unknown` in the prefilled support ticket for essentially every synchronous LN send failure (insufficient funds, duplicate payment, invalid invoice...). The `this::class.simpleName` fallback at `:61` is already unqualified, so the dot-stripping is vestigial on that path too.

Minimum regression test that pins the trigger without prescribing an implementation:

```kotlin
assertNotEquals("Unknown", Exception("Payment sending failed.").toCompactFailureType())
```

See the test-file comment for the fuller block. Getting a real type name out of an `LdkError` means reading the inner `NodeException` class rather than parsing its message — note `LdkError.inner` is currently `private` (`Errors.kt:31`), so that would need widening.

app/src/test/java/to/bitkit/ext/PaymentFailureReasonExtTest.kt
```diff
@@ -0,0 +1,63 @@
...
+        whenever(context.getString(R.string.wallet__payment_failed_description)).thenReturn(message)
+
+        assertEquals(message, Exception("  ").toSendFailureMessage(context))
+        assertEquals(message, Exception("Optional(NodeError(DuplicatePayment))").toSendFailureMessage(context))
+    }
+
+    @Test
+    fun `compact failure types omit optional and node error wrappers`() {
+        assertEquals("routeNotFound", PaymentFailureReason.ROUTE_NOT_FOUND.toCompactFailureType())
+        assertEquals("DuplicatePayment", Exception("Optional(NodeError(DuplicatePayment))").toCompactFailureType())
```
jvsena42: Both inputs here are iOS-shaped, so this passes for the wrong reason. `"Optional(NodeError(DuplicatePayment))"` is Swift `String(describing:)` output and happens to carry no trailing period — exactly the case `substringAfterLast(".")` handles correctly. The Android shape (`"LDK Node error: <sentence>."` from `Errors.kt:120`) is never asserted, and it returns `"Unknown"` today.

Suggested additions:

```kotlin
@Test
fun `compact failure types survive android ldk error messages`() {
    // Errors.kt formats NodeException as "LDK Node error: <sentence>."
    val withPeriod = LdkError(NodeException.DuplicatePayment("Duplicate payment.")).toCompactFailureType()
    val withoutPeriod = LdkError(NodeException.InvalidCustomTlvs("Invalid custom TLVs")).toCompactFailureType()

    assertNotEquals("Unknown", withPeriod)
    assertFalse(withoutPeriod.contains("LDK Node error"))
}

@Test
fun `compact failure types fall back to the exception class name`() {
    assertEquals("IllegalStateException", IllegalStateException().toCompactFailureType())
}

@Test
fun `compact failure types ignore sentence punctuation`() {
    assertNotEquals("Unknown", Exception("Payment sending failed.").toCompactFailureType())
}
```

The first and third fail on the current implementation; the second pins the `this::class.simpleName` path at `PaymentFailureReasonExt.kt:61`, which is what the ticket falls back to for app-level errors (note it only kicks in when `message` is null — `PaymentRoutingRefreshTimeoutError` etc. carry a message and so go down the parsing path).

They are deliberately written as `assertNotEquals`/`assertFalse` rather than exact strings so they do not lock in a specific naming scheme. If the fix ends up unwrapping `LdkError.inner`, these can tighten to `assertEquals("DuplicatePayment", ...)` / `assertEquals("InvalidCustomTlvs", ...)`.

Needs `kotlin.test.assertNotEquals` plus `to.bitkit.utils.LdkError` and `org.lightningdevkit.ldknode.NodeException` imports. `UNKNOWN_FAILURE_TYPE` is file-private in the main source, hence the `"Unknown"` literal.

pwltr:
I believe all comments have been addressed.

ovitrif (commented):
Pending QuickPay failures still drop the invoice from the Contact Support prefill after a timed-out payment later fails.

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
```diff
@@ -345,8 +373,16 @@ fun SendSheet(
...
                             )
                         },
-                        onPaymentError = {
-                            navController.navigateTo(SendRoute.Error()) {
+                        onPaymentError = { failure ->
+                            navController.navigateTo(
+                                SendRoute.errorFromFailure(
+                                    failure = failure.reason.toSendFailureDetails(
+                                        context = context,
+                                        paymentRequest = sendUiState.failurePaymentRequest(),
```
ovitrif: Pending QuickPay failures take the support payment request from `sendUiState.failurePaymentRequest()`, which returns null because QuickPay keeps `payMethod` as `ONCHAIN` and `decodedInvoice` null, so Contact Support prefill shows `Payment request: Unavailable` after a timed-out QuickPay later fails. Could we persist the `bolt11` or LNURL on `SendRoute.Pending`, or read it from `quickPayData`, when building that error route?

pwltr: Addressed and squashed [47aa9f4](https://github.com/synonymdev/bitkit-android/pull/1140/commits/47aa9f40199eb429847269151db8e50a2b681f73)

ovitrif (approved):
Approved. The pending QuickPay Contact Support prefill is fixed and now keeps the invoice after a timed-out payment later fails.

PR #1141, opened 2026-08-06
fix: establish private Paykit links promptly

ben-kaufman:
### Description

- Refresh each saved or scanned contact’s current Paykit receiver paths from Pubky during private sync and persist newly published paths such as `bitkit/server`.
- Start a bounded 28-second, two-second-cadence private link/message sync burst after contact updates, path discovery, app foregrounding, and network restoration, then return to the existing low-frequency polling cadence.
- Refresh incoming Payment Requests during the same bounded window so a newly linked request opens automatically without waiting for the normal 30-second poll.
- Preserve private-only resolution semantics; no public payment fallback was added.

### Preview

No UI changes.

### QA Notes

Regression coverage verifies that a contact saved with only `bitkit/wallet` later discovers `bitkit/server`, persists it, calls `ensureLinkWithPeer`, and never starts public payment resolution. Existing send-flow coverage now also verifies that Payment Request polling refreshes immediately when activated.

CI is the authoritative Android build/test run because the native GitHub Packages dependencies require repository package credentials.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
@@ -231,13 +231,16 @@ class PrivatePaykitRepo @Inject constructor(
 
     fun startInitialLinkBurst(publicKeys: Collection<String>, reason: String) {
         val publicKeys = normalizedPublicKeyBatch(publicKeys).normalizedKeys
-        if (publicKeys.isEmpty()) return
 
         synchronized(initialLinkBurstLock) {
-            initialLinkBurstPublicKeys += publicKeys
             initialLinkBurstGeneration += 1
-            val generation = initialLinkBurstGeneration
             initialLinkBurstJob?.cancel()
+            initialLinkBurstJob = null
+            initialLinkBurstPublicKeys.clear()
+            initialLinkBurstPublicKeys += publicKeys
+            if (publicKeys.isEmpty()) return
+
+            val generation = initialLinkBurstGeneration
             _initialLinkBurstStarted.tryEmit(Unit)
 
             initialLinkBurstJob = retryScope.launch {
--- a/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
@@ -1397,10 +1397,13 @@ class PrivatePaykitRepo @Inject constructor(
 
         return runSuspendCatching {
             val discoveredPaths = pubkyService.discoverRelevantReceiverPaths(publicKey)
-            val mergedPaths = supportedReceiverPaths(savedPaths + discoveredPaths)
-            if (mergedPaths == savedPaths) return@runSuspendCatching savedPaths
+            val currentRecord = paykitSdkService.contactRecord(publicKey)
+                ?: return@runSuspendCatching savedPaths
+            val currentSavedPaths = supportedReceiverPaths(currentRecord.receiverPaths)
+            val mergedPaths = supportedReceiverPaths(currentSavedPaths + discoveredPaths)
+            if (mergedPaths == currentSavedPaths) return@runSuspendCatching currentSavedPaths
 
-            val updatedRecord = pubkyService.saveContact(publicKey, record?.label, mergedPaths)
+            val updatedRecord = pubkyService.saveContact(publicKey, currentRecord.label, mergedPaths)
             _initialLinkBurstStarted.tryEmit(Unit)
             Logger.info("Discovered new Paykit receiver paths for '${redacted(publicKey)}'", context = TAG)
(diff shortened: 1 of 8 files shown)


app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -219,6 +229,35 @@ class PrivatePaykitRepo @Inject constructor(
         }
     }
 
+    fun startInitialLinkBurst(publicKeys: Collection<String>, reason: String) {
+        val publicKeys = normalizedPublicKeyBatch(publicKeys).normalizedKeys
+        if (publicKeys.isEmpty()) return
+
+        synchronized(initialLinkBurstLock) {
+            initialLinkBurstPublicKeys += publicKeys
```
ovitrif, line 237: I see `startInitialLinkBurst` accumulate with `initialLinkBurstPublicKeys += publicKeys` and return immediately on an empty normalized batch without cancelling the job or clearing that set. After `AppViewModel` removes deleted contacts it restarts the burst with the remaining keys (or empty when the last contact is deleted), so previously queued keys stay in the burst for the rest of the ~28s window and keep flowing through `refreshSavedContactEndpointsDuringInitialLinkBurst` / `receiverPathsForSavedContact`—including `ensureLinkWithPeer` and a possible `saveContact` after `removeSavedContact` already cleaned them. Could we replace the burst key set on each start and cancel/clear the burst when the provided key batch is empty?

ovitrif (dismissed):
Deleting a Paykit contact during an active private-link burst leaves that peer's key in the burst set, which allows later ticks to re-establish private links or republish endpoints for the rest of the 28s window.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in `1aa9b9da7`. `startInitialLinkBurst` now cancels the current job and replaces the tracked key set on every restart. Passing an empty set clears the keys and leaves no burst job running, so a removed contact cannot be re-linked or re-saved by later ticks. I added regression coverage for both replacing the key set and cancelling retries when the last contact is removed. The branch is also rebased onto current `master`; compile, the full unit suite, and detekt pass.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -1349,11 +1392,52 @@ class PrivatePaykitRepo @Inject constructor(
...
-        return paths.ifEmpty { listOf(PaykitReceiverPaths.WALLET) }
+        val record = paykitSdkService.contactRecord(publicKey)
+        val savedPaths = supportedReceiverPaths(record?.receiverPaths.orEmpty())
+
+        return runSuspendCatching {
+            val discoveredPaths = pubkyService.discoverRelevantReceiverPaths(publicKey)
+            val mergedPaths = supportedReceiverPaths(savedPaths + discoveredPaths)
+            if (mergedPaths == savedPaths) return@runSuspendCatching savedPaths
+
+            val updatedRecord = pubkyService.saveContact(publicKey, record?.label, mergedPaths)
```
ovitrif: I see `receiverPathsForSavedContact` persist merged paths with `saveContact` after the Paykit contact record has already been deleted. After a delete, `PubkyRepo.removeContact` drops that record first, then `removeSavedContact` waits on `serializedDispatcher` behind the in-flight burst tick, so the tick treats the missing record as wallet-only, discovers `bitkit/server`, and `saveContact` recreates the deleted contact. The next `loadContacts()` then returns that deleted contact to the list. Could we skip `saveContact` unless `contactRecord` is still present after discovery returns?

ovitrif (changes requested):
Deleting a contact while a private-link burst tick is in flight can bring that contact back, because path discovery still calls saveContact after the Paykit record is already gone.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in cfc308868. The refresh now re-reads the contact after receiver-path discovery and only persists the merged paths using that current record. If the contact was deleted while discovery was in flight, it returns without calling `saveContact`, so the contact is not recreated. I also added a regression test that simulates the record changing from present to absent during discovery.

ovitrif (approved):
The burst tick now skips saveContact when the Paykit contact is gone after discovery, so a deleted contact is not recreated.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ovitrif: After discovery the refresh now re-reads the contact and returns without saveContact when that record is gone, so the deleted contact is not recreated.

PR #1157, opened 2026-08-13
fix: gate screen deeplinks to debug

### LINKED ISSUE #1132: fix: gate screen deeplinks to debug
Parent: #1113
Refs: #1119

> [!WARNING]
> #1119 must NOT land in prod as-is, it would open a very risky vulnerability doorway.
> This issue describes the guard required to enforce that risk never reaches the prod userbase.

## Goal

Confine `bitkit://screen/...` deep-link registration and handling to debug builds via Android `src/debug` / `src/release` source sets, so release/store APKs cannot honor screen deep links even when Dev Mode is enabled.

Runtime `isDevModeEnabled` alone is not a durable control plane for this surface: any installed app can fire `bitkit:` intents, and a future change can weaken a boolean gate. Source-set separation fails closed at package time for prod users.

## Out of scope

- Do not change the `bitkit` scheme (or other hosts) in the manifest.
- Do not move `Routes.DeepLinkable` / sheet `DeepLinkStart` markers out of `main`; markers alone are harmless if release never attaches links or handles URIs.
- Agents, do not add APK-level asserts that scan shipped `NavDeepLink` patterns, it's irrelevant.
- Do not require `mainnet`-specific branching; the store boundary is `DEBUG=false` / release source sets.

## Approach

- Keep a thin API in `main` used by `AppViewModel`, `MainActivity`, `ContentView`, and nav helpers.
- Refactor to move the screen deep-link logic to `src/debug` (URI attach via `linksFor`, `SheetDeepLinks.sheetFor`, pending-URI queueing), practically turning `release` builds handling into a NOOP.
- Provide no-op implementations in `src/release` to keep compilation builds succeesful:
  - `linksFor` returns an empty list (no screen deep links on the nav graph).
  - `sheetFor` returns null.
  - Screen URIs are never queued for replay.
  - Intent detach for the `screen` host may remain as belt-and-suspenders.
- On debug builds, keep `isDevModeEnabled` as an additional gate so Dev Mode off still ignores screen URIs.

## Acceptance criteria

- [ ] Release/store builds (`src/release`) never register `bitkit://screen/...` deeplinks or handle such logic.
- [ ] Debug builds retain current developer behavior, still gated by Dev Mode.
- [ ] Existing non-screen `bitkit:` deep links remain unchanged (no manifest edits; no broad deeplink regression pass required beyond screen-namespace checks).
- [ ] Unit tests cover release no-ops and debug acceptance/rejection under Dev Mode.
- [ ] Journeys under `journeys/deeplinks/` still pass on debug builds.

ovitrif:
Fixes #1132

This PR confines `bitkit://screen/...` deep links to debug builds, so release and store APKs cannot honor them even when Dev Mode is enabled.

### Description

Screen deep-link registration, sheet lookup, and pending-URI replay now live behind debug/release source sets. Debug builds keep the current developer behavior and still require Dev Mode. Release builds attach no screen URIs to the nav graph, never open a sheet from a screen URI, and never queue one for replay. The activity still strips a `screen` URI from the intent as belt-and-suspenders. Other `bitkit:` hosts and payment URIs are unchanged.

### Preview

N/A

### QA Notes

Dev mode is on by default on debug builds (Settings ▸ Advanced ▸ Dev Settings). The app must be past onboarding.

#### Manual Tests
- [x] **1.** `adb shell am start -a android.intent.action.VIEW -d "bitkit://screen/settings" to.bitkit.dev` → Settings opens.
- [x] **2.** Deep-linked Settings → Back: returns to the wallet overview, not the launcher.
- [x] **3a.** `bitkit://screen/send` → Send sheet on the recipient picker.
  - [x] **3b.** `bitkit://screen/widgets/price-edit` → Bitcoin Price editor.
- [x] **4.** `bitkit://screen/recovery-mnemonic` and `bitkit://screen/backup/show-mnemonic` → screen unchanged, recovery phrase never shown, logcat carries `Unhandled screen deeplink`.
- [x] **5.** `bitkit://screen/send/fee-rate` → screen unchanged, no crash.
- [x] **6.** Settings ▸ Support → tap Version five times to turn Dev Mode off → `bitkit://screen/settings` on a warm start and a cold start: wallet overview stays, Settings does not open, logcat carries `Ignoring screen deeplink, not queued`.
- [x] **7.** `regression:` scan a `bitcoin:` / `lightning:` / `lnurl` URI → still decodes through the scanner path.
- [x] **8.** Release/store build with Dev Mode on → `bitkit://screen/settings`: ignored, Settings does not open.

#### Automated Checks
- Existing unit tests in `ScreenDeepLinksTest.kt`, `SheetDeepLinksTest.kt`, and `AppViewModelSendFlowTest.kt` still cover URI patterns, sheet lookup, and Dev Mode acceptance/rejection.
- Journeys under `journeys/deeplinks/` still apply on debug builds.


app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3565,7 +3565,7 @@ class AppViewModel @Inject constructor(
         }
 
         if (ScreenDeepLinks.isScreenDeepLink(uri)) {
-            if (!settingsStore.data.first().isDevModeEnabled) {
+            if (!ScreenDeepLinks.shouldQueue(settingsStore.data.first().isDevModeEnabled)) {
```
jvsena42, line 3721: should update the journey condition   <description>Precondition: onboarded dev wallet, dev mode on, at least one log file present.</description>

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3565,7 +3565,7 @@ class AppViewModel @Inject constructor(
...
                 Logger.warn("Ignoring screen deeplink, dev mode is off", context = TAG)
```
jvsena42: nit: Could update the log message now that the skip condition was widened

app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt
```diff
@@ -758,13 +759,17 @@ class AppViewModelSendFlowTest : BaseUnitTest() {
...
     @Test
-    fun `screen deeplink is held for replay when dev mode is on`() = test {
+    fun `screen deeplink is held for replay only on debug when dev mode is on`() = test {
         settingsData.value = SettingsData(isDevModeEnabled = true)
 
         sut.handleDeeplinkIntent(screenIntent("settings"))
         advanceUntilIdle()
 
-        assertNotNull(sut.pendingScreenDeepLink.value)
+        if (ScreenDeepLinks.isEnabled) {
```
jvsena42, line 1144: this is always true in unit tests, check if makes sense keep the test

jvsena42 (changes requested):

app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt
ovitrif: The idea was to prove and "lock" the fact that it works in case of accidental disables 🤷🏻

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Updated — journey now requires a debug build, matching the wider queue gate.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Updated the ignore log so it no longer says only Dev Mode is off.

jvsena42 (approved):

PR #1158, opened 2026-08-13
fix: secure shop payments and locked links

ben-kaufman:
### Description

This PR:

1. Restricts native shop payment messages to the exact `https://embed.bitrefill.com` origin through an origin-scoped WebMessage listener.
2. Disables native shop payments when origin-aware WebView messaging is unavailable, rejects non-string messages, and keeps broader Bitrefill navigation separate from the privileged payment origin.
3. Defers payment links and scans until authentication completes while preserving the intentional QuickPay opt-in behavior.
4. Keeps only the latest deferred external scan while incoming Paykit requests remain repository-backed until authentication.
5. Serializes scan execution and preserves each scan's contact context so concurrent requests cannot overlap or be attributed to the wrong contact.

Addresses VulnHunter 01b finding 02 and the unauthenticated deeplink handling from finding 03. The explicit QuickPay bypass of per-payment PIN confirmation remains accepted product behavior.

Companion iOS PR: https://github.com/synonymdev/bitkit-ios/pull/668

### Preview

N/A — security boundary and payment-flow state handling only.

#### Manual Tests

- [ ] **1.** Shop → Gift Cards → Bitrefill checkout → submit a payment: Send Confirm or QuickPay opens with the expected invoice.
- [x] **2a.** `regression:` Shop → navigate between legitimate Bitrefill pages: navigation remains available.
  - [x] **2b.** Shop → attempt off-Bitrefill main-frame navigation: navigation is blocked.
- [x] **3.** Settings → PIN on + PIN for payments on + QuickPay on → authenticate, then scan a small LN invoice: QuickPay opens without another PIN prompt, as designed.
- [x] **4.** Settings → PIN on + PIN for payments off + QuickPay on → scan a small LN invoice: QuickPay still opens.
- [x] **5.** Lock app → open a QuickPay-eligible `lightning:` link → authenticate: no payment flow opens before authentication, then QuickPay opens once.
- [x] **6.** Lock app → open two different payment links → authenticate: only the latest link opens and the earlier link is not replayed.
- [ ] **7.** Lock app with an incoming Paykit request → open a payment link → authenticate → dismiss the external send flow: the external payment opens first, then the Paykit request opens once with its contact context.
- [x] **8.** Lock app with a payment link deferred → open Scanner → authenticate → scan another invoice: the scanner result opens first and the deferred link opens after it is dismissed.

#### Automated Checks

- `ShopOriginTest.kt`: covers the exact payment origin, trusted port/user-info boundary, broader navigation allow-list, and bridge-page guard.
- `ShopWebViewInterfaceTest.kt`: covers feature detection, fail-closed fallback, exact listener origin, sibling-origin rejection, non-string messages, and payment parsing.
- `ShopWebViewClientTest.kt`: covers supported/unsupported bridge injection, sibling-page non-injection, and main-frame navigation boundaries.
- `AppViewModelSendFlowTest.kt`: covers authentication/QuickPay deferral, latest-wins replacement, context isolation, cancellation serialization, scanner ordering, sheet dismissal, and Paykit handoff races.
- Test coverage removed: deleted FIFO/capacity/stale-queue cases because deferred external scans now use one latest-wins slot and automatic Paykit requests remain repository-backed.
- Local verification: `compileDevDebugKotlin`, `testDevDebugUnitTest`, `detekt --rerun-tasks`, and `git diff --check`.


app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopWebViewInterface.kt
```diff
@@ -1,70 +1,93 @@
...
+                        Logger.warn("Received payment_intent with empty URI", context = TAG)
+                        return
+                    }
+                    onPaymentIntent(uri)
                 }
+                else -> Logger.debug("Ignored shop WebView event '${data.event}'", context = TAG)
             }
         }.onFailure {
-            Logger.error("Error parsing message: $message", it, context = "WebView")
+            Logger.error("Failed to parse shop WebView message", it, context = TAG)
```
jvsena42: nit: got noisily error logs `Failed to parse shop WebView message [MissingFieldException: Field 'event' is required] ` with full stack traces.

Bitrefill posts routine non-payment messages from the trusted origin, and the .`onFailure` treats each as an error. Logger.debug would fit better. 

pre-existent

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1578,31 +1591,146 @@ class AppViewModel @Inject constructor(
...
+
+    private fun isPaymentRequestPresentationBlocked() = !_isAuthenticated.value ||
+        currentSheet.value != null ||
+        hasActiveContactPaymentContext() ||
+        isScanPendingOrActive()
+
+    private fun flushDeferredScan() {
+        if (!_isAuthenticated.value) return
+        if (activeScanJob?.isActive == true) return
+        if (_currentSheet.value != null) return
```
jvsena42, line 1721: **`flushDeferredScan` can flush into a sheet that is mid-transition (medium)**

This guard treats `_currentSheet.value == null` as "UI is idle", but `showSheet()` sets the sheet from a fire-and-forget `viewModelScope.launch` that first nulls `_currentSheet` and then `delay(SCREEN_TRANSITION_DELAY)` before assigning the new one.

Scenario: a deferred link is flushed with `preserveUntilComplete = true` (scan A) while a sheet is visible, and a second scan B is parked in the deferred slot. A's `handleScan` reaches `navigateToSendRoute` → `showSheet(Sheet.Send(...))`, which nulls the sheet and suspends for 300 ms. A's job then completes and `invokeOnCompletion` posts `flushDeferredScan()`, which observes `_currentSheet.value == null`, pops B and launches it. B's `handleScan` calls `resetSendState`/`resetQuickPay` and overwrites `_sendUiState` before A's `showSheet` resumes and displays `Sheet.Send` with A's route but B's payment data.

That defeats the `mustComplete` guarantee this PR adds, and can surface a Send/Confirm sheet whose destination isn't the one that produced the route.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1578,31 +1591,146 @@ class AppViewModel @Inject constructor(
...
-            handleScan(data, routePubkyKeys)
-        }.also { it.invokeOnCompletion { if (activeScanInput == normalized) activeScanInput = null } }
+            scanMutex.withLock {
+                setActiveContactPaymentContext(contactPaymentContext)
+                if (startDelay > Duration.ZERO) delay(startDelay)
+                handleScan(data, routePubkyKeys)
+            }
+        }.also { job ->
+            job.invokeOnCompletion {
+                if (activeScanJob === job) {
```
jvsena42: **Race between cancelling the prior job and assigning the new one (medium)**

In `launchScan` the previous job is cancelled first, then `scheduledScan` is assigned, then `activeScanJob` is reassigned. The cancelled job's `handleScan` runs under `withContext(bgDispatcher)`, so its completion — and therefore this `invokeOnCompletion` handler — fires on a background thread at an arbitrary later point.

If it fires after `scheduledScan` is assigned but before `activeScanJob` is reassigned, `activeScanJob === job` is still true for the **old** job, so `scheduledScan = null` wipes the `ScheduledScan` just created for the **new** scan. The new scan then runs with no recorded `mustComplete` and no `normalizedInput`, so:

1. the `activeScanJob?.isActive && scheduled?.mustComplete` guard no longer protects a flushed deferred scan — an unrelated incoming scan cancels it and it is dropped; and
2. duplicate suppression stops working, so a repeated `payment_intent`/deeplink restarts the scan instead of being skipped.

Assigning `scheduledScan`/`activeScanJob` before cancelling, or capturing the job identity differently, closes the window.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -693,7 +706,7 @@ class AppViewModel @Inject constructor(
 
     private suspend fun presentIncomingPaymentRequestOrStop(request: PaykitPaymentRequest): Boolean {
         val result = privatePaykitRepo.beginPaymentRequest(request).getOrNull()
-        if (currentSheet.value != null || hasActiveContactPaymentContext()) return true
+        if (isPaymentRequestPresentationBlocked()) return true
```
jvsena42, line 709: **A deferred Paykit request can be marked presented but never shown (low)**

Refers to `presentedPaymentRequestIds += request.id` a few lines below (line 717): it happens before `openContactPayment(...)`, and `openContactPayment` now only *schedules* the scan.

If `isPending(request)` suspends long enough for a `preserveUntilComplete` scan to start, `launchScan` takes the `mustComplete` branch and parks the Paykit request in the single latest-wins deferred slot. Any subsequent scan then replaces it via `enqueueDeferredScan`. Because the request's context never became the active one, `setActiveContactPaymentContext` never runs its `presentedPaymentRequestIds -= replacedRequestId` compensation, and `retainPaymentRequestPresentationState` only drops ids that are no longer pending — so the request stays marked as presented but is never shown again for the lifetime of the ViewModel.

app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopOrigin.kt
```diff
@@ -0,0 +1,62 @@
...
+
+internal fun shopPaymentOriginRules(): Set<String> = setOf(Env.BITREFILL_URL)
+
+internal fun shopMessageBridgeScript(): String = """
+    window.ReactNativeWebView = {
+        postMessage: function(data) {
+            Android.postMessage(typeof data === 'string' ? data : JSON.stringify(data));
+        }
+    };
+    window.addEventListener('message', function(event) {
```
jvsena42: **Bridge script isn't idempotent across repeated `onPageFinished` (low)**

`shopMessageBridgeScript()` is re-evaluated on every `onPageFinished`, and `onPageFinished` is documented to fire more than once for a single document (frames, redirects). The listeners accumulate in the same JS context, so one `payment_intent` is forwarded to `Android.postMessage` N times, invoking `onPaymentIntent` → `launchScan` repeatedly.

This is mostly masked by the duplicate-scan check — but that check is exactly what the `scheduledScan` race (other comment on `AppViewModel.kt:1644`) can disable. A `if (window.__bitkitShopBridge) return; window.__bitkitShopBridge = true;` guard, or assigning `window.onmessage` instead of adding a listener, makes it idempotent.

Separately, while loading Gift Cards on device this produced 3× `ERROR ... Failed to parse shop WebView message [MissingFieldException: Field 'event' is required]` with full stack traces — Bitrefill posts routine non-payment messages from the trusted origin, and `handlePaymentMessage`'s `runCatching{}.onFailure { Logger.error }` treats each as an error. `Logger.debug` would fit better there.

app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopWebViewClient.kt
```diff
@@ -24,29 +28,17 @@ class ShopWebViewClient(
...
-                        Android.postMessage(data);
-                    } else {
-                        Android.postMessage(JSON.stringify(data));
-                    }
-                };
-            }
-            """.trimIndent(),
-            null
-        )
+    override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest?): Boolean {
```
jvsena42, line 37: **Blocked navigation dead-ends with no user feedback (low)**

This now returns `true` for every main-frame navigation that isn't `https://*.bitrefill.com`, and the only feedback is a log line. Reproduced on device: tapping the reCAPTCHA "Privacy Policy" link in the Bitrefill footer logs `Blocked shop navigation to untrusted origin 'https://policies.google.com/privacy'` and nothing else happens — no toast, no handoff to the browser.

Any `mailto:`, `tel:`, `intent:`, or third-party redirect the checkout emits (support links, external identity/payment providers) will behave the same way. Consider surfacing a toast, or routing non-HTTPS well-known schemes to the system handler via `Intent.ACTION_VIEW` while still blocking off-origin HTTP(S).

jvsena42 (commented):
Code review + manual verification on a regtest dev build (Pixel_9 emulator).

**Verified working on device:** QuickPay is skipped when PIN-for-payments is on (`Skipping QuickPay because PIN is required for payments` → Confirm sheet → PIN prompt on swipe); QuickPay still runs with PIN-for-payments off; payment links fired while locked are queued and opened exactly once after auth; a second link replaces the first (`Replacing deferred scan`) and the earlier one is never replayed; off-Bitrefill main-frame navigation is blocked (`Blocked shop navigation to untrusted origin 'https://policies.google.com/privacy'`) while in-shop navigation is unaffected.

**Not covered on device:** Bitrefill checkout (dev flavor talks to the real `embed.bitrefill.com`, so a checkout emits a mainnet invoice a regtest wallet can't act on), the Paykit handoff race, and the scanner-ordering case.

5 inline findings below.

app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopWebViewInterface.kt
ben-kaufman: Addressed in fb1dd89ea. Unrecognized trusted-origin messages are now ignored at debug level without attaching a throwable, so routine Bitrefill traffic no longer emits error stack traces. Added coverage for messages without an event.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in fb1dd89ea. Sheet transitions now have an explicit tracked job. Deferred scans and incoming Paykit presentation remain blocked while that job is active, and hiding a sheet cancels any pending transition. The new regression test holds a protected replay through a competing sheet transition and verifies the second scan cannot overwrite the first send state.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in fb1dd89ea. ScheduledScan now owns the Job, so job identity and scan metadata are published as one volatile value before the replacement starts and the predecessor is cancelled. Completion clears state only when its ScheduledScan instance is still current, removing the cancellation and assignment window.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Verified this interleaving is not possible in the current code. PaykitPaymentRequestRepo.isPending is a non-suspending function. presentIncomingPaymentRequestOrStop already rechecks isPaymentRequestPresentationBlocked immediately after the only suspension, beginPaymentRequest, and then synchronously checks pending state, marks the id, and calls openContactPayment. The endpoint-resolution race is covered by AppViewModelSendFlowTest. I left the code unchanged to avoid a redundant guard.

app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopOrigin.kt
ben-kaufman: Fixed in fb1dd89ea. The injected script now uses window.__bitkitShopBridgeInstalled to make listener installation idempotent. The parsing path also treats unrecognized trusted-origin messages as debug noise without stack traces. Tests cover the guard and missing-event traffic.

app/src/main/java/to/bitkit/ui/screens/shop/shopWebView/ShopWebViewClient.kt
ben-kaufman: Addressed in fb1dd89ea. Blocked main-frame navigation now surfaces a warning toast instead of silently dead-ending. The navigation remains fail-closed; arbitrary external URLs and schemes are not handed to the OS, preserving the security boundary introduced by this PR.

pwltr:
Note that QuickPay skipping PIN checks is assumed risk per design.

image

ben-kaufman:
@pwltr Thanks for clarifying. Agreed—QuickPay is the explicit opt-in exception to per-payment PIN confirmation. I removed the new PIN-for-payments gate in 32b59a538, kept the app-authentication deferral for locked links, and updated the regression test, changelog, title, and PR description to state the accepted behavior accurately. The focused test, full unit suite, compile, and detekt checks are green.

ovitrif:
Follow-up stacked PR that restores payment-PIN enforcement for QuickPay: https://github.com/synonymdev/bitkit-android/pull/1159

PIN for payments is a user-facing control. If it is on, every outgoing spend path should honor it — including QuickPay. The earlier “assumed risk” speed tradeoff is no longer acceptable: the automatic path settles real spending balance, and impact scales with the user’s QuickPay threshold and available spending. Manual send already asks for that PIN, and iOS already routes to Confirm in this case.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1579,14 +1583,23 @@ class AppViewModel @Inject constructor(
...
+            Logger.info(
+                "Queuing '${source.label}' scan until authenticated: '${scanLogId(data)}'",
+                context = TAG,
+            )
+            pendingLockedScan = PendingLockedScan(
+                source = source,
+                data = data,
+                startDelay = startDelay,
+                routePubkyKeys = routePubkyKeys,
+            )
```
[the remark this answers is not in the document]
ovitrif: Intentional latest-wins while locked: only one deferred payment input is kept, and a newer link replaces the older one (with a warn log). Flush still runs exactly once after unlock. Keeping a full queue would change the product behavior of this PR; out of scope here.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1605,6 +1618,26 @@
         }.also { it.invokeOnCompletion { if (activeScanInput == normalized) activeScanInput = null } }
     }
 
+    private fun scanLogId(data: String): String {
+        val scanLogInput = SamRockSetupRequest.sanitizedDescription(data.removeLightningSchemes()) ?: data
+        return if (scanLogInput.length > 24) {
+            "${scanLogInput.take(11)}…${scanLogInput.takeLast(11)}"
```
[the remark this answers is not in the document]
ovitrif: Already addressed on this branch — the flagged magic numbers are gone (alert instances report `fixed`). Local `just lint` is clean after the master rebase.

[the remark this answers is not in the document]
ovitrif: Already addressed on this branch — the flagged magic numbers are gone (alert instances report `fixed`). Local `just lint` is clean after the master rebase.

ovitrif (approved):
The shop payment origin gate, fail-closed WebView messaging, and locked-link deferral look correct on 49ab192, including the latest-wins deferred scan and the accepted QuickPay PIN bypass.

ovitrif:
Emulator QA on `to.bitkit.dev` @ `5ccc5c038` (Pixel emulator-5554). PIN `1234`, QuickPay on, spending 400000 sats. Self-pay of this wallet's invoices fails after QuickPay starts (`DuplicatePayment`); the check was that Confirm/QuickPay opened, not that the payment settled.

- [ ] **1.** Shop → Gift Cards → Bitrefill checkout → submit a payment: Send Confirm or QuickPay opens with the expected invoice.
  Reached Zalando €20 cart. Did not submit. Dev shop talks to real `embed.bitrefill.com` (mainnet invoice on a regtest wallet).
- [x] **2a.** Shop → navigate between legitimate Bitrefill pages: Gift Cards → eSIMs stayed on Bitrefill.
- [x] **2b.** Shop → off-Bitrefill main-frame nav: `https://evil.example/pay` blocked. Toast: "This link can’t be opened from the shop."
- [x] **3.** PIN on + PIN for payments on + QuickPay on → small LN invoice: QuickPay opened with no second PIN (500 sats).
- [x] **4.** PIN on + PIN for payments off + QuickPay on → small LN invoice: QuickPay still opened (600 sats).
- [x] **5.** Lock app → QuickPay-eligible `lightning:` link → authenticate: no send UI before PIN, then QuickPay once (700 sats).
- [x] **6.** Lock app → two payment links → authenticate: 800 sat replaced, only 900 sat opened.
- [ ] **7.** Lock + incoming Paykit request + payment link: not run. Android has no `proposePaymentRequest` UI. Mutual contacts only enable Contact Pay (outgoing). Incoming requests come from Paykit Server / Locks (`paykit-payment` invoice), not Bitkit ↔ Bitkit.
- [x] **8.** Lock + deferred `lightning:` link → open Scanner → authenticate → scan another invoice: scanner result opened first, deferred 400-sat QuickPay opened after dismiss.

Local artifacts: `.ai/pr/1158/review/`

PR #1162, opened 2026-08-17
fix: gate pubky auth handler

ben-kaufman:
### Description

This PR prevents Bitkit from advertising itself as a `pubkyauth` handler unless the Paykit UI is enabled and the current Pubky identity has a matching local secret key.

Previously, the main activity registered the scheme unconditionally, so Android could route Pubky App authorization links to Bitkit even when the production UI was unavailable or the identity was managed by Pubky Ring. The handler now lives on a disabled-by-default activity alias. Bitkit enables it only while local authorization is possible and disables it again when the feature or identity state changes.

### Preview

N/A — no UI changes.

#### Manual Tests

- [ ] **1.** With Paykit UI disabled, request Pubky authorization: Bitkit is not offered or selected.
- [ ] **2.** With Paykit UI enabled but no Pubky identity, request authorization: Bitkit is not offered or selected.
- [ ] **3.** With a Ring-managed identity, request authorization: Bitkit is not offered or selected.
- [ ] **4.** With a Bitkit-managed identity, request authorization: Bitkit opens the authorization screen.
- [ ] **5.** Remove the local identity or disable Paykit UI, then retry: Bitkit no longer handles the link.

#### Automated Checks

- `PubkyAuthHandlerRegistrarTest.kt`: covers eligibility, state transitions, idempotent startup, and PackageManager failure recovery.
- `PubkyAuthManifestTest.kt`: ensures the main activity no longer claims `pubkyauth` and the alias is disabled by default.
- Local compile, complete dev unit suite, and detekt passed.


ovitrif (approved):
Approved. This change prevents Bitkit from advertising the pubkyauth handler unless Paykit UI is enabled and the current identity has a matching local secret.

Issue #676, opened 2026-01-20
ldk-node Error


thisispav:
Setup a new RN wallet, did a CJIT (Phoenix to Bitkit) then updated to native v169 successfully. 

Went to check the LN connections and I've been stuck on this window for the past 5 min. Tried closing and re-opening the app multiple times to no avail.

Logs attached

[bitkit_logs_1768906123372.zip](https://github.com/user-attachments/files/24735571/bitkit_logs_1768906123372.zip)

jvsena42:
Test if it is the same case of #652

PR #1172, opened 2026-08-20
feat: add private payment requests

ben-kaufman:
This PR adds private Paykit Payment Requests to Bitkit.

### Description

1. Automatically opens incoming requests in the existing Send confirmation flow with the requesting contact, exact amount, and a Payment Request title.
2. Keeps dismissed requests actionable through a bell, preview sheet, and full request history, with manual reopen and explicit rejection.
3. Adds outgoing request creation from Receive for linked, saved contacts, including amount, note, expiry, queued delivery state, and sent history.
4. Drops expired or remotely unavailable requests, keeps presentation state scoped to the active Pubky identity, and protects account changes, sheet transitions, and overlapping actions.
5. Requires strict private resolution for requests: a consumed Private Payment List is never reused, another endpoint from that list is not attempted, and public details are never used as fallback while waiting for a newer list.
6. Updates Paykit to `0.1.0-rc44` and adds local E2E homeserver configuration plus safe cold-start restoration for externally managed Pubky sessions.

The request payload itself remains SDK-backed and durable; Bitkit persists only encrypted, identity-scoped presentation suppression, not a duplicate request queue. Payment proofs and receipts remain out of scope.

Dependencies:

- [Paykit 0.1.0-rc44](https://github.com/pubky/paykit-rs/releases/tag/v0.1.0-rc44)
- Local Locks interoperability was verified with [paykit-server#2](https://github.com/pubky/paykit-server/pull/2) at `41cda2567226a690a012770017d5e7c1d49e2a2b`.
- Companion iOS PR: [synonymdev/bitkit-ios#676](https://github.com/synonymdev/bitkit-ios/pull/676)

### Preview

N/A — proof recordings were completed locally and are intentionally not attached to the PR.

#### Manual Tests

- [x] **1.** Clean wallet → create Pubky profile → enable Paykit and Contact Payments → add the peer as a contact: private-capable request action appears once the Noise link is established.
- [x] **2.** Peer creates a private request → Home: Payment Request opens automatically with the correct contact and amount.
- [x] **3.** Payment Request → dismiss without rejecting → bell → request preview → Pay: the request remains queued, reopens, and pays successfully.
- [x] **4.** Peer creates a second request after payment → Pay: a newer Private Payment List is used; the consumed list is not reused and no public fallback occurs.
- [x] **5.** Receive → Send Payment Request → enter amount, note, and expiry → select linked contact → Send: request is queued once and appears in sent history.
- [x] **6.** Incoming request → Reject: only that request becomes terminal and the private payment list is not consumed.
- [x] **7.** Incoming request → dismiss → relaunch: request remains discoverable without automatically reopening again; expired requests disappear live.
- [x] **8.** Cross-platform E2E: iOS creates two requests that Android receives and pays, and Android creates two requests that iOS receives and pays, on clean regtest state.

#### Automated Checks

- `PaykitPaymentRequestRepoTest.kt`: covers mapping, eligibility, proposal delivery, rejection, expiry, identity-scoped presentation state, and action serialization.
- `PaykitSdkServiceTest.kt`: covers exact identity enforcement and safe deferred session restoration.
- `AppViewModelSendFlowTest.kt`: covers automatic/manual presentation, sheet transitions, identity changes, newer-list retry, strict private resolution, and payment lifecycle races.
- `PaymentRequestExpirationTest.kt`: covers expiry selection and retained draft state.
- `SheetHostTest.kt`: covers locked sheet dismissal and scrim input isolation during durable proposal creation.
- Local verification passed: app compile, Android-test compile, complete dev unit suite, detekt, and `git diff --check`.
- Both final cross-platform proof videos decoded end to end without errors.


app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3672,13 +3815,64 @@ class AppViewModel @Inject constructor(
...
+        draft: PaykitPaymentRequestDraft,
+        target: PaykitPaymentRequestTarget,
+        onCreated: (PaykitPaymentRequest) -> Unit,
+    ) {
+        viewModelScope.launch {
+            createPaymentRequest(draft, target)
+                .onSuccess { request ->
+                    if (sentPaymentRequests.value.any { it.id == request.id }) {
+                        onCreated(request)
+                    }
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in f42220d72. If the SDK has already durably queued the request but the active identity changed before local publication, Bitkit now closes only the stale Receive flow and shows an explicit queued confirmation. It does not publish the old request into the new identity state, and the user is no longer left on a screen where retrying could create a duplicate. Added regression coverage for this case.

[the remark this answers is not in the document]
ben-kaufman: Final signed fix is ba7f74577. It also scopes the cleanup to the exact originating Receive sheet, so a later replacement sheet is preserved; regression coverage includes both behaviors.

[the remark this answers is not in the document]
ben-kaufman: Final signed fix is e8183a594. The repository now returns an explicit durable-creation outcome with the creator identity and whether it was published into the still-active identity state. The ViewModel only opens success for that exact active identity; otherwise it closes only the originating Receive flow and confirms that the request is queued. Tests cover the paused proposal across A→B activation, the user feedback, and preserving a replacement sheet.

ovitrif (dismissed):
Looks good. Durably queued requests stay confirmed after an identity change, and a consumed private payment list is never reused or replaced by public details.

ovitrif:
> Local Locks interoperability was verified with https://github.com/pubky/paykit-server/pull/2 at 41cda2567226a690a012770017d5e7c1d49e2a2b.

Note this `paykit-server` needs to integrated into the bitkit-docker and e2e tests soonish 🙏🏻 

cc. @jvsena42 @piotr-iohk

piotr-iohk:
> > Local Locks interoperability was verified with [pubky/paykit-server#2](https://github.com/pubky/paykit-server/pull/2) at 41cda2567226a690a012770017d5e7c1d49e2a2b.
> 
> Note this `paykit-server` needs to integrated into the bitkit-docker and e2e tests soonish 🙏🏻
> 
> cc. @jvsena42 @piotr-iohk

As far as paykit-server it would be good to have a staging deployment since paykit e2e tests are run against our staging regtest and also using staging homeserver (as far as I understand that was the plan, see: https://github.com/synonymdev/bitkit-android/pull/1084#issuecomment-5045780919)

>Server staging
The intended next server-side QA step is to deploy Paykit Server on staging once its runtime/listener composition is implemented. There is no staging endpoint yet, and that work is outside these mobile PRs.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -102,6 +126,50 @@ fun ReceiveSheet(
                         onClickEditInvoice = { navController.navigateTo(ReceiveRoute.EditInvoice) },
                     )
                 }
+                composableWithDefaultTransitions<ReceiveRoute.PaymentRequestDetails> {
+                    PaymentRequestDetailsScreen(
+                        amountInputViewModel = paymentRequestAmountViewModel,
+                        initialDraft = paymentRequestDraft,
+                        onBack = { navController.popBackStack() },
```
ovitrif: Confirmed on-device issue: Payment Requests → Request Payment opens Receive on the Payment Request form, but that sheet still hard-codes `startDestination = ReceiveRoute.QR` and then `navigateTo(startRoute)`. The receive QR is therefore behind the sheet page we reach, and the in-sheet back arrow pops onto it instead of closing the sheet.

Could we start the Receive `NavHost` on the Payment Request subscreen, the same way Send already starts on `sheet.route`?

```kotlin
NavHost(
    navController = navController,
    startDestination = startRoute,
)
```

Send already does this:

```kotlin
SendSheet(
    ...
    startDestination = sheet.route,
)
```

```kotlin
NavHost(
    navController = navController,
    startDestination = startDestination,
)
```

PS. Fixing this will also reach behavior parity with how iOS does this (not device-confirmed, code-based reasoning by looking at how the nav back event is routed there).

ovitrif (commented):
Navigating back from Payment Request sheet should return you to Payment Requests (list) screen. It instead shows the Receive Bitcoin QR, with Payment Requests still behind the sheet.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3672,13 +3816,77 @@ class AppViewModel @Inject constructor(
...
-        toast(error)
-        val request = context?.incomingPaymentRequest
-        if (request != null && paykitPaymentRequestRepo.isPending(request)) {
-            presentedPaymentRequestIds.remove(request.id)
-            deferPaymentRequestPresentation(request)
+    fun showPaymentRequests() {
+        showSheet(Sheet.PaymentRequests)
+    }
+
+    fun openIncomingPaymentRequest(id: PaykitPaymentRequestId) {
```
jvsena42, line 3957: **`requestedPaymentRequestId` can stay set forever, wedging the Pay button.**

This bails out when `requestedPaymentRequestId != null`, and the only things that clear it are `onSheetVisible` (via `markPresented`), the request leaving `pendingRequests`, or the retry give-up.

If `openContactPayment` -> `handleScan` aborts *before* the Send sheet ever becomes visible (e.g. the peer's payload hits the duplicated-BIP21/decoding-error branch that toasts and calls `hideSheet`), nothing clears it: Pay becomes a silent no-op on every request, and `rejectIncomingPaymentRequest` returns `OperationInProgress` for that request.

Worse, `hideSheet` makes `currentSheet` null, which retriggers `presentNextIncomingPaykitPaymentRequest`, which re-selects the same `requestedId` and re-runs the failing scan - an error-toast loop.

The old code handled this via `handlePaymentPreparationFailure(context, error)` -> `deferPaymentRequestPresentation`; that call was dropped.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
```diff
@@ -0,0 +1,616 @@
...
+
+@Composable
+internal fun PaymentRequestCard(
+    request: PaykitPaymentRequest,
+    contact: PubkyProfile?,
+    compactSubtitle: String? = null,
+    onPay: (() -> Unit)? = null,
+    onReject: (suspend () -> Result<Unit>)? = null,
+) {
+    val scope = rememberCoroutineScope()
```
jvsena42: **Reject runs on the composition scope and gets cancelled mid-flight.**

`onReject()` is launched here on `rememberCoroutineScope()`, and it ends up in `PaykitPaymentRequestRepo.updateRequest`, which does:

```kotlin
_pendingRequests.update { requests -> requests.filterNot { it.id == current.id } }
discardExpiredRequestsLocked()
processPendingMessages()   // <- actually delivers the rejection to the counterparty
```

As soon as that `_pendingRequests` update lands, this card (and, in the sheet, the whole sheet via `LaunchedEffect(requests.isEmpty())`) leaves composition. That cancels the scope, and therefore the still-running `withContext(ioDispatcher)` body - so the outbound rejection delivery is dropped.

Launch this in `viewModelScope` instead.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3921,12 +4129,7 @@ class AppViewModel @Inject constructor(
         private const val PAYKIT_CHANNEL_USABILITY_REFRESH_DELAY_MS = 5_000L
         private val PAYKIT_PAYMENT_REQUEST_REFRESH_INTERVALS = listOf(30.seconds, 60.seconds, 120.seconds)
         private val INITIAL_PAYKIT_SYNC_RETRY_DELAYS = List(14) { 2.seconds }
-        private val PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS = listOf(
-            30.seconds,
-            60.seconds,
-            120.seconds,
-            300.seconds,
-        )
+        private val PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS = List(14) { 2.seconds }
```
jvsena42, line 4281: **Automatic presentation retry window shrank from ~8.5 min to 28 s.**

This went from `[30s, 60s, 120s, 300s]` to `List(14) { 2.seconds }`. The 2 s cadence makes sense for the new user-initiated "Pay" path, but the same constant also drives the *automatic* path that waits for the counterparty to publish an updated private payment list.

Once `retryAttempts > size`, `paymentRequestsForPresentation` filters that request out permanently (attempts are only cleared when the request leaves `pendingRequests` or is marked presented). So an incoming request whose details take >28 s to publish will never auto-surface again for the rest of the process lifetime - and the automatic branch has no user-visible fallback on give-up.

Worth splitting into two schedules.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
```diff
@@ -338,6 +347,15 @@ fun EditInvoiceContent(
 
                         FillHeight()
 
+                        if (showPaymentRequestButton) {
+                            SecondaryButton(
```
jvsena42, line 353: **"Send Payment Request" is not gated on an amount.**

This `SecondaryButton` has no `enabled` condition, and `onClickPaymentRequest` passes `amountInputUiState.sats.toULong()`, which is `0` when the user hasn't typed an amount (perfectly normal in the receive/QR flow).

The user then picks a recipient and taps Send, and `PaykitPaymentRequestRepo.propose` throws `RequestUnavailable` (`draft.amountSats == 0uL`), surfacing the generic "Payment request is unavailable" toast with no hint that the amount was the problem.

Gate the button on `amountInputUiState.sats > 0`.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -0,0 +1,542 @@
...
+    contacts: ImmutableList<PubkyProfile>,
+    isCreating: Boolean,
+    onEditExpiration: () -> Unit,
+    onPaste: () -> String,
+    onSend: (PaykitPaymentRequestTarget) -> Unit,
+) {
+    var selectedTarget by remember { mutableStateOf<PaykitPaymentRequestTarget?>(null) }
+    var query by remember { mutableStateOf("") }
+
+    val recipients = remember(targets, contacts, query) {
```
jvsena42, line 286: **Paste/search can never resolve a non-contact pubky.**

`recipients` is built by intersecting `targets` with `contacts`, and `targets` themselves derive from `pubkyRepo.contacts`. The Paste affordance only writes into `query`, which then filters that same contact-derived list.

So pasting a valid pubky that isn't a saved contact just empties the list with no message, while the placeholder ("Enter pubky") and the prominent Paste button imply arbitrary keys are accepted.

Either resolve a pasted key into a target directly, or show an explicit empty-state explaining that the recipient must be a saved contact.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -64,34 +87,109 @@ data class PaykitPaymentRequest(
...
+    val paymentRequestHistory: StateFlow<List<PaykitPaymentRequest>> = _paymentRequestHistory.asStateFlow()
+    private val _eligibleTargets = MutableStateFlow<List<PaykitPaymentRequestTarget>>(emptyList())
+    val eligibleTargets: StateFlow<List<PaykitPaymentRequestTarget>> = _eligibleTargets.asStateFlow()
+    private val _isCreatingRequest = MutableStateFlow(false)
+    val isCreatingRequest: StateFlow<Boolean> = _isCreatingRequest.asStateFlow()
 
-    suspend fun refresh(): Result<Unit> {
+    @Volatile
+    private var activeIdentity: String? = null
+    private var presentedRequestIds = emptySet<PaykitPaymentRequestId>()
```
jvsena42, line 165: **`presentedRequestIds` is read across threads without synchronization.**

It is written on the IO dispatcher under `operationMutex` (`activate`, `markPresented`, `prunePresentedRequestIds`) but read from `automaticPendingRequests()`, which `AppViewModel` calls off the mutex on the main thread.

Unlike `activeIdentity` it is not `@Volatile`, so a stale read can hand back a request that was just marked presented and re-present it.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3672,13 +3816,77 @@ class AppViewModel @Inject constructor(
...
+                paymentRequestSheetTransitionJob = null
+                presentNextIncomingPaykitPaymentRequest()
+            }
+            paymentRequestSheetTransitionJob = job
+        } else {
+            viewModelScope.launch { presentNextIncomingPaykitPaymentRequest() }
+        }
+    }
+
+    suspend fun rejectIncomingPaymentRequest(request: PaykitPaymentRequest): Result<Unit> {
```
jvsena42: **Silent failure on Dismiss.**

This returns `Result.failure(OperationInProgress)` without toasting, and both UI call sites discard the `Result`.

If the user taps Pay on a request that is deferring (waiting for details) and then taps Dismiss on the same row, the button spins and nothing happens, with no explanation. Same for the early `return` in `openIncomingPaymentRequest`.

Either toast here or surface the failure at the call sites.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -111,67 +284,177 @@ class PaykitPaymentRequestRepo @Inject constructor(
...
+            .forEach { peer ->
+                val publicKey = PubkyPublicKeyFormat.normalized(peer.counterparty) ?: return@forEach
+                if (peer.counterpartyReceiverPath in PaykitReceiverPaths.supported) {
+                    linkedPaths.getOrPut(publicKey, ::mutableSetOf) += peer.counterpartyReceiverPath
+                }
+            }
+
+        return savedKeys.mapNotNull { publicKey ->
+            val linked = linkedPaths[publicKey] ?: return@mapNotNull null
+            val capable = runSuspendCatching { paykitSdkService.paymentRequestReceiverPaths(publicKey) }
```
jvsena42: **O(contacts) SDK round trips on every poll.**

`paymentRequestReceiverPaths(publicKey)` is called serially for each saved contact inside `synchronizeLocked`, i.e. on every 30/60/120 s refresh, and each call takes `PaykitSdkService.operationMutex` (and may fetch the peer's receiver marker).

With a sizeable contact list this stalls other Paykit operations behind the refresh loop. The results are effectively static - worth caching, or scoping to linked peers only.

jvsena42 (commented):
Review of the payment request flows. `just compile` and `just test` pass; the `ComposeUi` instrumented lane was run on a Pixel 9 AVD and all 15 new tests pass (after 864f008 fixed `CreatePaymentRequestScreenTest` querying a merged-semantics node). The 8 other failures in that lane reproduce on `master` and are unrelated.

Findings inline, roughly most-severe first. The `SheetHost` `dismissEnabled` work looks sound to me: programmatic `hide()` bypasses `confirmValueChange`, the disabled scrim sits under the sheet, and the inner NavHost `BackHandler`s still take priority over the swallowed sheet back handler.

app/src/main/java/to/bitkit/ui/components/Money.kt
```diff
@@ -25,17 +30,67 @@ import java.math.BigDecimal
...
+            text = it.withAccent(accentColor = Colors.White64),
             modifier = Modifier
                 .clickableAlpha(onClick = onClick)
                 .testTag("MoneyText")
         )
     }
 }
 
+@Composable
+fun MoneyStack(
```
jvsena42: **`MoneyStack` looks like dead code.**

Nothing in `app/src` references it. Per the repo rules unused code should be removed rather than landed.

app/src/main/res/values/strings.xml
```diff
@@ -1168,7 +1169,56 @@
...
+    <string name="wallet__payment_request_timestamp">%1$s at %2$s</string>
+    <string name="wallet__payment_request_waiting">Waiting for payment</string>
+    <string name="wallet__payment_request_waiting_for_recipient">Waiting for %1$s to pay</string>
+    <string name="wallet__payment_request_waiting_for_details">Waiting for updated private payment details. Bitkit will retry automatically.</string>
+    <string name="wallet__payment_requests">Payment Requests</string>
+    <string name="wallet__payment_requests_empty_description">You have not made any payments to providers and don’t have any payment requests yet.</string>
+    <string name="wallet__payment_requests_empty_headline"><![CDATA[No Payment\n<accent>History</accent>]]></string>
+    <string name="wallet__payment_requests_earlier">Earlier</string>
+    <string name="wallet__payment_requests_not_now">Not Now</string>
+    <string name="wallet__payment_requests_pending_count">%1$d pending payment requests</string>
```
jvsena42: This is a plain `%1$d` string, so it renders "1 pending payment requests". It is only used as a `stateDescription` on the bell so the impact is cosmetic, but it should be a `<plurals>`.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -548,10 +580,63 @@ class PaykitSdkService @Inject constructor(
         }
     }
 
-    suspend fun actionableReceivedPaymentRequests(): List<PaymentRequestRecord> {
+    suspend fun paymentRequests(): List<PaymentRequestRecord> {
+        isSetup.await()
+        return operationMutex.withLock {
+            handle().listPaymentRequests(
+                com.synonym.paykit.PaymentRequestFilter(
```
jvsena42: Inline fully-qualified name - `com.synonym.paykit.PaymentRequestFilter` should be imported, per the repo rule "ALWAYS add imports instead of inline fully-qualified names".

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
```diff
@@ -0,0 +1,616 @@
...
+@Composable
+internal fun PaymentRequestsContent(
+    modifier: Modifier = Modifier,
+    requests: ImmutableList<PaykitPaymentRequest>,
+    pending: ImmutableList<PaykitPaymentRequest>,
+    contacts: ImmutableList<PubkyProfile>,
+    canRequestPayment: Boolean,
+    onBack: () -> Unit,
+    onRequestPayment: () -> Unit,
+    onPay: (to.bitkit.repositories.PaykitPaymentRequestId) -> Unit,
```
jvsena42: Inline fully-qualified name: `to.bitkit.repositories.PaykitPaymentRequestId` is already imported at line 51, so this (and the same parameter on `PaymentRequestsList`) can just use the simple name.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ben-kaufman: Fixed in 9c900c9f4.  now uses  as the  start destination, so opening Request Payment directly no longer leaves the QR screen behind it in the back stack.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: No indefinite cache is added here because receiver markers/paths are deliberately rediscovered: a peer may publish Payment Request support after the contact was saved. The implementation already skips unlinked contacts before the SDK lookup, and after the bounded discovery burst it uses the normal low-frequency cadence, preserving late discovery without permanent aggressive polling.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
ben-kaufman: Fixed in 9c900c9f4. The Send Payment Request button is disabled while the amount is zero, with an instrumentation assertion covering the zero-amount state.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: Fixed in 9c900c9f4. The immutable  snapshot is now , matching the existing cross-thread identity snapshot pattern and making completed writes visible to main-thread presentation reads.

app/src/main/java/to/bitkit/ui/components/Money.kt
ben-kaufman: Fixed in 9c900c9f4. The unused  component and its unused import were removed.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 9c900c9f4 (and the equivalent iOS fix in f76227e9). Both paths use the short 14×2s initial burst, but only a manually requested presentation gives up into the visible queue. Automatic presentation continues every 120s and is never permanently filtered out. Both behaviors have regression tests.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
ben-kaufman: Fixed in 9c900c9f4. Both signatures now use the imported  name.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 9c900c9f4. Clearing an interrupted contact-payment context now defers the still-pending incoming request instead of leaving  wedged. Manual resolution uses the bounded burst and then returns the request to the queue; automatic requests continue on the low-frequency schedule. Regression coverage includes the failed manual-presentation path.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
ben-kaufman: Fixed in 9c900c9f4 at the repository boundary. Accept/reject now queues the lifecycle update and processes the durable private outbox before removing the request from observable pending state, so composition cannot disappear while the delivery attempt is still running. Any transport failure remains in Paykit's durable outbox for later retry.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 9c900c9f4. Both the Pay/open early return and Dismiss/reject conflict now show the existing  toast instead of silently discarding the failure.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Fixed in 9c900c9f4.  is imported and the inline fully-qualified reference is gone.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
ben-kaufman: This is intentional for the private-only MVP. The field can search or paste an eligible saved contact's pubky, but an arbitrary unsaved key cannot have an established private/Noise link and must not fall back to public payment details. The separate arbitrary-key “Pay To” behavior was explicitly left out of scope, so no resolution path is added here.

app/src/main/res/values/strings.xml
ben-kaufman: Fixed in 9c900c9f4. The accessibility state description now uses Android plurals, including the singular “1 pending payment request” form.

ovitrif (commented):
The QR no longer sits under Request Payment, but in-sheet back still does not return to Payment Requests because that screen still only calls popBackStack().

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ovitrif: In-sheet back from list-driven Request Payment still leaves the sheet open: Payment Request is now the first screen, and onBack still only calls popBackStack(), which is a no-op at the root. Send dismisses the sheet when popBackStack() fails, so the back arrow actually leaves the flow:

```kotlin
onBack = {
    if (!navController.popBackStack()) {
        appViewModel.hideSheet()
    }
}
```

Could we do the same for PaymentRequestDetails?

ben-kaufman: Fixed in d82045ff7. The root Payment Request details back action now hides the sheet when `popBackStack()` returns false, so list-driven Request Payment returns to Payment Requests. App compile, Android-test compile, unit tests, and detekt pass.

ovitrif (dismissed):
List-driven Request Payment now starts on the form, and in-sheet back closes the sheet so you return to Payment Requests.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ovitrif: In-sheet back from list-driven Request Payment now hides the sheet when popBackStack() fails, so that path returns to Payment Requests.

piotr-iohk:
Flagging while QA’ing with the staging public-pay pubkys — not a prod issue (Paykit UI is still hidden).

Enabling payments with contacts fail-closes with **Unknown error** if a saved contact’s receiver marker doesn’t parse (here: missing `noise_public_key`). I hit it on:

- `pubky1jy146y751hkgqk9p69ds5cqf461aj5x7c8ptbca5rw6m9asotso`
- `pubkynekaxdyt5ktoqyufd7cgug6qppa7999hm46wfqdzm6jqnr8tj16o`

In principle the same thing could happen with any broken / outdated profile marker, not only these fixtures. Recording attached. Same surface on iOS after #676 (toast there is **Private Paykit is not available**).

Steps:
1. Enable Paykit UI. Create a profile. Leave payments with contacts off.
2. Add either (or both) staging pubkys above as contacts. They do not need to be live.
3. Enable payments with contacts.
4. Toast is **Unknown error** and the toggle stays off.


https://github.com/user-attachments/assets/072bfa60-3b7b-446e-af7f-0d8a83f49714

[logs.zip](https://github.com/user-attachments/files/31455742/logs.zip)

piotr-iohk:
Flagging from QA: paying a payment request that’s larger than the wallet balance.

Same setup, two directions, both recordings attached:

- **iOS → Android (this PR):** incoming sheet, **Pay** → toast **Insufficient Spending Balance** (e.g. ₿ 78,244 more needed to pay this Lightning invoice) and a red-bordered card. **Dismiss** then fails with **Payment request operation is already in progress**. Toast / error repeats; request stays on the sheet and is hard to dismiss.



https://github.com/user-attachments/assets/2f1f0650-af83-49a6-8e19-d61b0d1acb7e




- **Android → iOS (#676, already merged):** **Pay** → **Please try again. Coin selection failed.** Dismiss is fine; the copy is just the wrong error.


https://github.com/user-attachments/assets/cc0fe231-cee6-44f2-a756-c19344abee09



Steps:
1. Two live wallets with payments-with-contacts on. Put ~88k on the payer (split savings + spending is fine).
2. Other wallet requests more than that total (100k Android-as-payer, 90k iOS-as-payer).
3. On the incoming request, tap **Pay**, then try **Dismiss**.

Expected: don’t enter confirm if the amount is above spendable. On Android, **Dismiss** should still work after a failed Pay (no **already in progress** loop).

ben-kaufman:
Addressed both QA findings in signed commit `5d39cd96a`:

- A malformed remote receiver marker is still rejected and cannot be used, but its per-contact inspection error no longer aborts global private endpoint publication for valid contacts. The regression test now covers the immediate-publication path used when enabling Contact Payments.
- An unaffordable request now releases its manual presentation state after showing the existing insufficient-balance message. It remains pending and dismissible, is not accepted or rejected, and does not consume a Private Payment List. The regression test verifies Dismiss remains actionable and no retry starts from the old context.

The matching iOS fixes are isolated in synonymdev/bitkit-ios#684: malformed remote markers no longer block other contacts, and incoming request amounts are used during initial LN/on-chain balance validation so the standard insufficient Spending/Savings error appears before coin selection.

Local verification: 193 focused Android tests passed; detekt completed successfully with only unrelated pre-existing findings.

piotr-iohk:
Retested after [5d39cd9](https://github.com/synonymdev/bitkit-android/commit/5d39cd96a).

**Enable** — still fails. ❌ 

Toggle **Enable payments with contacts** → **Unknown error**, toggle does not stay on. Same staging contact as before (`pubky1jy146y751hkgqk9p69ds5cqf461aj5x7c8ptbca5rw6m9asotso`). Recording + logs attached.

This is not the `noise_public_key` inspect-fail anymore. Logs are `PrivateUnavailable`. In this run I turned the toggle off first; disable cleanup already failed with that error, then re-enable with cleanup still pending fail-closes again. Settings still maps it to **Unknown error**.

Same user-facing miss on iOS [#684](https://github.com/synonymdev/bitkit-ios/pull/684) (toast there is **Private Paykit is not available**).

Steps:
1. Paykit UI on. Profile exists. Staging pubky above saved as a contact (need not be live).
2. Payments with contacts off (or turn it off first).
3. Enable payments with contacts.
4. Toast is **Unknown error**; toggle stays off.

https://github.com/user-attachments/assets/a51ace48-b8e4-4192-bb0c-e22291329e51

[logs-android.zip](https://github.com/user-attachments/files/31508658/logs-android.zip)



**Over-balance** — looks good on Android. ✅ 

Incoming request larger than the wallet: one **Insufficient Savings / Spending** toast, request stays pending and dismissible. No **already in progress** loop.

iOS [#684](https://github.com/synonymdev/bitkit-ios/pull/684) still repeats that toast (see that PR).

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -80,7 +98,7 @@ fun ReceiveSheet(
         ) {
             NavHost(
                 navController = navController,
-                startDestination = ReceiveRoute.QR,
+                startDestination = startRoute,
```
jvsena42: **Back stack loses `ReceiveRoute.QR` when the sheet opens at a non-QR route.**

Replacing `LaunchedEffect(startRoute) { navController.navigateTo(startRoute) }` with `startDestination = startRoute` means `QR` is no longer the root of the inner NavHost stack.

`ContentView.kt:936` (FundingScreen -> "Fund") opens `Sheet.Receive(route = ReceiveRoute.Amount)`. In that flow:

1. `ReceiveAmountScreen`'s `onBack = { navController.popBackStack() }` now returns `false` — the back arrow is a dead control.
2. `ReceiveRoute.Confirm`'s `onContinue` does `navigateTo(QR) { popUpTo(QR) { inclusive = true } }`. With `QR` absent, the `popUpTo` is a no-op and the stack becomes `[Amount, Confirm, QR]`, so back from the QR screen lands on Confirm instead of closing the sheet.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -102,6 +120,52 @@ fun ReceiveSheet(
...
+                composableWithDefaultTransitions<ReceiveRoute.PaymentRequestRecipient> {
+                    PaymentRequestRecipientScreen(
+                        appViewModel = appViewModel,
+                        draft = paymentRequestDraft,
+                        onEditExpiration = {
+                            navController.navigateTo(ReceiveRoute.PaymentRequestExpiration)
+                        },
+                        onSent = {
+                            createdPaymentRequest = it
+                            navController.navigateTo(ReceiveRoute.PaymentRequestSent)
```
jvsena42: **Duplicate payment request via system back.**

`onSent` navigates to `PaymentRequestSent` without popping `PaymentRequestRecipient`, and `navigateTo` only sets `launchSingleTop`. The Sent screen has no `BackHandler` or back arrow, so system back returns to the recipient picker with `selectedTarget` still set and `isCreating == false` — "Send Request" is enabled and proposes a second request for the same draft.

Suggest `popUpTo(ReceiveRoute.PaymentRequestRecipient) { inclusive = true }`.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
```diff
@@ -765,7 +781,6 @@ class PrivatePaykitRepo @Inject constructor(
             val linkableReceiverPaths = receiverPathSelection.linkableReceiverPaths
             val publicationReceiverPaths = receiverPathSelection.publishableReceiverPaths
             receiverPathSelection.error?.let {
-                firstError = firstError ?: it
                 logPrivateReceiverPathSelectionFailure(publicKey, reason, it)
```
jvsena42, line 784: **Dropping `firstError = firstError ?: it` silently swallows total publication failure.**

This hunk removes the `firstError` assignment, so a `receiverPathSelection.error` is now only logged, never propagated.

If selection fails for every contact, `publicationReceiverPaths` is empty, so `preparation.updates.isEmpty()` and the `if (requireImmediatePublication) preparation.firstError?.let { throw it }` guard sees `null` and returns success. The `requireImmediatePublication = true` callers — `ContactPaymentSettingsRepo.kt:49`, `:79`, `:148`, the "share private payment details" toggles — then report success to the user while nothing was published.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -111,67 +286,177 @@ class PaykitPaymentRequestRepo @Inject constructor(
...
+        _paymentRequestHistory.update { history }
+        _eligibleTargets.update { targets }
+        prunePresentedRequestIds(incoming)
         scheduleExpirationLocked()
     }
 
+    private fun isCurrentState(generation: Long, expectedIdentity: String?): Boolean =
+        stateGeneration.get() == generation && PubkyPublicKeyFormat.matches(activeIdentity, expectedIdentity)
+
+    private suspend fun eligibleTargets(
```
jvsena42: **`eligibleTargets` now issues remote lookups on every `refresh()`.**

It runs from `synchronizeLocked`, and per saved contact does `paykitReceiverPaths(publicKey)` plus a `paykitReceiverMarker(publicKey, path)` remote read. `refresh()` is driven by `startInitialPaykitPaymentRequestPolling` (15 runs at 2s intervals, re-armed on every initial link burst) and then by the 30/60/120s poll loop.

With N contacts that is roughly 15 x N x 2 remote reads inside 30s, on a path that previously only listed local records. Worth caching per identity/generation or computing it off the refresh path.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -757,12 +846,38 @@ class AppViewModel @Inject constructor(
...
+                )
+                paymentRequestPresentationGeneration++
+                requestedPaymentRequestId = null
+                showSheet(Sheet.PaymentRequests)
+                viewModelScope.launch {
+                    paykitPaymentRequestRepo.markPresented(request)
+                }
+                return
+            } else {
+                PAYKIT_PAYMENT_REQUEST_REFRESH_INTERVALS.last()
```
jvsena42, line 897: **Automatic presentations now retry forever.**

The give-up branch above only applies when `requestedPaymentRequestId == request.id`. For a non-requested (automatic) presentation, once `PAYKIT_PAYMENT_REQUEST_PRESENTATION_RETRY_DELAYS` (now `List(14) { 2.seconds }`) is exhausted, `attempt` saturates at 14 via `coerceAtMost` and `retryDelay` falls back to `PAYKIT_PAYMENT_REQUEST_REFRESH_INTERVALS.last()` (120s) indefinitely.

An incoming request whose private payment details never resolve calls `beginPaymentRequest` every 2s for ~28s, then every 2 minutes for the request's whole lifetime (up to 30 days). The previous code stopped after 4 attempts.

Related: the requested-branch `return` at line 869 leaves `paymentRequestPresentationRetryAttempts[request.id]` at its saturated value. A later "Pay" tap on the same request therefore gives up on its first failure — no retry window, no "waiting for details" toast, and the log misreports the attempt count.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestPresentationStore.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val keychain: Keychain,
+) {
+    private val mutex = Mutex()
+
+    @Serializable
+    private data class State(
+        val idsByIdentity: Map<String, List<PaykitPaymentRequestId>> = emptyMap(),
+    )
+
+    fun load(identity: String): Set<PaykitPaymentRequestId> {
```
jvsena42, line 30: **Store can never repair a corrupt entry.**

Both `load` (line 28) and `save` (line 35) call `Json.decodeFromString<State>(...)` on the raw keychain value with no fallback. If that entry is ever corrupt, or written by a future/older schema, `load` throws (callers swallow it and see an empty set) and every subsequent `save` throws too — so the bad value is never overwritten.

Effect: presented-request ids stop persisting permanently, and already-surfaced payment requests are auto-presented again on every app start. `save` should treat a parse failure as `State()` and overwrite.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
```diff
@@ -0,0 +1,616 @@
...
+                    .fillMaxWidth()
+                    .background(Colors.Gray5)
+                    .padding(16.dp),
+            ) {
+                SecondaryButton(
+                    text = stringResource(R.string.wallet__payment_request_dismiss),
+                    onClick = {
+                        if (isRejecting || onReject == null) return@SecondaryButton
+                        isRejecting = true
+                        scope.launch {
```
jvsena42: **Reject runs in the card's `rememberCoroutineScope()`.**

`scope` is `rememberCoroutineScope()` from `PaymentRequestCard` (line 456), so it is tied to the card's composition. `onReject` is a `suspend () -> Result<Unit>`; closing the sheet mid-flight cancels it, which can leave the request rejected in the SDK but still shown as pending (and the outbound message unflushed) until the next refresh.

`createPaymentRequest` already runs this kind of work in `viewModelScope`; same is warranted here.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
jvsena42, line 286: **Pasting an unknown pubky gives a blank screen with no explanation.**

The field is labelled "RECIPIENT" with placeholder "Enter pubky" and a Paste button, but `query` only *filters* `targets`, which are derived from saved, linked contacts. Pasting a valid pubky that is not already a saved and linked contact yields an empty `recipients` list with no message, so the user cannot tell why the recipient is unusable.

Worth an empty-state explaining that the recipient must be a saved contact with private payment details linked.

app/src/main/java/to/bitkit/ui/components/Money.kt
```diff
@@ -25,17 +29,46 @@ import java.math.BigDecimal
 fun MoneyDisplay(
     sats: Long,
     onClick: (() -> Unit)? = null,
+    showSymbol: Boolean? = null,
```
jvsena42: **Dead parameter, and it is not part of the remember key.**

`showSymbol` has no callers — both existing call sites omit it.

If it is kept, line 34 is a problem: `if (showSymbol == null) rememberMoneyText(sats) else rememberMoneyText(sats, showSymbol = showSymbol)` creates two distinct composable call sites, so toggling it at runtime discards remembered state. More importantly `rememberMoneyText` keys on `remember(currencies, sats, unit)` — `showSymbol` is not a key, so a future caller that flips it for the same sats/unit gets the stale string.

Either drop the parameter or add `showSymbol` to the remember key.

app/src/main/res/values/strings.xml
```diff
@@ -1170,7 +1171,59 @@
...
+    <string name="wallet__payment_request_paste">Paste</string>
+    <string name="wallet__payment_request_queued_description">Your payment request is queued and will send automatically</string>
+    <string name="wallet__payment_request_recipient">RECIPIENT</string>
+    <string name="wallet__payment_request_request_payment">Request Payment</string>
+    <string name="wallet__payment_request_send">Send Payment Request</string>
+    <string name="wallet__payment_request_send_request">Send Request</string>
+    <string name="wallet__payment_request_sent_description">You have sent a payment request</string>
+    <string name="wallet__payment_request_sent_headline"><![CDATA[Payment <accent>Requested</accent>]]></string>
+    <string name="wallet__payment_request_sent_title">Sent</string>
+    <string name="wallet__payment_request_sending">Queued for delivery</string>
```
jvsena42: **Three new entries are out of alphabetical order** (CLAUDE.md: "ALWAYS add new localizable string resources in alphabetical order in `strings.xml`"):

- `wallet__payment_request_sending` (this line) sorts before `wallet__payment_request_sent_*`, not after
- `wallet__payment_request_waiting_for_details` (line 1209) sorts before `wallet__payment_request_waiting_for_recipient`
- `wallet__payment_requests_earlier` (line 1213) sorts before `wallet__payment_requests_empty_*`

jvsena42 (commented):
### Code review

Reviewed `git diff master...HEAD` at 56ce9908c (merge-base 6ae3cb7a8) — 34 files, ~3.8k added lines. 10 findings inline, roughly in severity order: the `ReceiveSheet` start-destination change and the dropped `firstError` in `PrivatePaykitRepo` look like the two most likely to bite in practice.

Checked and looked correct: the generation/identity guards in `PaykitPaymentRequestRepo` (`activate`/`refresh`/`synchronizeLocked`/`isCurrentState`), `creationMutex.tryLock` with `finally` unlock in `propose`, expiration rescheduling, the `hideSheet` re-entrancy guard, `SheetHost`'s `confirmValueChange` + `rememberUpdatedState` pairing, the scrim's pointer-consuming modifier, the `stopped`/recursion loop in `presentNextIncomingPaykitPaymentRequest`, and the sats/BTC `BigDecimal` conversions.

One judgement call I did not flag as a defect: the string-literal match in `canDeferStaleSession` is brittle against SDK message changes.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ben-kaufman: Fixed in 531578e4a. Receive now keeps QR as the graph root for ordinary deep routes and then navigates to the requested screen. List-driven Payment Request Details intentionally remains its own root and dismisses the sheet when back cannot pop. Instrumentation now exercises the real SheetHost and system back for Amount to QR and root Payment Request Details to dismissal.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ben-kaufman: Fixed in 531578e4a. Navigation to Sent now clears the previous inner Receive graph before installing Payment Request Sent, so system back cannot return to the selected recipient or submit a duplicate request. The regression test presses system back on the real sheet and verifies the sheet dismisses without restoring the form.

app/src/main/java/to/bitkit/repositories/PrivatePaykitRepo.kt
ben-kaufman: Fixed in 531578e4a. Receiver-path selection errors are retained separately and become the publication error when no contact produces a valid update, so total immediate-publication failure is propagated. A bad contact still does not block valid contacts from publishing. Tests cover both total failure and mixed valid/invalid contacts.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: Fixed in 531578e4a. Incoming request refresh no longer performs recipient capability discovery. Target discovery now has its own serialized path, caches complete identity/link contexts, retries unresolved capabilities during bounded bursts or forced refreshes, preserves known targets across transient reads, and bounds capability reads. Proposal-time validation checks only the selected saved contact, so unrelated stalled contacts cannot delay creation or incoming processing.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: The manual retry bug is fixed in 531578e4a: tapping Pay clears any saturated automatic-attempt counter, so the request receives the full 14 by 2-second window and the waiting feedback. The later 120-second automatic retry is intentional while the request is pending and unexpired: private details may be published later, and the request must then auto-surface without falling back to public details. Retry state is removed when the request leaves pending state. Both manual reset and automatic continuation have regression coverage.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestPresentationStore.kt
ben-kaufman: Fixed in 531578e4a. Decode failure now returns an empty State for both load and save, letting the next save overwrite and repair the corrupt keychain entry. The warning is generic so corrupt identity/request payload content is not logged. Added a regression test for repair after corruption.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
ben-kaufman: Fixed in 531578e4a. Rejection now launches in AppViewModel viewModelScope, with lifecycle-stable rejecting IDs, duplicate suppression, and try/finally cleanup. The card observes that StateFlow to disable both actions and show progress, so removing the composition cannot cancel delivery. Repository/ViewModel and UI regressions cover this path.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
ben-kaufman: Addressed in 531578e4a. This remains intentionally private-only: an arbitrary unsaved pubky has no established Noise link and must not fall back to public payment details. The UI now distinguishes no eligible private contacts from a search with no matching saved private contact, instead of showing a blank list. Instrumentation covers the no-match copy.

app/src/main/java/to/bitkit/ui/components/Money.kt
ben-kaufman: Fixed in 531578e4a by removing the unused showSymbol parameter and its alternate remembered call path.

app/src/main/res/values/strings.xml
ben-kaufman: Fixed in 531578e4a. The payment-request strings are now in exact alphabetical order.

ben-kaufman:
Addressed the remaining enable/disable QA path in signed, rebased commit 531578e4a.

The private setting now reflects the user choice independently of transient transport cleanup or immediate Noise availability:

- Disable commits private sharing off and records failed remote cleanup as durable deferred work instead of restoring the toggle.
- Re-enable supersedes pending removal and prepares saved contacts opportunistically; it no longer requires a live peer or immediate private endpoint publication to keep the toggle enabled.
- Recovery persists the cleanup marker before remote removal and local pruning, so a failure between those stages remains retryable. The cached-publication fallback still recovers older states where the marker was not stored.
- Malformed remote markers remain unusable, and there is still no private-to-public payment fallback.

Regression coverage includes deferred disable cleanup, re-enable with pending cleanup, recovery without a marker, failure after remote cleanup but before local pruning, and total versus partial marker-selection failure. The over-balance behavior that passed the retest is unchanged.

Post-rebase verification: full testDevDebugUnitTest, detekt, and compileDevDebugAndroidTestKotlin pass. The stack is rebased onto current master and all 13 PR commits are signed.

piotr-iohk (approved):
Retested. LGTM.

jvsena42 (approved):
tAck

Reviewed the code manually and tested with journeys

Issue #1126, opened 2026-07-31
feat: deep link the late transfer screens


guzino:
Refs #659
Refs #1119

#1119 denies six transfer destinations because their generated links do not reconstruct the state the screens read. This issue tracks direct support for them.

## Denied today

| Route | State it reads |
| - | - |
| `SavingsProgress` | the active transfer, and it can act on default or retained channel state |
| `SettingUp` | the paid order the funding step created |
| `SpendingAdvanced` | the order being adjusted, opened from `SpendingConfirm` or `SpendingHwSign` |
| `SpendingConfirm` | the composed order and fee estimate |
| `SpendingHwSign(deviceId)` | the composed transaction plus a paired device |
| `SpendingHwSigned` | the signed transaction and its broadcast result |

All six read `transferViewModel`, which is scoped to the activity rather than to the route. A fresh link therefore renders empty state, returns home, or acts on whatever the previous flow left behind.

## Setup contract

Each screen needs a per-flow contract that runs before navigation:

1. Prepare the state the preceding user and app actions would have produced.
2. Perform the server-side preparation the step depends on, using the `lsp` skill for Blocktank orders, deposits and channel opens.
3. Pass stable identifiers as route parameters rather than relying on retained view model state.
4. Reload and validate that data before navigating, and refuse the link when validation fails.

`SpendingHwSign` already carries `deviceId`, so it is the closest to the target shape and the smallest first slice.

## Done when

- Each of the six is reachable from a cold start and renders the state its identifiers name.
- A link whose identifiers no longer resolve is refused with the existing `Unhandled screen deeplink` warning rather than rendering empty.
- `SavingsProgress` cannot execute against retained channel state.
- The routes are removed from `DENIED` in `ScreenDeepLinks` and from `SENSITIVE_ROUTES` in `ScreenDeepLinksTest`, and `docs/deeplinks.md` moves them out of the excluded table.

PR #1176, opened 2026-08-22
feat: deep link spending hw sign

guzino:
Refs #1126
Refs #1119

This PR opens the hardware-wallet transfer Sign screen from `bitkit://screen/spending-hw-sign/{walletId}/{orderId}`.

### Description

#1119 left six transfer destinations `InternalOnly` because they read activity-scoped `TransferViewModel` state. `SpendingHwSign` was the closest: it already took a wallet id (the issue still says `deviceId`) and bounced home when `spendingUiState.order` was null. A generated link therefore could not reconstruct the Blocktank order.

`ContentView` now parses that URI and calls `prepareSpendingHwSign` before `navController.handleDeepLink`. Unknown wallet or missing order is refused with the existing `Unhandled screen deeplink` warning and does not navigate. A matching in-memory order is reused; otherwise `blocktankRepo.getOrder(orderId, refresh = true)` loads it and `adoptSpendingOrder` writes the same state `onOrderCreated` already wrote, without emitting `TransferEffect.OnOrderCreated`. The pending URI is consumed after that suspend, so the `LaunchedEffect` is not cancelled mid-fetch.

`OnOrderCreated` now carries `orderId`. Amount → Sign navigates `Routes.SpendingHwSign(walletId, orderId)` from the effect. The dest reads both route args and matches `state.order` to `orderId`.

- Promotes `Routes.SpendingHwSign` to `DeepLinkable` with path `bitkit://screen/spending-hw-sign/{walletId}/{orderId}`.
- Parses that path in `ScreenDeepLinks.spendingHwSignLink` via `kebabId(Routes.SpendingHwSign::class)`.
- Keeps `SavingsProgress`, `SettingUp`, `SpendingAdvanced`, `SpendingConfirm`, and `SpendingHwSigned` as `InternalOnly`. The rest of #1126 stays a follow-up.

### Preview

N/A

### QA Notes

Dev mode is on by default on debug builds (Settings ▸ Advanced ▸ Dev Settings). The app must be past onboarding. The Sign path needs a paired hardware wallet and a live Blocktank order id.

#### Manual Tests

- [ ] **1.** Hardware Wallet detail → Transfer To Spending → Amount → Continue → Sign: lands on Sign With Your Device with the created order.
- [ ] **2.** From that Sign screen, `adb shell am start -a android.intent.action.VIEW -d "bitkit://screen/spending-hw-sign/<walletId>/<orderId>" to.bitkit.dev` → Sign opens with the same order.
- [ ] **3.** Cold start the same URI → Sign opens with that order, not Home.
- [ ] **4.** Same path with an unknown wallet id or a missing order → screen unchanged, logcat carries `Unhandled screen deeplink`.
- [ ] **5.** `regression:` `bitkit://screen/spending-amount-hw/<walletId>` → Amount still opens.

#### Automated Checks

- Unit tests added in `ScreenDeepLinksTest.kt`: path pattern, wallet and order segments, missing order id.
- Unit tests added in `TransferViewModelTest.kt`: known wallet loads the named order, unknown wallet and missing order are refused, in-memory order is reused without a second fetch.


app/src/main/java/to/bitkit/ui/screens/transfer/hardware/SpendingHwSignScreen.kt
```diff
@@ -50,7 +51,7 @@ fun SpendingHwSignScreen(
 ) {
     val state by viewModel.spendingUiState.collectAsStateWithLifecycle()
 
-    val order = state.order ?: run {
+    val order = state.order?.takeIf { it.id == orderId } ?: run {
```
jvsena42, line 54: **Blocking — this guard breaks the existing Advanced flow.**

`orderId` is a route arg, frozen when `SpendingHwSign` was pushed. But `onAdvancedClick` pushes `Routes.SpendingAdvanced`, and `onSpendingAdvancedContinue` calls `blocktankRepo.createOrder(...)` and stores a **new order with a new id** as `spendingUiState.order` (the old one moves to `defaultOrder`). `SpendingAdvancedScreen`'s `onOrderCreated` is `navController.popBackStack()` — which lands back on `SpendingHwSign` still carrying the *old* id. `takeIf` yields null and the composable calls `onCloseClick()` → `navigateToHome()`.

Reproduced on device: HW detail → Transfer To Spending → 25% → Continue → Sign → Advanced → MAX → Continue lands on the **wallet home screen**, and the freshly created order is stranded. Logcat:

```
INFO [BlocktankRepo.kt:285] Buying channel with lspBalanceSat: '341987', ...
DEBUG [BlocktankRepo.kt:222] Orders refreshed: 7 orders, 0 cjit entries, 2 paid orders
```

and `blocktank.db` then holds `3c10c207-…  Created  341987  110227` with nothing pointing at it.

Suggest accepting `defaultOrder?.id == orderId` as a match too, or applying the id check only on first entry (deep-link admission) rather than on every recomposition.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -581,6 +581,31 @@ class TransferViewModel @Inject constructor(
...
+    suspend fun prepareSpendingHwSign(walletId: String, orderId: String): Boolean {
+        if (walletId.isBlank() || orderId.isBlank()) return false
+        if (hwWalletRepo.wallets.value.none { it.id == walletId }) {
+            Logger.warn("Refused spending hw sign deeplink, unknown wallet '$walletId'", context = TAG)
+            return false
+        }
+        val current = _spendingUiState.value.order
+        if (current?.id == orderId) return true
+
+        val order = blocktankRepo.getOrder(orderId, refresh = true).getOrNull()
```
jvsena42, line 597: Worth confirming this is the intent: `BlocktankRepo.getOrder` refreshes and then searches `_blocktankState.value.orders`, i.e. only orders **this install already tracks locally**. An order that exists on the LSP but was never created on this device is always refused.

Verified on device — I created a valid order via the Blocktank API with this node's `clientNodeId`, then deep-linked to it:

```
WARN [TransferViewModel.kt:599] Refused spending hw sign deeplink, missing order 'c464a24c-…'
```

while a locally-created order id opened the sign screen fine. That's the right behaviour if the link is only ever meant to resume a transfer started on this device; it does mean a link handed over from another device or from support tooling can never resolve. Fine to leave as-is — just flagging it so the constraint is deliberate.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -581,6 +581,31 @@ class TransferViewModel @Inject constructor(
 
     private suspend fun onOrderCreated(order: IBtOrder) {
         settingsStore.update { it.copy(lightningSetupStep = 0) }
+        adoptSpendingOrder(order)
+        setTransferEffect(TransferEffect.OnOrderCreated(order.id))
+    }
+
+    suspend fun prepareSpendingHwSign(walletId: String, orderId: String): Boolean {
```
jvsena42, line 588: `prepareSpendingHwSign` → `adoptSpendingOrder` clears `pendingHwFundingBroadcast` and `hasPendingHwBroadcast`. If the user has already signed a HW funding tx for order X that failed to broadcast, `onTransferToSpendingHwConfirm` relies on `pendingHwFundingBroadcast?.matches(...)` to retry without re-prompting the device — a deep link naming a different order Y silently discards that in-memory signed transaction, making it unrecoverable.

Dev-mode only, so low severity, but cheap to guard: refuse the link (or skip the clobber) while `hasPendingHwBroadcast` is set.

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -315,13 +315,22 @@ fun ContentView(
...
         SheetDeepLinks.sheetFor(uri)?.let {
             appViewModel.showSheet(it)
+            appViewModel.consumeScreenDeepLink()
             return@LaunchedEffect
         }
 
+        ScreenDeepLinks.spendingHwSignLink(uri)?.let { link ->
+            val prepared = transferViewModel.prepareSpendingHwSign(link.walletId, link.orderId)
+            if (!prepared) {
+                Logger.warn("Unhandled screen deeplink '$uri'", context = "ContentView")
```
jvsena42, line 328: This warn duplicates the specific reason already logged inside `prepareSpendingHwSign`, and reuses the exact wording of the generic `!handled` warn a few lines below. Observed on device — one refused link produces two WARN lines, the second of which says "Unhandled" when the link was in fact recognized and deliberately refused:

```
WARN [TransferViewModel.kt:591] Refused spending hw sign deeplink, unknown wallet 'foo'
WARN [ContentView.kt:328]       Unhandled screen deeplink 'bitkit://screen/spending-hw-sign/foo/bar'
```

Per the repo rule (*NEVER duplicate error logging in `.onFailure {}` if the called method already logs the same error internally*), drop this line or reword it so it doesn't collide with the generic one.

Separately, `consumeScreenDeepLink()` now appears three times in this effect. It can't be hoisted to the top — that's what f405cd3d fixed, since the effect is keyed on `pendingScreenDeepLink` and consuming early cancels the coroutine mid-`prepareSpendingHwSign` — but the three calls can collapse into a single one at the *end* by turning the two early returns into an if/else chain.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
```diff
@@ -38,6 +38,19 @@ object ScreenDeepLinks {
     fun isScreenDeepLink(uri: Uri): Boolean =
         uri.scheme?.lowercase() == SCHEME && uri.host?.lowercase() == HOST
 
+    fun spendingHwSignLink(uri: Uri): SpendingHwSignLink? {
```
jvsena42, line 41: Unlike `linksFor`/`sheetFor`, this isn't routed through `ScreenDeepLinkRuntime`/`isEnabled`, so it parses and returns a link in release builds too. Currently harmless because `AppViewModel.processDeeplink` gates queueing on `ScreenDeepLinks.shouldQueue(...)`, but that single call site is the only thing stopping a release build from mutating live transfer state from a dev-only URI. An `if (!isEnabled) return null` here would make it fail safe.

jvsena42 (commented):
Reviewed and validated on a regtest emulator against the deterministic Trezor Bridge emulator from `bitkit-docker` (T2T1, seed `all all ...`, paired as `BITKIT TEST TREZOR` with 26,890,661 sats).

**The deep link itself works.** Both branches of `prepareSpendingHwSign` were exercised end to end:
- fresh order id → `refreshOrders` runs, order is adopted, `isAdvanced` resets (Advanced button flips back from "Use Defaults"), sign screen opens with the right amounts;
- already-current order id → short-circuits on `current?.id == orderId` and opens directly;
- unknown wallet and unknown order are both refused, no navigation.

One blocking issue and three smaller ones inline.

ovitrif:
General note from briefly looking over review comments: it may not have been specified in the issues or past comments but this screen deeplinks work is more intended to aid in development with ai agents, for example: it could add the possibility to open a specific screen and continue from there. Maybe it should not be a requirement that the rest of the flow(s) would work correctly, or even if it tries to, it should only add it in logic specific to the handler of that screen deeplink; while the impact on production code should try to stay limited to parametrizing the screen inputs (strings, bools, etc, whatever is needed as starting state / to mutate UI)

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
```diff
@@ -330,6 +330,58 @@ class TransferViewModelTest : BaseUnitTest() {
         verify(hwWalletRepo, never()).signFunding(any(), any())
     }
 
+    @Test
+    fun `prepareSpendingHwSign loads the named order when the wallet is known`() = test {
+        val order = previewBtOrder()
+        whenever(hwWalletRepo.wallets)
+            .thenReturn(MutableStateFlow(persistentListOf(hwWallet(HARDWARE_WALLET_ID, connected = true))))
+        whenever { blocktankRepo.getOrder(eq(order.id), eq(true)) }.thenReturn(Result.success(order))
```
jvsena42, line 338: **⚪ Use whenever(mock.call()) for suspend stubs inside test {}**

AGENTS.md: inside `test {}` blocks suspend stubs use `whenever(blocktankRepo.getOrder(eq(order.id), eq(true)))`, not `whenever { }`. Same at lines 363 and 376. (The file has the same pattern pre-existing elsewhere - flagging only the new lines.)

jvsena42 (commented):
Reviewed this as a key-management change, and the trust boundary holds up: the link carries only `{walletId, orderId}`, both validated against local state (`hwWalletRepo.wallets`, and `getOrder` searching this client's own `_blocktankState.orders`). It cannot supply a PSBT, address, amount, fee rate or derivation path, the Trezor still confirms on-device, and `shouldQueue = isEnabled && devMode` keeps the whole surface out of release.

One substantive item, posted as a reply on the existing thread since the first half was already raised there.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: **🔵 Refuse the deep link while a hardware sign or broadcast is in flight**

`adoptSpendingOrder` sets `pendingHwFundingBroadcast = null` and swaps `order`, so a link for another of the user's orders discards a signed-but-unbroadcast funding tx (the user must re-sign on the device) and, mid-broadcast, lets the disposed Sign screen's `cancelHardwareTransfer` cancel `hwTransferSignJob` before `fundPaidOrder` records the paid order — the `if (pendingHwFundingBroadcast != null) return` guard there is bypassed because the field was already nulled. Return `false` from `prepareSpendingHwSign` when `hwTransferSignJob?.isActive == true || pendingHwFundingBroadcast != null`. Debug/dev-mode only, but the existing in-flow trigger (`onOrderCreated`) requires deliberate user navigation; this one is external.

Regression test:

```kotlin
@Test
fun `prepareSpendingHwSign keeps a pending hardware broadcast`() = test {
    val order = previewBtOrder()
    val otherOrder = previewBtOrder().copy(id = "order-b-id")
    val funding = HwFundingTransaction(
        psbt = "psbt",
        miningFeeSats = MINING_FEE,
        feeRate = FEE_RATE.toFloat(),
        totalSpent = order.feeSat + MINING_FEE,
        satsPerVByte = FEE_RATE,
    )
    val signed = signedFunding(funding)
    whenever(hwWalletRepo.wallets)
        .thenReturn(MutableStateFlow(persistentListOf(hwWallet(HARDWARE_WALLET_ID, connected = true))))
    whenever(hwWalletRepo.ensureConnected(HARDWARE_WALLET_ID))
        .thenReturn(Result.success(mock<TrezorFeatures>()))
    whenever(lightningRepo.getFeeRateForSpeed(any(), anyOrNull())).thenReturn(Result.success(FEE_RATE))
    whenever(hwWalletRepo.composeFundingTransaction(any(), any(), any(), any())).thenReturn(Result.success(funding))
    whenever(hwWalletRepo.signFunding(any(), any())).thenReturn(Result.success(signed))
    whenever(hwWalletRepo.broadcastFunding(signed))
        .thenReturn(Result.failure(AppError(BroadcastException.ElectrumException("DNS lookup failed"))))
    whenever(blocktankRepo.getOrder(eq(otherOrder.id), eq(true))).thenReturn(Result.success(otherOrder))

    sut.onTransferToSpendingHwConfirm(order, HARDWARE_WALLET_ID)
    advanceUntilIdle()
    assertTrue(sut.spendingUiState.value.hasPendingHwBroadcast)

    val prepared = sut.prepareSpendingHwSign(HARDWARE_WALLET_ID, otherOrder.id)

    assertFalse(prepared)
    assertTrue(sut.spendingUiState.value.hasPendingHwBroadcast)
    assertEquals(order.id, sut.spendingUiState.value.order?.id)
}
```

app/src/main/java/to/bitkit/ui/screens/transfer/hardware/SpendingHwSignScreen.kt
jvsena42: **Still stands, but the shape it describes is gone from master.** Re-checked at `origin/master` (`6d23c632`) — #1247 (`5620bef6 fix: create order only on swipe`) landed and removed the premise this PR is built on:

- `SpendingHwSignScreen` no longer takes an `orderId` at all; it keys off `state.feeSat == 0uL` and there is no order until the user swipes.
- `TransferEffect.OnOrderCreated(orderId)` became `TransferEffect.OnQuoteReady`.
- `prepareSpendingHwSign` and `adoptSpendingOrder` no longer exist in `TransferViewModel`.

So the Advanced-flow break I reproduced is no longer reachable as written — but only because the `orderId`-keyed route it lived on is gone. The deep link needs redesigning against the quote flow (there is no order id to name at this point), so I'm leaving this open as the blocker for that redesign rather than resolving it.

Merging `origin/master` into this branch conflicts in 7 files (`ContentView.kt`, `SpendingAdvancedScreen.kt`, `SpendingAmountScreen.kt`, `SpendingAmountHwScreen.kt`, `SpendingHwSignScreen.kt`, `TransferViewModel.kt`, `ContentViewTest.kt`), and the `TransferViewModel` ones are all this same order-vs-quote split.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: **Obsolete at `origin/master` (`6d23c632`) — resolving.** `prepareSpendingHwSign` and its `blocktankRepo.getOrder(orderId, refresh = true)` lookup no longer exist after #1247 (`5620bef6`), so the "only orders this install already tracks" constraint I flagged has no code left to apply to. Worth restating against whatever the redesigned link resolves, if it resolves anything server-side.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: **Addressed on master, by a different route — resolving.** #1247 (`5620bef6`) deleted `adoptSpendingOrder`, and `TransferViewModel` on `origin/master` now carries exactly the guard I asked for, at the flow entry points instead of the deep link:

```kotlin
fun onConfirmAmount(satsAmount: Long) {
    if (confirmPayJob?.isActive == true || hwTransferSignJob?.isActive == true) return
```

(same line at the head of `onSpendingAdvancedContinue`). The redesigned deep link should reuse that guard on whatever state it adopts, so an external link still cannot land mid-sign or mid-broadcast.

app/src/main/java/to/bitkit/ui/ContentView.kt
jvsena42: **Resolved by master — resolving.** The effect on `origin/master` is already the collapsed shape I suggested: one `consumeScreenDeepLink()` right after `navController.currentBackStackEntryFlow.first()`, and one `Logger.warn("Unhandled screen deeplink '$uri'")` at the end. The duplicate warn and the three consume calls only exist on this branch. Worth not reintroducing either when the link is redesigned.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
jvsena42: **Premise still holds on master — leaving open.** `ScreenDeepLinks.isScreenDeepLink` at `origin/master` is still a plain scheme/host check with no `isEnabled` gate, and `AppViewModel.processDeeplink`'s `shouldQueue(...)` is still the only thing keeping a release build from acting on a dev-only URI. The specific function I flagged (`spendingHwSignLink`) is this branch's own and will be rewritten, but the fail-safe point applies to whatever replaces it: route the parse through `ScreenDeepLinkRuntime`/`isEnabled` so it returns null in release rather than relying on one call site.

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
jvsena42: **Still applies — leaving open.** Cheap to apply when these tests are rewritten for the quote flow: inside `test {}` blocks, suspend stubs use `whenever(blocktankRepo.getOrder(eq(order.id), eq(true)))`, not `whenever { }`.

jvsena42:
Went through my six open review threads against `origin/master` (`6d23c632`) to check which still hold. Three are resolved, three are left open — details are in each thread.

The short version: **#1247 (`5620bef6 fix: create order only on swipe`) landed and removed the concept this PR is keyed on.** There is no order at the sign screen any more:

- `SpendingHwSignScreen` no longer takes an `orderId`; it gates on `state.feeSat == 0uL`.
- `TransferEffect.OnOrderCreated(orderId)` → `TransferEffect.OnQuoteReady`.
- `prepareSpendingHwSign` and `adoptSpendingOrder` are gone from `TransferViewModel`.

So `bitkit://screen/spending-hw-sign/{walletId}/{orderId}` has no order id to name. I merged `origin/master` into the branch locally to see the damage: 7 files conflict — `ContentView.kt`, `SpendingAdvancedScreen.kt`, `SpendingAmountScreen.kt`, `SpendingAmountHwScreen.kt`, `SpendingHwSignScreen.kt`, `TransferViewModel.kt`, `ContentViewTest.kt` — and the `TransferViewModel` ones are all the same order-vs-quote split. I did not push that merge: resolving it is a redesign of the link's payload, which is your call, not a conflict fix.

Two things from the old review are worth carrying into whatever replaces it:

- Master already added the in-flight guard I asked for, at the flow entry points rather than the link: `if (confirmPayJob?.isActive == true || hwTransferSignJob?.isActive == true) return` heads both `onConfirmAmount` and `onSpendingAdvancedContinue`. The redesigned link should sit behind the same guard.
- `ScreenDeepLinks.isScreenDeepLink` is still not routed through `ScreenDeepLinkRuntime`/`isEnabled` on master, so `AppViewModel.processDeeplink`'s `shouldQueue(...)` remains the only thing keeping a release build from acting on a dev-only URI. Worth making the new parse fail safe on its own.

Not blocking anything of mine — flagging it so the rebase isn't a surprise.

app/src/main/java/to/bitkit/ui/screens/transfer/hardware/SpendingHwSignScreen.kt
jvsena42: **Fixed in `105486a8a`, by removing the route argument this was about.**

The guard is gone because the link no longer names an order. After #1247, `SpendingHwSign` carries `{walletId}/{amountSats}` and `prepareSpendingHwSign` produces the same quote the amount screen does — `loadHwLimits(walletId)` then `quoteSpendingAmount(amountSats)` — so the screen opens on live state rather than a frozen id. Nothing is bought until the user swipes, so Advanced re-pricing has no stale id to disagree with.

Verified on a Pixel_9 emulator against a paired Trezor (device balance 24 831 128 sats, LSP cap 16 935 sats):

```
$ adb shell am start -a android.intent.action.VIEW \
    -d "bitkit://screen/spending-hw-sign/trezor:8df665ab…81be5/10000" to.bitkit.dev
```

opens the sign screen with `TO SPENDING ₿ 10 000`, `SERVICE FEES ₿ 1 158`, `TOTAL ₿ 11 158`, and the app log has **no** `Buying channel with lspBalanceSat` line — the order is not created by the link.

Sign screen opened by the deep link

The original repro no longer lands on the wallet home screen. I drove it on the same build — HW amount screen → 25% → Continue → Sign → Advanced — and the sign screen is reached with a quote (`TO SPENDING ₿ 4 197`) and Advanced opens normally.

One thing I could not exercise, and it is **not** caused by this change: on that Advanced screen the liquidity fee stays `—` and Continue does nothing. I checked it against a baseline build with `ContentView.kt`, `ScreenDeepLinks.kt`, `TransferViewModel.kt` and `SpendingHwSignScreen.kt` reverted to `origin/master`, and it stalls identically there, so it is pre-existing on master with this emulator's LSP state. Worth its own issue rather than holding this PR.

Regression cover in `TransferViewModelTest.kt`:

- `prepareSpendingHwSign quotes the linked amount when the wallet is known` — asserts `clientBalanceSat`, a non-zero `feeSat`, and `verify(blocktankRepo, never()).createOrder(...)`.
- `prepareSpendingHwSign refuses an unknown wallet` / `refuses a non positive amount` — both assert `getFundingAccount` is never reached.
- `prepareSpendingHwSign refuses an amount the device balance cannot fund`.

app/src/main/java/to/bitkit/ui/utils/ScreenDeepLinks.kt
jvsena42: **Fixed in `105486a8a`.** `spendingHwSignLink` now opens with the gate, so it fails safe on its own rather than relying on `AppViewModel.processDeeplink` remembering to call `shouldQueue(...)`:

```kotlin
fun spendingHwSignLink(uri: Uri): SpendingHwSignLink? {
    if (!isEnabled || !isScreenDeepLink(uri)) return null
```

While rewriting it I hit the failure mode that made this worth more than a fail-safe: the old parser returned `null` for a malformed link, which fell through to `navController.handleDeepLink(...)`. That still matched the route and navigated, and the sign screen then bounced the user to the wallet home on `state.feeSat == 0uL` — the same "lands on home" symptom as the blocking finding on this PR. So the parse result is now a sealed type: `null` means "not this screen, fall through", `Malformed` means "this screen, refuse it".

Covered in `ScreenDeepLinksTest.kt` by `spendingHwSignLink reports malformed arguments rather than falling through to nav`, over `…/hardware-wallet` (missing amount), `…/hw/abc`, `…/hw/0` and `…/hw/-1`, plus `returns null while screen deep links are disabled`.

On device, each refusal keeps the wallet overview up and logs exactly one line:

```
WARN [ContentView.kt:349]        Refused spending hw sign deeplink, malformed '…/hw/0'
WARN [ContentView.kt:349]        Refused spending hw sign deeplink, malformed '…/hw/abc'
WARN [TransferViewModel.kt:1020] Refused spending hw sign deeplink, unknown wallet 'not-a-wallet'
WARN [TransferViewModel.kt:1037] Refused spending hw sign deeplink, no quote for '999999999' sats
```

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
jvsena42: **Fixed in `105486a8a`.** Those tests were rewritten for the quote flow, and the new ones use `whenever(mock.call())` throughout inside `test {}` — e.g. `whenever(hwWalletRepo.getFundingAccount(HARDWARE_WALLET_ID))` and `whenever(hwWalletRepo.ensureConnected(HARDWARE_WALLET_ID))`. No `whenever { }` remains in the lines this PR adds.

jvsena42:
Rebased this onto `origin/master` and reworked the link, since #1247 (`5620bef6 fix: create order only on swipe`) removed the order the old design named. Pushed as `105486a8a`; all six of my review threads are now resolved.

### What changed

The link was `spending-hw-sign/{walletId}/{orderId}`. There is no order at the sign screen any more, so it is now `spending-hw-sign/{walletId}/{amountSats}` and `TransferViewModel.prepareSpendingHwSign` produces the same quote the amount screen does:

```kotlin
loadHwLimits(walletId)
if (!quoteSpendingAmount(amountSats)) { … return false }
```

To make that awaitable, `onConfirmAmount` and `updateHwLimits` keep their fire-and-forget signatures but delegate to suspend cores (`quoteSpendingAmount`, `loadHwLimits`), and `onEstimateReady` returns whether it applied. Nothing is bought until the user swipes, so the link cannot strand an order.

Three review points folded in:

- **In-flight guard.** The link is refused while `confirmPayJob`/`hwTransferSignJob` is active or `pendingHwFundingBroadcast != null`, so it cannot discard a signed-but-unbroadcast funding tx.
- **Fail-safe parsing.** `spendingHwSignLink` returns `null` when `!isEnabled`, so a release build cannot reach transfer state through a dev-only URI even if a call site forgets `shouldQueue(...)`.
- **One consume, one warn.** The deep-link effect in `ContentView` consumes `pendingScreenDeepLink` exactly once, at the end (consuming early cancels the coroutine mid-`prepareSpendingHwSign`), and no longer emits a second generic "Unhandled screen deeplink" for a link it recognised and deliberately refused.

A malformed link is now refused rather than navigated to. The old parser returned `null` for a bad amount, which fell through to `handleDeepLink`, matched the route anyway, and bounced the user to the wallet home on `feeSat == 0uL`. The parse result is a sealed type now: `null` means "not this screen", `Malformed` means "this screen, refuse it".

### QA

Pixel_9 emulator, dev build, paired Trezor emulator (device balance 24 831 128 sats, LSP cap 16 935 sats).

`…/spending-hw-sign/trezor:8df665ab…81be5/10000` opens the sign screen on a live quote, with no `Buying channel with lspBalanceSat` line in the app log:

Sign screen opened by the deep link, showing TO SPENDING 10 000 and TOTAL 11 158

Every refusal keeps the wallet overview up and logs exactly one line, with no duplicate "Unhandled screen deeplink":

| link | result |
| --- | --- |
| `…/not-a-wallet/10000` | `Refused … unknown wallet 'not-a-wallet'` |
| `…/<walletId>/0` | `Refused … malformed` |
| `…/<walletId>/abc` | `Refused … malformed` |
| `…/<walletId>/999999999` | `Refused … no quote for '999999999' sats` |

The in-app path still works: HW amount screen → 25% → Continue reaches the sign screen with `TO SPENDING ₿ 4 197`.

**One pre-existing issue, not from this PR.** On the Advanced screen reached from the sign screen, the liquidity fee stays `—` and Continue does nothing. I re-checked it on a baseline build with `ContentView.kt`, `ScreenDeepLinks.kt`, `TransferViewModel.kt` and `SpendingHwSignScreen.kt` reverted to `origin/master`, and it stalls identically, so it is master behaviour with this emulator's LSP state. Probably worth its own issue.

`just compile`, `just test` and `just lint` are green.

### Design

N/A — no design available.

### QA Notes

New cover in `ScreenDeepLinksTest.kt` (valid link, other-screen fall-through, malformed arguments, disabled gate) and `TransferViewModelTest.kt` (quotes the linked amount without creating an order; refuses unknown wallet, non-positive amount, unfundable amount; keeps a pending hardware broadcast). `journeys/deeplinks/spending-hw-sign-deeplink.xml` walks the same cases on a device.

PR #1178, opened 2026-08-24
feat: send paykit payment proofs

ben-kaufman:
This PR sends a Paykit payment proof back to the requester after Bitkit successfully pays an incoming private payment request.

### Description

1. Persists the payer identity, request, and selected endpoint before request acceptance, then associates the Lightning payment hash before sending so restart recovery keeps the correct correlation.
2. Persists successful on-chain transaction IDs and validated Lightning preimages before submitting them against the original one-time request and exact endpoint.
3. Uses Paykit's durable private outbox, retries incomplete submissions during normal request refresh, and suppresses duplicate proofs already recorded by the SDK.
4. Preserves earlier in-flight correlations across retries, clears the request after one proof is queued, and reads the secure store directly so wallet wipes cannot resurrect cached proof state.
5. Attempts immediate SDK delivery if persisting a completed proof fails, without adding any user-facing UI.

This PR is stacked on #1172. Companion iOS PR: https://github.com/synonymdev/bitkit-ios/pull/683

### Preview

N/A — there are no UI changes.

#### Manual Tests

- [x] **1.** Android → create a private request for iOS → iOS receives and pays on regtest: Android records the proof against the exact request, endpoint, and transaction ID.
- [x] **2.** iOS → create a private request for Android → Android receives and pays on regtest: iOS records the proof against the exact request, endpoint, and transaction ID.

#### Automated Checks

- `PaykitPaymentProofRepoTest.kt`: 8 tests cover restart retry, duplicate suppression, invalid Lightning preimage rejection, failed-payment cleanup, exact on-chain proof submission, overlapping retries, wallet-store clearing, and completed-proof persistence failure.
- `AppViewModelSendFlowTest.kt`: covers proof persistence before private-list consumption and request acceptance, plus successful and failed payment lifecycle handling.
- Full Android unit tests and focused proof/send-flow tests passed.
- Detekt passed with only pre-existing unrelated findings.
- `git diff --check` passed.


app/src/main/java/to/bitkit/repositories/PaykitPaymentProofRepo.kt
```diff
@@ -0,0 +1,308 @@
...
+    val method = MethodId.fromRawValue(identifier) ?: return false
+    return when (kind) {
+        PaykitPaymentProofKind.Lightning -> method == MethodId.Bolt11 || method == MethodId.Lnurl
+        PaykitPaymentProofKind.Onchain -> method.isOnchain
+    }
+}
+
+private fun proofJson(kind: PaykitPaymentProofKind, data: String): String = buildJsonObject {
+    put("data", JsonPrimitive(data))
+    put("type", JsonPrimitive(kind.type))
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in b2af2d9db. The process-scoped proof cache was removed; every operation now reads the current encrypted store, so a wallet wipe cannot be repopulated from stale memory. A same-process clear-and-prepare regression test verifies that old identity proof state is not restored.

app/src/test/java/to/bitkit/repositories/PaykitPaymentProofRepoTest.kt
```diff
@@ -0,0 +1,278 @@
...
+            if (shouldFailNextSave) {
+                shouldFailNextSave = false
+                error("temporary save failure")
+            }
+            storedProofs = it.getArgument(0)
+        }
+    }
+
+    @Test
+    fun `completed lightning proof retries after repository restart`() = test {
```
ovitrif, line 96: `completed lightning proof retries after repository restart` calls `completeLightningPayment` before the restarted `reconcile()`, so `proofData` is already stored and `lightningRepo.getPayments()` is never used. Crash recovery after an associated hash and an LDK outbound Bolt11 with no `proofData` therefore never runs, so a broken preimage lookup would not fail. Could we add a restart case that leaves `proofData` unset, stubs an outbound Bolt11 payment for the associated hash, and asserts the preimage is persisted and submitted?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3005,6 +3031,11 @@ class AppViewModel @Inject constructor(
 
                 // Extract payment hash from invoice for pre-activity metadata
                 val paymentHash = decodedInvoice.paymentHash.toHex()
+                associateLightningPaymentProof(incomingPaymentRequest, paymentHash).onFailure {
```
ovitrif, line 3433: `proceedWithPayment` now prepares a Bolt11 proof, associates the invoice hash before `payInvoice`, keeps that correlation on `PaymentPendingException`, and calls `failLightningPayment` plus `cancelPreparation` on send failure, but `AppViewModelSendFlowTest` never asserts those Lightning incoming-request calls. A regression that skips association or cancels a pending Lightning proof would leave restart recovery with no hash to reconcile. Could we cover an incoming Lightning payment-request send that verifies Bolt11 prepare, hash association before send, pending keep, and fail-plus-cancel cleanup?

ovitrif (commented):
Two Lightning test gaps are missing:
- Crash recovery after an associated hash never exercises the LDK lookup, so a broken preimage path would not fail.
- Incoming-request send never asserts prepare, associate, pending keep, or fail-plus-cancel, so a skipped hash association would not fail.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Added focused incoming Lightning request tests in f18953564 covering prepare/consume/accept/associate/send ordering, pending-correlation retention, and fail-plus-cancel cleanup. The affected repository and AppViewModel suites pass (166 tests).

app/src/test/java/to/bitkit/repositories/PaykitPaymentProofRepoTest.kt
ben-kaufman: Added a genuine restart recovery case in f18953564. It persists only the Lightning hash association, recreates the repository, stubs a successful outbound Bolt11 payment, and verifies LDK payment lookup, preimage submission, and cleanup.

ovitrif (dismissed):
Looks good: associated-hash Lightning crash recovery now submits from an LDK succeeded payment, and incoming-request send keeps or clears the proof correlation as required.

piotr-iohk:
Will test after conflicts are resolved.

jvsena42 (changes requested):
@ben-kaufman conflicts

app/src/main/java/to/bitkit/repositories/PaykitPaymentProofStore.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val keychain: Keychain,
+) {
+    @Serializable
+    private data class State(
+        val proofs: List<PendingPaykitPaymentProof> = emptyList(),
+    )
+
+    fun load(): List<PendingPaykitPaymentProof> {
+        val value = keychain.loadString(Keychain.Key.PAYKIT_PENDING_PAYMENT_PROOFS.name) ?: return emptyList()
+        return Json.decodeFromString<State>(value).proofs
```
jvsena42: non-blocking: this is unguarded, so a corrupted blob throws on every load() (Very unlikely to happen)

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -724,6 +728,7 @@ class AppViewModel @Inject constructor(
 
     private suspend fun refreshIncomingPaykitPaymentRequests(): Boolean {
         if (!isPaykitEnabled.value || pubkyRepo.publicKey.value == null || !walletRepo.walletExists()) return false
+        paykitPaymentProofRepo.reconcile()
```
jvsena42, line 730: nit: This does an SDK mutex acquire + keychain decrypt on every poll even with zero proofs

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3325,18 +3334,29 @@ class AppViewModel @Inject constructor(
...
 
         if (!validateIncomingPaymentRequest(contactPaymentContext)) return
 
+        val incomingPaymentRequest = contactPaymentContext?.incomingPaymentRequest
+        var preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).fold(
+            onSuccess = { it },
+            onFailure = {
+                handlePaymentPreparationFailure(it)
+                return
+            },
```
jvsena42, line 3349: The Trezor path bypasses preparePaymentProof/completeOnchainPaymentProof and don't produces a proof

jvsena42: can be addressed to a stacked branch, to reduce conflicts

jvsena42 (changes requested):
Can't pay, swipe is disabled

https://github.com/user-attachments/assets/2013e94e-9f19-4d7c-8667-1215b3e329cb

jvsena42:
Root cause and logs for the disabled swipe in the review above.

### Why it's disabled

`SendConfirmScreen.kt:400` gates the control on:

```kotlin
enabled = uiState.isAmountInputValid && !uiState.isFundingSourceLoading && !isHardwareFeeLoading
```

`isAmountInputValid` defaults to `false` (`AppViewModel.kt:4796`), and the incoming-payment-request branch navigates to the confirm sheet without ever setting it (`AppViewModel.kt:2798`):

```kotlin
navigateToSendRoute(fromMainScanner, SendRoute.Confirm, SendEffect.NavigateToConfirm)
refreshOnchainSendIfNeeded()
estimateLightningRoutingFeesIfNeeded()
return
```

The sibling branch eight lines below (`lnAmountSats > 0u`, `AppViewModel.kt:2806`) does set it before its identical `navigateToSendRoute` call.

### Confirmed on device

Two emulators on this branch build, each with its own pubky identity on the staging homeserver, added as contacts; B requested 16 000 sats from A.

- The sheet opens with the swiper dimmed, and injected drags do not move the thumb at all.
- One tap on the funding-source button — `selectHardwareFundingSource` and the savings branch both assign `isAmountInputValid` — flips it to full brightness, and the identical drag pays immediately.

Worth flagging that the funding-source toggle is only a workaround because that wallet had a paired Trezor. `canSwitchFundingSource = availableFundingSources(state).size > 1` (`AppViewModel.kt:2162`), so with a single funding source there is no way to pay the request at all: the only control on the screen is dead.

Looks like a one-liner — assign `isAmountInputValid = validateAmount(amount)` in that branch before navigating, matching :2806.

### The proof pipeline itself works once past this

Same run, payer side:

```
17:30:17.844 INFO [Keychain.kt:144]               Upserted value for key 'PAYKIT_PENDING_PAYMENT_PROOFS'
17:30:27.162 INFO [Keychain.kt:144]               Upserted value for key 'PAYKIT_PENDING_PAYMENT_PROOFS'
17:30:28.764 INFO [PaykitPaymentProofRepo.kt:270] Queued a Paykit payment proof for private delivery
17:30:32.407 INFO [Keychain.kt:144]               Upserted value for key 'PAYKIT_PENDING_PAYMENT_PROOFS'
```

That is `prepare()` writing the pending record, `completeOnchainPayment()` filling in txid + proofData, `submitReady()` queueing delivery, then cleanup. Receiver side picked it up over the encrypted link:

```
paykit_lib::encrypted_link::link: receive_private_application_messages;
paykit_lib::encrypted_link::link: restore_encrypted_link; remote_pubkey=PublicKey("pubkyej5obc…") local_receiver_path=PaykitReceiverPath("bitkit/wallet") remote_receiver_path=PaykitReceiverPath("bitkit/wallet")
```

and the request moved from "Waiting for … to pay" into the settled group. A force-stop and relaunch of the payer produced no re-submission and no leftover pending record, and there were no Paykit warnings or errors on either device.

ben-kaufman:
Fixed the disabled incoming-request swipe in 6d66ab669. The request branch validated the amount but did not persist that result, so Send Confirm kept the swipe disabled. It now writes the validation result before navigation, and the existing incoming payment request opens the existing confirm flow with its fixed amount test asserts that isAmountInputValid is true. The focused test and full unit suite pass.

app/src/main/java/to/bitkit/repositories/PaykitPaymentProofStore.kt
ben-kaufman: Fixed in 7f003c909. Corrupt JSON is now handled as empty state instead of throwing on every load, and saving an empty state removes the key. PaykitPaymentProofStoreTest covers corrupt data and key deletion.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Addressed in the conflict-isolated stacked PR #1199 as requested. The hardware path now prepares the proof immediately before broadcast and completes it with the returned txid before success navigation. The send-flow tests cover ordering, idempotency, and completion.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 7f003c909. reconcile now checks key existence before entering the operation mutex, asking the SDK for identity, or decrypting proof state. The zero-proof and persisted-empty cases are covered in PaykitPaymentProofRepoTest.

ben-kaufman:
The conflicts are resolved and the branch is current. It also now includes the proof-store/polling hardening and the disabled-swipe fix reported in the latest review. Please retest at 6d66ab669.

ovitrif (dismissed):
utACK

app/src/main/java/to/bitkit/repositories/PaykitPaymentProofRepo.kt
```diff
@@ -0,0 +1,382 @@
...
+                val publicKey = identityStatus.publicKey ?: return@runSuspendCatching
+                val identity = PubkyPublicKeyFormat.normalized(publicKey) ?: return@runSuspendCatching
+                val proofs = storedProofs.filter { PubkyPublicKeyFormat.matches(it.identity, identity) }
+                val payments = if (proofs.any { it.kind == PaykitPaymentProofKind.Lightning && it.proofData == null }) {
+                    lightningRepo.getPayments().getOrDefault(emptyList())
+                } else {
+                    emptyList()
+                }
+
+                proofs.forEach { reconcileProof(it, payments) }
```
jvsena42: nit: a single throw would cancel all `proofs reconcileProof`

jvsena42 (dismissed):
Tested against two emulators. Found one nit that can be adressed upstream

app/src/main/java/to/bitkit/repositories/PaykitPaymentProofRepo.kt
ben-kaufman: Fixed in ad132c893. Each proof reconciliation is now isolated with runSuspendCatching, so one failure is logged and later proofs still reconcile while coroutine cancellation continues to propagate. The regression verifies the first proof stays pending and the second is submitted and removed.

ben-kaufman:
The reconciliation finding is fixed in signed commit ad132c893. Each pending proof is now isolated, so one temporary submission failure does not stop later proofs, with a focused regression covering that sequence. @ovitrif please re-review the updated head when E2E finishes.

piotr-iohk:
Retested latest revision, Pixel emu ↔ iPhone 17 sim (same session as ios #683).

Incoming request pay on Android goes to Confirm with the amount locked. That matches this side — the amount-pad leak is iOS only (see #683).

**Request lists don’t match**
Same session, both on TODAY:

- Android: 2× 1000 + 26 000 + 25 000
- iOS: 3× 1000 + 26 000 (no 25 000)

Not amount-keyed overwrite in the app list — IDs are `paymentRequestId + counterparty + path`. This device sent the 25k on-chain; iOS activity shows +25 000, but iOS is missing that request-history row. Android is missing one of the 1k rows iOS still has. Hypothesis: after fulfill, one side drops the row when `localRole` is missing/unknown (history mapping skips those). Avatars I vs A are just the contact names.

Screenshot (Payment Requests list):
Screenshot 2026-09-01 at 16 12 09

Logs:

[logs-android.zip](https://github.com/user-attachments/files/31696609/logs-android.zip)

Proofs still queued on the successful pays (LN + on-chain both ways). List drift looks like a follow-up, not a proofs blocker.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2786,6 +2795,7 @@ class AppViewModel @Inject constructor(
                 return
             }
 
+            _sendUiState.update { it.copy(isAmountInputValid = validateAmount(amount)) }
```
ovitrif, line 2797: Incoming requests that open through `onScanOnchain` used to reach Confirm with `isAmountInputValid` still false, which keeps `SwipeToConfirm` disabled. This line is the fix, but the new assertion is on a lightning presentation test that already sets `isAmountInputValid = true` in `onScanLightning`, so deleting this assignment would not fail that test. Could we assert `isAmountInputValid` after an incoming-request scan that actually enters `onScanOnchain`?

ben-kaufman:
Thanks for confirming the Android amount path is correct. I also reran the clean cross-platform proof/history portion against the current Android and iOS heads: two requests succeeded in each direction, both apps ended with the same four history rows, and each remote request transitioned to paid after the payer logged durable proof-outbox acceptance. I could not reproduce a stuck proof or history divergence on clean state, so there is no speculative Android change for those observations. All Android checks are green on `ad132c893`.

ovitrif (commented):
The incoming-request swipe assignment in `onScanOnchain` still has no test that would fail if it were removed. The new `isAmountInputValid` assertion is on a lightning presentation that already sets that flag in `onScanLightning`.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Added a genuine `Scanner.OnChain` incoming-request regression. It verifies the request amount replaces the zero-amount decoded URI, is revalidated, and opens Confirm with the Paykit context intact.

ben-kaufman: Follow-up on the final implementation: the regression now drives a real `Scanner.OnChain` incoming request and stubs only the native address-validation boundary through `CoreService`. It asserts the request amount replaces the zero-amount URI and that `isAmountInputValid`, the on-chain type, and the Paykit context are all set, so removing the `onScanOnchain` assignment breaks the test. The final head (`10ff0df74`) passed the full CI build/unit suite: 2,283 tests, 0 failures, plus lint and detekt.

ovitrif (approved):
utACK

piotr-iohk (approved):
QA LGTM.

Retested latest (`10ff0df`) Pixel emu ↔ iPhone 17 sim, regtest. 2 requests from each side (1k LN + 25k on-chain), all 4 paid.

- Incoming request pay stays on Confirm, amount locked
- Proofs queued after both LN pays and both on-chain pays (`Queued a Paykit payment proof for private delivery`)
- No proof UI, as expected
- No `Failed to prepare/complete/reconcile` or ignored-proof warnings

Payment Requests lists still diverge after the same session (3 rows each, different missing amount). Follow-up: https://github.com/synonymdev/bitkit-android/issues/1219 (ios twin: https://github.com/synonymdev/bitkit-ios/issues/723). Not a proofs blocker.

jvsena42 (approved):
LGTM

PR #1179, opened 2026-08-24
fix: settle spending max on an affordable balance

### LINKED ISSUE #899: [Bug]: Unable to transfer max amount to spending
### What happened?

When trying to transfer max amount to spending a error toast appears.

"Error - The available funds are insufficient to complete the given operation"

### Expected behavior

It should let me transfer max amount

### Steps to Reproduce

1. Have onchain balance
2. Go to transfer to spending
3. Choose max amount
4. Swipe to transfer
5. See error toast

https://github.com/user-attachments/assets/568fc3aa-eeed-4986-96cf-2031bb9d95e1

[bitkit_2026-04-21_07-27-09.log](https://github.com/user-attachments/files/26927919/bitkit_2026-04-21_07-27-09.log)

### Bitkit Version

2.2.0

### Device / OS

Pixel 8

### Reproducibility

Always

### Additional context

The same scenario works on iOS

jvsena42:
Fixes #899

This PR fixes the maximum Savings → Spending transfer failing with an insufficient funds error.

### Description

**What the user hit**

A user moving their whole savings to spending got "The available funds are insufficient to complete the given operation", and the transfer never went through. They retried, and ended up with two Lightning channels they could receive into but not spend from — a spending balance of 0, after paying fees twice.

**Why the transfer failed**

Sizing a max transfer takes two fee quotes. The first prices the whole available balance; the second re-prices at the balance left after that fee. The max was then set to the available balance minus the second quote — a balance that quote never actually priced.

The order built at that max can cost more than the user has. Here it was one satoshi: the order came to 265,727 against 265,726 available. That single sat pushed the transfer past the drain shortcut into an ordinary send it could not fund, which surfaced the raw Lightning error above.

**Why they ended up with a 0 spending balance**

After the failure they left the transfer flow, returned to the wallet home screen, and started the transfer again. The order that followed carried a client balance of zero, which buys receiving capacity only: the channel opens entirely on the LSP's side, the user pays just the liquidity fee, and their spending balance stays 0. They did this twice, paying 1,809 and 2,345 sats, which is why they could receive but not send.

The zero was on screen, not submitted behind a populated field. `sats` and the rendered text are always written together in `AmountInputViewModel`, and a zero amount renders as the placeholder, so the amount screen and then the confirm screen both showed 0 before the swipe.

This PR does not change any of that. Continuing at zero is a supported purchase, and `SpendingAmountScreen.kt:265` still enables Continue while the amount is 0. What the fix removes is the reason to retry, since the max transfer no longer fails. Two related gaps are tracked separately: Max and ¼ stay tappable while the limits load and will set the amount to zero, and fiat entry with no exchange rate can submit zero behind a typed amount.

No funds were lost. Their savings stayed on-chain minus roughly 4,400 sats of fees.

**The fix**

An earlier attempt capped the max at the balance the fee was quoted for, assuming the second quote can never exceed the first. That assumption does not hold everywhere. The service fee moves with how a channel is split between the client and LSP sides, and the direction differs by deployment:

| Environment | Client balance | Service fee | Direction |
|---|---|---|---|
| Production | 261,561 → 265,726 | 3,072 → 3,109 | fee **rises** with client balance |
| Staging / regtest | 264,680 → 266,478 | 1,800 → 1,798 | fee **falls** with client balance |

Where the fee rises, the derived max sits above the quoted balance and capping is enough. Where it falls, it sits below, the cap does nothing, and the order still overshoots — confirmed on staging, where both the old code and the capping attempt produced an order one satoshi over.

Since that direction belongs to the LSP deployment rather than the app, it can differ between environments and change without a client release. So the max is now verified instead of assumed: check that the candidate balance covers its own order fee, and where it does not, re-quote at the shortfall-adjusted balance. This settles within two rounds. Where the fee rises the first check passes immediately and no extra quote is made, so that path keeps its current latency.

The same sizing path backs the hardware wallet transfer limits, so both are covered.

### Preview

https://github.com/user-attachments/assets/29f85ca3-3422-4225-b304-683e594bd9a6

https://github.com/user-attachments/assets/8ebfaf5a-dedb-4558-ba94-1fb3f7e65f86

https://github.com/user-attachments/assets/99326a99-d29c-4ad3-b422-6b77dd2e9c2f

### QA Notes

Reproducing the original failure needs the whole on-chain balance transferred at once — the gap only appears at exactly Max.

**Regtest cannot reproduce #899.** Under the regtest fee behaviour the vulnerable window is about 2 sats wide versus roughly 37 in production, and the regtest channel limits drift on their own — `maxChannelSizeSat` moved 1,388,280 → 1,403,872 within one session, shifting the target. Treat regtest as regression coverage only; the unit tests are the gate for the fix itself.

The reliable oracle on device is the log, not the screen. `Estimating order fee for spendingSats=` is the available amount, and the order's `feeSat` must not exceed it:

```
adb shell "run-as to.bitkit.dev cat files/logs/<newest>.log" \
  | grep -E "Estimating order fee|Buying channel|BT confirm|Failed to prepare"
```

#### Manual Tests

- [x] **1.** Transfer → Spending → tap Max → Continue → swipe: order funds, and the logged `feeSat` is at or below the available amount.
- [x] **2.** `regression:` Transfer → Spending → tap ¼ → Continue → swipe: still funds normally.

#### Automated Checks

- Unit test added in `TransferViewModelTest.kt`: reproduces the reported production quote pair and asserts the max stays within what the order will cost. Confirmed it fails against master.
- Unit test added in `TransferViewModelTest.kt`: covers the opposite fee direction with four live quotes, asserting the settled max funds its own order. Confirmed it fails against the capping attempt, so the two tests cover distinct fee directions rather than overlapping.
- Local `just compile`, `just test`, and `just lint` all pass, with only pre-existing detekt findings in untouched files.


app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -697,6 +699,57 @@ class TransferViewModel @Inject constructor(
...
+        }
+        if (candidate.safe() + fee.safe() <= availableAmount) return candidate
+
+        val fallback = availableAmount.safe() - fee.safe()
+        Logger.warn(
+            "Max '$candidate' still over budget '$availableAmount' after " +
+                "'$MAX_AFFORDABILITY_ROUNDS' rounds, advertising unverified '$fallback'",
+            context = TAG,
+        )
+        return fallback
```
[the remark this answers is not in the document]
jvsena42: guard added in https://github.com/synonymdev/bitkit-android/pull/1179/commits/688db3b8df2b5745f9b98981bd39e51a501183c3

jvsena42:
Thanks — both fallback branches are real, and the coverage gap was fair. Addressed in 688db3b8d, though not the way the review suggests.

**Where I disagree:** the fallbacks cannot "recreate the insufficient-funds failure". The send is independently guarded — `resolveSendAllSpendingConfirmFunding` throws `pay_insufficient_savings` and `paySpendingConfirmOrder` re-checks before broadcasting — so an unaffordable order is refused at confirm and #899's raw Lightning error on send does not come back.

**The real residual harm was different:** `onConfirmAmount` created the Blocktank order *before* anything checked affordability, guarding only `maxLspBalanceSat == 0uL`. So an imprecise max could leave a dangling order and a dead-end confirm screen. That is now guarded: `canFundOrder` resolves the budget live and prices the order with `quoteOrderFee`, which recomputes the same channel split order creation uses. An imprecise max is therefore display-only — nothing unfundable reaches the LSP.

**Why the fallback branches themselves are unchanged:** at both points, `availableAmount - fee` is already the most conservative value obtainable without another quote. Returning `candidate` would be strictly worse, since the check immediately above proved it unaffordable. And failing closed is actively harmful here: `maxAllowedToSend = 0` makes `sats <= maxAllowedToSend` true only at zero, which re-opens the zero-amount purchase path that gave the reporting user two unspendable channels. Both branches now log a warning naming the value being advertised.

On likelihood: exhaustion needs the iteration not to converge in two rounds, and since the fee moves well under a satoshi per satoshi of balance the error shrinks roughly two orders of magnitude per round — so that path is pathological. Quote failure is the reachable one; the reporting user's logs show repeated LSP DNS failures in a single session.

**Tests added** for both flagged branches: `updateLimits keeps the last candidate when a re-quote fails` and `updateLimits falls back to the shortfall balance when rounds are exhausted`, plus three covering the new guard. Each was confirmed to fail with its fix removed.

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
```diff
@@ -233,6 +235,224 @@ class TransferViewModelTest : BaseUnitTest() {
...
+        val spendable = 266_656uL
+        val miningFee = 178uL
+        val availableAmount = spendable - miningFee // 266_478
+        stubSpendableBalances(spendable)
+        blocktankState.value = BlocktankState(info = null)
+        whenever { lightningRepo.estimateSendAllFee(anyOrNull(), anyOrNull(), anyOrNull()) }
+            .thenReturn(Result.success(miningFee))
+        whenever(blocktankRepo.calculateLiquidityOptions(any()))
+            .thenReturn(Result.success(liquidityOptions(maxClientBalanceSat = spendable)))
+        // every quote stays 1_800, so no candidate ever becomes affordable and both rounds are used
```
ovitrif: This test never enters the exhausted-rounds fallback it is named for. A flat 1,800 sat quote makes the initial fee and the final fee the same, so quotedBalance plus quotedFee equals the 266,478 sat budget on the first check and resolveAffordableClientBalance returns immediately. The assertion would still pass if the unverified availableAmount minus fee fallback were deleted or returned the still-over candidate. Please stub a fee that keeps rising as the client balance steps down, so that after two re-quotes the candidate is still over budget, and then assert Max equals availableAmount minus that last fee.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -697,6 +705,96 @@ class TransferViewModel @Inject constructor(
...
+     * LSP will not give, leaves the decision to the confirm step rather than blocking the user here.
+     */
+    private suspend fun canFundOrder(clientBalance: ULong): Boolean {
+        val budget = _spendingUiState.value.fundingBudgetSats
+        if (budget == null) {
+            Logger.warn("Skipped funding check, no sized budget available", context = TAG)
+            return true
+        }
+        val fee = quoteOrderFee(clientBalance)
+        if (fee == null) {
```
ovitrif, line 778: When quoteOrderFee returns null, canFundOrder skips and onConfirmAmount still places the order. That confirm-time skip is new, and the only nearby test covers an unset budget rather than a failed quote. A change that blocked Continue whenever the estimate failed would not fail any current test. Please add a case that sizes fundingBudgetSats, then makes the confirm-time estimate fail, and still expects createOrder.

ovitrif (commented):
- Restub the exhausted-rounds case so the candidate stays over budget after two re-quotes, then assert Max equals availableAmount minus that last fee.
- Add a confirm case that sizes the budget, then makes the live estimate fail, and still expects createOrder.

app/src/test/java/to/bitkit/viewmodels/TransferViewModelTest.kt
jvsena42: Thanks, checking..

jvsena42: fixed in https://github.com/synonymdev/bitkit-android/pull/1179/commits/c0fe35ae0f03127a03d563d1d1ffd25f6e5439a3

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
jvsena42: fixed in https://github.com/synonymdev/bitkit-android/pull/1179/commits/83a5530901fdaf487a8f5c405ac2ae4c89db6ba6

ovitrif (approved):
utACK

PR #1187, opened 2026-08-26
feat: add trezor send

### LINKED ISSUE #1201: Send from hardware wallet
Part of #998.

**Related (iOS):** synonymdev/bitkit-ios#706

**Design:** [Figma — HW Send](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=48210-118203)

## Scope

Paired Trezor wallets are funding sources in the normal on-chain Send flow.

- Scanner, paste, manual entry, and contacts can fund a send from a paired Trezor.
- Amount, fee selection, confirmation, device signing, broadcast, and success all run with the selected hardware identity.
- Lightning and LNURL requests explain that a Bitcoin address is required.
- Switching among Savings, Spending, and Trezor updates available balance and the fee-aware maximum.
- Passphrase identities reconnect to the matching hidden wallet before signing.
- Cancel or disconnect during signing recovers the session without broadcasting a duplicate.
- Pending sent activity and contact metadata stay attached while watcher snapshots catch up.

Send UI polish:

- Switching the funding source to Trezor keeps the source chip filled and shows a small spinner until the hardware fee is ready.
- Confirm keeps the previous fee with a spinner while the estimate refreshes.
- Swipe-to-confirm stays visible and stays blocked until the hardware fee is present.

ben-kaufman:
### Description

This PR:
1. Adds paired Trezor wallets as funding sources throughout the normal on-chain Send flow, including scanner, paste, manual entry, contacts, fee selection, confirmation, signing, broadcasting, and success handling.
2. Hardens Trezor session recovery and preserves hardware-wallet activity and contact metadata while wallet snapshots catch up.

The Send UI follows the [Bitkit Wallet design](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=47426-195139&m=dev).

Receive support will follow in a separate stacked PR.

### Linked Issues/Tasks

Fixes #1201

#### Manual Tests

- [x] **1.** Trezor wallet → Send → enter a Bitcoin address and amount → Continue: the button loads once, Confirm opens, and repeated taps do not duplicate preparation.
- [x] **2.** Send Confirm → choose Trezor → Sign With Device → approve on Trezor: the transaction broadcasts and Success shows the hardware-wallet activity.
- [ ] **3.** Trezor passphrase wallet → Send → enter passphrase: the paired account reconnects and the operation resumes.
- [x] **4.** Trezor-funded Send → scan a Lightning or LNURL request: Bitkit explains that a Bitcoin address is required.
- [x] **5.** `regression:` Send → switch between Savings, Spending, and Trezor: available balance and fee-aware maximum update for each source.
- [ ] **6.** `regression:` cancel or disconnect during signing → retry: Bitkit reconnects without creating or broadcasting a duplicate transaction.

#### Automated Checks

- Unit tests added or extended in `HwFundingSignerTest.kt` and `TrezorSessionFailureTest.kt`: cover source coordination, timeouts, stale-session retry, retained signed transactions, and failure classification.
- Unit tests extended in `HwWalletRepoTest.kt`: cover pending sent activity and contact preservation during watcher reconciliation.
- Unit tests extended in `ShopPaymentRequestTest.kt`: cover hardware-wallet on-chain-only scan handling.
- Hardware-wallet tests, lint, and formatting checks pass.

### DIFF
--- a/app/src/main/java/to/bitkit/services/CoreService.kt
+++ b/app/src/main/java/to/bitkit/services/CoreService.kt
@@ -309,6 +309,7 @@ private fun OnchainActivity.mergedWith(stored: OnchainActivity?): OnchainActivit
         channelId = channelId ?: stored.channelId,
         transferTxId = transferTxId ?: stored.transferTxId,
         contact = contact ?: stored.contact,
+        seenAt = seenAt ?: stored.seenAt,
     )
 }
 
--- a/app/src/main/java/to/bitkit/services/CoreService.kt
+++ b/app/src/main/java/to/bitkit/services/CoreService.kt
@@ -240,6 +240,7 @@ class CoreService @Inject constructor(
 
 // region Activity
 private const val CHUNK_SIZE = 50
+private const val HW_PENDING_SEND_GRACE_PERIOD_SECONDS = 86_400L
 
 /**
  * Outcome of replacing a hardware wallet's on-chain snapshot.
@@ -261,6 +262,9 @@ internal data class HwSnapshotMerge(
 /**
  * Builds the delete/upsert plan for a hardware wallet's on-chain snapshot.
  *
+ * Recent pending sends remain during the watcher's eventual-consistency window. Their persisted
+ * creation timestamp keeps that protection across process restarts.
+ *
  * @param transferChannelIdsByFundingTxId funding tx id to channel id for transfers Bitkit recorded
  * itself. Re-pairing a wallet that was removed leaves nothing to carry forward, because removal
  * deleted its activities, and [to.bitkit.repositories.TransferRepo.syncTransferStates] only re-marks
@@ -271,10 +275,15 @@ internal data class HwSnapshotMerge(
 internal fun mergeHwSnapshot(
     existing: List<Activity.Onchain>,
     incoming: List<Activity>,
+    currentTimestamp: ULong,
     transferChannelIdsByFundingTxId: Map<String, String>,
 ): HwSnapshotMerge {
     val incomingIds = incoming.map { it.rawId() }.toSet()
-    val toDelete = existing.filter { !it.v1.isTransfer && it.v1.id !in incomingIds }
+    val toDelete = existing.filter {
+        !it.v1.isTransfer &&
+            !it.v1.isRecentPendingSend(currentTimestamp) &&
+            it.v1.id !in incomingIds
+    }
     val existingByTxId = existing.associateBy { it.v1.txId }
     val toUpsert = incoming.map { activity ->
         val onchain = activity as? Activity.Onchain ?: return@map activity
@@ -286,12 +295,20 @@ internal fun mergeHwSnapshot(
     return HwSnapshotMerge(toDelete = toDelete, toUpsert = toUpsert)
 }
 
+private fun OnchainActivity.isRecentPendingSend(currentTimestamp: ULong): Boolean {
+    val createdAt = createdAt ?: return false
+    if (txType != PaymentType.SENT || confirmed || !doesExist) return false
+    val age = (currentTimestamp.toLong() - createdAt.toLong()).coerceAtLeast(0)
+    return age <= HW_PENDING_SEND_GRACE_PERIOD_SECONDS
+}
+
 private fun OnchainActivity.mergedWith(stored: OnchainActivity?): OnchainActivity = when (stored) {
     null -> this
     else -> copy(
         isTransfer = isTransfer || stored.isTransfer,
         channelId = channelId ?: stored.channelId,
         transferTxId = transferTxId ?: stored.transferTxId,
+        contact = contact ?: stored.contact,
     )
 }
 
@@ -380,6 +397,7 @@ class ActivityService(
         val merge = mergeHwSnapshot(
             existing = existingActivities,
             incoming = activities,
+            currentTimestamp = nowTimestamp().epochSecond.toULong(),
             transferChannelIdsByFundingTxId = transferChannelIdsByFundingTxId,
         )
         merge.toDelete.forEach {
@@ -389,7 +407,6 @@ class ActivityService(
 
         if (merge.toUpsert.isNotEmpty()) upsertActivities(merge.toUpsert)
         if (transactionDetails.isNotEmpty()) upsertTransactionDetails(transactionDetails)
-
         HwSnapshotResult(
             activities = getActivities(
                 walletId = walletId,
--- a/app/src/main/java/to/bitkit/ui/screens/wallets/send/HwSendViewModel.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/wallets/send/HwSendViewModel.kt
@@ -1,0 +1,340 @@
+package to.bitkit.ui.screens.wallets.send
+
+import android.content.Context
+import androidx.compose.runtime.Immutable
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import dagger.hilt.android.lifecycle.HiltViewModel
+import dagger.hilt.android.qualifiers.ApplicationContext
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeout
+import to.bitkit.R
+import to.bitkit.ext.isBroadcastConnectivityFailure
+import to.bitkit.ext.isTrezorDeviceBusy
+import to.bitkit.ext.isTrezorFirmwareError
+import to.bitkit.ext.isTrezorSessionFailure
+import to.bitkit.ext.isTrezorUserCancellation
+import to.bitkit.ext.runSuspendCatching
+import to.bitkit.models.HwFundingBroadcastResult
+import to.bitkit.models.HwFundingSignedTx
+import to.bitkit.models.HwFundingTransaction
+import to.bitkit.models.Toast
+import to.bitkit.repositories.ActivityRepo
+import to.bitkit.repositories.HwPassphraseMismatchError
+import to.bitkit.repositories.HwPassphraseRequiredError
+import to.bitkit.repositories.HwWalletRepo
+import to.bitkit.repositories.PreActivityMetadataRepo
+import to.bitkit.services.CoreService
+import to.bitkit.ui.shared.toast.ToastEventBus
+import to.bitkit.utils.Logger
+import javax.inject.Inject
+import kotlin.time.Duration.Companion.seconds
+
+@HiltViewModel
+class HwSendViewModel @Inject constructor(
+    @ApplicationContext private val context: Context,
+    private val hwWalletRepo: HwWalletRepo,
+    private val preActivityMetadataRepo: PreActivityMetadataRepo,
+    private val coreService: CoreService,
+    private val activityRepo: ActivityRepo,
+) : ViewModel() {
+    private companion object {
+        const val TAG = "HwSendViewModel"
+        val RECONNECT_TIMEOUT = 30.seconds
+        val COMPOSE_TIMEOUT = 45.seconds
+        val SIGN_TIMEOUT = 120.seconds
+        val BROADCAST_TIMEOUT = 120.seconds
+    }
+
+    private val _uiState = MutableStateFlow(HwSendUiState())
+    val uiState = _uiState.asStateFlow()
+
+    private val pendingResult = MutableStateFlow<HwSendResult?>(null)
+    val results = pendingResult.filterNotNull()
+
+    private var pendingBroadcast: PendingHwSendBroadcast? = null
+    private var signingWalletId: String? = null
+    private var signingJob: Job? = null
+    private var passphraseJob: Job? = null
+
+    fun warmUp(walletId: String) {
+        hwWalletRepo.warmUpKnownDevice(walletId)
+    }
+
+    fun signAndBroadcast(
+        request: HwSendRequest,
+        beforeBroadcast: suspend () -> Boolean = { true },
+    ) {
+        if (_uiState.value.isSigning || signingJob?.isActive == true) return
+        if (pendingBroadcast?.matches(request) == false) return
+        signingWalletId = request.walletId
+        _uiState.update { it.copy(isSigning = true) }
+        signingJob = viewModelScope.launch {
+            try {
+                runSuspendCatching {
+                    var pending = pendingBroadcast?.takeIf { it.matches(request) }
+                    if (pending == null && hwWalletRepo.needsPassphrase(request.walletId)) {
+                        _uiState.update { it.copy(isPassphraseRequired = true) }
+                        return@runSuspendCatching
+                    }
+                    if (pending == null) {
+                        val signedTx = prepareSignedTransaction(
+                            walletId = request.walletId,
+                            address = request.address,
+                            amountSats = request.amountSats,
+                            satsPerVByte = request.satsPerVByte,
+                        )
+                        pending = PendingHwSendBroadcast(request, signedTx)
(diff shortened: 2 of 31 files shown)


app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -310,6 +316,7 @@ class ActivityService(
     private val privatePaykitContactResolver: Provider<PrivatePaykitContactResolver>,
 ) {
     private val defaultWalletId = WalletScope.default
+    private val pendingHwSendTxIds = mutableMapOf<String, MutableSet<String>>()
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in 98139e8fc. Pending sends are now protected by their persisted createdAt timestamp during a bounded 24-hour watcher consistency window, so process restart no longer drops the protection; expired missing sends remain prunable. Added coverage for both recent and expired pending sends.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -55,6 +59,13 @@ fun ReceiveSheet(
     val showCreateCjit = remember { mutableStateOf(false) }
     val cjitEntryDetails = remember { mutableStateOf<CjitEntryDetails?>(null) }
     val lightningState: LightningState by wallet.lightningState.collectAsStateWithLifecycle()
+    val hardwareWallets by hwReceiveViewModel.wallets.collectAsStateWithLifecycle()
+    val hwReceiveState by hwReceiveViewModel.uiState.collectAsStateWithLifecycle()
+    val selectedHardwareWalletId = hardwareWalletId ?: hardwareWallets.singleOrNull()?.id
```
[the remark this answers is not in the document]
ben-kaufman: This is intentional and matches the iOS flow. Global Receive exposes the hardware tab only when the wallet identity is unambiguous. With multiple paired identities, the user opens the desired hardware-wallet screen and taps Receive; that entry point passes its explicit hardwareWalletId and opens the Trezor tab. Choosing an arbitrary wallet would be unsafe, and adding a wallet picker is separate UX outside this PR.

jvsena42 (commented):
Worth updating the journeys with the new flows

ovitrif:
Could be split into 2 stacked PRs, one for Send, one for Receive 🙏🏻

ben-kaufman:
Done — #1187 is now Send-only and conflict-free. Receive is split into the stacked #1189. iOS is split the same way: Send in synonymdev/bitkit-ios#688 and Receive in synonymdev/bitkit-ios#693.

ovitrif:
> Done — #1187 is now Send-only and conflict-free. Receive is split into the stacked #1189. iOS is split the same way: Send in [synonymdev/bitkit-ios#688](https://github.com/synonymdev/bitkit-ios/pull/688) and Receive in [synonymdev/bitkit-ios#693](https://github.com/synonymdev/bitkit-ios/pull/693).

Thanks, recommending to use `gh-stack` extension so this is done automatically from local machine via `gh` cli, we can also do it from the GitHub web UI but that's more streamlined.

You might need to install it first:
```
gh extension install github/gh-stack
```

ben-kaufman:
@jvsena42 Done. The journeys are split with the implementation: `send-onchain.xml` is included in #1187, and `receive-onchain.xml` is included in the stacked #1189. The hardware-wallet journey README is updated in each layer as well.

@ovitrif Thanks for the `gh-stack` recommendation. I’ll use it for future stacked PR work.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1912,51 +1967,122 @@ class AppViewModel @Inject constructor(
...
+            val amount = state.amount
+            val balance = walletRepo.balanceState.value
+            amount > Defaults.dustLimit.toULong() &&
+                amount <= balance.maxSendOnchainSats &&
+                amount <= balance.maxSendLightningSats
+        }
+        _sendUiState.update {
+            it.copy(
+                canSwitchWallet = canSwitchWallet,
+                canSwitchFundingSource = availableFundingSources(state).size > 1,
```
ovitrif, line 2160: I can reach an underfunded state here because `canSwitchFundingSource` only checks how many entries `availableFundingSources()` returns; it does not check whether the next source covers the current amount. `SendConfirmScreen` still allows `SwipeToConfirm` when `isAmountInputValid` is false, so switching on the review screen to an empty `SendFundingSource.Savings` or underfunded Trezor proceeds into a doomed payment/sign attempt. Could we restrict the available sources to those that can fund the amount, or disable confirmation until the newly selected source is valid?

app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -271,10 +275,15 @@ internal data class HwSnapshotMerge(
 internal fun mergeHwSnapshot(
     existing: List<Activity.Onchain>,
     incoming: List<Activity>,
+    currentTimestamp: ULong,
```
ovitrif, line 278: I don't think `currentTimestamp` makes this retention truly bounded. After the first watcher snapshot omits a recent send, `replaceHwSnapshot()` retains it, but `HwWalletRepo` caches that persisted result and every identical `HwSnapshot.toCacheKey()` bypasses `replaceHwSnapshot()` entirely. If no other transaction changes, the 24-hour condition is never reevaluated, so a missing or evicted send can remain pending until restart. Could we expire or bypass the snapshot cache while it contains a locally retained pending send?

app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -286,12 +295,20 @@ internal fun mergeHwSnapshot(
...
+    return age <= HW_PENDING_SEND_GRACE_PERIOD_SECONDS
+}
+
 private fun OnchainActivity.mergedWith(stored: OnchainActivity?): OnchainActivity = when (stored) {
     null -> this
     else -> copy(
         isTransfer = isTransfer || stored.isTransfer,
         channelId = channelId ?: stored.channelId,
         transferTxId = transferTxId ?: stored.transferTxId,
+        contact = contact ?: stored.contact,
```
ovitrif, line 311: I see `mergedWith()` preserve `contact` but not `seenAt`. The local activity created for a successful hardware send sets `seenAt`, while the watcher row has no seen marker; the next snapshot therefore upserts `seenAt = null`, and `isActivitySeen()` treats the user's outgoing payment as unseen again. Could we carry forward `stored.seenAt` when the watcher has no value and add a snapshot-merge regression test?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2196,27 +2354,39 @@ class AppViewModel @Inject constructor(
...
-        is Scanner.NodeId -> handleNonPaymentScan { onScanNodeId(scan) }
-        is Scanner.Gift -> handleNonPaymentScan { onScanGift(scan.code, scan.amount) }
-        else -> {
-            hideSheet()
-            Logger.warn(
-                if (scan == null) "Failed to decode scan data" else "Received unhandled scan data '$scan'",
-                context = TAG,
-            )
+    ) {
+        if (activeHardwareWalletId != null && scan != null && scan !is Scanner.OnChain) {
```
ovitrif, line 2539: I couldn't find an automated test that exercises this new hardware-scoped rejection branch or the parallel pasted-input branch. The added journey only enters a valid on-chain address, so removing this condition would not fail a test and `Sheet.Send(hardwareWalletId = ...)` could accept Lightning or LNURL data again. Could we add `AppViewModel` tests for scanned and pasted Lightning/LNURL input while a hardware-scoped `Sheet.Send` is active?

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
```diff
@@ -187,7 +216,11 @@ fun SendSheet(
                     val lightningState by walletViewModel.lightningState.collectAsStateWithLifecycle()
                     SendAmountScreen(
                         uiState = uiState,
-                        nodeLifecycleState = lightningState.nodeLifecycleState,
+                        nodeLifecycleState = if (uiState.hardwareWalletId != null) {
+                            to.bitkit.models.NodeLifecycleState.Running
```
ovitrif: I don't see a name collision or interop constraint that requires the inline `to.bitkit.models.NodeLifecycleState.Running` reference here. The repository's Kotlin rules require an import for this case. Could we import `NodeLifecycleState` and use `NodeLifecycleState.Running`?

ovitrif (changes requested):
Two correctness defects block this change: switching funding sources can leave confirmation enabled for an underfunded source, and cached watcher snapshots can prevent pending-send expiry from being reevaluated. I also found regressions in activity seen-state preservation and hardware-recipient coverage, plus one Kotlin import violation.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 686479dad. The review swipe is now disabled while the selected funding source is invalid, and the ViewModel also ignores invalid confirmation events. Added regression coverage for the guard.

app/src/main/java/to/bitkit/services/CoreService.kt
ben-kaufman: Fixed in 686479dad. Cached snapshots are no longer reused while they contain a pending send that is absent from the watcher source, so identical watcher updates re-run persistence and reevaluate the expiry window. Added coverage for this cache path.

app/src/main/java/to/bitkit/services/CoreService.kt
ben-kaufman: Fixed in 686479dad. Snapshot merging now keeps the stored seenAt value when the watcher does not provide one, with a regression test.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Added in 686479dad. The AppViewModel tests now cover both scan and paste entry paths with Lightning and LNURL Pay data while a hardware-scoped send is active.

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
ben-kaufman: Fixed in 686479dad. NodeLifecycleState is now imported and referenced directly.

ovitrif (dismissed):
Code LGTM, found some UI nits during testing:

## 1. Available (Savings) text
This was either updated in figma long ago or only in v62, but since we're here we could remove the `(Savings)` part as it's no longer there and a bit redundant (button already says and means the same thing, while when showing trezor funds, it disappears).

 

## 2. **Funding source switch lag** 
when switching to/between HWWs there is an unexpected delay.

Could we add a loading spinner while switching? Button text could fade out while spinner fades in, or even without transition, the UX would feel more snappy, less clumsy?

| Amount | Confirm |
| - | - |
| <video src="https://github.com/user-attachments/assets/aa9d18b7-e1a3-4c15-b42c-e62203054fe6"> | <video src="https://github.com/user-attachments/assets/f8ada66f-5f9e-4297-aa6f-cd2cdf3660ae" > |

I don't see this delay that much on confirm for some reason, but there's a small strangeness if tapping continuously multiple times on the button, just a FYI, don't think it's in scope of this PR to over optimize all small details.

## QA Notes

Tested on a Pixel 9a emulator using regtest:

- Sent from the standard Trezor through source selection, amount entry, duplicate-preparation protection, confirmation, device signing, broadcast, and success.
- Verified the hardware-wallet transaction appeared once in recent activity.
- Exercised source switching across Savings and multiple named Trezor identities; balances updated correctly, with the non-blocking delay noted above.

Approve.

ben-kaufman:
Addressed both UI notes in 0cf15dbc6. The amount screen now uses Available for Savings. Source switching shows a spinner, ignores repeated taps, cancels stale fee refreshes, and disables Continue and confirmation until the selected source is ready.

piotr-iohk:
Testing...

piotr-iohk (dismissed):
LGTM

Tested on emulator + Trezor emu, and on a Samsung S22 with a real Trezor 7 over BLE. Standard HW send, passphrase HW send, tags, and cancel-during-sign all worked. S22 logs look fine on that path.

Nit: Android never shows the hourglass on pending on-chain txs (HW or software) — still the send/transfer arrow. Figma uses HourglassSimple on the activity list ([this](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-147643) and [this](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-147744)). iOS already has the hourglass. Would be good to add it on Android in general, and use blue for pending HW sends so it matches the rest of the hardware styling.

piotr-iohk:
Note: conflicts appeared after recen merges to master I guess.

jvsena42 (changes requested):
@ben-kaufman conflicts

piotr-iohk (dismissed):

ovitrif:
fyi I'm polishing a bit the UI/UX of the new components, there's a few issues with https://github.com/synonymdev/bitkit-android/commit/0cf15dbc6479fe3db0a3ec81ead8ed73e56ba62a, non-blocking though IMO.

I'll either open a stacked PR targeting this or a new one if this gets merged.

jvsena42:
Starting review...

app/src/main/java/to/bitkit/ui/screens/wallets/send/HwSendViewModel.kt
```diff
@@ -0,0 +1,340 @@
...
+                    }
+                    _uiState.update { it.copy(isBroadcastUnresolved = true) }
+                    val result = withTimeout(BROADCAST_TIMEOUT) {
+                        hwWalletRepo.broadcastFunding(payment.signedTx).getOrThrow()
+                    }
+                    runSuspendCatching { persistResult(request, result) }
+                        .onFailure { Logger.error("Failed to persist hardware send result", it, context = TAG) }
+                    pendingResult.update { HwSendResult(request.walletId, result.txId, request.amountSats) }
+                }.onFailure {
+                    handleFailure(it, request.walletId)
```
jvsena42, line 113: `TimeoutCancellationException` never reached this branch, because of runSuspendCatching

ovitrif:
This is probably an issue, on `SendConfirmScreen`, after the swiper disables and then re-enables, sometimes it loses its bg:

https://github.com/user-attachments/assets/0487e83d-57cb-4696-9440-2e814e1c77e1

Note that the UI changes to the funding source button and the fee cell are because the recording is from my local WIP polish, but the swiper issue was there already and I didn't touch it.

jvsena42 (changes requested):

ben-kaufman:
@ovitrif Fixed the swipe background issue in 9484f7852. Funding-source switching now disables gestures and lowers opacity without changing the track background style, so re-enabling it keeps the background intact.

app/src/main/java/to/bitkit/ui/screens/wallets/send/HwSendViewModel.kt
ben-kaufman: Fixed in 9484f7852. The outer send operation now catches timeout cancellation as a failure while rethrowing ordinary cancellation, so `TimeoutCancellationException` reaches `handleFailure`. Added composition-timeout coverage in `HwSendViewModelTest`.

jvsena42 (approved):
tAck

ovitrif (approved):
Approved

I still see the issue I flagged earlier, that was supposed to be fixed after the last commit (https://github.com/synonymdev/bitkit-android/pull/1187#issuecomment-5455998259).



| Expected | Actual |
| - | - |
|  |  |

I'll fix it together with the other polishing changes.

ovitrif:
@ben-kaufman can you resign the commits:
> Commits must have verified signatures.

I can't merge otherwise 🙏🏻

PR #1189, opened 2026-08-27
feat: add trezor receive

### LINKED ISSUE #1202: Receive to hardware wallet
Part of #998.

**Related (iOS):** synonymdev/bitkit-ios#707

## Scope

Receive opens on a Trezor tab from a paired hardware wallet.

- The tab shows the watcher next-unused native SegWit address, with account-scan fallback.
- Amount and note editing stay on-chain only.
- Copy, share, and Show Details use the hardware payment request.
- Verify on Device compares the displayed address to the device and completes only on an exact match.
- Passphrase wallets reconnect and resume verification.
- A watcher address change cancels a stale in-flight verification and shows the new address.
- Hardware receive uses the white selected-tab underline and blue QR/action accents.

ben-kaufman:
### Description

This PR:
1. Adds a Trezor tab to the Receive flow with the current next-unused native SegWit watcher address, falling back to the existing account scan when watcher state is unavailable.
2. Supports on-chain-only amount and note editing without starting Lightning receive work.
3. Lets users copy or share the Trezor payment request and verify the displayed address on the connected device.
4. Matches hardware receive styling with a white selected-tab underline, blue Bitcoin mark, and blue action icons.

### Stack

- Base: #1187
- Requires [bitkit-core #147](https://github.com/synonymdev/bitkit-core/pull/147) / [v0.5.10](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.10)

### Linked Issues/Tasks

Fixes #1202

#### Manual Tests

- [x] **1.** Paired Trezor → Home → Receive → Trezor: the current native SegWit address and QR load with the hardware-wallet styling.
- [x] **2.** Trezor Receive → Show Details → Verify on Device: the button shows a loading state, the device address exactly matches Bitkit, and approval completes without changing the displayed address or showing an error.
- [x] **3.** Trezor Receive → edit amount and note → Show QR → Show Details → Copy/Share: the edit returns directly to the Trezor QR without tags or Lightning setup, and the payment request contains the Trezor address with the entered amount and note.
- [ ] **4.** Passphrase Trezor → Receive → Show Details → Verify on Device → enter the passphrase: the wallet reconnects and address verification resumes.
- [x] **5.** `regression:` Start Verify on Device → while device approval is pending, fund the displayed regtest address and advance the watcher: the stale verification is cancelled and Bitkit displays and copies the new receive address.

#### Automated Tests

- Unit tests added or extended in `HwWalletRepoTest.kt`: cover watcher-backed receive-address selection, stored-xpub fallback, and device-address mismatch rejection.
- Unit tests added in `HwReceiveViewModelTest.kt`: cover address loading and reset, passphrase reconnection, live watcher-address updates, and cancellation of stale verification.
- Unit tests extended in `ReceiveInvoiceUtilsTest.kt`: cover the hardware-only BIP21 destination, shared amount and note details, and zero-amount omission.
- Compose UI coverage added in `EditInvoiceContentTest.kt`: verifies the on-chain-only hardware edit callback and hidden tag actions.
- Hardware-wallet journey coverage added in `receive-onchain.xml`: defines the Trezor tab, QR, details, and exact on-device address-verification flow.
- GitHub CI build and the full `testDevDebugUnitTest` suite pass; lint and detekt pass.
- `git diff --check` passes.

### DIFF
--- a/app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
+++ b/app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
@@ -15,6 +15,7 @@ import kotlinx.collections.immutable.toImmutableList
 import kotlinx.collections.immutable.toImmutableSet
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharedFlow
@@ -47,6 +48,7 @@ import to.bitkit.models.HwFundingAddressType
 import to.bitkit.models.HwFundingBroadcastResult
 import to.bitkit.models.HwFundingSignedTx
 import to.bitkit.models.HwFundingTransaction
+import to.bitkit.models.HwReceiveAddress
 import to.bitkit.models.HwWallet
 import to.bitkit.models.HwWalletReceivedTx
 import to.bitkit.models.KnownDevice
@@ -330,13 +332,89 @@ class HwWalletRepo @Inject constructor(
             }
             val balanceSats = _watcherData.value
                 .values
-                .filter { it.addressType == addressType.settingsKey && it.walletId == walletId }
+                .filter { it.addressType == addressType && it.walletId == walletId }
                 .fold(0uL) { acc, watcher -> acc + watcher.balanceSats }
             HwFundingAccount.Trezor(
                 xpub = xpub,
                 addressType = addressType,
                 balanceSats = balanceSats,
             )
+        }
+    }
+
+    /** Resolves the next unused external address from watcher state, falling back to an account scan. */
+    suspend fun getReceiveAddress(
+        walletId: String,
+        addressType: HwFundingAddressType = HwFundingAddressType.DEFAULT,
+    ): Result<HwReceiveAddress> = withContext(ioDispatcher) {
+        runSuspendCatching {
+            watcherReceiveAddress(walletId, addressType)?.let { return@runSuspendCatching it }
+            val account = getFundingAccount(walletId, addressType).getOrThrow()
+            val accountInfo = trezorRepo.getAccountInfo(
+                extendedKey = account.xpub,
+                network = Env.network.toCoreNetwork(),
+                scriptType = account.accountType,
+            ).getOrThrow()
+            val unused = requireNotNull(accountInfo.account.addresses.unused.firstOrNull()) {
+                "No unused external address returned for hardware wallet '$walletId'"
+            }
+            val scannedAddress = HwReceiveAddress(
+                address = unused.address,
+                path = unused.path,
+                addressType = addressType,
+            )
+            watcherReceiveAddress(walletId, addressType) ?: scannedAddress
+        }
+    }
+
+    fun observeReceiveAddress(
+        walletId: String,
+        addressType: HwFundingAddressType = HwFundingAddressType.DEFAULT,
+    ): Flow<HwReceiveAddress?> = _watcherData
+        .map { watcherData -> watcherData.receiveAddress(walletId, addressType) }
+        .distinctUntilChanged()
+
+    private fun watcherReceiveAddress(
+        walletId: String,
+        addressType: HwFundingAddressType,
+    ): HwReceiveAddress? = _watcherData.value.receiveAddress(walletId, addressType)
+
+    /** Displays the exact address currently shown by Bitkit on the device and rejects a mismatch. */
+    suspend fun verifyReceiveAddress(
+        walletId: String,
+        receiveAddress: HwReceiveAddress,
+    ): Result<Unit> = withContext(ioDispatcher) {
+        runSuspendCatching {
+            suspend fun readOnDevice() = trezorRepo.getAddress(
+                path = receiveAddress.path,
+                showOnTrezor = true,
+                scriptType = receiveAddress.addressType.trezorScriptType,
+                coin = Env.network.toTrezorCoinType(),
+            ).getOrThrow()
+
+            ensureConnected(walletId).getOrThrow()
+            val firstAttempt = runSuspendCatching { readOnDevice() }
+            val firstError = firstAttempt.exceptionOrNull()
+            val response = if (firstError == null) {
+                firstAttempt.getOrThrow()
+            } else {
+                if (!firstError.isTrezorSessionFailure()) throw firstError
+                disconnectStaleSession(walletId).getOrThrow()
+                ensureConnected(walletId).getOrThrow()
+                runSuspendCatching { readOnDevice() }
+                    .onFailure {
+                        if (it.isTrezorSessionFailure()) {
+                            disconnectStaleSession(walletId).getOrThrow()
+                        }
+                    }
+                    .getOrThrow()
+            }
+            if (response.address != receiveAddress.address) {
+                throw HwReceiveAddressMismatchError(
+                    "Address verification failed: Trezor returned '${response.address}' for " +
+                        "'${receiveAddress.path}', expected '${receiveAddress.address}'"
+                )
+            }
         }
     }
 
@@ -585,7 +663,7 @@ class HwWalletRepo @Inject constructor(
                 val ids = devices.map { it.id }.toSet()
                 val walletWatchers = watcherData.values.filter { it.walletId == walletId }
                 val fundingBalanceSats = walletWatchers
-                    .filter { it.addressType == HwFundingAddressType.DEFAULT.settingsKey }
+                    .filter { it.addressType == HwFundingAddressType.DEFAULT }
                     .fold(0uL) { acc, watcher -> acc + watcher.balanceSats }
                 HwWallet(
                     id = walletId,
@@ -649,11 +727,13 @@ class HwWalletRepo @Inject constructor(
                     val transactionDetails = event.transactionDetails
                         .filter { it.walletId == walletId }
                         .toImmutableList()
+                    val addressType = watcherId.toFundingAddressType() ?: return@withLock emptyList()
                     val watcher = HwWatcherData(
                         walletId = walletId,
-                        addressType = watcherId.toAddressTypeKey(),
+                        addressType = addressType,
                         balanceSats = event.balance.total,
                         activities = activities,
+                        receiveAddress = event.toReceiveAddress(addressType),
                     )
                     _watcherData.update { it + (watcherId to watcher) }
                     val previousIds = persistedActivityIds.getOrPut(watcherId) {
@@ -686,7 +766,6 @@ class HwWalletRepo @Inject constructor(
                     val persistedWatcher = watcher.copy(activities = immutablePersistedActivities)
                     val updatedWatcherData = _watcherData.value + (watcherId to persistedWatcher)
                     _watcherData.update { updatedWatcherData }
-
                     persistedActivityIds[watcherId] = persistedActivities.map { it.scopedId() }.toSet()
                     buildReceivedTxs(previousIds, persistedActivities, updatedWatcherData)
                 }
@@ -915,6 +994,9 @@ class HwWalletRepo @Inject constructor(
     private fun String.toWalletId(): String = substringBefore(WATCHER_ID_SEPARATOR)
 
     private fun String.toAddressTypeKey(): String = substringAfter(WATCHER_ID_SEPARATOR)
+
+    private fun String.toFundingAddressType(): HwFundingAddressType? =
+        HwFundingAddressType.entries.firstOrNull { it.settingsKey == toAddressTypeKey() }
 }
 
 private data class WatcherSettings(
@@ -957,6 +1039,8 @@ class HwPassphraseRequiredError : AppError("Passphrase needed to reopen this wal
 /** The entered passphrase opened a different wallet than the one being signed from. */
 class HwPassphraseMismatchError : AppError("Passphrase opened a different wallet")
 
+class HwReceiveAddressMismatchError(message: String) : AppError(message)
+
 /**
  * A removal asked to keep the wallet's backup data, but its tags could not be read. Raised before
  * anything is deleted, so the wallet is untouched and the removal can be retried or repeated without
@@ -966,9 +1050,25 @@ class HwBackupDataUnreadableError(cause: Throwable) : AppError("Could not read t
 
 private data class HwWatcherData(
     val walletId: String,
-    val addressType: String,
+    val addressType: HwFundingAddressType,
     val balanceSats: ULong,
     val activities: ImmutableList<Activity>,
+    val receiveAddress: HwReceiveAddress,
+)
+
+private fun Map<String, HwWatcherData>.receiveAddress(
+    walletId: String,
+    addressType: HwFundingAddressType,
+): HwReceiveAddress? = values.firstOrNull {
+    it.walletId == walletId && it.addressType == addressType
+}?.receiveAddress
+
+private fun WatcherEvent.TransactionsChanged.toReceiveAddress(
+    addressType: HwFundingAddressType,
+) = HwReceiveAddress(
+    address = nextUnusedExternalAddress.address,
+    path = nextUnusedExternalAddress.path,
+    addressType = addressType,
 )
 
 private data class HwSnapshot(
--- a/app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
+++ b/app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
@@ -148,13 +148,10 @@ fun EditInvoiceScreen(
             }
         },
         onContinueKeyboard = { keyboardVisible = false },
-        onContinueGeneral = {
-            if (onchainOnly) {
-                updateOnchainInvoice(amountInputUiState.sats.toULong())
-                onBack()
-            } else {
-                editInvoiceVM.onClickContinue()
-            }
+        onContinueGeneral = editInvoiceVM::onClickContinue,
+        onContinueOnchain = { amountSats ->
+            updateOnchainInvoice(amountSats)
+            onBack()
         },
         isLoading = isLoading,
         onClickAddTag = onClickAddTag,
@@ -164,7 +161,7 @@ fun EditInvoiceScreen(
         onClickPaymentRequest = {
             onClickPaymentRequest(amountInputUiState.sats.toULong(), walletUiState.bip21Description)
         },
-        allowsTags = !onchainOnly,
+        onchainOnly = onchainOnly,
     )
 }
 
@@ -174,13 +171,14 @@ fun EditInvoiceContent(
     amountInputViewModel: AmountInputViewModel,
     noteText: String,
     isSoftKeyboardVisible: Boolean,
-    allowsTags: Boolean = true,
+    onchainOnly: Boolean = false,
     keyboardVisible: Boolean,
     tags: ImmutableList<String>,
     onBack: () -> Unit,
     onContinueKeyboard: () -> Unit,
     onClickBalance: () -> Unit,
     onContinueGeneral: () -> Unit,
+    onContinueOnchain: (ULong) -> Unit = {},
     onClickAddTag: () -> Unit,
     onTextChanged: (String) -> Unit,
     onClickTag: (String) -> Unit,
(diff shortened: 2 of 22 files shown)


app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -127,13 +140,18 @@ fun ReceiveQrScreen(
                 cjitInvoice = cjitInvoice,
                 isNodeRunning = lightningState.nodeLifecycleState.isRunning(),
                 onchainAddress = walletState.onchainAddress,
+                hardwareAddress = hardwareReceiveState.address?.address.orEmpty(),
+                hardwareAmountSats = walletState.bip21AmountSats,
+                hardwareMessage = walletState.bip21Description,
```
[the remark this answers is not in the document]
ben-kaufman: This is intentional. The amount and note are payment-request metadata shared across the Receive tabs, not metadata owned by the Savings account; Savings and Auto already preserve them when the user changes receive method. For Trezor we rebuild the BIP21 URI with the Trezor address, so the destination cannot cross accounts while the user’s requested amount and note remain intact. ReceiveInvoiceUtilsTest also covers the hardware address plus request details case.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
```diff
@@ -0,0 +1,203 @@
...
+    fun loadAddress(walletId: String) {
+        val state = _uiState.value
+        if (state.walletId == walletId && (state.address != null || state.isLoadingAddress)) return
+        loadJob?.cancel()
+        addressUpdatesJob?.cancel()
+        _uiState.update { HwReceiveUiState(walletId = walletId, isLoadingAddress = true) }
+        addressUpdatesJob = viewModelScope.launch {
+            hwWalletRepo.observeReceiveAddress(walletId).collect { address ->
+                if (address != null && _uiState.value.walletId == walletId) {
+                    _uiState.update {
```
ovitrif, line 62: I see `verifyAddress()` snapshots `address`, but this collector can replace the displayed `address` while `isVerifyingAddress` is true. If `nextUnusedExternalAddress` advances after an inbound payment during the device prompt, Bitkit shows and copies the new address while `verifyReceiveAddress()` still validates Trezor's response against the old snapshot and completes successfully. Could we cancel and invalidate the active verification before applying a different watcher address?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -575,11 +642,53 @@ private fun ReceiveDetailsView(
                         }
                     }
                 }
+
+                ReceiveTab.TREZOR -> {
+                    hardwareAddress?.let { address ->
+                        CopyAddressCard(
+                            title = stringResource(R.string.wallet__receive_bitcoin_invoice),
+                            address = address,
```
ovitrif: I see the Trezor Details card passes the bare `hardwareAddress` as `CopyAddressCard.address`, which that component uses for both Copy and Share. After an amount or note is set, the QR view carries the full BIP21 request, but Details silently drops those terms and shares only the destination. Could we pass the built hardware BIP21 URI as `address` and the plain hardware address as `body`, matching the Savings card?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
```diff
@@ -142,11 +145,19 @@ fun EditInvoiceScreen(
             }
         },
         onContinueKeyboard = { keyboardVisible = false },
-        onContinueGeneral = { editInvoiceVM.onClickContinue() },
+        onContinueGeneral = {
+            if (onchainOnly) {
```
ovitrif: I don't see a test that exercises the new `onchainOnly` Continue branch. The added `HwReceiveViewModelTest`, `ReceiveInvoiceUtilsTest`, and `receive-onchain.xml` coverage never drives this UI path, so they would not catch it triggering `editInvoiceVM`, failing to save the hardware amount, or leaving tags visible. Could we add a focused screen test that verifies `updateOnchainInvoice` receives the amount without triggering `editInvoiceVM`, with tags hidden?

ovitrif (commented):
The receive flow can complete verification for an address that is no longer displayed. I also found two smaller gaps in the Details actions and editor coverage.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
ben-kaufman: Fixed in e9afa2a4d. A different watcher address now cancels the active verification and passphrase work before the new address is applied. Added a regression test that holds verification open, advances the watcher address, and confirms the old verification is cancelled.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
ben-kaufman: Fixed in e9afa2a4d. The Trezor Details card now uses the built hardware BIP21 URI for Copy and Share while continuing to display the plain hardware address, matching the Savings behavior.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
ben-kaufman: Added in e9afa2a4d. The focused Compose test verifies that the hardware path passes the current amount to the on-chain callback, does not call the regular Continue callback, and hides tag actions.

ovitrif:
PLS add QA notes and Manual Tests + Automated Tests sections with details like we usually do for apps PRs. (this is more about formatting than content, content is a bit already there, just not structured in the same format we usually do)

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -99,7 +111,20 @@ fun ReceiveSheet(
...
-                        onClickEditInvoice = { navController.navigateTo(ReceiveRoute.EditInvoice) },
+                        onClickEditInvoice = {
+                            isEditingHardwareInvoice.value = false
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        onClickHardwareEditInvoice = {
+                            isEditingHardwareInvoice.value = true
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        initialTab = if (hardwareWalletId != null) ReceiveTab.TREZOR else null,
```
ovitrif: I see the global one-wallet flow resolves `selectedHardwareWalletId`, but `initialTab` stays `null` because no explicit `hardwareWalletId` was supplied. After the user selects `TREZOR`, opens the hardware editor, and pops back, the `ReceiveRoute.QR` destination is recreated and the state stored with `remember` defaults to `SAVINGS` (then may auto-switch to `AUTO`), so the edited request is shown for the software wallet instead of the Trezor address. Could we preserve the selected receive tab across navigation, or return to `TREZOR` when `isEditingHardwareInvoice` initiated the editor?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -220,6 +247,14 @@ fun ReceiveSheet(
             }
         }
 
+        if (hwReceiveState.isPassphraseRequired) {
+            HwPassphrasePromptSheet(
```
ovitrif: This receive-only branch reuses `HwPassphrasePromptSheet`, whose body reads `hardware__passphrase_sign_text` and tells the user the device will “sign the transaction.” Address verification does not sign a transaction, so hidden-wallet users see a misleading spend-authorization prompt before a read-only check. Could we supply receive-specific body copy here that explains the passphrase is reopening the wallet to verify the address?

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -340,6 +342,82 @@ class HwWalletRepo @Inject constructor(
...
+                coin = Env.network.toTrezorCoinType(),
+            ).getOrThrow()
+
+            ensureConnected(walletId).getOrThrow()
+            val firstAttempt = runSuspendCatching { readOnDevice() }
+            val firstError = firstAttempt.exceptionOrNull()
+            val response = if (firstError == null) {
+                firstAttempt.getOrThrow()
+            } else {
+                if (!firstError.isTrezorSessionFailure()) throw firstError
```
ovitrif: I couldn't find coverage for the session-failure retry branch in `verifyReceiveAddress`: `HwWalletRepoTest` exercises watcher lookup, scan fallback, and final address mismatch, but never makes the first `getAddress` call fail with a session error. If the disconnect/reconnect or second device read regresses, the claimed recovery path fails while the current tests still pass. Could we add a focused test that makes the first device read fail with a session error and verifies reconnect plus a successful second read?

ben-kaufman:
Updated the PR description with structured QA Notes, Manual Tests, and Automated Tests sections. I left the manual checks unchecked until they are run on Android.

coreyphillips (dismissed):

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -120,7 +132,20 @@ fun ReceiveSheet(
...
-                        onClickEditInvoice = { navController.navigateTo(ReceiveRoute.EditInvoice) },
+                        onClickEditInvoice = {
+                            isEditingHardwareInvoice.value = false
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        onClickHardwareEditInvoice = {
+                            isEditingHardwareInvoice.value = true
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        initialTab = if (hardwareWalletId != null) ReceiveTab.TREZOR else null,
```
ovitrif: I see the global one-wallet flow resolves `selectedHardwareWalletId`, but `initialTab` stays `null` because no explicit `hardwareWalletId` was supplied. After the user selects `TREZOR`, opens the hardware editor, and pops back, the `ReceiveRoute.QR` destination is recreated and the state stored with `remember` defaults to `SAVINGS` (then may auto-switch to `AUTO`), so the edited request is shown for the software wallet instead of the Trezor address. Could we preserve the selected receive tab across navigation or return to `TREZOR` when `isEditingHardwareInvoice` initiated the editor?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -296,6 +323,14 @@ fun ReceiveSheet(
             }
         }
 
+        if (hwReceiveState.isPassphraseRequired) {
+            HwPassphrasePromptSheet(
```
ovitrif: I see this receive-only branch reuses `HwPassphrasePromptSheet`, whose body reads `hardware__passphrase_sign_text` and says the device will “sign the transaction.” Address verification does not sign anything, so hidden-wallet users are shown a misleading spend-authorization prompt before a read-only check. Could we pass receive-specific body copy that explains the passphrase is reopening the wallet to verify its address?

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
ovitrif, line 401: I couldn't find coverage for the session-failure retry branch in `verifyReceiveAddress`: `HwWalletRepoTest` exercises watcher lookup, scan fallback, and final address mismatch, but never makes the first `getAddress` call fail with a session error. If the disconnect, reconnect, or second device read regresses, the claimed recovery path fails while the current tests still pass. Could we add a focused test that makes the first device read fail with a session error and verifies reconnect plus a successful second read?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
```diff
@@ -0,0 +1,212 @@
...
+                .onFailure { error ->
+                    if (error is CancellationException) return@onFailure
+                    if (_uiState.value.walletId == walletId) {
+                        _uiState.update { it.copy(isLoadingAddress = false, addressLoadFailed = true) }
+                    }
+                }
+        }
+    }
+
+    fun retryAddress() {
```
ovitrif, line 92: I don't see coverage that makes `getReceiveAddress` fail and then exercises `retryAddress`; `HwReceiveViewModelTest` only covers successful loading, cancellation, watcher updates, and passphrase verification. If the new `addressLoadFailed` state or retry behavior regresses, users can remain stuck without an address while the current tests still pass. Could we add a focused test that fails the first load, asserts `addressLoadFailed`, retries, and verifies the recovered address?

ovitrif (changes requested):
The global hardware edit path can return on a software-wallet QR. I also found misleading passphrase copy and two untested recovery paths.

piotr-iohk:
Tested on emulator and device (Samsung S22 + Trezor Safe 7).

Generally all good. Able to present the receive address, verify the address on the device, etc.

One observation (looks intentional in code, but may be misleading to the user): with 2+ paired hardware wallets (e.g. standard + passphrase), Home → Receive does not show the Trezor tab, so there is no hardware receive address from that entry point. Fine to defer to a follow-up PR — if that's the case, let's create a ticket for this.

### Steps to reproduce
1. Pair a Trezor (standard wallet). Home should show one hardware wallet tile.
2. Add a passphrase wallet on the same device (Paired → Passphrase → enter passphrase). Home should now show two hardware wallet tiles.
3. From the main/home screen, tap Receive.
4. Only Savings / Spending tabs are shown. No Trezor tab, so you cannot get a hardware receive address.

### Expected (or less misleading)
Home → Receive either shows a Trezor tab (or a wallet picker) when more than one hardware wallet is paired, or makes it clear that hardware receive is only available from the specific wallet screen.

### Workaround
Open the specific hardware wallet screen first, then Receive. That path still shows the address.

### Recording

https://github.com/user-attachments/assets/b6a4dc8c-feda-49f4-a98e-9a930daf9dce

piotr-iohk (dismissed):
Approve. Single-wallet receive looks good on emulator and device. See the QA comment for the multi-wallet Home Receive note (fine to defer).

ben-kaufman:
The single-wallet hardware Receive flow is green and the earlier change requests are resolved. I filed #1204 for the requested Home Receive wallet picker when multiple hardware wallets are paired. @ovitrif please re-review the current head so the stale change-request state can be cleared.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
```diff
@@ -0,0 +1,212 @@
...
+                        it.copy(address = address, isLoadingAddress = false, addressLoadFailed = false)
+                    }
+                }
+            }
+        }
+        loadJob = viewModelScope.launch {
+            hwWalletRepo.getReceiveAddress(walletId)
+                .onSuccess { address ->
+                    if (_uiState.value.walletId == walletId) {
+                        _uiState.update { it.copy(address = address, isLoadingAddress = false) }
```
ovitrif: The fallback result can overwrite a newer watcher address. If the watcher emits address B after the repository's final check but before this continuation runs, the collector stores B and this update then restores scanned address A; the `StateFlow` need not emit B again, and verification can still succeed for the stale path. Could we preserve an address already supplied by `observeReceiveAddress`, and add an interleaving test that emits B before completing the load with A?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -120,7 +132,24 @@ fun ReceiveSheet(
...
-                        onClickEditInvoice = { navController.navigateTo(ReceiveRoute.EditInvoice) },
+                        onClickEditInvoice = {
+                            isEditingHardwareInvoice.value = false
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        onClickHardwareEditInvoice = {
+                            isEditingHardwareInvoice.value = true
+                            navController.navigateTo(ReceiveRoute.EditInvoice)
+                        },
+                        initialTab = if (hardwareWalletId != null || isEditingHardwareInvoice.value) {
```
ovitrif: This restoration branch has no focused regression test. The new Compose test covers the on-chain editor callback, and the journey manually selects Trezor without entering the editor, so neither would fail if returning from Edit Invoice opened Savings again. Could we add coverage that opens the Trezor editor from global Receive, returns, and asserts that the Trezor tab is still selected?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -296,6 +327,15 @@ fun ReceiveSheet(
             }
         }
 
+        if (hwReceiveState.isPassphraseRequired) {
+            HwPassphrasePromptSheet(
+                isVerifying = hwReceiveState.isVerifyingPassphrase,
+                onSubmit = hwReceiveViewModel::submitPassphrase,
+                onDismiss = hwReceiveViewModel::dismissPassphrase,
+                bodyText = stringResource(R.string.hardware__passphrase_verify_address_text),
```
ovitrif: The receive-specific `bodyText` override is not exercised by a test. The ViewModel test proves reconnect behavior but never renders this sheet, and the receive journey uses a standard wallet. Could we add focused Compose coverage asserting that address verification shows the receive-specific passphrase explanation rather than the transaction-signing copy?

ovitrif (changes requested):
The initial hardware receive load can still revert to a stale watcher address. I also found two regression-coverage gaps in the addressed UI fixes.

coreyphillips (dismissed):

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
ben-kaufman: Fixed in signed commit d42064c. The fallback now fills the address only while watcher state is still empty, so a newer watcher address cannot be overwritten. Added a deferred interleaving regression that publishes B before the fallback returns A.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ben-kaufman: Added in signed commit d42064c. The receive edit source is now retained in ReceiveInvoiceEditState, and the Compose regression selects Trezor, opens the hardware editor, returns to the QR screen, and verifies the hardware address is loaded again from the reselected Trezor tab.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
ben-kaufman: Added in signed commit d42064c. The receive-specific prompt now renders through a focused production wrapper, with Compose coverage asserting that the address-verification copy is shown and the transaction-signing copy is absent.

ben-kaufman:
Fixed all three review findings in signed commit `d42064c`: preserved newer watcher addresses against the fallback race, added the hardware Edit Invoice return-path regression, and added receive-specific passphrase-copy coverage. The focused unit suite passes, the Compose test suite compiles, and detekt passes. @ovitrif please re-review when fresh CI is ready.

ovitrif (dismissed):
utACK

ovitrif (approved):
### QA Notes

Pixel_8_Pro emulator (sdk_gphone16k_arm64), regtest, Trezor User Env T2T1 via Bridge.

1. Home → Receive → Trezor: the QR loaded with hardware-wallet styling.
2. Show Details → Verify on Device showed a loading state; after device approval the same `bcrt1` address remained and no error appeared.
3. Hardware Edit Invoice hid tags, accepted 12,345 sats plus a note, and returned to the Trezor QR and details with the same hardware address.
4. Passphrase Trezor: Verify opened the receive-specific passphrase sheet. Submitting after an emulator restart returned “Device not found nearby”, so this box is left unchecked.
5. With Verify pending, funding the displayed address and mining made Bitkit show a 200,000 sat receive, switch to a new `bcrt1` address, and leave Verify idle.

Approve.

piotr-iohk (approved):
Retested current head `fd877013` on Samsung S22 + Trezor Safe 7 and on the Trezor emulator.

I do not see the missing `Tab-trezor` issue that was raised on the iOS counterpart PR. Opening Receive from the hardware-wallet screen still shows the Trezor tab, the hardware address, and Verify on Device.

E2E now covers this path as well (`@hardware_wallet_2`): it opens Receive, waits for `Tab-trezor`, and funds the Trezor-tab address.

LGTM.

PR #1195, opened 2026-08-28
fix: send source & fee progress ui/ux

### LINKED ISSUE #1203: Trezor send UI polish
Part of #998.

**Related (iOS):** synonymdev/bitkit-ios#696

## Scope

Send UI polish after hardware send:

- Switching the funding source to Trezor keeps the source chip filled and shows a small spinner until the hardware fee is ready.
- Confirm keeps the previous fee with a spinner while the estimate refreshes.
- Swipe-to-confirm stays visible and stays blocked until the hardware fee is present.

ovitrif:
### Description

This PR polishes the Trezor send UI from #1187, where switching to hardware left a rough loading chip: the track vanished, the label disappeared behind a spinner, and a Material ripple flashed while the device warmed up.

This PR:

1. Rebuilds that funding-source chip so switching to Trezor keeps the filled track, animates the label width as the name appears, and swaps only the transfer icon for a small tinted spinner. Savings and Spending switches stay instant, with no loader flash.
2. Replaces the fee-rate icon with a matching spinner and keeps the previous fee text until the next estimate lands.
3. Keeps swipe-to-confirm's gradient track visible after switching to Trezor.
4. Refactors onchain fee UI logic to group fee rate, sats, estimates, and loading into one state.
5. Pops Speed immediately after a rate is chosen, matching iOS's tap-to-select flow, while blocking conflicting hardware send actions until the new maximum and fee are ready.
6. Keeps the Trezor source chip filled while Amount Continue prepares the hardware fee.

### Preview

https://github.com/user-attachments/assets/66562b0e-ef67-4fff-aabe-211452c03883

### Linked Issues/Tasks

Fixes #1203

#### Manual Tests

- [x] **1.** Send Amount or Confirm → switch to Trezor: the chip keeps its filled track, the label stays visible and animates to the Trezor name, and a small tinted spinner replaces only the transfer icon until the device fee is ready.
- [x] **2a.** Send Confirm → switch Trezor → Spending: no source loader flash.
  - [x] **2b.** Send Confirm → switch Trezor → Savings: no source loader flash.
- [x] **3.** Send Confirm → switch Spending → Savings: swipe-to-confirm does not dim or flash.
- [x] **4.** `regression:` after Trezor is selected and loading ends: swipe-to-confirm keeps its gradient track and can still be swiped.
- [x] **5.** Trezor Send Amount → Continue: the Trezor chip stays filled with no blue outline until Confirm.
- [x] **6.** Trezor Send Confirm → Speed → pick another rate: Speed closes immediately like iOS, Confirm keeps the previous fee with a spinner, and speed, source, and swipe actions remain blocked until the new fee lands.
- [x] **7a.** `regression:` Savings Send Confirm → Speed: Fast, Normal, and Slow still show estimates like master, and the current rate stays selected.
  - [x] **7b.** `regression:` Savings Send Confirm → Speed → pick Fast, Normal, or Slow: returns to Confirm and the fee/speed row updates.
  - [x] **7c.** `regression:` Savings Send Confirm → Speed → Custom → set a valid rate → Continue: Confirm shows the custom fee.
  - [x] **7d.** `regression:` Savings Send Fee Custom → rate below min or above max: toast blocks continue, same as master.
  - [x] **7e.** `regression:` Savings Send Confirm → Speed → back without changing: Confirm fee and speed stay the same.
- [x] **8.** `regression:` unified Send Confirm → Speed → Instant: still switches to Spending like master.

#### Automated Checks

- Unit tests added or extended in `AppViewModelSendFlowTest.kt`: cover hardware-only source loading, keeping a valid amount while Savings refreshes, retaining the previous on-chain fee while estimates refresh, returning from Speed before hardware max finishes, resetting selected UTXOs from the previous rate, and confirming that a missing hardware fee does not block navigation to signing.
- Instrumented tests added or updated in `SendAmountContentTest.kt`: use the funding-source loading flag to disable source and continue, and ignore source-chip taps while Amount Continue is loading.
- Unit tests updated in `SendFeeViewModelTest.kt`: fee-sheet init still disables unaffordable rates and validates custom min/max against the unified on-chain fee estimates.


app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3758,30 +3762,33 @@ class AppViewModel @Inject constructor(
...
+        val rate = FeeRate.fromSpeed(currentState.speed)
 
-        _sendUiState.update {
-            it.copy(
-                fees = feesMap.toImmutableMap(),
-                fee = if (it.payMethod == SendMethod.ONCHAIN) SendFee.OnChain(currentFee) else it.fee,
+        updateOnchainFeeUi {
+            OnchainFeeUi(
+                rate = rate,
+                sats = estimates[rate]?.takeIf { sats -> sats > 0 },
```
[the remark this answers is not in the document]
ovitrif: The confirmation is not blocked here. The `?: return` exits only `handleSanityChecks`; `onSwipeToPay` then sees no warning dialog and sets `shouldConfirmPay = true`, after which `PayConfirmed` navigates to hardware signing. This is now covered by `missing hardware fee does not block confirmation` in `AppViewModelSendFlowTest.kt` (`1dcc90bc9`). Ben's base already represented failed estimates as zero, so this PR does not introduce a blocked-confirmation regression.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2098,48 +2098,40 @@ class AppViewModel @Inject constructor(
...
     fun setTransactionSpeed(speed: TransactionSpeed) {
+        val previous = _sendUiState.value
+        _sendUiState.update {
+            it.copy(
+                payMethod = SendMethod.ONCHAIN,
+                speed = speed,
+                onchainFeeUi = it.onchainFeeUi.copy(isLoading = true),
+            )
+        }
+        setSendEffect(SendEffect.PopBack(SendRoute.Confirm))
```
ben-kaufman, line 2110: Returning to Confirm immediately is fine, but the hardware speed preparation has no interaction or ownership guard. While it runs, confirmation, fee selection, and source switching remain active. A swipe can use the old fee and maximum with the new speed, while another selection can let an older result overwrite current state. Could we disable these hardware interactions until preparation finishes and make the job latest-wins?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2169,17 +2161,22 @@ class AppViewModel @Inject constructor(
         onchainSendRefreshJob?.cancel()
         val selected = current.selectedFundingSource()
         val selectedIndex = sources.indexOf(selected).takeIf { it >= 0 } ?: 0
-        when (val nextSource = sources[(selectedIndex + 1) % sources.size]) {
+        val nextSource = sources[(selectedIndex + 1) % sources.size]
+        _sendUiState.update {
+            it.copy(isFundingSourceLoading = nextSource is SendFundingSource.Hardware)
```
ben-kaufman, line 2167: Only hardware destinations enter the loading state here. When switching from hardware to Savings, the old hardware `isAmountInputValid` keeps Continue and swipe enabled until the Savings refresh completes. A queued Continue can then proceed without rechecking the newly invalid amount. Could we update Savings validity immediately, or keep these controls disabled for the switch, and cover an amount above the Savings balance?

app/src/main/java/to/bitkit/ui/components/NumberPadActionButton.kt
```diff
@@ -34,54 +34,47 @@ fun NumberPadActionButton(
...
             modifier = modifier
                 .requiredHeight(height)
                 .primaryButtonStyle(
-                    isEnabled = enabled && !isLoading,
+                    isEnabled = enabled || isLoading,
                     shape = buttonShape,
                 )
-                .alphaFeedback(enabled = enabled && !isLoading)
+                .animateContentSize(animationSpec = tween(durationMillis = 200))
+                .clickableAlpha(enabled = enabled && clickable && !isLoading, onClick = onClick)
```
ben-kaufman, line 55: When `isLoading` is true or `clickable` is false, `clickableAlpha` returns the modifier unchanged. Since this Row replaces Material `Button`, the chip no longer exposes a button role or disabled semantics while it remains visible. Could we preserve those semantics here, or retain a Material button surface while changing only the content?

app/src/main/java/to/bitkit/ui/components/GradientCircularProgressIndicator.kt
```diff
@@ -35,7 +36,7 @@ fun GradientCircularProgressIndicator(
         label = "rotation"
     )
 
-    val brush = remember { Brush.sweepGradient(listOf(Color.Transparent, Colors.White)) }
+    val brush = remember { Brush.sweepGradient(listOf(Color.Transparent, tint)) }
```
ben-kaufman, line 39: The brush is remembered without `tint` as a key. During a Savings to Trezor switch, the same loader can receive the blue tint after first composing with the Savings color, but its brush remains unchanged. Could we use `remember(tint)` here?

ovitrif: I could not reproduce stale tint after repeated Savings → Trezor transitions, and the attached recording shows the correct spinner color gets updated for every of the many switches. Without a reproduction or failing visual test, changing the implementation for this hypothetical case is not warranted.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Reproduced this with distinct hardware fee rates: Confirm could show the previous fee while signing used the newly selected rate. Fixed in 9673a170f. Confirmation, fee selection, and source switching are now disabled while hardware fee preparation runs. Returning to Confirm immediately remains intentional and matches iOS's tap-to-select flow.

I did not add latest-wins job ownership. Selecting a rate immediately leaves Speed, and fee selection stays disabled on Confirm until preparation finishes, so ownership for a second selection the UI does not expose would be defensive overengineering. The existing fee refresh is still cancelled before new preparation starts so stale fee work cannot clear the loading state.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: The Savings switch is intentionally immediate and does not need a loading state. Manual testing did not reproduce an invalid transition or stale enabled action. Reintroducing blanket disabling would restore the UI regression this PR removes. Please provide reproducible steps if this occurs in practice; otherwise, no change.

app/src/main/java/to/bitkit/ui/components/NumberPadActionButton.kt
ovitrif: Accessibility semantics are not an acceptance criterion currently enforced for this component or consistently implemented by the app. Adding isolated semantics work is outside this visual loading-state fix. No change.

piotr-iohk (approved):
LGTM

Source chip stays filled with an icon-only spinner on Trezor, no flash switching back to Spending/Savings, swipe keeps its track after load, and Speed pops immediately while the previous fee stays until the new estimate lands.

PR #1196, opened 2026-08-30
fix: polish payment request ui

ovitrif:
This PR polishes the Payment Request screens so spacing, contact rows, amount keypad, and success layout match the `Bitkit - Handoff v62` frames, using the existing Bitkit components (Contacts rows, Receive amount keypad, backup success) instead of the GPT layout from #1172.

Figma is intent. Existing Bitkit components win when Figma disagrees. The screens were captured with a Maestro flow on a Pixel emulator and measured against the handoff frames band by band. Frames: [Payment Requests](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48056-274346), [Payment Request from Receive > Contacts, Contact profile or Payment Requests](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206543), [Payment Request from Receive > Edit invoice](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48049-273376).

### Description

- Splits the Payments tab list into Incoming and Outgoing, with the contact name as row title, the note as subtitle, glow cards on incoming and a pending label on outgoing amounts.
- Aligns contact rows with Contacts (`Text13Up` + `BodyMSB`) and runs the recipient list to the sheet floor; tapping a row chooses the recipient. Following the two handoff flows, the CONTACTS label shows when Choose Recipient opens from the Receive invoice editor and is hidden from the Payment Requests list, Contacts, and contact profile paths.
- Rebuilds the amount step on the Receive amount layout: 352dp keypad as in the amount frame, unit toggle under the amount, leftover space split 4:1 around it, keypad shrinking on short windows so Continue stays on screen.
- Details step drops the header back arrow (system back still pops) and uses the frame spacing between amount, note, recipient, and expiry. Reached from the amount step it shows the fiat value above the amount; reached from the invoice editor it shows the AMOUNT label, as the two details frames specify.
- Pins the sent layout to check, PAYMENT REQUESTED, card, then OK, with the illustration capped by the space left for the content under it.
- Adds a Maestro flow at `.maestro/payment-request/payment-request.yaml` that seeds the frame fixture through a debug-only deep link on the regtest flavor and captures one screenshot per frame. `bitkit://dev-fixture/payment-request?clear` removes the fixture; release builds compile it out.

### Preview

| Screen | Figma (Handoff v62) | Before | After |
| --- | --- | --- | --- |
| [List](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48056-274346) |  |  |  |
| [Recipient](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206543) |  |  |  |
| [Recipient (from invoice)](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48049-273376) |  |  |  |
| [Amount empty](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206614) |  |  |  |
| [Amount filled](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206606) |  |  |  |
| [Details](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=47993-199959) |  |  |  |
| [Sent](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206596) |  |  |  |

#### Manual Tests

- [ ] **1.** Home → menu → Subscriptions → Payments: list shows Incoming cards with Dismiss and Pay, then Outgoing rows with contact name, note, amount and `pending`, Request Payment button under the list.
- [ ] **2.** Payments → Request Payment: Choose Recipient in-sheet shows the RECIPIENT field and the contact rows to the sheet floor, no CONTACTS label, no Send button.
- [ ] **3.** Choose Recipient → tap a contact: Amount opens with the contact avatar in the header, keypad under the amount, Continue disabled at 0.
- [ ] **4.** Amount → enter 50 000 → Continue: Payment Request details show the fiat value above the amount, note, recipient card, expiry tabs; header has no back arrow; system back returns to Amount.
- [ ] **5.** Details → Send Request: Sent shows check, PAYMENT REQUESTED, the recipient card, OK; OK returns to the Payments tab.
- [ ] **6.** Receive → invoice editor → contact picker: Choose Recipient opens with the CONTACTS label; picking a contact lands on details with the AMOUNT label and the pencil.
- [ ] **7.** `regression:` Payments tab → Incoming card → Pay / Dismiss: still works.
- [ ] **8.** `regression:` Send → contact picker: rows still show the truncated key above the name.

#### Automated Checks

- Instrumented tests modified: `CreatePaymentRequestScreenTest.kt` covers the CONTACTS header per entry path and the details fiat caption vs AMOUNT label; `PaymentRequestsScreenTest.kt` covers the Incoming section with Pay and Dismiss and the Outgoing `pending` label.
- Maestro flow added: `payment-request.yaml` seeds the fixture and captures the seven handoff frames on a Pixel emulator; layout measured against the frames with a band-by-band ledger.
- `just compile`, `just test`, and `just lint` pass locally on the pushed head.


app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
```diff
@@ -363,55 +383,22 @@ private fun paymentRequestSections(
     now: Instant,
 ): PaymentRequestSections {
     val pendingIds = pending.mapTo(mutableSetOf()) { it.id }
-    val active = requests.filter { request ->
-        request.id in pendingIds ||
-            request.direction == PaykitPaymentRequestDirection.Outgoing &&
+    val incoming = requests.filter { it.id in pendingIds }
```
jvsena42, line 386: Low: the new INCOMING section has no positive test coverage.

Every `PaymentRequestsContent` test passes `pending = persistentListOf()` (`PaymentRequestsScreenTest.kt:138`, `:168`, `:196`), so `incoming` is always empty, and the only assertion about it is the negative one at `:155` (`"INCOMING"` `assertDoesNotExist()`). The three tests that do exercise incoming requests render `PaymentRequestsSheetContent` — a different composable.

Invert this predicate to `it.id !in pendingIds` and every incoming request silently vanishes from the Payments tab, with no Pay and no Dismiss — and all six androidTests still pass, `just test` and CI stay green.

One `PaymentRequestsContent` test with a non-empty `pending` asserting `"INCOMING"` is displayed and the row's Pay/Dismiss exist would close it.

(The `.maestro` flow doesn't cover this: nothing in `.github/workflows/` references maestro, and it runs against the fixture — `AppViewModel.kt:4970-4973` short-circuits `createPaymentRequest` and calls `onCreated` without touching the repo. Fine for its stated visual purpose, just not regression protection.)

app/src/main/java/to/bitkit/ui/components/PubkyContactRow.kt
```diff
@@ -32,7 +32,7 @@ fun PubkyContactRow(
     modifier: Modifier = Modifier,
     isSelected: Boolean? = null,
     isEnabled: Boolean = true,
-    verticalPadding: Dp = 12.dp,
+    verticalPadding: Dp = 24.dp,
```
jvsena42, line 35: Low: this restyle is shared, so it silently changes `SendContactSelectScreen` — which has its own Figma frame that this PR doesn't check.

`SendContactSelectScreen.kt:119` is the only other caller and passes no `verticalPadding`, so it inherits all three changes: `12.dp` → `24.dp` (row height roughly 68dp → 92dp), `BodyS` → `Text13Up`, and `BodySSB` → `BodyMSB`. `Text13Up` uppercases at `Text.kt:387`, so the truncated pubkey now renders in caps there too.

Net effect: the Send contact picker shows ~2 fewer contacts per screen and a differently-cased key, with no design pass — and `docs/screens-map.md:208` gives that screen its own frame (`Send (Contact) (Lightning) › Send to contact`).

The change is *correct* for the payment-request screens — `ContactsScreen.kt:221-231` already uses `Text13Up` + `BodyMSB` with `vertical = 24.dp`, and frame `48185:303457` shows the uppercased key. It's only the spillover that's unreviewed. Either verify the Send frame too, or keep the default at 12dp and pass `verticalPadding = 24.dp` explicitly from the payment-request call site.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -85,6 +85,18 @@ import kotlin.time.Duration.Companion.hours
 import kotlin.time.ExperimentalTime
 import kotlin.time.Instant
 
+/** Keypad height for four 88dp key rows, matching the Figma amount frame. */
+private val NUMBER_PAD_HEIGHT = 352.dp
```
jvsena42, line 91: Low: giving the keypad a *fixed* height bypasses the 75dp key cap every other Bitkit keypad honours.

`NumberPad.kt:116-119`:
```kotlin
constraints.hasFixedHeight -> maxHeight / ROWS
else -> (availableHeight / ROWS).coerceIn(minButtonHeight, idealButtonHeight)
```
Passing `Modifier.height(keypadHeight)` at `:181-183` takes the first branch, so keys are 352/4 = 88dp and `idealButtonHeight = 75.dp` (`NumberPad.kt:63`) never applies. `ReceiveAmountScreen.kt:242-248` passes `availableHeight = maxHeight` with no fixed height, so it takes the `else` branch and clamps to ≤75dp.

On any phone tall enough not to clamp, this keypad's rows are 13dp taller each — 52dp overall — than the Receive keypad this was rebuilt on, so the two screens are visibly inconsistent.

Either pass `availableHeight` to match Receive, or if 88dp is what the amount frame actually specifies, raise `idealButtonHeight` so both agree. I can't adjudicate which is intended — the amount frame wasn't among the three referenced node ids.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -250,17 +265,10 @@ internal fun PaymentRequestDetailsContent(
...
-        rememberMoneyText(
-            sats = initialDraft.amountSats.coerceAtMost(Long.MAX_VALUE.toULong()).toLong(),
-            reversed = true,
-            showSymbol = true,
-        )?.let {
-            Caption13Up(text = it.removeAccentTags(), color = Colors.White64)
-        }
+        SheetTopBar(titleText = stringResource(R.string.wallet__payment_request))
+        VerticalSpacer(16.dp)
+        Caption13Up(text = stringResource(R.string.wallet__payment_request_amount), color = Colors.White64)
```
jvsena42: Low: the details step no longer shows the fiat value of the amount being requested.

This caption replaced `rememberMoneyText(sats, reversed = true, showSymbol = true)`, which rendered the *secondary* unit. `MoneyDisplay` just below (`:278`) renders only `currencies.primaryDisplay`, and there's no `UnitButton` on this step.

So a user on BTC primary display, reviewing a request before sending it, sees `₿ 14 500` and no fiat anywhere — and no way to toggle. They have to back out to the amount step via the pencil at `:282` to see it.

The PR body says this is Figma intent, and it may well be, but the details frame isn't among the referenced node ids so I can't confirm. If the frame specifies the caption, ignore this; otherwise restore the secondary money text alongside the label.

Separately, one measurable delta on a frame I *could* check: the OUTGOING header at `PaymentRequestsScreen.kt:311` has `padding(top = 24.dp)` inside a `LazyColumn` with `spacedBy(16.dp)`, so the gap is 40dp. In `48185:303376` the last incoming card ends at y=338 and `Header 7` starts at y=354, with 16dp baked into the header's own 34dp height — an effective 32dp. `padding(top = 16.dp)` matches.

jvsena42 (commented):
Ran the Figma comparison — the PR body's "Figma" column links to PNG exports on a gist rather than frames, but `PaymentRequestFixtureRuntime.kt:15` names three node ids (`48185:303457`, `48185:303440`, `48185:303376`) and `docs/screens-map.md:3` gives the file key, so I resolved them directly.

**The three available frames largely match.** `48185:303376` confirms the INCOMING → 2 glow cards with Dismiss/Pay → OUTGOING structure, with `pending` replacing the fiat line. `get_variable_defs` returns only `Gray/Gray 6`, `Gray/Gray 5`, `Brand/Bitkit` and the `Alpha/White` set — and the code uses `Colors.Gray6`/`Gray5`/`Brand`/`White64|16|10` throughout, so **no hardcoded colours or type were introduced**. `48185:303457` vs `48185:303440` confirm the CONTACTS-header toggle is real design intent, and `showContactsHeader` matches both.

**Not compared:** the amount, details and sent steps — no frame for those was referenced anywhere in the PR, so I can't adjudicate F3 or F4 below. If you have those node ids, worth adding them next to the existing comment.

Screens-map rows at head: `CreatePaymentRequestScreen.kt` → `Payment Request (…) › Request payment enter amount`; `PaymentRequestsScreen.kt` → `Payment Requests › Payment Requests`; `SendContactSelectScreen.kt` → `Send (Contact) (Lightning) › Send to contact`. `ReceiveSheet.kt` isn't a `*Screen.kt` so it's out of the map's scope.

Also checked clean: `Money.kt` has no unit or rounding change (`secondaryText` only swaps which composable fills the existing secondary slot); the details back-arrow removal is safe since `SheetTopBar.onBack` is nullable and system back still pops; both `navigateTo(PaymentRequestRecipient)` sites set `showRecipientContactsHeader` immediately before navigating, so there's no stale-value path; `incoming ∪ outgoing` is exactly the old `active` set so history is unchanged; and conventions are right throughout (PascalCase testTags, `ImmutableList` at the boundary, no trailing comma after `modifier =`, strings alphabetical, one changelog fragment).

For context on greptile's 24dp thread — the `Contact` instance in both recipient frames is 96dp tall, closer to the 30dp override than to 24dp. Design data only; not reopening it.

Four LOW notes inline. Nothing blocking — Paykit UI is behind `isPaykitEnabled` (default off).

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -474,7 +496,7 @@ internal fun PaymentRequestRecipientContent(
                 PubkyContactRow(
                     profile = contact,
                     onClick = { onSelected(target) },
-                    verticalPadding = 16.dp,
+                    verticalPadding = 30.dp,
```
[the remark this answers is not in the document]
ovitrif: The recipient frames on the `Bitkit - Handoff v62` page (`48030:206543`, `48049:273376`) place a 96dp `Contact` row instance: a 40dp avatar with 28dp above and below. The 24dp Contacts default gives an 84dp row, so this list now passes 28dp explicitly to land on the frame's row height.

Fixed in e878f65

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -151,49 +163,55 @@ internal fun PaymentRequestAmountContent(
...
+                        modifier = Modifier.padding(end = 8.dp)
                     )
                 }
             },
         )
         BoxWithConstraints(modifier = Modifier.weight(1f)) {
-            val availableHeight = this.maxHeight
-
-            Column(modifier = Modifier.fillMaxSize().padding(horizontal = 16.dp)) {
+            val keypadHeight = (maxHeight - AMOUNT_STEP_FIXED_HEIGHT).coerceIn(NUMBER_PAD_MIN_HEIGHT, NUMBER_PAD_HEIGHT)
```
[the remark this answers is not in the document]
ovitrif: Dropped the 200dp floor. The keypad now takes whatever height is left under the amount and above Continue, capped at the frame's 352dp, so the column never overflows.

Fixed in e878f65

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -511,51 +533,56 @@ internal fun PaymentRequestSentContent(
...
-            },
-        )
-        VerticalSpacer(32.dp)
-        PrimaryButton(
-            text = stringResource(R.string.common__ok),
-            onClick = onDone,
-        )
-        VerticalSpacer(16.dp)
+        BoxWithConstraints(modifier = Modifier.weight(1f)) {
+            val checkSize = maxHeight * SENT_CHECK_HEIGHT_FRACTION
```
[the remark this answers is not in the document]
ovitrif: The illustration is now capped by the height left after the headline, subtitle, request card and OK button, so the button stays reachable on short windows.

Fixed in e878f65

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -5084,6 +5093,12 @@ class AppViewModel @Inject constructor(
     }
 
     private fun processDeeplink(uri: Uri) = viewModelScope.launch {
+        PaymentRequestFixtureRuntime.fixtureFor(uri)?.let { fixture ->
+            settingsStore.setIsPaykitEnabled(true)
+            paymentRequestFixture.update { fixture }
+            return@launch
+        }
```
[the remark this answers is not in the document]
ovitrif: The fixture link is now accepted only on the regtest dev flavor, and `bitkit://dev-fixture/payment-request?clear` removes an installed fixture.

Fixed in e878f65

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
ovitrif: Added a `PaymentRequestsContent` test with a non-empty pending list that asserts INCOMING is displayed and the row's Pay and Dismiss exist.

Fixed in e878f65

app/src/main/java/to/bitkit/ui/components/PubkyContactRow.kt
ovitrif: Checked the Send to contact frame on the handoff page (`46920:148718`): it uses the same Contacts row style as the recipient frames, the uppercase truncated key above the name at the same row height, so the shared restyle is what that screen's design asks for as well. Added a regression line for it to the QA Notes.

Skipped in e878f65

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
ovitrif: The amount frame (`48030:206614`) places a 347dp `Keyboard/NumPad` instance for four rows, about 87dp per key, so the 352dp keypad follows that frame rather than the Receive keypad's 75dp cap. It shrinks below that when the sheet is shorter (thread above).

Skipped in e878f65

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
ovitrif: You were right about the fiat line. On the handoff page the details frame reached from the amount step (`47993:199959`) shows the fiat value in the label slot above the amount, and only the invoice-editor flow's details frame (`48049:273665`) carries the AMOUNT label. The step now shows the fiat caption when the draft came through the amount step and the AMOUNT label when it came from the invoice editor. The OUTGOING header top padding is 16dp now.

Fixed in e878f65

ovitrif:
> jvsena42: Not compared: the amount, details and sent steps — no frame for those was referenced anywhere in the PR.

The three node ids you resolved were on a scratch page that no longer exists. Everything now points at the `Bitkit - Handoff v62` page: the Preview table links each frame, and the fixture KDoc lists them. The two flows there differ, and this pass follows them: [from Receive > Contacts, Contact profile or Payment Requests](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206543) has no CONTACTS header, the amount steps ([empty](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206614), [populated](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206606)), [details with the fiat caption](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=47993-199959) and [sent](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48030-206596); [from Receive > Edit invoice](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48049-273376) has the CONTACTS header and [details with the AMOUNT label](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48049-273665). The [list frame](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit?node-id=48056-274346) is unchanged.

jvsena42 (commented):
Clean — no findings this pass, and the Figma gap I left open last time is now closed.

**Figma: I compared the five frames I previously couldn't.** Your follow-up superseded the three scratch-page ids I'd resolved, so I went through the `Bitkit - Handoff v62` frames you linked.

- **Details, both variants** (`47993:199959` fiat caption, `48049:273665` AMOUNT label) — structurally identical in the file: `Amount` group with `Header 6` at y=0 h=18 and the amount at y=34 h=36, text field at y=94, recipient at y=196 h=98, expiry at y=318 h=62. Only the `Header 6` *text* differs between them, which is exactly what `fromInvoiceEditor` switches at `CreatePaymentRequestScreen.kt:281-282`. Matches.
- **This settles F3 from my last review, in your favour.** The fiat caption is design intent. More to the point, my concern was that a BTC-primary user would see `₿ 14 500` and no fiat anywhere with no way to toggle — that's wrong: both frames render the recipient card with *both* units stacked (`₿ 14 500` / `$ 14.50`), and `MoneyCell` (`Money.kt:58-77`) renders primary plus `primaryDisplay.not()` in exactly that shape. Fiat is present on both variants. Nothing to change.
- **Amount** (`48030:206606`) — back arrow, "Amount" title, contact avatar in the action slot, fiat caption above the amount, unit toggle right-aligned below it, divider, keypad, Continue. `PaymentRequestAmountContent:161-215` matches band for band, including the toggle sitting under the amount rather than beside it.
- **Sent** (`48030:206596`) — check, 32dp, Display headline with the second word accented, 8dp, body copy, 24dp, card, then OK. `PaymentRequestSentContent:553-597` matches. The `deliveryStatus` queued/sent branch at `:571-591` is a state the frame doesn't specify — that's an addition, not an omission, so no delta.
- **No hardcoded colours or type introduced**, consistent with the token check from the earlier pass.

**Code, traced clean at head:**

- **`Money.kt` is additive only.** The one change is `secondaryText: String? = null` (`:50`); non-null fills the existing secondary `CaptionB` slot, null runs the old `primaryDisplay.not()` branch unchanged. Primary line and `rememberMoneyText` untouched — no formatting, rounding, unit-switch, sign or `ULong` change. All three `MoneyCell` call sites are Paykit-only surfaces, so despite `Money.kt` being shared, no ungated screen's rendering changes.
- **`ReceiveSheet.kt` — the ordinary receive path is byte-for-byte unchanged.** Exactly two lines changed (`:225`, `:235`), both inside `PaymentRequestRecipient`/`PaymentRequestDetails`, which are `InternalOnly` and reachable only when `paymentRequestTargets.isNotEmpty()` — and `PaykitPaymentRequestRepo.kt:333-339` empties that whenever `isAvailable()` is false. QR / EditInvoice / CJIT amount / address-copy are untouched. `skipPaymentRequestAmount` is set immediately before both `navigateTo` calls, so neither flag can read stale.
- **The release fixture split genuinely compiles out.** `app/build.gradle.kts:242-257` declares only `debug` and `release`, so no third build type misses a source set; `release/…/PaymentRequestFixtureRuntime.kt:7` returns `null`. The debug variant additionally requires `Env.network == REGTEST`, so mainnet and tnet debug builds reject it too. `PaymentRequestFixture.kt` stays in `main` but is inert with no writer.
- **Long/hostile counterparty text can't overlay the amount** — name, subtitle and note are all `maxLines = 1` with ellipsis, and the amount sits in its own weighted column. `Text13Up` uses Kotlin `uppercase()`, which is `Locale.ROOT`-invariant, so no Turkish-dotted-I on z-base-32's `i`.
- **Screens-map rule satisfied** — rows exist at `docs/screens-map.md:73` and `:75`; the PR adds and removes no `*Screen.kt`, so `ScreensMapTest` is not at risk.

**Pre-existing, noted not filed:** `PubkyImage.kt:59-60` does `AsyncImage(model = uri)` with no host allowlist for contact avatars. Not in this diff — the diff only changes padding and typography around it.

**Worth a re-run before merge:** CI shows `e2e-tests-local - lnurl_transfer`, `e2e-tests-staging - hardware_wallet` and `multi_address_2_regtest` red. None touch these files and `pubky_paykit` passes, so they read as unrelated, but I'd rather you confirm than assume.

jvsena42 (changes requested):
On first row of preview, the "Subscriptions" title is not matching "Payment Requests" Figma. Also, the badge is breaks the tab indicator UI

ovitrif:
Dropped the ui fix skill in my local setup, concluded it makes the job harded than it has to be.

Closing this and leaving it non-assigned to me.

PR #1174, opened 2026-08-21
chore: bump version 2.4.1

piotr-iohk:
Bump version to 2.4.1 (build 188) for release.

This is a hotfix branch cut from `v2.4.0`, not from master. **Do not merge this branch into master.** Master already contains these commits plus 2.5.0 work. After store ship, open a small follow-up on master to insert `## [2.4.1]`, drop consumed `changelog.d/next/` fragments, and bump master's version.

### Description

- `versionCode`: 187 → 188
- `versionName`: 2.4.0 → 2.4.1

### Cherry-picked from master

- #1107 fix: update DataStore for 16 KB compatibility
- #1100 fix: harden ldk node teardown on stop
- #1122 fix: integrate ldk crash fixes (`ldk-node` 0.7.0-rc.62)
- #1155 fix: stop auto-accepting stale monitors
- #1146 feat: increase background stop debounce time
- #1158 fix: secure shop payments and locked links
- #1162 fix: gate pubky auth handler
- #1156 fix: fail closed without vss auth
- #1164 fix: consume vss-client 0.5.23
- #1092
- https://github.com/synonymdev/bitkit-android/pull/1175
- #1094

Also on this branch: `fix: compile release-2.4.1 cherry-picks` (strip 2.5.0 leftovers so the cherry-picks compile on 2.4.0).

### Preview

N/A

### QA Notes

N/A

Made with [Cursor](https://cursor.com)

PR #1198, opened 2026-08-31
fix: cap electrum probe response lines

piotr-iohk:
This PR caps Electrum probe response lines so a user-supplied endpoint cannot grow `readLine()` until the process OOMs.

Follow-up to the Greptile note on #1174: https://github.com/synonymdev/bitkit-android/pull/1174#discussion_r3828223628

We left it off the 2.4.1 hotfix on purpose. The risk is a local crash after the user taps Connect on an Electrum URL they typed or scanned — not remote, not funds, not unlock. `soTimeout` already covers a silent peer. This only bounds the remaining “stream bytes without a newline” case before the next store build.

### Description

- Replaces unbounded `readLine()` in the Electrum probe with a 16 KiB byte-capped line read
- Treats an oversized or incomplete line as `NotElectrum`, same as any other failed handshake
- Adds coverage for the oversized-line path

### Preview

N/A

#### Manual Tests

- [x] **1.** Settings → Electrum → Connect to a working server: probe succeeds and the node keeps the server.
- [x] **2.** `regression:` Settings → Electrum → Connect to a closed or non-Electrum port: still rejected, wallet stays up.

#### Automated Checks

- Unit tests added: oversized-line rejection and cause text in `ElectrumProbeServiceTest.kt`.


ovitrif (approved):
utAck; looks good

PR #1186, opened 2026-08-26
feat: add Paykit subscriptions

ben-kaufman:
### Description

Adds the payer side of Paykit subscriptions, building on the merged payment-proof work in [#1199](https://github.com/synonymdev/bitkit-android/pull/1199) and [#1178](https://github.com/synonymdev/bitkit-android/pull/1178).

- Receives private subscription proposals and lets the payer review and confirm them.
- Discloses the first billing period’s end date/time and full-period charge on Review, including short initial intervals.
- Requires an immediately due first period to be paid before showing the subscribed success state; definitely failed payments remain due and can be retried, while uncertain submissions stay protected until their outcome is known.
- Turns each unpaid billing period into the existing incoming payment-request flow, including automatic presentation, the bell queue, and the Payments tab.
- Schedules local due-payment notifications and opens the corresponding request when the notification is selected.
- Adds the Figma-aligned Subscriptions and Payments tabs, subscription detail/status views, cancellation, empty states, and fixed/open-ended summaries.
- Shows a normalized Monthly Cost across all active subscriptions, independent of whether the current periods are already paid.
- Associates payment proofs with their exact billing period so recurring periods remain independent while retaining private-only resolution and fresh Private Payment List requirements.
- Reuses the shared amount-entry component without duplicating its secondary currency amount.
- Keeps queued delivery visible even when a request has a note, and lets long request details scroll while Pay/Dismiss stay accessible.
- Keeps subscribe-and-pay inside one continuously presented sheet, including initial-payment failure and retry, without the modal dismissal/reappearance gap.
- Retains the settled payment rail in completed request and subscription history so Lightning feedback is purple and on-chain feedback is orange, including success confetti.

This PR intentionally covers payer-side receipt only. Subscription creation, Discover, automatic payment without confirmation, and renewal are not included because they are outside this MVP or are not currently exposed by Paykit.

Dependencies:

- Paykit `0.1.0-rc51`
- [Hardware-wallet payment proofs #1199](https://github.com/synonymdev/bitkit-android/pull/1199) (merged)
- [Payment proof delivery #1178](https://github.com/synonymdev/bitkit-android/pull/1178) (merged)
- [iOS companion PR #685](https://github.com/synonymdev/bitkit-ios/pull/685)

### Preview

Final Android and cross-platform proof recordings were completed locally; they are not attached to the PR.

#### Manual Tests

- [x] Receive an immediate-start private subscription proposal and verify review → payment → success stays in one continuously presented sheet.
[...]

PR #1199, opened 2026-08-31
fix: send hardware payment proofs

ben-kaufman:
This PR sends Paykit payment proofs for on-chain payment requests paid from a hardware wallet. It is stacked on #1178 so the proof-delivery changes remain isolated from the subscription work.

### Description

- Prepares the payment proof immediately before hardware broadcast, after signing succeeds.
- Preserves the existing private Payment List consumption and request acceptance order.
- Completes the proof with the broadcast transaction ID before showing payment success.
- Keeps preparation idempotent across hardware broadcast retries.

### Preview

N/A

#### Manual Tests

- [ ] **1.** Incoming on-chain Payment Request → choose a hardware wallet → review → sign and broadcast: payment succeeds and the requester receives its transaction proof.
- [ ] **2.** `regression:` hardware broadcast retry → retry the same signed transaction: the request is not accepted twice and the private Payment List is not consumed twice.

#### Automated Checks

- `AppViewModelSendFlowTest.kt`: verifies hardware proof preparation ordering and idempotency, plus proof completion with the broadcast transaction ID.
- `./gradlew testDevDebugUnitTest --tests 'to.bitkit.viewmodels.AppViewModelSendFlowTest'`

### DIFF
--- a/app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
+++ b/app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
@@ -127,6 +127,7 @@ fun SendSheet(
             val navController = rememberNavController()
             LaunchedEffect(hwSendViewModel, navController) {
                 hwSendViewModel.results.collect { result ->
+                    appViewModel.completeHardwareContactPayment(result.txId)
                     appViewModel.onSendSuccess(
                         details = NewTransactionSheetDetails(
                             type = NewTransactionSheetType.ONCHAIN,
--- a/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
+++ b/app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
@@ -3341,13 +3341,7 @@ class AppViewModel @Inject constructor(
         if (!validateIncomingPaymentRequest(contactPaymentContext)) return
 
         val incomingPaymentRequest = contactPaymentContext?.incomingPaymentRequest
-        var preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).fold(
-            onSuccess = { it },
-            onFailure = {
-                handlePaymentPreparationFailure(it)
-                return
-            },
-        )
+        var preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).getOrNull()
 
         consumePrivatePaymentListIfNeeded(contactPaymentContext).onFailure {
             cancelPaymentProofPreparation(preparedPaymentProofRequest)
@@ -3392,7 +3386,7 @@ class AppViewModel @Inject constructor(
                 sendOnchain(address, amount, tags = tags)
                     .onSuccess { txId ->
                         preparedPaymentProofRequest = null
-                        completeOnchainPaymentProof(incomingPaymentRequest, txId)
+                        completeOnchainPaymentProofInBackground(incomingPaymentRequest, txId)
                         Logger.info("Onchain send result txid: $txId", context = TAG)
                         onSendSuccess(
                             NewTransactionSheetDetails(
@@ -3432,8 +3426,7 @@ class AppViewModel @Inject constructor(
                 val paymentHash = decodedInvoice.paymentHash.toHex()
                 associateLightningPaymentProof(incomingPaymentRequest, paymentHash).onFailure {
                     cancelPaymentProofPreparation(preparedPaymentProofRequest)
-                    handlePaymentPreparationFailure(it)
-                    return
+                    preparedPaymentProofRequest = null
                 }
 
                 // Create pre-activity metadata before sending
@@ -3528,12 +3521,14 @@ class AppViewModel @Inject constructor(
     ): Result<Unit> = request?.let { paykitPaymentProofRepo.associateLightningPayment(it, paymentHash) }
         ?: Result.success(Unit)
 
-    private suspend fun completeOnchainPaymentProof(request: PaykitPaymentRequest?, txId: String) {
-        request?.let {
+    private fun completeOnchainPaymentProofInBackground(request: PaykitPaymentRequest?, txId: String) {
+        val paymentRequest = request ?: return
+        val endpointIdentifier = paymentProofPreparation().endpointIdentifier
+        viewModelScope.launch {
             paykitPaymentProofRepo.completeOnchainPayment(
-                request = it,
+                request = paymentRequest,
                 txid = txId,
-                paymentEndpointIdentifier = paymentProofPreparation().endpointIdentifier,
+                paymentEndpointIdentifier = endpointIdentifier,
             )
         }
     }
@@ -4346,13 +4341,7 @@ class AppViewModel @Inject constructor(
         if (isPreparedContactPayment(contactPaymentContext)) return true
 
         val incomingPaymentRequest = contactPaymentContext?.incomingPaymentRequest
-        val preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).fold(
-            onSuccess = { it },
-            onFailure = {
-                handlePaymentPreparationFailure(it)
-                return false
-            },
-        )
+        val preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).getOrNull()
         if (!prepareContactPayment(contactPaymentContext)) {
             cancelPaymentProofPreparation(preparedPaymentProofRequest)
             return false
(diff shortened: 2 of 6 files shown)


ovitrif (approved):
utACK

ovitrif (dismissed):
utACK

jvsena42:
Starting review

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -4344,7 +4343,28 @@ class AppViewModel @Inject constructor(
...
     suspend fun prepareHardwareContactPayment(): Boolean {
         val contactPaymentContext = synchronized(contactPaymentContextLock) { activeContactPaymentContext }
-        return prepareContactPayment(contactPaymentContext)
+        if (isPreparedContactPayment(contactPaymentContext)) return true
+
+        val incomingPaymentRequest = contactPaymentContext?.incomingPaymentRequest
+        val preparedPaymentProofRequest = preparePaymentProof(incomingPaymentRequest).fold(
+            onSuccess = { it },
+            onFailure = {
+                handlePaymentPreparationFailure(it)
```
jvsena42: A proof preparation failure now vetoes the payment itself. `pendingProof` throws `RequestUnavailable` whenever `identityStatus()?.liveSessionAvailable != true` or the public key can't be normalized, so a momentarily unavailable relay session cancels an otherwise valid on-chain payment before signing even starts.

Proof delivery is auxiliary to the payment. Could preparation degrade to best-effort here (log and continue) instead of `return false`? Without a persisted row `completeOnchainPayment` simply finds no match and the proof is skipped, which seems strictly better than refusing to pay.

Same shape on the non-hardware path at :3347, so probably worth handling in one place. Note iOS has the identical abort in `SendConfirmationView.performPayment`, so whatever we settle on should be mirrored there.

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
```diff
@@ -127,6 +127,7 @@ fun SendSheet(
             val navController = rememberNavController()
             LaunchedEffect(hwSendViewModel, navController) {
                 hwSendViewModel.results.collect { result ->
+                    appViewModel.completeHardwareContactPayment(result.txId)
```
jvsena42, line 130: This awaits the full Paykit round-trip before `onSendSuccess`, `navigateTo(SendRoute.Success)` and `hwSendViewModel.completeBroadcast()`. `completeHardwareContactPayment` reaches `submitReady`, which calls `identityStatus()`, `paymentRequests()`, `submitPaymentProof()` and `processPendingPrivateMessages()` — all behind `isSetup.await()` and none of them time-bounded.

While that runs `HwSendViewModel` still has `isBroadcastUnresolved = true`, and `HwSendSignScreen`'s back handler is gated on `!uiState.isBroadcastUnresolved`, so a slow or unreachable relay leaves the user stuck on the signing screen with back disabled after their money has already moved. Measured ~2s against a local regtest relay; against a real one it's unbounded.

iOS hands this off with `submitInBackground(...)` so only the keychain write sits on the critical path. Could we launch it off `viewModelScope` (or wrap it in a `withTimeout`) and navigate to success as soon as the broadcast resolves?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -4344,7 +4343,28 @@ class AppViewModel @Inject constructor(
...
+            return false
+        }
+        return true
+    }
+
+    suspend fun completeHardwareContactPayment(txId: String) {
+        val incomingPaymentRequest = synchronized(contactPaymentContextLock) {
+            activeContactPaymentContext?.incomingPaymentRequest
+        }
+        completeOnchainPaymentProof(incomingPaymentRequest, txId)
```
jvsena42: Pre-existing, but this PR makes it reachable from the hardware path, so flagging it here: in `PaykitPaymentProofRepo.completeOnchainPayment`, `persistAndSubmit` calls `submitReady(proof)` (which submits to the SDK) and `submitReady` ends with `removeProofsLocked`, whose `loadProofs()`/`persist()` can throw. That throw escapes `persistAndSubmit` and makes `completion.isFailure` true, so the `if (completion.isFailure)` branch rebuilds the proof and calls `submitReady` a second time.

The `alreadyQueued` guard only helps if `paykitSdkService.paymentRequests()` already reflects the submission that just happened; otherwise the counterparty receives the same proof twice. Could the trailing cleanup be wrapped so a removal failure doesn't read as a delivery failure? iOS avoids this by keeping the persist failure separate from the submit path.

ben-kaufman: Fixed in 1d5007060. The trailing persisted-proof cleanup is now handled separately from successful SDK submission, so a cleanup failure is logged and retains the completed proof for later reconciliation instead of entering the immediate delivery fallback/retry. The regression verifies the submit call happens once when cleanup persistence fails.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Fixed in 1d5007060. Proof preparation is now best-effort for both software and hardware payments, and a Lightning proof-association failure cancels only the stale proof preparation instead of vetoing the payment. Request validation, private-list consumption, and request acceptance remain payment gates. I added regressions for on-chain, Lightning, and hardware degradation. The equivalent iOS behavior should be mirrored in the iOS proof PR rather than in this Android PR.

app/src/main/java/to/bitkit/ui/sheets/SendSheet.kt
ben-kaufman: Fixed in 1d5007060. The request and endpoint identifier are captured synchronously, then proof completion is launched on `viewModelScope`; the hardware success navigation and `completeBroadcast()` no longer wait for Paykit/relay calls. The same non-blocking completion path is used by software on-chain sends. The regression test suspends repository completion and verifies `completeHardwareContactPayment` returns while delivery is still pending.

ovitrif (dismissed):
utACK

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3529,12 +3521,14 @@ class AppViewModel @Inject constructor(
     ): Result<Unit> = request?.let { paykitPaymentProofRepo.associateLightningPayment(it, paymentHash) }
         ?: Result.success(Unit)
 
-    private suspend fun completeOnchainPaymentProof(request: PaykitPaymentRequest?, txId: String) {
-        request?.let {
+    private fun completeOnchainPaymentProofInBackground(request: PaykitPaymentRequest?, txId: String) {
```
jvsena42, line 3530: Thanks, the three fixes look right and the regressions are real ones — `onchain proof cleanup failure does not retry delivery` in particular pins the single submit properly.

One consequence of making preparation best-effort, anchoring here since this is the completion side: when `prepare()` failed there is no persisted row, so `completeOnchainPayment` hits `indexOfLast` → -1 and returns at `PaykitPaymentProofRepo.kt:153` **successfully**. The immediate-delivery fallback just below (`:168`) is gated on `completion.isFailure`, so it never runs, and the proof is dropped permanently and silently — nothing is persisted for `reconcile()` either, so not even the `paymentStarted` recovery in #1186 can pick it up.

So a transient relay hiccup at prepare time now costs the proof outright, where before the user at least got a toast and could retry once the session recovered. Since the payment is already broadcast by then and the session has usually come back, could the "no prepared row" case fall through to the same `pendingProof(...)` + `submitReady(...)` path as the load failure, rather than returning early? That would make the degradation genuinely best-effort.

The Lightning path has the same shape — `associateLightningPayment` failing calls `cancelPreparation`, so `completeLightningPayment` later finds nothing to complete — though that one is lower priority since it needs the preimage anyway.

Non-blocking: `viewModelScope.launch` here means a ViewModel clear right after broadcast cancels the submission and orphans the pending row. Same window #1186 closes with `reconcileOnchainProof`, just slightly wider now that the call isn't awaited.

ben-kaufman: Fixed in `16c7158`. On-chain completion now creates, persists, and submits a ready proof when best-effort preparation left no stored row. Lightning association now likewise creates a durable row using the actually selected endpoint, so the later preimage can complete and submit it. I added repository regressions for both missing-row paths and kept completion non-blocking; the ViewModel cancellation window remains covered by the #1186 reconciliation path, as noted.

ben-kaufman:
Follow-up: current signed head is `1769cb2`. The first CI run compiled the production code and ran 2,290 tests; its only failure was the new assertion mixing one raw endpoint with Mockito matchers. That assertion now uses `eq(endpoint)`, static analysis remains clean for all changed files, and fresh CI is running.

ovitrif (approved):
utACK

jvsena42:
Doing another round of review and testing

jvsena42 (approved):
tAck

Issue #697, opened 2026-01-21
Migration screen infinitely showing on network loss


piotr-iohk:
There seems to be similar issue in bitkit-android as there is currently being fixed in https://github.com/synonymdev/bitkit-ios/pull/381

Tested as follows:
 - have wallet in RN
 - Disable network using Network Link Conditioner
 - Install Bitkit-android

Logs:
[logs.zip](https://github.com/user-attachments/files/24770202/logs.zip)

https://github.com/user-attachments/assets/d996b860-6ab7-465b-8869-c075d6cbe3fa

_Originally posted by @piotr-iohk in https://github.com/synonymdev/bitkit-ios/issues/381#issuecomment-3778489741_

PR #1200, opened 2026-08-31
feat: upgrade paykit to rc51

### LINKED ISSUE #1219: [Bug]: payment request lists diverge after cross-device pay
## What happened?

Two live wallets (Android emu `andi` ↔ iOS sim `ios`) each created 2 private payment requests (1k Lightning + 25k on-chain) and paid the other side’s requests. All 4 payments settled. Each Payment Requests list then showed **3** rows, and they did not match:

- Android: 1× 25 000 + 2× 1 000 (missing one 25k)
- iOS: 2× 25 000 + 1× 1 000 (missing one 1k)

The missing 25k *did* pay (Android → iOS on-chain `e418fd9b…`, iOS activity +25 000, proof queued). The missing 1k *did* pay (iOS → Android LN, proof queued). So this is a **history-row drop**, not a failed payment.

Not same-amount overwrite in the app list: IDs are `paymentRequestId + counterparty + receiverPath`. Hypothesis: after fulfill, one side drops the row when `localRole` is missing/unknown (history mapping skips those).

Seen twice (2026-09-01 and 2026-09-02) on #1178 / #683 builds.

## Expected behavior

After 4 created + 4 paid requests between the same pair, both devices show the same 4 rows (2× 1k + 2× 25k).

## Steps to Reproduce

1. Two wallets, private Paykit + contact payments enabled, same contact pair.
2. Android creates 1k Lightning request; iOS pays it.
3. iOS creates 1k Lightning request; Android pays it.
4. Android creates 25k on-chain request; iOS pays it (or the reverse order used in the recording).
5. iOS creates 25k on-chain request; Android pays it.
6. Open Payment Requests on both devices.

- Recording: 

https://github.com/user-attachments/assets/4b88c2c2-aec1-41b3-9bb2-9caef4ce2637

Image

- Android log: [logs-android.zip](https://github.com/user-attachments/files/31728720/logs-android.zip)

- iOS log: [bitkit_logs_2026-09-02_08-09-31.zip](https://github.com/user-attachments/files/31728729/bitkit_logs_2026-09-02_08-09-31.zip)

## Bitkit Version

Android #1178 (`10ff0df`). Cross-tested vs iOS #683 (`489b1e6`).

## Device / OS

Pixel emulator (paired with iPhone 17 simulator)

## Reproducibility

Always (2/2 sessions)

## Additional context

- New wallets, regtest
- Lightning + on-chain
- Incoming request pay on Android stays on Confirm with amount locked (expected).
- Proofs queued on all 4 pays (`Queued a Paykit payment proof for private delivery`). This bug is list/history only.
- iOS counterpart: https://github.com/synonymdev/bitkit-ios/issues/723
- Found while QA of https://github.com/synonymdev/bitkit-android/pull/1178

ben-kaufman:
Fixes #1219

This PR:
1. Upgrades Paykit from `0.1.0-rc46` to `0.1.0-rc51`.
2. Adopts app-scoped Pubky grants using Bitkit's stable client ID.
3. Revokes Bitkit's grant on normal sign-out while preserving local state when remote revocation fails.
4. Restricts local-only session forgetting to destructive reset and backup-replacement flows.
5. Fixes payment requests disappearing from history after requests are exchanged and paid in both directions.

### Description

Bitkit now identifies itself as `bitkit.to` on mainnet and `staging.bitkit.to` elsewhere. Normal sign-out remotely revokes only Bitkit's current grant. If revocation cannot be confirmed, the profile and private Paykit state remain available so the user can retry.

Completed Ring authentication that is later canceled, and identity creation that fails after activating a session, also attempt secure revocation. Explicit wallet wipe and backup replacement forget the local session. Existing local E2E homeserver behavior remains unchanged.

Paykit has not launched, so migration from earlier development builds is intentionally unsupported. There is no upgrade handling for cookie-backed sessions or grants using old client IDs. Normal recovery of current-format grant sessions remains supported.

Paykit rc51 also fixes replay ordering when incoming and outgoing Payment Request lifecycles interleave. Each request's events are derived in the correct order, preserving its payer/payee role so the existing history mapping retains the row. The fix is in [paykit-rs #151](https://github.com/pubky/paykit-rs/pull/151) and is available in the published release before that PR merges.

### Preview

N/A — no visual changes.

#### Manual Tests

- [x] **1.** Pubky profile → Sign Out while online: Bitkit signs out and returns to the disconnected profile state.
- [x] **2.** Pubky profile → interrupt network access → Sign Out: Bitkit shows an error and keeps the profile and private Paykit state; restore network access and retry successfully.
- [ ] **3.** Pubky auth → approve another app → Sign Out of Bitkit: Bitkit's session is revoked while the other app remains authorized.
- [ ] **4.** `regression:` restore a wallet backup with different Pubky state: the previous local session is forgotten and the backup identity is installed.
- [ ] **5.** `regression:` two privately linked wallets → each creates a 1,000-sat Lightning request and a 25,000-sat on-chain request → the peer pays all four → Payment Requests: both wallets retain all four rows after refresh and app restart.

#### Automated Checks

- `PubkyRepoTest.kt`: covers canceled completed authentication revocation, revocation failure, wallet wipe, and backup replacement.
- `ProfileViewModelTest.kt` and `EditProfileViewModelTest.kt`: cover preserving private Paykit state when sign-out fails.
- `PaykitSdkServiceTest.kt`: covers the stable Bitkit client ID and Pubky client configuration.
- Upstream rc51 regression covers interleaved bidirectional request events; Paykit tests and both binding builds passed.
- Local rc51 compilation, all 2,316 unit tests, and lint pass.
- Added a regression for canceled approval where remote revoke and SDK forget both fail, verifying both stored credentials are removed and the app remains unauthenticated.

### Linked Issues/Tasks

- [Paykit rc51](https://github.com/pubky/paykit-rs/releases/tag/v0.1.0-rc51)
- [paykit-rs #143](https://github.com/pubky/paykit-rs/pull/143)
- [paykit-rs #146](https://github.com/pubky/paykit-rs/pull/146)
- [paykit-rs #151](https://github.com/pubky/paykit-rs/pull/151)


app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -534,40 +534,72 @@ class PubkyRepo @Inject constructor(
...
-            }.getOrElse {
-                Logger.warn("Retrying sign in after sign up failed", it, context = TAG)
-                pubkyService.signIn(secretKeyHex)
+                    runSuspendCatching {
+                        pubkyService.signUp(secretKeyHex, signupDetails.first, signupDetails.second)
+                    }.getOrElse {
+                        Logger.warn("Retrying sign in after sign up failed", it, context = TAG)
+                        pubkyService.signIn(secretKeyHex)
+                    }
+                    didActivateSession = true
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in b2377c16a. The incomplete-session cleanup guard is now armed before queued sign-up/sign-in starts, so cancellation after the SDK persists a session always triggers the non-cancellable revocation attempt. The focused Pubky/AppViewModel suites and the full unit suite pass.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -1045,19 +1077,20 @@ class PubkyRepo @Inject constructor(
...
-        }.fold(
-            onSuccess = { Result.success(it) },
-            onFailure = {
-                Logger.warn("Forcing local sign out after server sign out failed", it, context = TAG)
-                runSuspendCatching { withContext(ioDispatcher) { pubkyService.forceSignOut() } }
-            },
-        )
+        }.onFailure { Logger.error("Failed to revoke Pubky session during sign out", it, context = TAG) }
+
+        if (result.isFailure) return result
```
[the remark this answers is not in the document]
ben-kaufman: Fixed in b2377c16a. A failed grant revocation now preserves local auth and sharing settings while marking Paykit reconciliation pending. The retry republishes public endpoints or restores the private-only receiver marker from those saved settings, instead of leaving the cleanup state visible. Added regression coverage; focused and full unit suites pass.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -333,7 +333,7 @@ class PubkyRepo @Inject constructor(
                 loadContacts()
             }.map { }
         } catch (e: CancellationException) {
-            clearCompletedAuthSessionIfNeeded(didCompleteAuth)
+            revokeCompletedAuthSessionIfNeeded(didCompleteAuth)
```
ovitrif: Cancellation of `completeAuthentication` still sets `didCompleteAuth` only after `pubkyService.completeAuth()` returns. `PubkyChoiceViewModel.cancelAuth()` cancels that job first, and `completeAuth()` runs on `ServiceQueue.CORE` under a `SupervisorJob`, so Paykit can persist the session in `activateBootstrapResult` after the caller has already thrown `CancellationException`. The handler then sees a false guard and skips `signOut()`, leaving Bitkit's grant active. Identity creation now arms its cleanup flag before `signUp`/`signIn` for this same window. Could we set `didCompleteAuth` before calling `completeAuth()` so a canceled Ring completion still attempts `signOut`?

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -534,40 +534,72 @@ class PubkyRepo @Inject constructor(
...
+                        Logger.warn("Retrying sign in after sign up failed", it, context = TAG)
+                        pubkyService.signIn(secretKeyHex)
+                    }
+
+                    val imageUrl = avatarBytes?.let { uploadAvatar(it).getOrNull() }
+                    writeProfile(name, bio, links, tags, imageUrl)
+                    finishIdentityCreation(publicKeyZ32, name, bio, links, tags, imageUrl)
+                }
             }
+            if (result.isFailure) revokeIncompleteIdentitySessionIfNeeded(shouldRevokeSessionOnFailure)
```
ovitrif, line 579: `createIdentity` now calls `pubkyService.signOut()` when sign-up or sign-in has started and later work fails or the coroutine is cancelled, but no unit test would fail if that cleanup stopped happening. The canceled Ring path already asserts `signOut()`; this new grant-cleanup path does not. Could we add a `PubkyRepoTest` that fails after `signUp` or cancels after the session is activated and asserts `signOut()` is invoked?

ovitrif (changes requested):
Canceled Ring completion can still persist Bitkit's grant. `didCompleteAuth` is set only after `completeAuth()` returns, so canceling the approval job can skip `signOut()` after Paykit has already saved the session.

`createIdentity` now revokes on post-activation failure, but `PubkyRepoTest` has no case that would fail if that cleanup stopped running.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Added in dd3b7bad. PubkyRepoTest now forces profile publication to fail after sign-up activates the session and verifies signOut is called, so this cleanup path cannot regress silently.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in dd3b7bad. completeAuth and the success marker now run in one minimal NonCancellable section, so caller cancellation cannot land between session activation and arming cleanup. The new regression cancels during gated completion and verifies signOut runs after activation.

ben-kaufman:
The cancellation and cleanup findings are fixed in signed commit `dd3b7bad`. Successful `completeAuth()` and its completion marker now cross a minimal `NonCancellable` boundary, so cancellation cannot leave a persisted grant while skipping cleanup; ordinary completion failures still preserve existing sessions. The focused tests cover cancellation after SDK completion and post-activation `createIdentity()` failure cleanup. All CI, lint, detekt, build, and E2E checks are green on the current head. @ovitrif please re-review.

ovitrif (dismissed):
utACK

piotr-iohk (dismissed):
e2e ACK.

Latest (`dd3b7ba`) with matching e2e branch `codex/paykit-rc50-auth` ([#212](https://github.com/synonymdev/bitkit-e2e-tests/pull/212)). Recreated the two staging Paykit fixture pubkys. `e2e-tests-staging - pubky_paykit` green. Full CI green.

Manual on Pixel emu: online Delete/Disconnect clears session + secret for `pubkyif…ejhsa1y`. Offline Delete: `transport_error`, profile kept. Offline Disconnect from that dialog: endpoint cleanup WARN, then ~22s later network returned and session cleared. Offline Disconnect waited for network instead of a clean fail-closed error. Not a blocker.

Did not retest other-app grant stays authorized, or backup replace.

jvsena42:
Non blocking: I got this error when trying to sign in with Pubky Ring

```
{
  "action": "unknown",
  "rawInput": "pubkyauth://signin_grant?caps=%2Fpub%2Fpaykit%2Fv0%2Fbitkit%2Fwallet%2F%3Arw%2C%2Fpub%2Fpaykit%2Fv0%2Fprivate%2Fbitkit%2Fwallet%2F%3Arw%2C%2Fpub%2Fstaging.bitkit.to%2Fbitkit%2Fwallet%2F%3Arw&relay=https%3A%2F%2Fhttprelay.pubky.app%2Finbox&secret=9P-JtlM1i5HvaJnmru5fZhgYU5dvpK3sL-KNi-1eAng&cid=staging.bitkit.to&cpk=cc4amfpzrugsxkbqgia9he9yy9jezq4kcjo3wdsgoozar4zt1b5o&x-success=bitkit%3A%2F%2Fpubky-auth%2Fsuccess%3Fnonce%3D90e53840-9b30-4e4d-9751-b6b018beca25&x-cancel=bitkit%3A%2F%2Fpubky-auth%2Fcancel%3Fnonce%3D90e53840-9b30-4e4d-9751-b6b018beca25&x-error=bitkit%3A%2F%2Fpubky-auth%2Ferror%3Fnonce%3D90e53840-9b30-4e4d-9751-b6b018beca25&x-source=Bitkit",
  "error": "Unrecognized format. Expected a recovery phrase, invite code, auth URL, or session request."
}
```

Probably same issue of https://github.com/jvsena42/loopky/issues/130


https://github.com/user-attachments/assets/d2144a1e-1fe6-45ee-bb64-30774a596b9a

Will probably work on next Ring release

gradle/libs.versions.toml
```diff
@@ -22,7 +22,7 @@ appcompat = { module = "androidx.appcompat:appcompat", version = "1.7.1" }
 barcode-scanning = { module = "com.google.mlkit:barcode-scanning", version = "17.3.0" }
 biometric = { module = "androidx.biometric:biometric", version = "1.4.0-alpha05" }
 bitkit-core = { module = "com.synonym:bitkit-core-android", version = "0.5.5" }
-paykit = { module = "com.synonym:paykit-android", version = "0.1.0-rc46" }
+paykit = { module = "com.synonym:paykit-android", version = "0.1.0-rc50" }
```
jvsena42: **rc50 emits identity error contexts that `canDeferStaleSession` does not match.**

`PaykitSdkService.initialize()` has a recovery path for a session it cannot restore (suspend it, `resetRuntime()`, re-init), but it is gated on `canDeferStaleSession`, which matches a single literal — `STALE_SESSION_IMPORT_CONTEXT = "import Pubky session from platform provider"`. rc50 produces contexts that do not match it, so the recovery never engages and the exception propagates out of `ensureServiceInitialized()`. `PubkyRepo.initialize()` then hits its `?: return@withContext` before the `initializeMutex` block, so `_sessionRestorationFailed` is never set (no `profile__session_expired` toast) and `_publicKey` stays null (Profile renders the "Join the Pubky Web" onboarding), while the home header still shows the cached name/avatar.

**Offline app start with a valid rc50 grant-backed session.** Reproduced on a dev build with a profile freshly created *on this branch* — no upgrade or pre-existing session involved. Opening the app with no connectivity looks exactly like the profile was deleted, and the screen offers "Create profile with Bitkit":
```
ERROR [PubkyRepo.kt:169] Failed to initialize paykit
  [AppError='code=identity_error, context=restore Pubky grant session from platform provider']
WARN  [AppViewModel.kt:606] Failed to refresh public Paykit endpoints [same error]
```
Restoring connectivity recovers it: `INFO [PubkyRepo.kt:198] Restored paykit session for 'pubky65…npipe8y'`.

**Sign-out while offline** hits a third context — see the separate comment on `PubkyRepo.signOut()`:
```
ERROR [PubkyRepo.kt:1082] Failed to revoke Pubky session during sign out
  [AppError='code=identity_error, context=restore Pubky grant before revocation']
```

Suggest matching the set of unrestorable-session contexts rather than one literal, and having `PubkyRepo.initialize()` surface `_sessionRestorationFailed` when `ensureServiceInitialized()` fails with an identity error, so a transient failure shows the existing "session expired" toast instead of a silent logout.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -1045,19 +1079,29 @@ class PubkyRepo @Inject constructor(
...
         val result = runSuspendCatching {
             withContext(ioDispatcher) { pubkyService.signOut() }
-        }.fold(
-            onSuccess = { Result.success(it) },
-            onFailure = {
-                Logger.warn("Forcing local sign out after server sign out failed", it, context = TAG)
-                runSuspendCatching { withContext(ioDispatcher) { pubkyService.forceSignOut() } }
-            },
-        )
+        }.onFailure { Logger.error("Failed to revoke Pubky session during sign out", it, context = TAG) }
```
jvsena42, line 1108: **A failed revoke leaves the user permanently signed in, with no local fallback.**

`forceSignOut()` was removed, and this path now returns early on any failure, so local state and the keychain session are both retained. Reproduced on device: with the app offline, Edit Profile → Delete Profile → "Unable to Delete Profile" → **Disconnect** left the user still signed in on Edit Profile with the profile intact.

```
WARN  [PubkyRepo.kt:1078] Failed to remove Bitkit payment endpoints
       [AppError='Failed to remove Paykit payment endpoints']
ERROR [PubkyRepo.kt:1082] Failed to revoke Pubky session during sign out
       [AppError='code=identity_error, context=restore Pubky grant before revocation']
```

Note it reached `pubkyService.signOut()` and failed at the revoke itself, not at the earlier `removePublishedEndpointsForCleanup` gate. Once connectivity returned the same flow completed cleanly (`Deleted value for key 'PAYKIT_SESSION'` / `'PUBKY_SECRET_KEY'` / `'PAYKIT_SDK_STATE'`), so transient failures do recover — but there is no path out when the grant is unrevocable server-side (e.g. the user revokes Bitkit's authorization from Pubky Ring), which would make every future sign-out fail. Worth keeping a local forget as a fallback after a failed remote revoke.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -333,7 +335,7 @@ class PubkyRepo @Inject constructor(
                 loadContacts()
             }.map { }
         } catch (e: CancellationException) {
-            clearCompletedAuthSessionIfNeeded(didCompleteAuth)
+            revokeCompletedAuthSessionIfNeeded(didCompleteAuth)
```
jvsena42: **Cancellation after a *successful* auth revokes the grant the user just approved.**

`loadProfile()` / `loadContacts()` run inside `result.onSuccess { … }`, which is still inside the `try`, after `_publicKey` and `_authState = Authenticated` are set. `didCompleteAuth` is still `true` there, so cancelling during those two network calls lands in this `catch (CancellationException)` and calls `revokeCompletedAuthSessionIfNeeded(true)` → `pubkyService.signOut()`.

Concrete path: the user approves in Pubky Ring, returns to Bitkit, auth completes, then navigates back off `PubkyChoiceScreen` while contacts are still loading — `approvalJob` is a `viewModelScope.launch` and is also cancelled explicitly by `cancelAuthAndShowRingDialog`.

It also leaves inconsistent state: `restoreAuthStateAfterAuthFlow()` is `if (_publicKey.value == null) Idle else Authenticated`, and `_publicKey` was just populated, so the app settles into `Authenticated` against a grant that was revoked server-side.

Suggest clearing `didCompleteAuth` (or moving the revoke guard) once the success path is reached, so it only covers the genuinely incomplete window.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -534,40 +536,72 @@ class PubkyRepo @Inject constructor(
...
-                ?: fetchHomegateSignupCode().let { it.homeserverPubky to it.signupCode }
+                    val signupDetails: Pair<String, String?> = Env.e2eHomeserverPubky?.let { it to null }
+                        ?: fetchHomegateSignupCode().let { it.homeserverPubky to it.signupCode }
 
-            runSuspendCatching {
-                pubkyService.signUp(secretKeyHex, signupDetails.first, signupDetails.second)
-            }.getOrElse {
-                Logger.warn("Retrying sign in after sign up failed", it, context = TAG)
-                pubkyService.signIn(secretKeyHex)
+                    shouldRevokeSessionOnFailure = true
```
jvsena42, line 565: **Same shape as the auth-cancellation issue: a fully created identity can be revoked.**

`shouldRevokeSessionOnFailure` is set here and never cleared, while `finishIdentityCreation(...)` — which sets `_authState = Authenticated`, writes `cacheMetadata`, then does `loadProfile()` / `loadContacts()` — is still inside the guarded `runSuspendCatching`.

So a back-navigation during the contact load (`createIdentity` is called from `viewModelScope` in `CreateProfileViewModel`), or any throw from `cacheMetadata`, reaches the `catch (CancellationException)` / `result.isFailure` branch and revokes the session for an identity that was in fact created and published to the homeserver.

Suggest resetting the flag right after `writeProfile` / `finishIdentityCreation` succeeds.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -792,28 +792,16 @@ class PaykitSdkService @Inject constructor(
...
 
-    suspend fun forceSignOut() {
-        operationMutex.withLock {
-            clearSessionAccessLocked()
-            clearStateLocked()
-        }
-    }
-
-    suspend fun clearSessionAccess() {
+    suspend fun forgetSessionAccess() {
```
jvsena42, line 802: **`forgetSessionAccess()` no longer clears `activeAuthRequest`.**

The `clearSessionAccessLocked()` this replaces also reset `activeAuthRequest = null`; this path only calls `handle.forgetSessionAccess()` + `resetRuntime()`.

If a `startAuth()` is in flight when `PubkyRepo.restoreSessionBackupState` or `wipeLocalState` runs, the stale `PubkyAuthRequest` survives, and a later `completeAuth()` can activate a session belonging to the discarded flow instead of failing with "No active Pubky auth request".

Not reproduced on device — raised from reading the diff.

jvsena42 (commented):
Review notes from testing this branch on a dev build (regtest, Pixel_9 emulator).

The first two comments were reproduced on device with logs attached inline; the rest come from reading the diff. The happy paths all worked: create identity, restore across restart, and online profile delete with full keychain teardown.

gradle/libs.versions.toml
ben-kaufman: Fixed in 54a620df0. `canDeferStaleSession` now matches rc50’s exact initialization context, `restore Pubky grant session from platform provider`, so the existing deferred recovery path runs for a stored grant that cannot be restored at startup. I kept this deliberately scoped to the SDK initialization context: the revocation context is a separate sign-out failure, and classifying every identity error as “session expired” would mislabel transient/network failures.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Fixed in 54a620df0. `forgetSessionAccess()` again clears `activeAuthRequest` inside the same operation mutex before forgetting SDK access and resetting runtime state, restoring the previous stale-auth invalidation invariant.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in 54a620df0. The cleanup guard is now disarmed immediately after the authenticated identity/state is committed, before profile and contact hydration. A regression test cancels during profile loading and verifies that the committed session remains authenticated and is not revoked.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Investigated and kept this fail-closed intentionally. rc50 treats a server-confirmed revoked or expired grant as a successful revocation, so revoking Bitkit in Ring does not create a permanent trap. An offline/transport failure leaves the remote grant status unknown; forgetting it locally at that point could abandon a still-live grant. The user can retry once connectivity returns, as your test also confirmed, which matches the explicit sign-out behavior documented and tested by this PR.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in 54a620df0. The cleanup guard is now disarmed immediately after successful profile publication, before cache/profile/contact hydration. The existing publication-failure cleanup remains covered, and a new cancellation-during-contact-load test verifies that a created identity is not revoked.

ben-kaufman:
@jvsena42 Confirmed the remaining Ring observation separately: rc50 intentionally emits the grant-auth request shape, while the currently installed Ring build does not yet recognize it. That is a Ring compatibility gap for its next release, not another Bitkit change. We will retest the end-to-end handoff with a Ring build that supports grant auth.

ovitrif (dismissed):
utACK

jvsena42 (commented):
Follow-up review of `65491308d`/`54a620df0`.

app/src/main/java/to/bitkit/ui/screens/profile/PubkyAuthApprovalViewModel.kt
```diff
@@ -165,7 +166,12 @@ class PubkyAuthApprovalViewModel @Inject constructor(
             handleApprovalFailure(it, authUrl)
             return
         }
-        if (_uiState.value.authUrl != authUrl) return
+        val approvalState = _uiState.value
+        if (approvalState.authUrl != authUrl) return
+        if (approvalState.clientId != request.clientId) {
+            handleApprovalFailure(IllegalArgumentException("Pubky auth requester changed"), authUrl)
```
jvsena42: **Raw exceptions with hardcoded English reach the user.**

`handleApprovalFailure` puts `error.localizedPubkyAuthMessage(context)` in the toast description, and that mapper (`ui/utils/PubkyAuthErrorMessage.kt:28`) walks the cause chain looking for a `PubkyAuthRequestError` / `WatchOnlyAccountError` and otherwise falls through to `return message`. A plain `IllegalArgumentException` matches nothing, so its developer text is what the user sees, untranslated in every locale.

Two new sites do this:

- here — `IllegalArgumentException("Pubky auth requester changed")`
- `PaykitSdkService.kt:928` — `require(...) { "Approved Pubky client ID does not match auth request" }`, which propagates through `PubkyRepo.approveAuth`'s `runSuspendCatching` into the same `handleApprovalFailure`

This also runs against two repo conventions: "NEVER use `Exception` directly, use `AppError` instead" and "NEVER hardcode strings and always preserve string resources".

Suggest an `AppError` subtype for the client-ID mismatch plus a branch in `localizedPubkyAuthMessage` mapping it to a new `profile__auth_error_*` string, so both sites surface a localized message.

Worth noting the severity is currently theoretical: as far as I can tell neither site can fire, since both compare a client ID against a re-parse of the same URL (see my comment on `PaykitSdkService.kt:352`). But if the client-ID check is made load-bearing, these become the user-visible failure path — so they are worth fixing alongside it rather than after.

jvsena42 (commented):
Follow-up on error localization in `65491308d`/`54a620df0`.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -1038,26 +1081,36 @@ class PubkyRepo @Inject constructor(
 
     // region Sign out
 
-    suspend fun signOut(): Result<Unit> {
+    suspend fun signOut(): Result<Unit> = withContext(NonCancellable + ioDispatcher) {
```
jvsena42, line 1101: **`NonCancellable` here covers the network calls, not just the state cleanup it needs to protect.**

Making sign-out atomic against caller cancellation is the right goal — it avoids being cancelled between revoking the grant and clearing local state, which would strand the keychain session. But wrapping the whole body puts three suspension points beyond cancellation that do not need to be:

- `settingsStore.data.first()`
- `removeBitkitPaymentEndpoints()` — a homeserver round-trip
- `pubkyService.signOut()` — relay revocation, serialized behind `ServiceQueue.CORE` and `PaykitSdkService.operationMutex`

Only `clearLocalState(...)` (and the `markPaykitStateForReconciliation` write in the failure branch) actually needs the guarantee.

Both callers launch this in `viewModelScope` — `ProfileViewModel.signOut()` and `EditProfileViewModel.disconnectProfile()`. So with a stalled homeserver, or while another operation holds `operationMutex`, the job stays alive and unabortable after the user has left the screen and the ViewModel is cleared; it can only end when the network layer times out.

Suggest keeping the network calls cancellable and wrapping just the local teardown:

```kotlin
suspend fun signOut(): Result<Unit> = withContext(ioDispatcher) {
    // … endpoint cleanup + revoke, cancellable …
    withContext(NonCancellable) { clearLocalState(...) }
}
```

I did not observe a hang on device — in my offline testing these calls failed fast on DNS rather than stalling — so this is from reading the change, not a reproduction.

jvsena42 (commented):
Follow-up on the sign-out cancellation hardening in `54a620df0`.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -923,7 +918,21 @@ class PaykitSdkService @Inject constructor(
         ).also { sdk = it }
     }
 
-    private fun bootstrap() = PubkySessionBootstrap.withPubkyClientConfig(pubkyClientConfig)
+    private fun bootstrap() = PubkySessionBootstrap.withPubkyClientConfig(
+        clientId = BitkitPaykitSdkConfig.clientId,
+        pubkyClient = pubkyClientConfig,
+    )
+
+    private fun approvalBootstrap(authUrl: String, approvedClientId: String): PubkySessionBootstrap {
```
jvsena42, line 930: **This check can never fail.**

- `:927` parses the client ID out of `authUrl`.
- `:928` compares it to `approvedClientId`.
- But `approvedClientId` came from parsing that same `authUrl`: `PubkyRepo.kt:944` sets it, `PubkyAuthApprovalViewModel.kt:92` stores it, `:214` passes it back here.

Parsing is deterministic, so the two values are always equal.

`approvedClientId.isNotBlank()` is always true too — rc50 rejects empty client IDs when parsing the URL.

Same for the guard at `PubkyAuthApprovalViewModel.kt:171`: `:170` already returns when `authUrl` differs, so the re-parse gives back an identical client ID. Its test only passes because `parseAuthUrl` is stubbed to return two different results for one URL.

The important part: `bootstrap()` at `:921` pinned approvals to `BitkitPaykitSdkConfig.clientId`. This function replaces that pin with a check that does nothing, and both call sites (`:337`, `:352`) now accept whatever client ID the request carries.

jvsena42 (commented):
One more on the client-ID change.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
```diff
@@ -339,11 +344,12 @@ class PaykitSdkService @Inject constructor(
     suspend fun approveAuthWithCompanionClaim(
         authUrl: String,
         expectedCapabilities: String,
+        approvedClientId: String,
         secretKeyHex: String,
         claim: PubkyAuthCompanionClaim,
     ) {
         isSetup.await()
-        bootstrap().approveAuthWithCompanionClaim(
+        approvalBootstrap(authUrl, approvedClientId).approveAuthWithCompanionClaim(
```
jvsena42, line 353: **This path hands out the account xpub, and it is no longer pinned to Bitkit.**

- `:352` builds the bootstrap from the request's client ID, so the pin to `BitkitPaykitSdkConfig.clientId` in `bootstrap()` (`:921`) no longer applies here. See my comment on `:926` for why the replacement check does not catch anything.
- The payload is `WatchOnlyAccountClaimCodec.encode(..., xpubSerializer::serialize)` — `WatchOnlyAccountRepo.kt:82` and `:103` — which carries the account xpub.
- So any app whose `pubkyauth://` URL requests `PubkyAuthClaim.WATCH_ONLY_ACCOUNT_CAPABILITIES` with `x-bitkit-claim=watch-only-account-v1` can now obtain that signed claim. The only remaining gate is the user approving on the consent sheet.

Suggest keeping this call on `bootstrap()`, and letting only the generic `approveAuth` (`:337`) use the request's client ID.

jvsena42 (commented):
The consequence of the client-ID change on the companion-claim path.

app/src/main/java/to/bitkit/ui/screens/profile/PubkyAuthApprovalViewModel.kt
ben-kaufman: Fixed in 447e7819e. The impossible ViewModel mismatch branch and its mock-only test are removed. The service boundary now throws PubkyAuthRequestError.RequesterChanged, which maps to the existing localized invalid-authorization message.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: I am keeping this non-cancellable intentionally. Sign-out is one transaction, and cancellation after remote revocation starts can leave the server grant revoked while the keychain and local session remain active. The network work is bounded by the configured 30-second request timeout. Completing endpoint cleanup, revocation, and local teardown after the initiating ViewModel leaves avoids an ambiguous half-sign-out.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: Agreed for the ViewModel check. Both values came from reparsing the same immutable URL, and its test manufactured an impossible result, so I removed both in 447e7819e. I kept one service-boundary invariant because that API receives the raw request and the user-approved client identity separately, and it should fail closed if a future caller wires them inconsistently. The new tests cover accepting an external requester and rejecting a mismatched approved identity without mocking the parser.

app/src/main/java/to/bitkit/services/PaykitSdkService.kt
ben-kaufman: This is intentional. Bitkit is acting as a general Pubky authenticator here, so an explicitly approved requester is not restricted to Bitkit’s own client ID. The watch-only account is disclosed only when the request contains the exact claim marker and capabilities, and the user separately approves that privacy-sensitive disclosure. Pinning this to bitkit.to would prevent approved external services such as Paykit Server from receiving the claim.

jvsena42: Indeed, got it

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -996,23 +996,18 @@ class AppViewModel @Inject constructor(
...
-                publicPaykitRepo.syncPublishedEndpoints(publish = false)
-                    .onSuccess {
-                        settingsStore.update { it.copy(publicPaykitCleanupPending = false) }
-                    }
-                    .onFailure {
-                        Logger.warn("Failed to retry public Paykit endpoint removal for '$reason'", it, context = TAG)
-                    }
+            val reconciliationResult = when {
+                settings.sharesPublicPaykitEndpoints -> publicPaykitRepo.syncCurrentPublishedEndpoints()
+                settings.sharesPrivatePaykitEndpoints -> publicPaykitRepo.syncLocalReceiverMarker()
```
jvsena42: **Non-blocking:** this branch looks unreachable.

Reaching it needs `sharesPublicPaykitEndpoints == false` and `sharesPrivatePaykitEndpoints == true`. Every writer of the pair:

- `ContactPaymentSettingsRepo.kt:38-39` — public `true`, private `canUsePrivateContactPayments`
- `ContactPaymentSettingsRepo.kt:63-64` — restores both from `previous`
- `ContactPaymentSettingsRepo.kt:91-92` — both `false`
- `ContactPaymentSettingsRepo.kt:112` — restores public to `previous`, leaves private `false`
- `PubkyRepo.kt:306` — private `false` only
- `PubkyRepo.kt:1298-1299` and `SettingsStore.kt:191-192` — both `false`

None produces that combination. Both paths that set `publicPaykitCleanupPending = true` (`ContactPaymentSettingsRepo.kt:132`, `PubkyRepo.kt:1096`) leave public `true`, so the first branch always wins.

If it ever does become reachable it would also drop the retry it is meant to perform: `syncLocalReceiverMarker()` computes `isDiscoverable = publicSharing || privateSharing` (`PublicPaykitRepo.kt:264`), which is `true` here, so it refreshes the local marker and leaves the published public endpoints in place — and then `onSuccess` clears `publicPaykitCleanupPending`.

Is this intentionally defensive, or is there a private-only sharing state that the settings writers do not produce yet?

jvsena42 (commented):
One non-blocking question on the reconciliation branch.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Good catch. There is no current writer for that private-only combination, and the branch would not have removed stale public endpoints if it were reached. Cleanup now republishes only when public sharing is enabled; otherwise it always removes the public endpoints. I also updated the defensive private-only test to assert removal and ensure the marker-only path is not used. Fixed in a53d912d1.

app/src/main/java/to/bitkit/ui/screens/profile/PubkyAuthApprovalSheet.kt
```diff
@@ -379,6 +380,11 @@ private fun ColumnScope.ApprovalDetails(
         VerticalSpacer(26.dp)
 
         DescriptionText(serviceName = uiState.serviceName)
+        VerticalSpacer(8.dp)
+        BodyS(
+            text = stringResource(R.string.profile__auth_approval_requester, uiState.clientId),
```
jvsena42, line 386: **Non-blocking:** two notes on rendering `uiState.clientId` here.

**1. It is unauthenticated, requester-chosen text, shown as "Requested by %s".** Now that approvals are no longer pinned to Bitkit's own client ID, a third-party requester can set `client_id=bitkit.to` and this line reads "Requested by bitkit.to". A label like "Requester ID" would claim less than "Requested by".

**2. It is unbounded.** `BodyS` defaults to `maxLines = Int.MAX_VALUE, overflow = TextOverflow.Clip` (`ui/components/Text.kt:287-288`), and rc50 permits client IDs up to 253 characters (`ClientId must be at most 253 characters`, in `libpaykit.so`). `ApprovalDetails` is a `Column(Modifier.weight(1f))` at `:379` and there is no `verticalScroll` anywhere in this file, so a long client ID wraps far enough to push `PermissionsSection` (`:390`) and `TrustWarning` (`:393`) past the bottom of that fixed-height column. The Authorize button is a sibling after the weighted column, so it stays visible — the user can approve without seeing what was requested.

`maxLines = 1` plus `TextOverflow.Ellipsis` covers the second point; making the column scrollable would be the fuller fix.

Raising it because the watch-only claim disclosure is gated on this sheet — if the permissions and trust warning can be pushed out of view, that gate is weaker than it looks.

jvsena42 (commented):
One non-blocking note on the requester line.

jvsena42 (dismissed):
Approved with one non-blocking comment

app/src/main/java/to/bitkit/ui/screens/profile/PubkyAuthApprovalSheet.kt
ben-kaufman: Fixed in 83b97d619. The copy now says `Requester ID`, and the requester-controlled value is constrained to one line with a trailing ellipsis so it cannot displace the permissions or trust warning. I kept the existing fixed-height layout rather than adding a broader scrolling refactor.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -345,14 +348,14 @@ class PubkyRepo @Inject constructor(
...
     }
 
-    private suspend fun clearCompletedAuthSessionIfNeeded(didCompleteAuth: Boolean) {
-        if (!didCompleteAuth) return
+    private suspend fun revokeCompletedAuthSessionIfNeeded(shouldRevokeSession: Boolean) {
+        if (!shouldRevokeSession) return
         runSuspendCatching {
             withContext(NonCancellable + ioDispatcher) {
-                pubkyService.clearSessionAccess()
+                pubkyService.signOut()
```
ovitrif, line 360: I noticed that after `completeAuth()` persists a grant, `revokeCompletedAuthSessionIfNeeded` calls `pubkyService.signOut()` and only logs if that fails. `completeAuthentication` then restores `PubkyAuthState.Idle` because `_publicKey` was never set. `revokeIncompleteIdentitySessionIfNeeded` uses the same cleanup after `createIdentity` has already signed up. `PaykitSdkService.signOut()` does not delete `PAYKIT_SESSION` itself, and `PubkyRepo.signOut()` treats a failed revoke as local preservation, so a canceled Ring completion or failed profile publication can leave a keychain session behind an idle UI. The next `initialize()` or `snapshotSessionBackupState()` can then restore or back up a grant the in-app flow already abandoned. Could we fall back to `forgetSessionAccess()` when this cleanup `signOut()` fails so `PubkyAuthState.Idle` matches the keychain?

ovitrif (changes requested):
Canceled or failed incomplete auth can still leave a keychain session behind an idle UI. Cleanup `signOut()` only logs on failure, then `restoreAuthStateAfterAuthFlow()` sets Idle because `_publicKey` was never committed. The next `initialize()` or backup snapshot can restore or back up that abandoned grant.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in 21d5a06a9. Both abandoned-flow cleanup paths now attempt remote sign-out first and fall back to forgetSessionAccess if revocation fails, so Idle cannot hide a persisted session. Normal user-initiated sign-out remains unchanged. The existing canceled-auth and failed-profile-publication tests now cover the fallback and preservation of the original failure.

piotr-iohk (dismissed):
lgtm

jvsena42 (dismissed):

ben-kaufman:
Follow-up cross-platform lifecycle audit is fixed in `a681dfc6c`: auth cleanup is armed before activation can persist a session, profile deletion clears publication intent before a failed sign-out can queue republishing, backup restore proceeds safely when old-session revocation fails, and keychain teardown now attempts both credentials in session-first order. Regression tests cover each edge case. The branch was also rebased onto current master `4bb9d929f`; detekt succeeds with only pre-existing repository warnings.

ben-kaufman:
One final self-review follow-up is in `d2634de17`: the explicit forget-session path now resets the in-memory Paykit runtime in `finally`, even when SDK/keychain teardown throws, so restore cannot continue with the old runtime instance. Detekt remains green apart from the existing baseline warnings.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -345,14 +349,25 @@ class PubkyRepo @Inject constructor(
...
+            withContext(NonCancellable + ioDispatcher) {
+                pubkyService.signOut()
+            }
+        }.exceptionOrNull() ?: return
+
+        Logger.warn("Failed to revoke abandoned Pubky session", revocationError, context = TAG)
         runSuspendCatching {
             withContext(NonCancellable + ioDispatcher) {
-                pubkyService.clearSessionAccess()
+                pubkyService.forgetSessionAccess()
```
ovitrif, line 367: This fallback calls `pubkyService.forgetSessionAccess()` when the remote revoke fails, but its result only reaches `.onFailure { Logger.warn(...) }`, so a second failure is swallowed and nothing else clears the session. `forgetSessionAccess()` in `PaykitSdkService.kt:802` only calls `handle.forgetSessionAccess()` and `resetRuntime()`; it never touches `Keychain.Key.PAYKIT_SESSION`. The base's `clearSessionAccessLocked()` did delete both keys directly on this path, and this PR removed that. What makes the gap concrete is that `PubkyRepo` still has a network-independent local teardown: `clearLocalState()` at `:1299` deletes `PAYKIT_SESSION` and `PUBKY_SECRET_KEY` under `runCatching`, and `signOut()` calls it at `:1118` while `wipeLocalState()` calls it at `:1128`. `discardAbandonedSession()` is the only cleanup path that never does, so it depends entirely on the SDK calling back into `PaykitSdkSessionProvider.clearSessionAccess()` (`:1100`), which has no in-app caller to guarantee it. When both calls fail, `completeAuthentication()` still reaches `restoreAuthStateAfterAuthFlow()` and shows `PubkyAuthState.Idle` while the grant survives in the keychain, and `initialize()` reads that key on the next launch and restores it as `InitResult.Restored`. That is the same ghost session this fallback was added to close, one level deeper. Could we call the existing `clearLocalState()` as a last resort when the `forgetSessionAccess()` fallback also fails, so the keychain always matches `PubkyAuthState.Idle`?

app/src/test/java/to/bitkit/repositories/PubkyRepoTest.kt
```diff
@@ -233,19 +247,71 @@ class PubkyRepoTest : BaseUnitTest() {
...
+        whenever(pubkyService.signOut()).thenAnswer { throw TestAppError("Server error") }
 
         val authRequest = startAuthForTesting()
         approveAuthForTesting(authRequest)
         val result = sut.completeAuthentication()
 
         assertTrue(result.isFailure)
-        verifyBlocking(pubkyService) { clearSessionAccess() }
+        verifyBlocking(pubkyService) { signOut() }
+        verifyBlocking(pubkyService) { forgetSessionAccess() }
```
ovitrif, line 264: This test stubs only `pubkyService.signOut()` to throw and then asserts that `forgetSessionAccess()` was called, so it proves the fallback is invoked but not that the abandoned session is actually gone. Nothing here stubs `forgetSessionAccess()` to fail as well, and nothing asserts `keychain.delete(Keychain.Key.PAYKIT_SESSION.name)` — the sibling cases at `:828-830` and `:1089-1091` do make that assertion, so the style is already established in this file. That matters because the whole point of this change is that `PubkyAuthState.Idle` can no longer hide a persisted session, and as written both added tests would still pass if `forgetSessionAccess()` left `PAYKIT_SESSION` behind. Could we add a case where `pubkyService.forgetSessionAccess()` also throws and assert the session credentials are cleared anyway?

ovitrif (dismissed):
The ghost-session fix is close but stops one level short. `discardAbandonedSession()` now falls back to `forgetSessionAccess()` when the remote revoke fails, and that is the right shape, but `forgetSessionAccess()` does not delete the keychain credentials the way the code it replaced did. `PubkyRepo` already has a network-independent `clearLocalState()` that both `signOut()` and `wipeLocalState()` use; the abandoned-session path is the only cleanup that never reaches it, so a canceled Ring completion or a failed identity creation can still leave a live grant behind `PubkyAuthState.Idle` when both calls fail.

One non-blocking note follows: the new fallback has no test that would fail if it stopped clearing the session.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in b748938b6. If both revoke and SDK forget fail, we now run the existing local cleanup under NonCancellable on IO, deleting both persisted credentials and clearing authenticated state.

app/src/test/java/to/bitkit/repositories/PubkyRepoTest.kt
ben-kaufman: Added the double-failure regression in b748938b6. It makes both signOut and forgetSessionAccess fail after cancellation and verifies both keychain deletions and unauthenticated state. The full local unit suite passes.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
```diff
@@ -345,14 +349,28 @@ class PubkyRepo @Inject constructor(
...
             }
         }.onFailure {
-            Logger.warn("Failed to clear canceled Pubky auth session", it, context = TAG)
+            Logger.warn("Failed to forget abandoned Pubky session access", it, context = TAG)
+            withContext(NonCancellable + ioDispatcher) {
+                clearLocalState()
```
jvsena42: Question: should this last-resort cleanup preserve `publicPaykitCleanupPending`?

`clearLocalState()` defaults it to `false` (:1302 → :1305 → :1320), so it overwrites a pending flag an earlier partial-failure sign-out set at :1121 (server sign-out OK, endpoint removal failed). The automatic background retry via `AppViewModel.retryPendingPaykitEndpointRemoval` (:998) is then skipped.

Not as bad as it sounds — `signOut():1103` calls `removeBitkitPaymentEndpoints()` unconditionally and `applyPublishedEndpoints` reconciles, so the stale endpoints still go on the next online sign-out, on `deleteProfile()`, or get replaced on the next re-enable. So it's a lost retry, not a permanent leak. Reachability is also thin: it needs the prior partial-failure sign-out, a cancelled auth after `completeAuth()` persisted, a failing network `signOut()`, *and* a failing local `forgetSessionAccess()`.

Note :1131 (`wipeLocalState()`) has the same defaulting, which makes me think the `false` default is a deliberate "full local reset" convention rather than an oversight here — in which case ignore this.

If you do fix it, don't add a bare `settingsStore.data.first()`: this `.onFailure {}` block isn't inside a `runSuspendCatching`, so a DataStore read failure would propagate out and displace the original auth error. `runSuspendCatching { ... }.getOrDefault(true)`, or just pass `true` here.

jvsena42 (commented):
Went through the auth/session/credential lifecycle at head against full-file context, and pulled the rc51 AAR to check the error-context strings rather than assume them. Notes:

- **rc51 context strings match.** The native lib contains the exact literal `restore Pubky grant session from platform provider`, so `canDeferStaleSession` (`PaykitSdkService.kt:1088-1089,1108`) matches it. rc51 also adds `grant credential has expired` and `stored grant credential client key does not match the grant cnf`, and confirms `signOut` fails with `cannot revoke Pubky grant without persisted session access` when nothing is persisted — which is what the `discardAbandonedSession` fallback chain is built around.
- **Cancellation semantics are sound.** `ServiceQueue.CORE.background` runs under its own `SupervisorJob`, so cancelled callers still wait for the Rust call and then get `CancellationException`; the `NonCancellable` placement at :297-300, the flag clear at :334/:575, and the `catch (CancellationException)` paths behave as intended. No path leaves a persisted session with `_publicKey == null` except through logged double failures, which now fall back to local credential deletion.
- **`completeAuthentication` arms its revoke flag correctly** — `shouldRevokeSessionOnFailure` is set inside `withContext(NonCancellable)` *after* `waitForAuthApproval` returns, so a cancelled or timed-out Ring prompt sits outside the armed region and cannot revoke a pre-existing identity. (iOS diverged here and does have that problem; worth not "aligning" the two in that direction.)
- **Backup restore ordering is fine:** `restoreSessionBackupState` (METADATA, `BackupRepo.kt:677`) runs before `paykitSdkService.restoreBackupState` (WALLET, `PrivatePaykitRepo.kt:523`), so `forgetSessionAccess()` doesn't wipe a restored SDK blob.
- **Requester pinning holds** — `approvalBootstrap` uses the request's own client ID, and the displayed and approved `clientId` come from the same immutable URL, so no TOCTOU.
- `createIdentity`'s revoke-while-live-session path is only reachable via `bitkit://screen/create-profile`, and `ScreenDeepLinkRuntime.isEnabled` is hard-coded `false` in `app/src/release` — debug-only.

**One migration gap, non-blocking** (Paykit is behind `isPaykitEnabled`, default false in `SettingsStore.kt:42`, so this reaches dev/QA devices rather than shipped users — but it becomes user-facing the moment the flag flips on with rc46-era sessions still on disk).

`canDeferStaleSession` (`PaykitSdkService.kt:1088-1089`, `:1108`) matches a single literal:

```kotlin
fun canDeferStaleSession(errorContext: String): Boolean =
    errorContext == STALE_SESSION_RESTORE_CONTEXT && hasSessionAccess()

const val STALE_SESSION_RESTORE_CONTEXT = "restore Pubky grant session from platform provider"
```

consumed at `:178` as `if (!sessionProvider.canDeferStaleSession(e.context)) throw e`.

In rc51 that literal (`paykit-ffi/src/session.rs:372`) wraps only an error *from* `restore_session`. A session that **restores successfully** but isn't grant-backed produces a different, non-deferrable context: `"Pubky session must be grant-backed"` (`session.rs:431` `validate_grant_session_client_id`, also `paykit-sdk/src/identity.rs:289`).

Why an upgrade lands there:

1. `libs.versions.toml` goes `0.1.0-rc46` → `0.1.0-rc51`, i.e. **pubky 0.8.0 → 0.11.0**.
2. rc46 persists `export_secret()` — pubky 0.8.0's `<pubkey_z32>:<cookie>`. pubky 0.8.0 has zero occurrences of `as_grant`/`GrantCredential` and rc46 zero of `grant-backed`.
3. pubky 0.11.0 `pubky.rs:382 restore_session` still accepts legacy tokens by design (no `pubky-grant-credential-` prefix → `import_secret` → cookie credential), and `grant/view.rs:28` documents that cookie-backed sessions return `None` from `as_grant()`.
4. `loadSessionAccess()` (`:1063-1081`) builds `PubkySessionAccess` from the raw keychain string with no format check.
5. → non-deferrable throw → Paykit never initialises.

I grepped the full `origin/master...origin/codex/paykit-rc50-auth` diff: no legacy/cookie/migration handling anywhere.

Android degrades more gracefully than iOS but is still non-functional: `PubkyRepo.initialize()` (`:166-170`) wraps this in `runSuspendCatching { }.onFailure { Logger.error(...) }.getOrNull() ?: return@withContext`, so it logs and returns, and `ContentView.kt:1914-1920 navigateToProfile` has no error branch, so the user does reach `Routes.PubkyChoice`. But `startAuth()`/`signUp()` all begin with `isSetup.await()` (`PaykitSdkService.kt:206+`), and `initialize()` completed that deferred exceptionally, so re-auth fails too. `PubkyRepo` is constructed once per process, so every cold start reproduces it.

Caveat that decides severity: this depends on the homeserver still revalidating the stored cookie. Valid cookie → grant-backed error → broken. Expired → `RequestExpired` → deferrable → re-sign-in recovers.

Fix options, smallest first: widen the predicate to the grant-backed and client-ID-mismatch contexts (or invert it — defer on any identity error when a stored session exists); or clear the stale session on the non-recoverable branch so the re-sign-in fallback runs.

app/src/main/java/to/bitkit/repositories/PubkyRepo.kt
ben-kaufman: Fixed in c4611067d. The last-resort local cleanup now preserves publicPaykitCleanupPending=true, so failed remote endpoint removal is retried after authentication is restored. It still clears both persisted credentials even if SDK sign-out and forgetting fail. The double-cleanup-failure regression now asserts the pending flag as well.

ben-kaufman:
Addressed the legacy-session restore gap in c4611067d. The stale-session predicate now recognizes rc51's cookie-backed and grant-client mismatch contexts in addition to restore failure, while still requiring stored session access and rejecting unrelated identity errors. This lets the existing re-authentication path run instead of leaving startup permanently failed. Mirrored on iOS #697. The cleanup-flag follow-up is also fixed and regression-tested; full Android unit tests passed.

jvsena42 (commented):
Re-reviewed at `c4611067d`. **No HIGH/MEDIUM — nothing to file.** One candidate died on verification and I'm recording why below so it doesn't get re-chased.

**Migration was the top risk for this bump and the fix is correct — traced through the rc51 Rust, not assumed.** For a stored rc46 cookie: `paykit-ffi/src/session.rs:367-375` `restore_session` succeeds (pubky 0.11 still accepts legacy tokens) → `validate_grant_session_client_id:430` → `Identity{"Pubky session must be grant-backed"}` → `runtime/mod.rs:355` propagates with `?` and no re-wrap → `sdk.rs:182-188` → `errors.rs:40-44` → Kotlin `PaykitException.Identity.context` → `canDeferStaleSession:1088-1093` returns true. The client-ID mismatch format at `session.rs:437-439` is ``Pubky grant client ID `{actual}` did not match `{expected}` ``, so the `startsWith` prefix is exact rather than an equality miss. After deferral, local-seed users re-sign-in and `persistSessionAccess` overwrites the cookie with a grant secret — a transparent migration — while Ring users get `RestorationFailed` plus the expired-session toast and re-auth overwrites it.

**`PAYKIT_SDK_STATE` is format-stable across the bump**, which matters because `runtime/mod.rs:358-377` writes an identity blob even with no session, so every rc46 device has one. `IdentityState` is field-identical between the tags including `sign_out_generation`, `paykit-ffi/src/storage.rs` is byte-identical with the same blob version constant, and the `storage/records.rs` serde shapes are unchanged.

**Blast radius is nil.** Paykit has never shipped — `v2.3.2`, `v2.4.0` and `v2.4.1` `libs.versions.toml` carry no paykit dependency, and rc46 exists only on `master`. So every upgrade concern here is dev/QA by construction.

**Binding API diff rc46→rc51** (javap): `PubkyClientConfig` gained `authRelayUrl`, and since the app uses `defaultPubkyClientConfig()` it picks up the SDK default — the same relay rc46 URLs already carried — with the E2E override unchanged. `PubkySessionAccess`'s constructor now throws, but only validates `ClientId`, which is the constant `bitkit.to`/`staging.bitkit.to`. `resumeAuth`'s signature changed and is unused. No casts, `!!`, or catch-alls were introduced to keep anything compiling, which was the specific failure I was looking for.

**Sign-out semantics check out.** rc51 `sign_out` is revoke → `clear_local_identity_state` → provider `clearSessionAccess` → `clearPubkySessionCredentials`, and `Keychain.delete` is `edit { remove }` which doesn't throw on a missing key, so Ring users without `PUBKY_SECRET_KEY` can't fail post-revocation. `deleteProfile`'s `markPublicCleanupPending = hasPaykitState()` keeps `signOut`'s `hadPaykitState` true because `hasPaykitState` includes `publicPaykitCleanupPending`, so the retry flag survives `clearLocalState`. The rc50→rc51 delta itself is one 12-line patch in `payment_requests/derivation.rs` changing an event sort key — no amount or destination derivation touched.

The candidate I dropped: rc51 rejects pre-grant `pubkyauth://` URLs and that rejection reaches the approval toast as a raw `code=protocol_error, context=…` string. Real mechanically, but pre-existing — `master` at rc46 has the identical unwrapped `parseAuthUrl` and the same `PubkyAuthErrorMessage` fallthrough, so any malformed URL already surfaced a raw string; rc51 only adds one more trigger. It's also the same mechanism as my existing `:172` thread, which you closed by mapping the two new `IllegalArgumentException` sites. And the grant-only behaviour is intentional SDK design per your note about the Ring-side compatibility gap, so there's nothing here for Bitkit to fix.

ben-kaufman:
Thanks @jvsena42. One explicit scope correction to the migration part of your review: Paykit has not launched, so we intentionally do not support migrations from earlier development builds. d8882611f removes the cookie-backed and old client-ID session recovery cases and their legacy-only test cases. Normal offline/expired recovery for current-format grants and the pending endpoint-cleanup fix remain intact. The same cleanup is in iOS #697, and signup #1224 is restacked on top. Compilation, all 2,316 unit tests, and lint pass here; the restacked signup branch also passes all 2,346 tests.

This is intentional, not an outstanding migration issue to fix before launch.

jvsena42 (commented):
Reviewed `d8882611f` ("refactor: remove paykit migration handling"). **No HIGH/MEDIUM on Android — safe.**

I checked this carefully because I flagged the same removal as a problem on the iOS twin (#697), and because my own note that Paykit has never shipped may be what prompted it. The removed clauses were `"Pubky session must be grant-backed"` and the `"Pubky grant client ID \`"` prefix. The first is genuinely migration-only — it fires for an rc46 cookie session, and an rc46-era QA device now gets a silent init failure and the unauthenticated choice screen rather than a deferral, which is consistent with dropping migration support on a feature that never shipped.

The second is the one that isn't a migration condition — `validate_grant_session_client_id` raises it whenever a stored grant's client id differs from the one the build passes, so on iOS a Debug build run over a TestFlight build hits it. **Android is not exposed**, and the reason is structural rather than luck: `app/build.gradle.kts` gives each flavour an `applicationIdSuffix` (`.dev` / `""` / `.tnet`), so the flavours have separate keychains and a grant can't cross between them. `Env.network` is compile-time from `BuildConfig.NETWORK`, `BitkitPaykitSdkConfig.clientId` is identical at every commit on this branch, and `vssStoreIdPrefix` is network-scoped so a backup can't carry a grant across networks either. `approvalBootstrap` uses the requester's client id but never persists to `PAYKIT_SESSION`.

So the mismatch can't arise here and the removal costs nothing on this platform. I've raised the client-id half on #697, where the single shared bundle id makes it reachable — you may want to keep the prefix clause on Android purely for parity if iOS restores it, but there's no defect here either way.

Test narrowing is fine: `missing session or unrelated identity failures are not deferred` still asserts the surviving context with a null keychain, so no live coverage was lost.

jvsena42 (approved):
Manually reviewed
✅ Tested signin and signet flows on emulator

Issue #721, opened 2026-01-23
Android channel closed via Connections > Detail > Close Connection didn't result in balance update on home


piotr-iohk:
### What happened?

Android channel closed via Connections > Detail > Close Connection didn't result in balance update on home screen (likely UI issue requiring call to updateBalances or smth… the receive flow was aware there are no channels so I could create CJIT)

### Steps to Reproduce

- Have wallet with 2 channels
- close 1 of them Connections > Detail > Close Connection

It seems spending balance is not updated.
Even after restarting (force stopping) the app the balance is still not updated.

### Logs / Screenshots / Recordings

[bitkit_logs_1769173425325.zip](https://github.com/user-attachments/files/24821360/bitkit_logs_1769173425325.zip)

PR #1217, opened 2026-09-02
fix: expose rejected incoming payment requests

### LINKED ISSUE #1209: fix: expose rejected incoming payment requests
Counterpart: synonymdev/bitkit-ios#714

## Scope

- Add redacted, reason-specific diagnostics for parse-time rejection.
- Implement the shared requested/open-time failure UX after retries exhaust.

## Acceptance criteria

- Behavior and localized user outcomes match the iOS contract.
- Unit tests cover parse and resolution failures.

ovitrif:
Fixes #1209

### Description

- Records stable parse, resolution, and presentation failure reasons while redacting payment-request counterparties and targets.
- Retries explicitly opened payment requests 15 times before showing localized terminal feedback and keeping the request actionable.
- Adds accessibility tags, regression coverage, documentation, and a reproducible rejected-request journey.

### Preview

https://github.com/user-attachments/assets/03351b25-6dda-4433-a52e-fb21fa1c5baa

#### Manual Tests

- [ ] **1.** Sender Receive → Payment Request Details → request 1 sat from a saved contact → Payment Request Sent: request reaches the receiver.
- [x] **2.** Sender Settings → disable Paykit → receiver Home → Payment Requests in-sheet → tap Pay: 15 attempts run at two-second intervals, then the localized unavailable toast appears.
- [x] **3.** Payment Requests in-sheet after terminal feedback → use the same row: it remains available for retry or dismissal.

#### Automated Checks

- Unit tests added in `PaykitPaymentRequestDiagnosticsTest.kt`: verify parse and resolution diagnostics redact valid and invalid counterparties and Throwable messages while retaining a stable error type.
- Unit tests modified in `PaykitPaymentRequestRepoTest.kt` and `PublicPaykitRepoTest.kt`: cover stable parse and resolution failure reasons and suppress repeated expired-record diagnostics.
- Unit tests modified in `AppViewModelSendFlowTest.kt`: cover 15 explicit attempts, final redacted resolution diagnostics, decoded-target log redaction, localized terminal feedback, expiration during backoff or resolution, automatic-batch continuation, and request-sheet restoration without replacing unrelated sheets.
- Instrumented tests modified in `PaymentRequestsScreenTest.kt`: cover stable request, Pay, and Dismiss accessibility tags; the focused class passes 5/5 on API 37 and 5/5 on API 36.
- Prior-head verification before the latest master merge: `AppViewModelSendFlowTest` passed 202/202; `just compile`, `just test`, and `just lint` passed.
- Latest master merge: tests, builds, lint, device QA, CI triggers, and CI observation were not run.

The full two-wallet journey passed on API 37 before the latest master merge: a delivered 1-sat request became unresolvable after its sender disabled Paykit, produced 15 redacted `resolution_failed` attempts, showed terminal feedback, and returned to the request sheet with the same row actionable. The prior head passed GitHub's full local and staging E2E matrix, and the focused payment-request UI class passed on API 37 and API 36.


piotr-iohk (dismissed):
QA LGTM.

Tested latest (`95788b9`) Pixel emu against iOS `codex/paykit-payment-proofs`, regtest. Incoming 1 sat + 27k from iOS. Tap Pay on an unresolvable request:

- 15 `resolution_failed` attempts at ~2s
- `Stopped retrying requested incoming Paykit payment request after '15' presentation attempts`
- Toast: "Payment Request" / "The payment request is no longer available."
- Row stays with Pay and Dismiss

Opening a still-resolvable request goes to Confirm with swipe disabled. That is the master `isAmountInputValid` hole, not this PR. Already fixed on #1178 (`e04115003`); standalone: #1218 / #1221. Not a blocker for this toast/retry path.

piotr-iohk:
Please resolve conflicts.

piotr-iohk (dismissed):
QA LGTM.

Latest (`800b902d`) after the conflict rebase.

I already ran the full unresolvable-request journey on `95788b9` (15 × ~2s, redacted `resolution_failed`, unavailable toast, row stays with Pay/Dismiss). The only commit since that QA is the valid-pubky redaction test. Rebase onto master (incl. #1178) does not change the toast/retry path.

This pass:

- Installed `800b902d` on Pixel_6 emu. Wallet restored (Alice / `pubkyff…qyqnsuy`), contact payments on, Paykit session re-signed and published `btc-regtest-p2wpkh`.
- Focused unit tests pass: `PaykitPaymentRequestDiagnosticsTest`, `PaykitPaymentRequestRepoTest`, `PublicPaykitRepoTest`, `AppViewModelSendFlowTest`.
- Journey XML parses.

CI green on this head, including local + staging E2E.

Happy to approve.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2567,6 +2602,9 @@ class AppViewModel @Inject constructor(
             is Scanner.NodeId -> handleNonPaymentScan { onScanNodeId(scan) }
             is Scanner.Gift -> handleNonPaymentScan { onScanGift(scan.code, scan.amount) }
             else -> {
+                clearActiveContactPaymentContext(
```
ben-kaufman: Could we keep the request-sheet restore from being canceled on invalid target failures? On the fifteenth attempt this call reaches `deferPaymentRequestPresentation()`, which starts `showSheet(Sheet.PaymentRequests)`, and the following `hideSheet()` cancels that transition or closes the restored sheet. The toast appears, but requests opened from the sheet end on the underlying screen. A test where `beginPaymentRequest()` returns `Opened` with an invalid target would cover this path.

ovitrif: Skipped hideSheet after deferPaymentRequestPresentation on invalid target failures so the restored PaymentRequests sheet stays open. Added a test where beginPaymentRequest() returns Opened with an invalid target.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2567,7 +2602,11 @@ class AppViewModel @Inject constructor(
             is Scanner.NodeId -> handleNonPaymentScan { onScanNodeId(scan) }
             is Scanner.Gift -> handleNonPaymentScan { onScanGift(scan.code, scan.amount) }
             else -> {
-                hideSheet()
+                val hasIncomingPaymentRequest = activeIncomingPaymentRequest() != null
+                clearActiveContactPaymentContext(
+                    failureReason = IncomingPaykitPaymentRequestFailureReason.InvalidPaymentTarget,
+                )
+                if (!hasIncomingPaymentRequest) hideSheet()
```
ben-kaufman: Could we also skip the generic QR toast when this was an incoming request? On the fifteenth failure, `clearActiveContactPaymentContext()` enqueues `PaymentRequestUnavailableToast`, then the toast below immediately replaces it because `ToastQueueManager.enqueue()` dismisses the current toast. The test only searches the mocked call history, so it passes even though the terminal toast is not the final visible one. Could we assert that the terminal toast is the last enqueue?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3600,6 +3661,9 @@ class AppViewModel @Inject constructor(
             description = context.getString(R.string.wallet__payment_request_mismatch),
             testTag = "PaymentFailedToast",
         )
+        clearActiveContactPaymentContext(
```
ben-kaufman: Could we preserve the request-sheet restore in decoded-target failures too? A mismatched Bolt11 reaches this branch on every retry. On the fifteenth attempt, `clearActiveContactPaymentContext()` calls `showSheet(PaymentRequests)`, then `hideSheet()` cancels that transition. The same clear-then-hide order appears in the duplicate-BIP21, invalid-address, and expired-invoice branches. The new guard only covers decode failure.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -873,19 +878,24 @@ class AppViewModel @Inject constructor(
         request: PaykitPaymentRequest,
         generation: Long,
     ): Boolean {
-        val result = privatePaykitRepo.beginPaymentRequest(request).getOrNull()
+        val presentationResult = privatePaykitRepo.beginPaymentRequest(request)
+        val result = presentationResult.getOrNull()
```
ben-kaufman, line 1019: Could we preserve explicit-request expiration before discarding the result here? If the request expires during `beginPaymentRequest()` or the retry window, `isPending()` becomes false and the requested state is cleared without restoring the sheet or showing a terminal result. The linked iOS contract records this case and shows the localized expired toast exactly once, while Android currently leaves the user after the initial Waiting toast with no outcome.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -75,6 +76,23 @@ data class PaykitPaymentRequest(
     val direction: PaykitPaymentRequestDirection = PaykitPaymentRequestDirection.Incoming,
     val lifecycleState: PaymentRequestLifecycleState = PaymentRequestLifecycleState.PROPOSED,
 ) {
+    enum class ParseFailure(
+        val logValue: String,
+        val shouldLogIncomingRejection: Boolean = true,
+    ) {
+        MissingLocalRole("missing_local_role"),
+        UnsupportedLocalRole("unsupported_local_role", shouldLogIncomingRejection = false),
```
ben-kaufman: Could we suppress only the expected outgoing `PAYEE` case here? `parseIncomingPaykitPaymentRequest()` returns `UnsupportedLocalRole` for both `PAYEE` and `UNKNOWN`, so disabling logging on the reason also silently drops malformed `UNKNOWN` records. The linked iOS contract distinguishes a suppressed outgoing request from a logged `unsupported_local_role`. A test with `PaymentRequestLocalRole.UNKNOWN` would cover this.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Skipped the generic QR toast when this was an incoming request so PaymentRequestUnavailableToast stays the last enqueue. The invalid-target test now asserts that.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Applied the same skip-hideSheet guard to mismatched Bolt11, duplicate BIP21, invalid-address, and expired-invoice so the restored PaymentRequests sheet is not canceled. Covered mismatched Bolt11 from the request sheet.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Handle RequestExpired from beginPaymentRequest before discarding the result. Explicit requests now log request_expired and show PaymentRequestExpiredToast once, then restore the request sheet.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ovitrif: PAYEE now maps to outgoing_request and stays silent. UNKNOWN still maps to unsupported_local_role and is logged. Added a PaymentRequestLocalRole.UNKNOWN test.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: The immediate `RequestExpired` result is handled now, but real expiration can clear the state before that branch runs. During retry backoff this collector cancels the retry and clears the requested ID, and during suspended resolution the generation change discards the later error. Both paths still skip the expired toast and sheet restore. Could we preserve the requested expiration here and add clock-driven backoff coverage?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2470,14 +2524,13 @@ class AppViewModel @Inject constructor(
 
         // TODO Workaround for https://github.com/synonymdev/bitkit-core/issues/63
         if (Bip21Utils.isDuplicatedBip21(input)) {
-            hideSheet()
+            clearIncomingPaymentRequestTarget()
```
ben-kaufman: Could we apply the incoming-request early return to these branch-specific errors too? On the final attempt, `clearIncomingPaymentRequestTarget()` enqueues `PaymentRequestUnavailableToast`, then `DuplicatedBip21Toast` replaces it. The same order remains for invalid addresses and expired invoices. Could we skip those scan-error toasts when the helper reports an incoming request and assert the terminal toast is last?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3600,7 +3683,7 @@ class AppViewModel @Inject constructor(
             description = context.getString(R.string.wallet__payment_request_mismatch),
             testTag = "PaymentFailedToast",
         )
-        hideSheet()
+        clearIncomingPaymentRequestTarget()
```
ben-kaufman, line 4058: Could we still close `Sheet.Send` when this runs from confirmation? `clearIncomingPaymentRequestTarget()` skips `hideSheet()` whenever a request context existed. During confirmation that clears the context but leaves the Send sheet open in stale payment-request state, so the user needs another Pay tap to close it. Could we distinguish an already-visible Send sheet and cover it in the confirmation mismatch tests?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Applied the incoming-request early return to duplicate BIP21, invalid-address/network, and expired-invoice failures, so branch-specific scan toasts no longer replace terminal feedback. Added duplicated-BIP21 regression coverage that asserts the terminal toast is last.

Fixed in cbfe1c6

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: The helper now snapshots whether `Sheet.Send` is visible and closes it after clearing the request context, while request-sheet restoration remains intact. The confirmation mismatch regression now opens Send Confirm and asserts it closes.

Fixed in cbfe1c6

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -873,19 +878,28 @@ class AppViewModel @Inject constructor(
         request: PaykitPaymentRequest,
         generation: Long,
     ): Boolean {
-        val result = privatePaykitRepo.beginPaymentRequest(request).getOrNull()
+        val presentationResult = privatePaykitRepo.beginPaymentRequest(request)
+        val result = presentationResult.getOrNull()
         if (!isCurrentPaymentRequestPresentation(request, generation) || isPaymentRequestPresentationBlocked()) {
             return true
         }
+        if (presentationResult.exceptionOrNull() is PaykitPaymentRequestError.RequestExpired) {
```
jvsena42: **🔵 Log the Throwable from a failed beginPaymentRequest at the handling layer**

This PR deletes the only `Logger.warn(..., it, ...)` on this path (`PrivatePaykitRepo.beginPaymentRequest`'s `.onFailure`), and here `exceptionOrNull()` is inspected only for `RequestExpired`. Every other Throwable — SDK decrypt/network, `PrivatePaykitError.InvalidPublicKey` — is dropped and appears in logs only as `reason='resolution_failed'`, repeated 15x. `beginContactPayment` passes `allowPublicResolution = paymentRequest == null`, so for a payment request `preparePrivateContactPayment` rethrows unlogged. AGENTS.md: log errors at the final handling layer. iOS #721 went the other way and added `errorType=` in `PrivatePaykitService+Payments.swift:125`. Add:

```kotlin
presentationResult.onFailure {
    if (it !is PaykitPaymentRequestError.RequestExpired) {
        Logger.warn("Failed to resolve incoming Paykit payment request", it, context = TAG)
    }
}
```

jvsena42 (commented):
Diffed this against the iOS counterpart (#721) semantically. The port is faithful in the parts that carry weight: same 12-reason parse taxonomy and wire strings, same two suppressed reasons, same `ULong.MAX/1000` cap, exhaustive `when` on the presentation taxonomy with no permissive default, same redaction (`counterparty` is the only peer-supplied value that reaches a log, and it is bounded). No rejection reason is sent back to the sender, so there is no privacy leak to the requesting party, and no attacker-supplied error text reaches the UI unescaped.

The one behavioural gap I found was already filed by @ben-kaufman on the `AppViewModel.kt:882` thread, so I have not duplicated it. One low convention item below.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Added final-layer diagnostics with a stable error type and redacted counterparty. The Throwable message is deliberately excluded so arbitrary payment data cannot enter logs, and tests cover both the handling call and redaction.

Fixed in b79c43d

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ben-kaufman: Rechecked at `b79c43d` and this still applies. `retainPaymentRequestPresentationState()` clears the requested ID and cancels its retry as soon as the expired request leaves `pendingRequests`, before `finishExpiredPaymentRequestPresentation()` can run. Expiry during backoff or an in-flight resolution still skips the expired toast and sheet restore. Could we keep this open until a clock-driven case is covered?

docs/payment-requests.md
```diff
@@ -0,0 +1,41 @@
...
+
+Bitkit distinguishes Payment Requests rejected while reading a Paykit record from requests that
+parse successfully but cannot be opened.
+
+## Failure contract
+
+- Parse-time rejection emits a warning with category `parse`, a stable reason, and only the
+  redacted counterparty. It excludes the request id, amount, note, endpoint identifier, and
+  endpoint payload.
+- Open-time rejection emits a warning with category `resolution` or `presentation`, a stable
```
ben-kaufman, line 11: This says open-time rejection logs only the redacted counterparty, but invalid data from an incoming request is passed to `handleScan()`, and the decode failure logs `safeLogInput ?: input`. For normal Paykit targets that is the peer-supplied invoice or address, and the retry path can log it on every attempt. Could we suppress or redact the scan input when an incoming request is active and add coverage for the invalid-target path?

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -939,6 +967,29 @@ class AppViewModel @Inject constructor(
...
 
+    private fun finishExpiredPaymentRequestPresentation(request: PaykitPaymentRequest) {
+        paykitPaymentRequestDiagnostics.logPresentationRejection(
+            request.counterparty,
+            IncomingPaykitPaymentRequestFailureReason.RequestExpired,
+        )
+        val restorePaymentRequestSheet =
+            requestedPaymentRequestId == request.id && shouldRestorePaymentRequestSheet
+        val showExpiredToast = requestedPaymentRequestId == request.id
+        paymentRequestPresentationGeneration++
```
jvsena42: **Regression vs base — this bump is unconditional, so an expired request in an *automatic* batch discards the resolution of every request behind it.**

Base bumped the generation only when `requestedPaymentRequestId == request.id`; here `paymentRequestPresentationGeneration++` sits above that `if` at :979.

Concrete: A and B are presented automatically, A expires. `beginPaymentRequest(A)` throws `RequestExpired` (`PrivatePaykitRepo.kt:370`) → bump → the loop moves to B and does B's full network round trip, after which `isCurrentPaymentRequestPresentation(B, generation)` (:916-920) fails on the stale generation and the result is thrown away. Deterministic whenever an automatic batch has an expired request ahead of a payable one.

The tail re-entry at :848-856 then re-runs the pass, and A is still eligible — `automaticPendingRequests()` filters on `presentedRequestIds` only, not expiry, and :982 just cleared A's backoff. It terminates once `discardExpiredRequestsLocked` prunes A, so it's ~1-2 wasted round trips rather than an indefinite spin, but B's presentation is delayed and re-resolved for no reason.

Fix is just moving the bump inside the `if` at :979. (No need to add `markPresented` for the automatic case — expiry pruning handles A, and `markPresented` no-ops for a request not in `_pendingRequests` anyway.)

Regression test attached in the PR conversation.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -75,6 +76,24 @@ data class PaykitPaymentRequest(
...
+        UnsupportedLocalRole("unsupported_local_role"),
+        NonActionableState("non_actionable_state", shouldLogIncomingRejection = false),
+        MissingTerms("missing_terms"),
+        RecurringRequest("recurring_request"),
+        UnsupportedAsset("unsupported_asset"),
+        InvalidAmount("invalid_amount"),
+        AmountOutOfRange("amount_out_of_range"),
+        NoSupportedEndpoint("no_supported_endpoint"),
+        InvalidExpiration("invalid_expiration"),
+        Expired("expired"),
```
jvsena42: Low: `Expired` inherits `shouldLogIncomingRejection = true`, so every expired request re-logs a rejection warning on every refresh, forever.

An expired proposal is a normal retained record — nothing writes expiry back, `record.state` stays `PROPOSED` (which is exactly why `withExpiredLifecycle:693-699` derives `PROPOSAL_EXPIRED` locally), so it falls past `NonActionableState` into `Expired` and hits `logParseRejection` at :468 on every pass. No dedup or throttle anywhere in `refresh()` or `Logger`.

Polling is foreground-only and settles at 120s, so it's roughly 300 warn lines/hour of use with 10 expired records — not huge, but unbounded in history size and it's noise in the log this PR added for triage.

`Expired("expired", shouldLogIncomingRejection = false)`, same as `OutgoingRequest`/`NonActionableState`. Worth keeping it scoped to `Expired` — `NoSupportedEndpoint` has the same repetition but flags genuinely malformed requests, so that one is real signal.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -3000,7 +3085,9 @@ class AppViewModel @Inject constructor(
                 title = context.getString(R.string.other__lnurl_pay_error),
                 description = context.getString(R.string.other__scan__error__generic),
             )
-            clearActiveContactPaymentContext()
+            clearActiveContactPaymentContext(
```
jvsena42, line 3347: Question, not a defect — this ordering is pre-existing, the PR only added the `failureReason` arg here.

The PR converted five sibling sites to the new `if (clearIncomingPaymentRequestTarget(...)) return` shape (duplicated BIP21 :2529, invalid address :2777, network mismatch :2788, expired bolt11 :3014, unhandled scan :2631) but left this branch toasting before `clearActiveContactPaymentContext`. Since `ToastQueueManager.enqueue` (:42-53) dismisses and shows synchronously, the LNURL error is replaced by the "waiting for updated private payment details" INFO toast (:955-961) in the same stack, then the same out-of-range endpoint is re-resolved 15× over 28s.

Mild — the clobbered description is `other__scan__error__generic`, and the PR's new `PaymentRequestUnavailableToast` (:939-944) does give terminal feedback base didn't have. Was skipping this one deliberate? `:2615` (hardware on-chain-only) has the same shape.

jvsena42 (commented):
The `parsePaykitPaymentRequest` refactor is behaviour-preserving against base for every reason (role/state/terms/asset/amount/endpoints/expiry all produce the same accept-reject set), `IncomingPaykitPaymentRequestFailureReason` covers all `PublicPaykitPaymentResult` cases, and the diagnostics logger correctly redacts the counterparty and never emits the `Throwable` message.

One regression worth fixing before merge, plus two low notes.

jvsena42:
### Regression test — expired request discards the payable request behind it

Reproduces the `finishExpiredPaymentRequestPresentation` generation bump. Test-only, applies cleanly onto `codex/1209-payment-request-errors`.

```
./gradlew :app:testDevDebugUnitTest --tests 'to.bitkit.viewmodels.AppViewModelSendFlowTest.expired request does not discard a later payable request'
```

**Fails on head:**
```
java.lang.AssertionError: expired request invalidated the automatic presentation, so the payable
request was resolved again instead of being shown expected:<1> but was:<2>
  at AppViewModelSendFlowTest.kt:1543
```

**Passes** once `paymentPresentationGeneration++` moves inside the `if (requestedPaymentRequestId == request.id)` block — full class green at 196 tests. `detekt` clean.

Harness note: on unfixed code the re-entry loop recurses forever because the mocked `pendingPaykitPaymentRequests` never prunes the expired entry, so `advanceUntilIdle()` would hang rather than fail. The test bounds it inside the mock instead — `beginPaymentRequest(payableRequest)` uses a `doSuspendableAnswer` that parks on `awaitCancellation()` past the first attempt, so the count assertion reports the regression in plain language.

<details><summary>pr1217-regression.diff</summary>

```diff
diff --git a/app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt b/app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt
index 30fa70582..3a8a36478 100644
--- a/app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt
+++ b/app/src/test/java/to/bitkit/viewmodels/AppViewModelSendFlowTest.kt
@@ -1509,6 +1509,49 @@ class AppViewModelSendFlowTest : BaseUnitTest() {
         assertEquals(Sheet.Send(SendRoute.Confirm), sut.currentSheet.value)
     }
 
+    @Test
+    fun `expired request does not discard a later payable request`() = test {
+        val expiredRequest = paymentRequest()
+        val payableRequest = expiredRequest.copy(paymentRequestId = "payable-request")
+        val bolt11 = "lnbcrt1payableafterexpired"
+        val privateContext = PrivatePaykitPaymentContext("bitkit/server", 7uL)
+        var payableAttempts = 0
+        whenever(privatePaykitRepo.beginPaymentRequest(expiredRequest))
+            .thenReturn(Result.failure(PaykitPaymentRequestError.RequestExpired))
+        whenever(privatePaykitRepo.beginPaymentRequest(payableRequest)).doSuspendableAnswer {
+            payableAttempts++
+            if (payableAttempts > 1) awaitCancellation()
+            Result.success(
+                PublicPaykitPaymentResult.Opened(
+                    paymentRequest = bolt11,
+                    privatePaymentContext = privateContext,
+                ),
+            )
+        }
+        stubLightningScan(bolt11 = bolt11, amountSats = 0u)
+        balanceState.value = BalanceState(maxSendLightningSats = 100_000u)
+        pendingPaykitPaymentRequests.value = listOf(expiredRequest, payableRequest)
+        isPaykitEnabled.value = true
+        pubkyPublicKey.value = testPublicKey
+        whenever(paykitPaymentRequestRepo.refresh()).thenReturn(Result.success(Unit))
+
+        sut.startPaykitPaymentRequestPolling()
+        advanceTimeBy(30.seconds.inWholeMilliseconds)
+        runCurrent()
+        sut.stopPaykitPaymentRequestPolling()
+
+        assertEquals(
+            expected = 1,
+            actual = payableAttempts,
+            message = "expired request invalidated the automatic presentation, so the payable request " +
+                "was resolved again instead of being shown",
+        )
+        verify(privatePaykitRepo).beginPaymentRequest(expiredRequest)
+        verify(privatePaykitRepo).beginPaymentRequest(payableRequest)
+        assertEquals(payableRequest, activeContactPaymentContext()?.incomingPaymentRequest)
+        assertEquals(Sheet.Send(SendRoute.Confirm), sut.currentSheet.value)
+    }
+
     @Test
     fun `cancelled request resolution releases the presentation guard`() = test {
         val request = paymentRequest()
```

</details>

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ovitrif: Expired records are now treated as expected filtering and do not emit repeated parse-rejection warnings. The repository test and failure-contract documentation cover this behavior.

Fixed in 73df07f

docs/payment-requests.md
ovitrif: Incoming request targets now use a fixed scan identifier, and decode failures omit both the target and Throwable details. The invalid-target regression captures the emitted logs and verifies the peer-supplied value is absent.

Fixed in 73df07f

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: This was not deliberate. I aligned the out-of-range LNURL and hardware-only branches with the other incoming-request failures so retry feedback is not replaced by a generic toast, while ordinary hardware-send behavior remains unchanged. The LNURL path now has terminal-feedback coverage.

Fixed in 73df07f

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: I now retain the explicit request until its expiration transition is handled. Expiration during retry backoff or suspended resolution cancels the stale work, restores the request sheet when applicable, and emits `PaymentRequestExpiredToast` exactly once. Both paths have regression coverage.

Fixed in 73df07f

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: The generation now advances only for an explicitly requested expiration. I added the attached automatic-batch regression, which verifies the payable request behind an expired request is resolved and presented once.

Fixed in 73df07f

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -939,18 +975,52 @@ class AppViewModel @Inject constructor(
...
         val requestIds = requests.mapTo(mutableSetOf()) { it.id }
         paymentRequestPresentationRetryAttempts.keys.retainAll(requestIds)
         paymentRequestPresentationRetryJobs.keys.filter { it !in requestIds }.forEach {
             paymentRequestPresentationRetryJobs.remove(it)?.cancel()
         }
-        if (requestedPaymentRequestId?.let { it !in requestIds } == true) {
+        val requestedRequest = requestedPaymentRequest
+        if (requestedRequest != null && requestedRequest.id !in requestIds) {
+            if (paykitPaymentRequestRepo.isExpired(requestedRequest)) {
+                finishExpiredPaymentRequestPresentation(requestedRequest)
```
jvsena42, line 1165: Non-blocking. This `finishExpiredPaymentRequestPresentation` call can restore the PaymentRequests sheet on top of an unrelated sheet the user is mid-flow in.

The other two callers (`:870`, `:896`) run downstream of `isPaymentRequestPresentationBlocked()`, so `currentSheet == null` is guaranteed there. This one runs inside the `pendingRequests.drop(1).collect` collector with no sheet check. Sequence: user taps Pay on R, it doesn't resolve, goes to backoff; user then pastes an unrelated invoice and is in `Sheet.Send(Confirm)`; the retry fires and is dropped as blocked, but `requestedPaymentRequestId`/`shouldRestorePaymentRequestSheet` stay set; R's `expiresAt` passes, the repo prunes, and this path reaches `showSheet(Sheet.PaymentRequests)` — which nulls `_currentSheet` and tears down the user's Send flow. No fund impact (a swipe already in `proceedWithPayment` continues in `viewModelScope`), they just lose the result screen.

One line: `if (restorePaymentRequestSheet && currentSheet.value == null) showSheet(Sheet.PaymentRequests)`. The toast still fires, and if the open sheet was R's own Send sheet the `:808-815` collector already hides it. The tests at `AppViewModelSendFlowTest.kt:866/906` only cover the no-other-sheet case.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -2567,23 +2655,25 @@ class AppViewModel @Inject constructor(
             is Scanner.NodeId -> handleNonPaymentScan { onScanNodeId(scan) }
             is Scanner.Gift -> handleNonPaymentScan { onScanGift(scan.code, scan.amount) }
             else -> {
-                hideSheet()
+                val hasIncomingPaymentRequest = clearIncomingPaymentRequestTarget()
                 Logger.warn(
                     if (scan == null) "Failed to decode scan data" else "Received unhandled scan data '$scan'",
```
jvsena42: Non-blocking, and it's the adjacent branch to the one already fixed. The decode-failure log at `:2612` is now redacted, but this unhandled-scan branch still interpolates the whole `Scanner` value.

A counterparty can publish, under a supported `MethodId`, a string that decodes successfully to a variant this `when` doesn't handle — a lightning address landing as `Scanner.LnurlAddress(address=…)`, say. `beginContactPayment → openContactPayment → handleScan → coreService.decode` succeeds, falls through to `else`, and the peer-supplied payload gets written to the log on each of the 15 retries. Same class as the thread you closed on `docs/payment-requests.md:11`, so the doc's claim is slightly ahead of the code here.

`if (hasIncomingPaymentRequest) "Received unhandled incoming Paykit payment request target" else …` covers it. Separately and pre-existing: `:2618`'s `Logger.info("Handling decoded scan data: $it")` logs the full decoded target on the success path too, which is outside that doc's rejection claim — noting only.

jvsena42 (commented):
Re-reviewed at `73df07f`. **No HIGH/MEDIUM — not blocking.** Two LOW observations inline; both are dev-gated, take or leave them.

I spent most of this pass confirming your pushed fixes are actually correct rather than merely present, and they are:

- **Generation bump** is now inside `if (requestedPaymentRequestId == request.id)`. I traced the case I was worried about — an automatic batch with expired A ahead of payable B: `finishExpired(A)` no longer bumps, `clearPaymentRequestPresentationRetry(A)` returns false, the loop reaches B, `isCurrentPaymentRequestPresentation(B)` passes, `openContactPayment(B)` runs. B is no longer swallowed.
- **Expiry during backoff and in-flight** both resolve to exactly one toast. Backoff: the retry job is cancelled, then one toast + restore. In-flight: the generation bump makes `beginPaymentRequest` return early so `openContactPayment` is never called and no second toast fires. The reverse race (`RequestExpired` thrown before the repo prunes) clears the requested id, so the later emission finds `requestedRequest == null`. No duplicate either way.
- **`hideSheet` after restore**: `clearIncomingPaymentRequestTarget` snapshots `currentSheet is Send` before clearing, and retry attempts only run with `currentSheet == null`, so the 15th-failure `showSheet(PaymentRequests)` is never followed by a `hideSheet()` that would undo it.
- **Final-layer logging**: `logPresentationFailure` now emits only the error class name plus the redacted pubkey, no `Throwable.message`, and `scanLogId` returns a fixed string whenever the context carries a request — so `safeLogInput ?: input` is only reachable when `isPaymentRequest == false`. That closes what I raised.

**Fund safety traced clean.** Amount and counterparty are pinned at open time in `ContactPaymentContext`; at pay time `onConfirmPay` single-flights on `isSubmittingPaymentRequest`, `validateIncomingPaymentRequest` re-checks `acceptsPaymentAmount` by equality plus the bolt11 msat match plus `isPending`, and `accept() → updateRequest` adds to `processingRequestIds` under `operationMutex` and removes from `_pendingRequests` before the send. A request can't be paid twice, after expiry, or at an amount other than the one shown — structurally equivalent to what iOS #721 does with `markPresentedIfPending` + `processingRequestIds`. A 15th-failure request stays in the sheet but any re-tap goes through the full Send confirm and pay-time validation again.

Also clean: both new toasts are fixed string resources, so no counterparty or error text reaches the UI; nothing seed-derived is touched; `ParseFailure` and the failure-reason enum aren't persisted, so there's no migration concern; the new `runCatching` uses are all non-suspend (`Instant.parse`, `Bolt11Invoice.fromStr`) with the suspend paths on `runSuspendCatching`; and `synchronizePaykitContacts` clears requested state before `repo.clear()`, so the new expiry path can't toast for a previous identity's request.

One pre-existing thing I'm noting rather than filing: `handleScan` runs on `bgDispatcher`, so `clearIncomingPaymentRequestTarget → deferPaymentRequestPresentation` mutates the plain `mutableMapOf` retry maps and the generation counter off-main while the collectors run on main. Base already did this; this PR adds two more fields to the same unsynchronised set without changing the shape. Worth a separate look someday, not here.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Expiration now restores Payment Requests only when no sheet is open. The clock-driven backoff regression opens an unrelated Send Confirm sheet, expires the request, verifies the Send sheet remains, and still receives the expired toast.

Fixed in 759ca61

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Unhandled decoded targets from incoming requests now log a fixed identifier instead of the Scanner payload. The regression decodes a lightning address and verifies warning logs contain the fixed message without the target.

Fixed in 759ca61

docs/payment-requests.md
jvsena42: Worth a follow-up here rather than a new thread, since it is the same ask only half-applied. The decode-*failure* branch is redacted correctly now, but `AppViewModel.kt:2618` `.onSuccess { Logger.info("Handling decoded scan data: $it") }` is not split on `isPaymentRequest` the way `.onFailure` at `:2610-2617` is. So a counterparty target that decodes fine but is an unhandled variant gets logged in full before `handleDecodedScan` reaches the redacted WARN at `:2658` — e.g. `alice@example.com` becomes `LnurlAddress(data=LnurlAddressData(uri=alice@example.com, domain=example.com, username=alice))`, a uniffi data class — and it repeats across the 15 deferral retries, reaching `files/logs/` in every variant since `LoggerImpl.info` saves unconditionally.

The new test passes only because it filters `it.type == Log.WARN` (`AppViewModelSendFlowTest.kt:742-743`); the sibling invalid-target test at `:711-713` checks all levels. I ran the test with that filter dropped and it fails on head.

The emission itself is pre-existing on master for all scans (`origin/master:2538`), so this is not a regression — it just means the claim in `docs/payment-requests.md:10-11` does not hold for the success path. Splitting it the same way as the failure branch, plus dropping the WARN filter in the test, closes it.

jvsena42 (commented):
Reviewed the delta at `759ca617a`. **No HIGH/MEDIUM.** One LOW as a follow-up on ben's `docs/payment-requests.md:11` thread rather than a new one, since it's the same ask half-applied.

**Both of my earlier findings are addressed, and I checked the guard still restores when it should** — that was the risk with the fix I asked for, since a guard that's too broad silently stops doing its job. It doesn't: immediate `RequestExpired`, expiry during backoff, and expiry during in-flight resolution all still restore the sheet, each asserted. And the new `explicit request expiring during backoff keeps an unrelated send sheet open` pins the case I raised — `Sheet.Send(Confirm)` survives with the expired toast last of exactly two enqueues.

The unhandled-target WARN branch selection is sound too: `clearIncomingPaymentRequestTarget()` captures `activeIncomingPaymentRequest() != null` before clearing the context and returns it, so the boolean at `:2658` reflects the state before teardown rather than after. The remaining leak is the INFO line, on the thread.

One candidate I chased and dropped: I thought the restore guard might miss the window where `showSheet` has started a transition but `_currentSheet` is still null. It doesn't — `getBolt11()` and `getOnchainAddress()` are plain `_walletState.value` reads with no suspension, and `viewModelScope` is `Main.immediate`, so a user tap sets `_currentSheet` synchronously before `showSheet` returns and the guard correctly rejects. The only residue is the 300 ms `SCREEN_TRANSITION_DELAY` that runs when *replacing* an already-open sheet, which needs three conditions to line up and ends in a replaced sheet with no fund impact. Not worth changing.

Nothing else in the scoped delta — no new `runCatching`, no new flows, no off-main map mutation, no key material logged. The three `strings.xml` hunks in the range are already on master, so they're rebase noise rather than part of this PR.

docs/payment-requests.md
ovitrif: The success path now uses a fixed incoming-request log before decoded scans are handled, while normal scans retain their existing diagnostics. The unhandled-target regression now checks every log level and verifies the peer target is absent.

Fixed in 0838607

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -939,18 +975,52 @@ class AppViewModel @Inject constructor(
...
             paymentRequestPresentationRetryJobs.remove(it)?.cancel()
         }
-        if (requestedPaymentRequestId?.let { it !in requestIds } == true) {
+        val requestedRequest = requestedPaymentRequest
+        if (requestedRequest != null && requestedRequest.id !in requestIds) {
+            if (paykitPaymentRequestRepo.isExpired(requestedRequest)) {
+                finishExpiredPaymentRequestPresentation(requestedRequest)
+                return
+            }
             invalidatePaymentRequestPresentation()
```
jvsena42, line 1170: **A requested request that leaves `pendingRequests` for a non-expiry reason during an explicit retry is cleared silently** — no terminal toast, no sheet restore. Only the expiry exit (:1009-1011) got terminal feedback; this third path didn't.

Trace: user taps Pay from the PaymentRequests sheet, so `openIncomingPaymentRequest` sets `shouldRestorePaymentRequestSheet = true` and hides the sheet. First attempt returns `WaitingForUpdatedPaymentList`, which toasts *"Bitkit will retry automatically"* and schedules a 2s retry. Inside the 28s window a poll refresh runs; the payee has meanwhile cancelled, so the record comes back `state = CANCELED` -> `NonActionableState` -> absent from `incoming` -> the collector reaches :1008 true, :1009 `isExpired` false (the request still had a future `expiresAt`), and falls to these two lines. That clears everything including `shouldRestorePaymentRequestSheet`.

Outcome: the user was told Bitkit would retry, the sheet is gone, and nothing ever follows — the retry job was already cancelled at :1004-1006, so the 15th-attempt terminal path can't fire either.

Pre-existing shape rather than a regression (base cleared silently too), but it's a gap in the feature this PR is delivering. iOS #721 covers exactly this case: `requestedPresentationUnavailableTrigger` -> `.requestedPresentationEnded` -> `PaymentRequestUnavailableToast`, with a test asserting the trigger fires once.

Narrowest fix: in this `else`, when `requestedRequest.id == requestedPaymentRequestId`, snapshot `shouldRestorePaymentRequestSheet`, log `logPresentationRejection(counterparty, ResolutionFailed)`, enqueue the existing unavailable toast, and restore the sheet if none is open — i.e. the same shape as `finishExpiredPaymentRequestPresentation` with the unavailable copy.

jvsena42 (commented):
Reviewed the delta since my last pass (`08386077a` plus two clean master merges). The redaction commit closes my `docs/payment-requests.md` thread as asked — `isPaymentRequest` is captured at :2529 before `handleDecodedScan` can tear the context down, so the branch reflects pre-teardown state, and the normal-scan path keeps its existing log line. Nothing over-corrected.

One LOW observation inline. Everything here is behind `PaykitFeatureFlags.isUiEnabled` (`!BuildConfig.FEATURE_PAYKIT_UI_DISABLED && localFlag`), and `isAvailable()` short-circuits `refresh()` to `clearStateLocked()` when it's off, so no incoming request is parsed, presented or logged in a release build. Dev/QA-facing today. Not blocking.

**Checked and clean:**

- **Rejection durability.** The SDK write in `updateRequest` lands before `_pendingRequests` is updated, under `operationMutex`, so durable state precedes the UI. On the next refresh a `REJECTED` record maps to `NonActionableState` and never re-enters pending. The `PaykitPaymentRequestId(paymentRequestId, counterparty, counterpartyReceiverPath)` triple means a counterparty can't re-mint the same record under a new id without it being a genuinely new proposal that still needs Pay + Confirm. No auto-pay path exists in the presentation machinery.
- **A 15th-failure request** is `markPresented`, which removes it only from *automatic* presentation; it stays payable by design, and `presentedRequestIds` is persisted per identity so a restart doesn't re-auto-present.
- **Amount TOCTOU.** The object handed to `openContactPayment` is the one the loop started with and is pinned in `ContactPaymentContext`. A refresh that swapped in a same-id/different-amount record can't change what's paid.
- **Diagnostics logging.** All three `PaykitPaymentRequestDiagnostics` methods emit only enum `logValue`s, `error::class.simpleName`, and a redacted counterparty — no `Throwable.message`, no note/amount/endpoint/request id. `PaykitPaymentRequestDiagnosticsTest` pins the raw pubky, an arbitrary counterparty and the Throwable message all absent.
- **Trust boundary.** The only counterparty-supplied text rendered is `request.note`, single-line and ellipsised, in a separate column from `MoneyCell` — a note can't fake the amount cell. Both new toasts are fixed string resources.
- **No leak back to the counterparty.** Failure reasons are consumed only by `Logger`; `reject()`/`accept()` carry no reason.
- **Journey + docs match the code.** 15 attempts x 2s = 28s inside the journey's 35s wait; the silent reasons (`outgoing_request`, `non_actionable_state`, `expired`) match `shouldLogIncomingRejection = false`; accessibility ids and toast copy match `strings.xml`. The journey correctly starts on the full-screen route, where `shouldRestorePaymentRequestSheet` is false.

**Cross-repo note** (https://github.com/synonymdev/bitkit-ios/pull/721): iOS dedupes its parse-rejection warn per (record, reason, counterparty) via `IncomingPaykitPaymentRequestRejectionLog`, and gates presentation-failure warns to terminal-only for requested presentations plus first-per-reason for automatic ones. Android's `logParseRejection` and `logPresentationRejection` both fire on every sync/retry. That's the per-attempt log noise ben-kaufman had removed on the iOS side — worth matching, but it's noise, not a defect, so I'm not filing it.

Also worth recording: iOS's `PaykitResolutionFailureDiagnostics` classifies resolution errors into stable `storage/<code>`-style reasons behind a 64-byte `[a-z0-9_-]` allowlist, where Android logs `error::class.simpleName` only. Android's is strictly less informative but equally safe — no attacker-influenced bytes can reach the log at all — so no change needed.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: Requested payment requests that leave the pending list now emit the redacted `resolution_failed` diagnostic, show the unavailable toast once, and restore Payment Requests only when no other sheet is open. Added a regression covering cancellation during retry backoff.

Fixed in 17c4e57

jvsena42: Confirmed fixed in `17c4e57cc`. `finishUnavailablePaymentRequestPresentation` mirrors the expiry path as suggested — logs `ResolutionFailed`, snapshots `shouldRestorePaymentRequestSheet` and the toast condition *before* clearing, then restores the sheet only when none is open. Retry cleanup runs unconditionally, so the cancelled job can't strand state. Test covers it. That closes the third exit path.

piotr-iohk:
Pls resolve conflicts

ovitrif:
@piotr-iohk Merged current `master` and resolved the payment-request/subscription conflicts while retaining both behaviors.

Fixed in 36c4404

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -969,77 +1056,125 @@
...
+            ?: return PaykitPaymentRequestParseResult.Rejected(PaykitPaymentRequest.ParseFailure.InvalidExpiration)
     }
-    val isExpiredProposal = state == PaymentRequestLifecycleState.PROPOSED && expiresAt?.let { it <= now } == true
-    if (requiresActionableRequest && isExpiredProposal) {
-        return null
+    if (
+        requiresActionableRequest &&
+        state == PaymentRequestLifecycleState.PROPOSED &&
+        expiresAt != null &&
+        expiresAt <= now
```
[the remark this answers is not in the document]
ovitrif: I extracted the proposal-expiry calculation into `isExpiredProposal`, preserving the behavior while reducing the condition complexity.

Fixed in 486fbf7

jvsena42 (commented):
All three of my earlier findings are **still fixed** after the master merge — I checked rather than assumed, since `36c440497` brought in #1186 and touched these exact blocks:

- The generation bump is still inside the guard (`AppViewModel.kt:1116-1120`); the merge only renamed `clearRequestedPaymentRequestPresentation()` → `clearRequestedPaymentRequest()`, and `finishUnavailablePaymentRequestPresentation:1132` mirrors it. Automatic-batch behaviour intact: expired A → no bump → B still passes `isCurrentPaymentRequestPresentation` and presents.
- `Expired("expired", shouldLogIncomingRejection = false)` survives at `:97`.
- `if (clearIncomingPaymentRequestTarget()) return` still precedes the generic toast at `:3317`.

**`fix: simplify expiry condition` is behaviour-identical**, which I traced end to end: `synchronizeLocked` samples `clock.now()` once (`:643`) for both paths, and `expiresAt != null && expiresAt <= now` ≡ `expiresAt?.let { it <= now } == true` — same operator, same clock. It actually restores verbatim what master already had; the merge had hand-expanded it, which is what tripped detekt. `PaykitPaymentRequest.isExpired:111-112` uses the same predicate, and `scheduleExpirationLocked` delays by `(nextExpiration - clock.now()).coerceAtLeast(ZERO)` so the discard fires at or after `expiresAt`, never before. No boundary shift, and iOS's `parsedExpiration > now` keep-condition is the same boundary. Injected `Clock` throughout, no `System.currentTimeMillis()` in the Paykit files.

Two LOW notes inline. Both port to the iOS twin (https://github.com/synonymdev/bitkit-ios/pull/721).

app/src/test/java/to/bitkit/repositories/PaykitPaymentRequestRepoSubscriptionTest.kt
```diff
@@ -143,6 +146,15 @@ class PaykitPaymentRequestRepoSubscriptionTest : BaseUnitTest(StandardTestDispat
         assertEquals(Instant.parse("2027-02-01T08:00:00Z"), request.billingPeriod?.endsAt)
     }
 
+    @Test
+    fun `refresh does not report subscription proposals as one time parse failures`() = test {
+        whenever(paykitSdkService.paymentRequests()).thenReturn(listOf(paymentRequestRecord()))
+
+        sut.refresh().getOrThrow()
```
jvsena42, line 153: Low: this asserts only an absence, so it passes with the filtering it's named for removed.

`verify(diagnostics, never()).logParseRejection(any(), any())` pins `RecurringRequest.shouldLogIncomingRejection == false`, but not that the record stays *out of* the one-time queue — which is the half with a user-visible failure mode.

Delete the `if (requestTerms.recurrence != null) return Rejected(RecurringRequest)` guard (repo `:1096-1098`) and the default `paymentRequestRecord()` — PAYER, PROPOSED, `btc`, `0.001`, `[bolt11]`, `expiresAt = null` — clears every remaining guard and parses as a normal one-time request. So `oneTimeIncoming` contains it, nothing is logged, and **this test still passes** while a subscription proposal shows up simultaneously in `subscriptions` *and* as a payable one-time row at the first period's amount with no billing period attached. Paying it is a real send.

No other test closes it: every `pendingRequests.isEmpty()` assertion in the file (`:239`, `:243`, `:313`, `:362`) uses an `ACTIVE_RECURRING` record, and the PROPOSED-record tests (`:176`, `:192`, `:205`) only assert `subscriptions.value.single()`.

Two lines, no new fixture:

```kotlin
assertTrue(sut.pendingRequests.value.isEmpty())
assertEquals(1, sut.subscriptions.value.size)
```

The second also removes the vacuous-pass risk — a `never()` assertion holds equally if `refresh()` short-circuits and never parses the record at all.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -78,6 +79,24 @@ data class PaykitPaymentRequest(
...
+    enum class ParseFailure(
+        val logValue: String,
+        val shouldLogIncomingRejection: Boolean = true,
+    ) {
+        MissingLocalRole("missing_local_role"),
+        OutgoingRequest("outgoing_request", shouldLogIncomingRejection = false),
+        UnsupportedLocalRole("unsupported_local_role"),
+        NonActionableState("non_actionable_state", shouldLogIncomingRejection = false),
+        MissingTerms("missing_terms"),
+        RecurringRequest("recurring_request", shouldLogIncomingRejection = false),
```
jvsena42, line 93: Low: silencing `recurring_request` wholesale also silences malformed recurring records that no surface displays.

The premise is that the subscription surface picks these up, but it doesn't always. `RecurringRequest` is returned for *any* record with `terms.recurrence != null` (`:1096`), while `toPaykitSubscription` (`PaykitSubscription.kt:243-276`) returns null for an unknown `recurrence.unit`, `every == 0` or `> Int.MAX_VALUE`, an unparseable `startsAt`/`anchor`/`endsAt`/`proposalExpiresAt`, `endsAt <= startsAt`, a non-`btc` asset, or an unrepresentable amount. Such a record is dropped from `_pendingRequests`, from `_subscriptions`, and from history — with zero log lines. Before the merge it produced one `recurring_request` warning.

So a counterparty (or an SDK version skew) sending a proposal with a recurrence unit Bitkit doesn't know leaves the payer seeing nothing and QA with nothing to triage "my proposal never arrived" from — which is what this PR's diagnostics exist for.

This is the same shape as the `UnsupportedLocalRole` split ben already landed at `:84`, and the fix is that same split rather than a reversal:

```kotlin
if (requestTerms.recurrence != null) {
    return PaykitPaymentRequestParseResult.Rejected(
        if (toPaykitSubscription() != null) ParseFailure.RecurringRequest else ParseFailure.UnsupportedRecurrence,
    )
}
```

or more cheaply, keep one reason and pass `shouldLog = toPaykitSubscription() == null` at the `refresh()` call site (`:698-704`). `docs/payment-requests.md:31-33` currently claims the stronger "expected filtering of outgoing, completed, subscription, or elapsed records", which would want the same qualification either way.

app/src/test/java/to/bitkit/repositories/PaykitPaymentRequestRepoSubscriptionTest.kt
ovitrif: I added assertions that the proposed subscription remains absent from the one-time queue and present exactly once in subscriptions, alongside the diagnostics assertion.

Fixed in 9c1fa8c

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ovitrif: I split invalid recurring records into `unsupported_recurrence`, which logs by default, while valid subscriptions keep the silent `recurring_request` filter. I added a malformed-unit regression and qualified the diagnostics documentation.

Fixed in 9c1fa8c

app/src/main/res/values/strings.xml
```diff
@@ -1280,6 +1281,8 @@
     <string name="wallet__payment_request_status_rejected">Rejected</string>
     <string name="wallet__payment_request_status_unavailable">Unavailable</string>
     <string name="wallet__payment_request_time">Time</string>
+    <string name="wallet__payment_request_timestamp">%1$s at %2$s</string>
```
jvsena42: Low: the master merge resurrected a string master deliberately deleted.

`wallet__payment_request_timestamp` was removed as unused in `ae762113` ("fix: clean up cancelled hardware payments"), which is an ancestor of `origin/master`. It has zero callers at this head — `git grep payment_request_timestamp -- app/src` matches only this line. `git log -S` over the PR's own commits finds no add, so it came back through the conflict resolution in the master merge `36c4404`.

Not a security issue: a format string with no call site can't be reached with remote values. But it undoes a master cleanup and will trip `UnusedResources` lint.

Delete the line.

jvsena42 (commented):
No HIGH, no MEDIUM. One LOW inline — a merge artefact, not a security issue.

This pass focused on the delta since my last one (`486fbf7` expiry simplification, `9c1fa8c` `unsupported_recurrence` split) and on whether the new logic behaves over the state the master merge brought in.

**What I traced clean:**

- **`toPaykitSubscription()` now runs inside `parsePaykitPaymentRequest` on every recurring record** (`PaykitPaymentRequestRepo.kt:1099`). It cannot throw and take down `refresh()`: every validation exits `return null`, `every.toInt()` is guarded by the `> Int.MAX_VALUE.toUInt()` check, and the only throwing call (`metadata.exportText()`) sits inside `runCatching`. A hostile recurring record cannot DoS the pending list.
- **The recurring split is behaviour-preserving for the accept set.** Both `RecurringRequest` and `UnsupportedRecurrence` return `Rejected`; a malformed recurrence still never reaches `oneTimeIncoming`. The history path maps to `UnsupportedRecurrence` → `null`, identical to base's `recurrence != null → null`. Only logging changed.
- **Fund safety re-traced after the merge.** Amount and counterparty are pinned at open time (`beginContactPayment` passes `PaymentAmountContext(it.amountValue, "btc")` from the pinned request, `:555`); pay time checks equality against `_sendUiState.amount` (`:3994`), bolt11 msat match (`:4003-4006`), `isPending` (`:4019`), single-flight (`:4770`), and `updateRequest` moves the id into `processingRequestIds` and out of `_pendingRequests` before the send (`:844-861`). A failed SDK reject leaves the request pending and visible; nothing auto-pays.
- **Redaction is closed on every path.** `scanLogId` returns a fixed string whenever the context carries a request (`:2183`, `:2227`); decode-failure, decode-success and unhandled-variant branches all use fixed messages; `logPresentationFailure` emits only `error::class.simpleName`. No peer-supplied bytes reach the log, both toasts are fixed resources, and no failure reason is sent back to the counterparty.
- **StateFlow conflation on `pendingRequests` is safe** — `retainPaymentRequestPresentationState` is state-based (`id !in requestIds`), not event-based, so a dropped intermediate emission cannot produce a false unavailable toast.
- **Docs and journey match head.** 1 + 14 × 2 s (`:5408`) = 15 attempts / 28 s, inside the journey's 35 s; automatic presentation falls back to `REFRESH_INTERVALS.last()` = 120 s; silent reasons match `shouldLogIncomingRejection = false`; toast title and `PaymentRequestsScreen` tag both resolve. The journey starts on the full-screen route where `shouldRestorePaymentRequestSheet` is false, so "remains visible" is the correct assertion.
- `runCatching` in the diff is only on non-suspend calls (`Instant.parse`, `Bolt11Invoice.fromStr`); suspend paths use `runSuspendCatching`.

**Parity with https://github.com/synonymdev/bitkit-ios/pull/721:** iOS has **no `unsupported_recurrence`** — its `ParseFailure` still has 12 cases (`PaykitPaymentRequestService.swift:5-22`), and its `shouldLogIncomingRejection` silences only `outgoingRequest`/`nonActionableState`, so it logs `recurring_request` for every recurring record and leans on per-(record, reason, counterparty) dedupe to stay quiet. Net: iOS has no triage blind spot for malformed recurrences, but the two wire taxonomies now diverge (13 vs 12 reasons) and Android is silent where iOS logs once. Worth a decision on which side moves — not a defect on either.

app/src/main/res/values/strings.xml
ovitrif: I removed the resurrected string after confirming it had no callers.

Fixed in 20f8c55

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1061,13 +1106,66 @@ class AppViewModel @Inject constructor(
...
+    private fun finishExpiredPaymentRequestPresentation(request: PaykitPaymentRequest) {
+        paykitPaymentRequestDiagnostics.logPresentationRejection(
+            request.counterparty,
+            IncomingPaykitPaymentRequestFailureReason.RequestExpired,
+        )
+        val restorePaymentRequestSheet =
+            requestedPaymentRequestId == request.id && shouldRestorePaymentRequestSheet
+        val showExpiredToast = requestedPaymentRequestId == request.id
+        if (requestedPaymentRequestId == request.id) {
+            paymentRequestPresentationGeneration++
```
ben-kaufman: This expiration branch now skips `invalidatePaymentRequestPresentation()`, so an incoming-request scan already waiting in `decode()` or `waitForUsableChannels()` is not canceled. It can resume after the expired toast and reopen Send confirmation over the restored request sheet. Could we keep the scan cancellation when handling explicit expiration and add a regression that expires the request while its target scan is suspended?

jvsena42 (commented):
Fix confirmed at `20f8c55e`. `wallet__payment_request_timestamp` is gone from `strings.xml:1284`, `git grep payment_request_timestamp -- app/src` returns zero hits at the new head, and alphabetical ordering is preserved around the deletion (`..._time` → `..._unavailable`). Nothing else changed in the commit.

That was the only finding I had. Clean from my side.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: I now invalidate explicit expiration before clearing the request, which cancels any request-owned target scan before restoring the sheet. I added a regression source case that expires while decode is suspended and verifies Send does not reopen.

Fixed in 9afdc19

jvsena42 (commented):
`fix: cancel expired payment request scan` reviewed — this is new work rather than a response to my findings (the dead string was the only one and it's already gone). The change is right in intent; one LOW inline about collateral damage, gated.

The swap at `:1118` from a bare `paymentRequestPresentationGeneration++` to `invalidatePaymentRequestPresentation()` means an expired requested payment request now also tears down a scan that's still queued for it, instead of leaving one to fire later against a request the user was just told had expired. That's the correct fix for the case the new test pins.

I checked the three remaining bare `paymentRequestPresentationGeneration++` sites to see whether the fix was applied incompletely, and I don't think it was:
- `:929` — the Send sheet is visible and `markPresented` just succeeded. This is the success path; there shouldn't be a queued scan to cancel, and cancelling one would be wrong.
- `:1076` — retries exhausted, "unavailable" toast. Arguable, but the request stays pending and the automatic presentation path can still pick it up, so a hard scan teardown would be more aggressive than the state warrants.
- `:2953` — interrupted request with `retryIncomingRequest == false`, which immediately `markPresented`s. Same reasoning.

So the single call-site change looks deliberate rather than partial. The helper itself and its other eight references pre-date this commit; only `:1118` changed.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1061,13 +1106,66 @@ class AppViewModel @Inject constructor(
...
+    private fun finishExpiredPaymentRequestPresentation(request: PaykitPaymentRequest) {
+        paykitPaymentRequestDiagnostics.logPresentationRejection(
+            request.counterparty,
+            IncomingPaykitPaymentRequestFailureReason.RequestExpired,
+        )
+        val restorePaymentRequestSheet =
+            requestedPaymentRequestId == request.id && shouldRestorePaymentRequestSheet
+        val showExpiredToast = requestedPaymentRequestId == request.id
+        if (requestedPaymentRequestId == request.id) {
+            invalidatePaymentRequestPresentation()
```
jvsena42: Low, gated: this now cancels a queued scan for a *different* request than the one that expired.

`invalidatePaymentRequestPresentation()` (`:1204-1214`) cancels the scheduled scan and clears the deferred scan whenever either carries **any** incoming payment request — it never compares against the request being invalidated:

```kotlin
1206        scheduledScan
1207            ?.takeIf { it.contactPaymentContext?.incomingPaymentRequest != null }
1208            ?.job
1209            ?.cancel()
1210        synchronized(deferredScanLock) {
1211            if (deferredScan?.contactPaymentContext?.incomingPaymentRequest != null) {
1212                deferredScan = null
```

Both are single slots (`:289`, `:292`), so there's no per-request bookkeeping to fall back on.

Failure scenario: request A is the `requestedPaymentRequestId` (user tapped Pay on A). A scan carrying request B is deferred because the app wasn't authenticated yet — B sits in `deferredScan`. A then expires, `finishExpiredPaymentRequestPresentation(A)` passes the `:1117` guard, and `:1118` nils B's deferred scan. B's replay-after-unlock never happens.

Not severe: B stays pending, so `presentNextIncomingPaykitPaymentRequest` should surface it on a later refresh. The user loses the deferred-scan flow for B, not the request itself — which is why this is LOW rather than higher.

But it is introduced here. Before this commit `:1118` was a bare generation bump and didn't touch either scan slot; the other eight call sites of the helper are places where a broad teardown is plausibly intended, whereas this one is explicitly per-request (it's inside `if (requestedPaymentRequestId == request.id)`).

Minimal fix: scope the cancel to the request being invalidated, e.g. give the helper an optional `requestId` and compare `contactPaymentContext?.incomingPaymentRequest?.id == requestId` before cancelling, passing `request.id` from here. Existing callers that want the broad behaviour keep it by passing nothing.

Worth noting the new test wouldn't catch this — `explicit request expiration cancels a suspended target scan` uses the same request for both, so the cross-request case isn't covered.

ben-kaufman: Canceling the scan here leaves `activeContactPaymentContext` set. If Pay was opened from the full-screen request page and the request expires while decoding or waiting for channels, no sheet is restored to clear that context. The next request's Pay action then stops at `hasActiveContactPaymentContext()`, and further taps report an operation already in progress. Could we clear the expired request's own context when canceling its scan and cover opening another request afterward from the full-screen page?

ovitrif: I scoped invalidation by request ID. Expiring A still bumps the presentation generation and cancels A's scan, while an unrelated scheduled or deferred B scan remains replayable. The cross-request regression keeps B queued through A's expiration and verifies B opens afterward.

Fixed in ca1144f

ovitrif: I now clear active and prepared contact context only when it belongs to expired A. The cross-request regression starts A from the full-screen path, expires it during suspended decoding, and verifies queued B opens with its own context afterward.

Fixed in ca1144f

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -1097,21 +1195,41 @@ class AppViewModel @Inject constructor(
...
         synchronized(deferredScanLock) {
-            if (deferredScan?.contactPaymentContext?.incomingPaymentRequest != null) {
+            if (targetsRequest(deferredScan?.contactPaymentContext)) {
                 deferredScan = null
             }
         }
+        if (requestId != null) {
+            synchronized(contactPaymentContextLock) {
+                if (activeContactPaymentContext?.incomingPaymentRequest?.id == requestId) {
+                    activeContactPaymentContext = null
```
ben-kaufman, line 1228: Clearing this context makes the pending-request collector return before its `hideSheet()` call. If an explicit request expires after `Sheet.Send` is set but before `onSheetVisible()` finishes marking it presented, the expired toast appears while its confirmation stays open without a request context. Could we close the expired request's own Send sheet as part of this cleanup and cover expiration between opening Send and marking it presented?

jvsena42:
@ovitrif conflicts again — this one re-conflicted after your 09-09 resolution, so the earlier "resolved" note no longer holds.

`mergeable=CONFLICTING` at head `ca1144f4`. Master moved since (#1216 and #1236 merged), which is the likely cause.

Flagging because I reviewed `20f8c55e` earlier today and there's an open LOW on `AppViewModel.kt:1118` — worth folding into the same push rather than a separate one.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
ovitrif: I now snapshot whether Send belongs to the expiring request, clear only that request's active and prepared context plus requested state, then close its Send sheet. Unrelated contexts and sheets remain untouched. I added regression source coverage for expiration after Send opens but before onSheetVisible marks it presented.

Fixed in 97a5c34

ben-kaufman (approved):

jvsena42 (commented):
Rebased and mergeable again at `97a5c34c` — and the LOW I raised is fixed, in exactly the narrow shape rather than a broad one.

`invalidatePaymentRequestPresentation` now takes a `requestId` and scopes the teardown through a local predicate (`:1205-1223`):

```kotlin
1209        fun targetsRequest(context: ContactPaymentContext?): Boolean {
1210            val scanRequestId = context?.incomingPaymentRequest?.id ?: return false
1211            return requestId == null || scanRequestId == requestId
1212        }
```

The `requestId == null` branch preserves the old broad behaviour, so the seven pre-existing call sites are unchanged — that was the part I cared about, since a blanket narrowing there could have been its own regression. And the expiry site now passes the id (`:1118`), so request A expiring can no longer nil a deferred scan belonging to request B. The scenario I described is closed.

You also added behaviour I didn't ask for and it looks right: the helper returns whether the expired request owned the active context with the Send sheet open, and `:1120` hides the sheet in that case. `ownsActiveContext` compares against the same `requestId`, so it can't dismiss a sheet belonging to a different request — worth stating explicitly since that was the obvious way to get it wrong.

Rebase looks clean: the PR-authored delta against master is intact (14 files), and nothing from the earlier review was dropped.

PR #1220, opened 2026-09-02
test: add marketplace wallet journey

### LINKED ISSUE #1212: test: prove the android pubky marketplace wallet leg
Counterpart: synonymdev/bitkit-ios#718

## Scope

- Exercise the existing `pubkyauth` deep link, watch-only consent and xpub claim, linked-contact prerequisite, incoming Payment Request, payment approval, and regtest confirmation.
- Record the first shipped release or tag containing the claim and incoming-request surfaces.
- Add the matching journey with identifiers aligned to iOS.
- Run the journey against the current implementation first so it establishes the baseline and exposes remaining gaps.

## Acceptance criteria

- The full flow passes end to end on Android.
- Any discovered code gap is tracked as a blocking subissue.

ovitrif:
Closes #1212

This PR:

1. Adds the Android wallet journey for a Pubky marketplace purchase, aligned with the iOS journey in synonymdev/bitkit-ios#720.
2. Adds stable identifiers for watch-only authorization, contact payments, incoming requests, and payment approval.
3. Records the live 2026-09-02 exact-head acceptance baseline and the first-release provenance for the existing claim and incoming-request surfaces.

Acceptance evidence covers the full controlled journey: seller watch-only claim, reciprocal linked-contact setup, fresh buyer funding, canonical Paykit delivery, payment approval, zero-confirmation detection, broadcast, one-block confirmation, paid request history, and Locks completion. This branch merges Android master `e2fa21f86`; the approved original recording and acceptance evidence were captured at `0a0f7c8b`. A later replay at pre-rebase head `cbc88a40` reached two fresh Android identities, a funded buyer, and seller authorization, where Paykit Android rc51 rejected the pinned fixture's legacy `pubkyauth://signin` request because rc51 requires grant auth. Pre-merge head `04d0720f8` corrects the reviewed request-bell sequence. Merge head `b122caa6f` preserves the request-row navigation from merged #1186 while retaining the reviewed hyphenated row and Pay-action identifiers. Current head `031314514` incorporates master `e2fa21f86`, including the #1242 screens-map update, without further conflicts. The current-head manual replay remains unchecked.

### Description

- Documents the controlled two-wallet integration fixture contract, emulator port mappings, evidence requirements, and exact Android buyer actions.
- Aligns request-row identifiers to `PaymentRequestRow-<id>` and adds `PaymentRequestPay-<id>`.
- Adds `PubkyAuthAuthorize`, `PubkyAuthOK`, and `ContactPaymentsToggle` identifiers.
- Pins the marketplace driver and Locks harness at `ed03a32e`, cites merged upstream Paykit Server #2 for canonical request behavior, and cites merged Android #1178 for incoming-request swipe behavior.
- Records successful live acceptance identifiers, both device identifiers, zero-confirmation and one-confirmation boundaries, and the intended first release milestone 2.6.0.

### Preview

https://github.com/user-attachments/assets/1a4f009f-1579-4760-a327-2a97cb73ec0b

#### Manual Tests

- [x] **1.** Fresh funded Android buyer → receive request `5db852f6-d47a-43c4-8b17-fb7a654ac4dd` → review the accepted fallback seller identity `pubky…apfiy`, 15,000 sats, and 141-sat fee → swipe once → verify transaction `a3a427ca69a3f1d3f26fb47f38182c8543fdbd997227ddb8a2dec1ef0e927166` at zero confirmations → mine one regtest block → verify confirmed Android activity, paid request history without a Pay action, confirmed Paykit delivery, and completed Locks purchase.
- [ ] **2.** Re-run the full seller authorization and buyer purchase/payment journey for the current PR head. The latest replay at pre-rebase head `cbc88a40` created seller `pubky86…gfhdz4y` and buyer `pubkyr7…tu75yby`, and the buyer received 1,000,000 sats in transaction `21709c35a60fe081b1de0faeaae1ca403ec67cee76587262ccc6e4b37be6e840`, confirmed at fixture height 16,424. At seller authorization, Android resolved and delivered the pinned server's `pubkyauth://signin` request, then Paykit Android `0.1.0-rc51` rejected it with `code=protocol_error, context=only Pubky grant auth URLs are supported`. The pinned server at `867fc883` uses Paykit source `6b241878` / Pubky `0.8.0` and has no `signin_grant` setup path, while rc51 embeds Pubky `0.11.0` grant auth. The request, payment, and confirmation boundaries require a compatible marketplace fixture.

The checked replay is separate from the initial and selector-specific acceptance runs in `README.md`; its request and transaction identifiers are intentionally distinct. The Preview remains historical evidence until manual test 2 replaces it with a current-head recording.

#### Automated Checks

- `PaymentRequestsScreenTest.kt`: all five focused Compose tests passed at the original reviewed head, including the incoming row and Pay identifiers.
- `just compile`, `just test`, and `just lint`: passed for pre-merge head `04d0720f87d2a11b0cbbb5f12426016baa6c80b2`; detekt reported only existing baseline findings outside this PR.
- `git diff --check` and `xmllint --noout journeys/pubky-marketplace/wallet-leg.xml`: passed at pre-merge head `04d0720f87d2a11b0cbbb5f12426016baa6c80b2`.
- All nine pre-merge branch commits have valid signatures; signed merge commits `b122caa6f` and `031314514` bring in master through `e2fa21f86` and resolve the request-row conflict.
- The E2E debug build passed at pre-merge head `04d0720f87d2a11b0cbbb5f12426016baa6c80b2` with the isolated local-backend routing; APK SHA-256: `8180be69e01c93c6e41f0f54a3e9f3a042e3565cd77c7ad358e0a39c47667fa0`.
- `just e2e` at the pre-merge head reached `validateSigningDevRelease` and stopped because no release keystore was configured in that worktree. The debug-signed E2E variant above is the device-test artifact.
- No tests, validation builds, manual QA, recordings, device launches, CI triggers, reruns, or watching were performed for merge heads `b122caa6f` and `031314514`, at user direction.
- Preview video SHA-256: `e21ce4b0bea14e42c0bb437cd8d443deebe6e2ba40488c0b24dbaaad54ec05ec`.


piotr-iohk (dismissed):
QA LGTM on the identifiers + journey contract. Did not re-run the isolated marketplace fixture.

Checked latest (`0a0f7c8b`):

- Cited IDs exist on this branch: `PubkyAuthWatchOnlyConsent` / `Approve` / `Authorize` / `OK`, `ContactPaymentsToggle`, `PaymentRequestsSheet`, `PaymentRequestRow-<id>`, `PaymentRequestPay-<id>`, `ReviewAmount`, `ReviewContactRecipient`, `GRAB`, `SendSuccess`, `ActivityAmount`, `ActivityTxDetails`.
- `xmllint --noout journeys/pubky-marketplace/wallet-leg.xml` passes.
- Compose identifier tests in `PaymentRequestsScreenTest` match the hyphenated tags.
- CI green, including `e2e-tests-staging - pubky_paykit`.

Notes, not blockers:

- Merge **#1217 first**. This PR’s `PaymentRequestRow-` / `PaymentRequestPay-` tags are a subset of #1217 (that one also adds `PaymentRequestDismiss-<id>`).
- PR body QA notes cite request `5db852f6-…` / tx `a3a427ca…`. README records different accepted runs (`ad1a8463-…`, then selector replay `b7f67854-…`). Fine if those are separate replays — worth one line so we do not treat it as a mismatch later.
- README says #1209 “preserves rejected incoming requests as visible history.” On #1217, parse rejects stay log-only; the visible-history / retry path is the explicit Pay terminal toast. Wording is slightly off.

I am not blocking on not replaying the Locks/Paykit Server fixture here. The journey is documentation of an accepted 2026-09-02 run, and the app-side selectors it needs are present.

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,40 @@
...
+    <action>Verify the fixture completes setup with the seller account xpub and no spending authority</action>
+    <action>Open General Settings in both wallets and verify contact payments (id "ContactPaymentsToggle") are enabled</action>
+    <action>Open Contacts in the buyer wallet and save the fixture's seller public key</action>
+    <action>Verify the seller contact (id "Contact_&lt;seller-public-key&gt;") is visible</action>
+    <action>Open Contacts in the seller wallet and save the buyer public key</action>
+    <action>Verify the buyer contact (id "Contact_&lt;buyer-public-key&gt;") is visible</action>
+    <action>Verify the fixture reports the seller and buyer as linked peers</action>
+    <action>Have the fixture create one marketplace purchase for the buyer and record its Payment Request id</action>
+    <action>Verify the request uses lowercase "btc", a regtest endpoint identifier, and a JSON endpoint value</action>
+    <action>Wait for the incoming requests sheet (id "PaymentRequestsSheet")</action>
```
ben-kaufman: A fresh incoming request opens payment review automatically, so waiting here never produces `PaymentRequestsSheet`. Could we dismiss the automatic review and tap `PaymentRequestsBell` before checking the row and Pay identifiers?

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,40 @@
...
+    receives, approves, pays, and confirms the resulting marketplace Payment Request on regtest.
+    Precondition: two clean wallets and the integration fixture runtime described in this suite's
+    README.
+    Configure the fixture Electrum endpoint, Android emulator port mappings, and Paykit UI before
+    either wallet's first launch. Start with the seller wallet open and the fixture's fresh setup
+    auth URL available.
+  </description>
+  <actions>
+    <action>Open the fixture setup auth URL in the seller wallet</action>
+    <action>Verify the watch-only consent screen (id "PubkyAuthWatchOnlyConsent") is visible</action>
```
jvsena42: Use `testTag "X"` instead of `id "X"`. The 32 existing journey specs use `testTag "` 147 times and `id "` zero times.

`s/(id "/(testTag "/` on lines 13, 14, 16, 17, 19, 21, 23, 27-33 and 36.

jvsena42 (commented):
Selector-parity pass looks right — I checked every tag the journey uses against `bitkit-ios` and all of them resolve there except one. Two notes below.

Also: the README pins Paykit Server `867fc883`, the adb port maps and the homeserver key, but not the marketplace driver / Locks harness, so nobody else can replay this end to end yet. Is there a ref you can pin, like `hardware-wallet/README.md` does with `bitkit-docker`?

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
```diff
@@ -553,7 +553,9 @@ internal fun PaymentRequestCard(
                         )
                     },
                     size = ButtonSize.Small,
-                    modifier = Modifier.weight(1f),
+                    modifier = Modifier
+                        .weight(1f)
+                        .testTag("PaymentRequestPay-${request.paymentRequestId}")
```
jvsena42, line 604: `PaymentRequestPay-<id>` is the only selector in this PR with no iOS counterpart — the iOS pay button (`PaymentRequestsView.swift:82-90`) has no `accessibilityIdentifier`, unlike the row at :104.

`wallet-leg.xml:29` is written platform-neutrally and the README replays an Android buyer against an iOS seller, so the step won't resolve if the buyer leg runs on iOS. Either add the identifier on iOS or note in the README that this selector is Android-only.

journeys/pubky-marketplace/wallet-leg.xml
ovitrif: Updated the journey to wait for automatic payment review, dismiss it with Android back, then open the requests sheet through `PaymentRequestsBell` before checking the row and Pay selectors.

Fixed in 419503dcb

journeys/pubky-marketplace/wallet-leg.xml
ovitrif: Changed every journey selector reference from `id` to the established `testTag` terminology.

Fixed in 419503dcb

app/src/main/java/to/bitkit/ui/screens/paymentrequests/PaymentRequestsScreen.kt
ovitrif: Documented `PaymentRequestPay-<id>` as Android-only and specified accessible-label lookup for an iOS buyer until that platform adds a counterpart.

Fixed in 419503dcb

ovitrif:
@jvsena42 I pinned the marketplace driver and Locks harness to `BitcoinErrorLog/pubky-marketplace` commit `ed03a32e`. That immutable `payments-env` ref includes `scripts/verify.sh` for the Locks, Paykit, Pubky, bitcoind, and Fulcrum path.

jvsena42 (commented):
Re-reviewed at `419503d`. No HIGH/MEDIUM, nothing to file.

The thing I wanted to rule out on a "test:" PR that touches three production files is a consent gate loosened to make a journey pass. It isn't: the `PubkyAuthApprovalSheet` diff is two `testTag` additions (`PubkyAuthAuthorize` on the Authorize button, `PubkyAuthOK` on the success button) with no `onClick`, state, or effect change. `onAuthorize` is still guarded by `if (uiState.authUrl == authUrl)`, `Content` still renders `ApprovalState.Loading` unless `isCurrentRequest`, and local auth still sits between `requestAuthorize` and `confirmAuthorize` — so no auto-approve, no approving a stale or different request than the one displayed. I also grepped `app/src/main` for the diff looking for anything keyed off `BuildConfig`, a flag, or an injected fake that only exists for tests: nothing, only tag strings changed.

`PaymentRequestsScreen` is the `PaymentRequestRow-${id}` / `PaymentRequestPay-${id}` hyphenation only — `lazyListKey` is untouched, so no reorder or item loss, and no other reference to the old un-hyphenated tag exists at head. `SettingsScreen` is a `ContactPaymentsSwitch` → `ContactPaymentsToggle` rename, still inside `if (state.isPaykitEnabled && state.isPubkyAuthenticated)`; the remaining `ContactPaymentsSwitch` hits are in `SettingsSwitchRowTest`, which passes its own `switchTestTag`, so that's self-consistent rather than a dangling rename.

On the earlier threads — I checked the pushed fixes are correct, not just present. ben-kaufman's point on `wallet-leg.xml:27`: the journey now waits for auto-review, backs out, taps the bell, then asserts the sheet/row/Pay. That holds because `pendingPaymentRequests` reads the raw `_pendingRequests`, not the presented-filtered view, and dismissing the send sheet neither accepts nor rejects (accept only fires on send completion) — so the request really is still pending when the bell is tapped. My own two: all 15 `id "` are now `testTag "` with none left, and the README correctly documents `PaymentRequestPay-<id>` as Android-only with the accessible-label fallback for an iOS buyer.

Every selector in `wallet-leg.xml` resolves against real code at head, so the journey isn't encoding a flow the app doesn't have. All dev/QA-facing today — `isPaykitEnabled` defaults false and is Dev-Settings-only.

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,42 @@
...
+    <action>Open General Settings in both wallets and verify contact payments (testTag "ContactPaymentsToggle") are enabled</action>
+    <action>Open Contacts in the buyer wallet and save the fixture's seller public key</action>
+    <action>Verify the seller contact (testTag "Contact_&lt;seller-public-key&gt;") is visible</action>
+    <action>Open Contacts in the seller wallet and save the buyer public key</action>
+    <action>Verify the buyer contact (testTag "Contact_&lt;buyer-public-key&gt;") is visible</action>
+    <action>Verify the fixture reports the seller and buyer as linked peers</action>
+    <action>Have the fixture create one marketplace purchase for the buyer and record its Payment Request id</action>
+    <action>Verify the request uses lowercase "btc", a regtest endpoint identifier, and a JSON endpoint value</action>
+    <action>Verify the fresh request opens payment review automatically (testTag "ReviewAmount")</action>
+    <action>Press Android back to dismiss the automatic review, then tap the requests bell (testTag "PaymentRequestsBell")</action>
```
ben-kaufman: After the contact setup, the buyer is still on Contacts. Dismissing the automatic review only closes the send sheet, while `PaymentRequestsBell` is on Home, so this step cannot reach the bell as written. Could we explicitly return the buyer to Home before creating the purchase?

ovitrif: Fixed in `6df11ed9`. The journey now returns the buyer to Home and verifies `PaymentRequestsBell` before the fixture creates the purchase, so dismissing the automatic review can reach the bell as written.

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,43 @@
...
+    <action>Tap Authorize (testTag "PubkyAuthAuthorize")</action>
+    <action>Verify authorization succeeds (testTag "PubkyAuthOK")</action>
+    <action>Verify the fixture completes setup with the seller account xpub and no spending authority</action>
+    <action>Open General Settings in both wallets and verify contact payments (testTag "ContactPaymentsToggle") are enabled</action>
+    <action>Open Contacts in the buyer wallet and save the fixture's seller public key</action>
+    <action>Verify the seller contact (testTag "Contact_&lt;seller-public-key&gt;") is visible</action>
+    <action>Open Contacts in the seller wallet and save the buyer public key</action>
+    <action>Verify the buyer contact (testTag "Contact_&lt;buyer-public-key&gt;") is visible</action>
+    <action>Verify the fixture reports the seller and buyer as linked peers</action>
+    <action>Return the buyer wallet to Home and verify the requests bell (testTag "PaymentRequestsBell") is visible</action>
```
ben-kaufman: The requests bell is only shown when there is a pending payment request. With the clean buyer required by this journey, it is still hidden here because the fixture creates the first request in the next step. Could we keep the return to Home here and check the bell after dismissing the automatic review?

ovitrif: I kept the explicit return to Home before purchase creation and moved the `PaymentRequestsBell` visibility check to after the automatic review is dismissed, when the request is pending.

Fixed in 04d0720

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,43 @@
+<journey name="pubky marketplace wallet leg">
+  <description>
+    Verifies a Bitkit seller grants a watch-only account claim and a separate linked Bitkit buyer
+    receives, approves, pays, and confirms the resulting marketplace Payment Request on regtest.
+    Precondition: two clean wallets and the integration fixture runtime described in this suite's
+    README.
+    Configure the fixture Electrum endpoint, Android emulator port mappings, and Paykit UI before
```
jvsena42: **"Configure … Paykit UI before either wallet's first launch" describes an enable path that doesn't exist on Android.**

The datastore default is `false` and nothing sets it from a build, env var or intent — `FEATURE_PAYKIT_UI_DISABLED` can only make the UI *unavailable*, never enable it. The sole setter is the in-app Dev Settings toggle plus the "Enable Paykit UI?" confirmation dialog, which requires an existing wallet and dev mode. So there is no pre-first-launch configuration to perform, and an operator will look for one.

The knock-on is that with Paykit UI off, `CreateProfile` is unreachable behind `PaykitRouteGuard`, so the seller never gets an identity and step 12 ends in the `pubky_auth__no_identity` toast.

The iOS twin gets this right: "enable Paykit UI in Dev Settings (id `PaykitUiToggle`) and confirm the warning". Suggest matching that wording — after wallet creation, enable Paykit UI in Dev Settings (`testTag "PaykitUiToggle"`) and confirm the dialog (`PaykitUiEnabledToast`) in both wallets before opening the auth URL.

journeys/pubky-marketplace/wallet-leg.xml
```diff
@@ -0,0 +1,43 @@
...
+    either wallet's first launch. Start with the seller wallet open and the fixture's fresh setup
```
jvsena42: **The precondition also omits the Pubky identity both wallets need**, which is a separate gap from the one above and stops an operator at step 12 even after Paykit UI is correctly enabled.

`handlePubkyAuth` (`AppViewModel.kt:4655-4663`) returns a `pubky_auth__no_identity` toast when `pubkyRepo.publicKey.value == null`, and `PubkyAuthHandlerRegistrar` only enables the auth alias under `isPaykitUiEnabled && hasIdentity && hasSecretKey`. A clean wallet has neither: `createIdentity` has exactly one caller, `CreateProfileViewModel.kt:170`, so an identity is never created automatically on wallet creation.

The README hints at it obliquely — "map its TCP services into each emulator **before creating a profile**" — but never states profile creation as a precondition step, so it reads as background rather than an instruction.

Fix: one clause here in the `<description>` and in the README, saying each wallet must first create a Bitkit-generated Pubky identity (it has to come *after* the Paykit-UI step, since the route is behind `PaykitRouteGuard`), and that a Pubky Ring–imported identity can't approve the setup auth URL.

This one is shared with https://github.com/synonymdev/bitkit-ios/pull/720 — same gate, same string, same single-caller `createIdentity` — and I've filed it there too.

jvsena42 (commented):
Two findings inline, both on the precondition block. Doc-only, so neither blocks.

**Scope note:** the title says `test:` but three production files change. I checked each against master rather than against my previously reviewed head, because the rebase makes the raw diff misleading:
- `PubkyAuthApprovalSheet.kt` — **only two `testTag` additions** (`PubkyAuthAuthorize`, `PubkyAuthOK`). The extra ~10 lines versus what I last reviewed are the requester-label row from master's `ecb0499e9`/`75e1b4805`, carried in by the rebase, not by this PR. `git range-diff` shows all six previously reviewed commits patch-equivalent.
- `PaymentRequestsScreen.kt` — hyphenated row tag plus a new `PaymentRequestPay-<id>`; `onClick` and `enabled` unchanged.
- `SettingsScreen.kt` — `switchTestTag` rename only, still inside the same `isPaykitEnabled && isPubkyAuthenticated` guard.

So it is genuinely test-hook work and the consent gate is untouched. Worth stating explicitly since a consent-sheet change under a `test:` heading is exactly what gets waved through.

**Delta since my last pass** (`ba75c85e8`, `04d0720f8`): XML only — "Return the buyer wallet to Home" and moving the bell check after the auto-review is dismissed. Both address ben-kaufman's threads exactly without over-correcting, and the code agrees: auto-presentation is blocked while `currentSheet != null` but not by the current route, and the bell exists only on Home, so being on Home before the purchase is what makes it reachable after back.

**Checked and clean:**

- **Consent gate unchanged.** `onAuthorize` still guarded by `uiState.authUrl == authUrl`; `Content` renders `Loading` unless `isCurrentRequest`; Authorize -> `requestAuthorize` -> `RequestLocalAuth` -> PIN/biometrics -> `confirmAuthorize`, with `LocalAuthMode.None` only when neither PIN nor biometrics is enabled. The requester `clientId` and the full permission list still render, nothing truncated. No `BuildConfig`, flag or injected fake anywhere in the diff — only tag strings. No auto-approve or auto-dismiss path: `Dismiss` fires only from `viewModel.dismiss()` on user Cancel/OK/back.
- **Scope pinning.** `matchesWatchOnlyAccountCapabilities` rejects any capability set other than exactly the two watch-only paths, with `InvalidBitkitClaimCapabilities` thrown before the Authorize screen renders — so step 15's vaguer wording can't wave through an over-scoped grant even though the text is loose.
- **Back on the auto-review behaves as the journey says.** `canGoBack` is false at `Confirm`, so back routes to `hideSheet()` -> `clearActiveContactPaymentContext(retry = true)`, whose defer condition is already false because `onSheetVisible` called `markPresented` on show. Nothing re-presents ahead of the bell tap. `dismissEnabled` is false only during HW signing or an unresolved broadcast, neither of which applies pre-swipe. Dismissal never rejects — `reject` is called only from the Dismiss button, `accept` only after send — so the request stays in `_pendingRequests` and the bell, row and Pay all render.
- **Pay tag placement** matches the README's "paid request remained as history without Pay or Dismiss": the button row renders only when `onPay != null || onReject != null`, and history rows pass neither.
- **Every tag in `wallet-leg.xml` resolves at head** — the four `PubkyAuth*` ids, `ContactPaymentsToggle`, `Contact_<pk>`, `PaymentRequestsBell`, `PaymentRequestsSheet`, `PaymentRequestRow-`/`PaymentRequestPay-`, `ReviewAmount`, `ReviewContactRecipient` (inside `OnChainDetails`, behind the show-details toggle, so step 34 is accurate), `GRAB`, `SendSuccess`, `ActivityAmount`/`ActivityTxDetails`. A sweep found no un-hyphenated `PaymentRequestRow` reference left anywhere.
- **Step 19 is accurate**: a clean wallet completing in-app profile creation ends with `setEnabled(true)` on contact payments, so "verify enabled" holds for the standard flow.
- **Trust boundaries**: nothing newly rendered from counterparty data; the requester row that arrived from master is bounded with `maxLines = 1` and ellipsis.

**Cross-repo** with https://github.com/synonymdev/bitkit-ios/pull/720: 30/30 actions in identical order once `id`->`testTag` is normalised. The remaining divergences are the platform-appropriate back wording, and action 4 — iOS now enumerates `/pub/paykit/v0/bitkit/server` and `/pub/paykit/v0/private/bitkit/server` with `READ, WRITE`, where this file still says "only the requested Paykit capabilities". The behaviour is identical on both platforms, so **this is now the weaker file** and the enumeration is worth porting here rather than weakening iOS.

Also stale here: the README still says `PaymentRequestPay-<payment-request-id>` is Android-only and that an iOS buyer must find its Pay control by accessible label. https://github.com/synonymdev/bitkit-ios/pull/721 supersedes that.

On why this PR needs production changes when iOS needs none — iOS master already carried every identifier the journey uses and Android didn't, so this is pure instrumentation catch-up. That asymmetry is expected, not a smell.

journeys/pubky-marketplace/wallet-leg.xml
ovitrif: The setup now keeps Electrum and emulator port mapping before first launch, then explicitly enables Paykit UI through `PaykitUiToggle`, confirms the warning, and verifies `PaykitUiEnabledToast` before the setup auth URL is opened.

Fixed in 425d62a

journeys/pubky-marketplace/wallet-leg.xml
ovitrif: The XML and README now require a Bitkit-generated Pubky identity in both wallets after Paykit UI is enabled and state that a Pubky Ring-imported identity cannot approve the setup auth URL. The same commit also enumerates the exact requested capabilities and updates the iOS Pay-selector reference to #721.

Fixed in 425d62a

journeys/pubky-marketplace/wallet-leg.xml
jvsena42: Confirmed fixed in `425d62a4e` — the precondition now enables Paykit UI in Dev Settings after wallet creation with the `PaykitUiEnabledToast` activation check, dropping the impossible before-first-launch framing.

journeys/pubky-marketplace/wallet-leg.xml
jvsena42: Confirmed fixed in `425d62a4e` — both files now require a Bitkit-generated identity via **Create profile with Bitkit**, correctly sequenced after the Paykit-UI step, with the Ring caveat.

You also picked up the two side notes from my review body: action 4 now enumerates `/pub/paykit/v0/bitkit/server` and `/pub/paykit/v0/private/bitkit/server` with READ, WRITE, matching iOS, and the stale "`PaymentRequestPay-<id>` is Android-only" claim now points at bitkit-ios#721. Nothing left open from my side on this one.

jvsena42 (commented):
Clean — no findings.

Reviewed as an **authorization change**, not a test change: the `test:` label understates it, since three production files change and one of them is the auth consent sheet. Worth stating so it isn't waved through on the label — but also so it isn't over-escalated, because the verdict is that only `testTag` strings changed.

**Consent sheet (`PubkyAuthApprovalSheet.kt`) — two tag additions, no behaviour change:**
- `:355` `.testTag("PubkyAuthAuthorize")` on the Authorize button, `:434` on the success button. No `onClick`, state, effect, flag, `BuildConfig` or injected-fake path changed.
- Requester pinning intact end to end: `onAuthorize` guards `uiState.authUrl == authUrl` (`:85`); `Content` renders `Loading` unless `isCurrentRequest` (`:208`); `requestAuthorize` rejects unless `state.authUrl == authUrl && state == Authorize` (VM `:104`); `transitionToAuthorizing` re-checks the pin under CAS (VM `:237-243`); `authorize()` re-parses the same pinned string and bails if it moved (VM `:170`); `validatedApprovalClientId` throws `RequesterChanged` if the approved clientId differs from the displayed one (`PaykitSdkService.kt:1021-1026`).
- Local auth still sits between the tap and the grant (`RequestLocalAuth` → PIN/biometrics → `confirmAuthorize`); `None` only when neither PIN nor biometrics is configured (`:192-196`). Disclosure unchanged: requester row `:388`, full permission list `:465`, `TrustWarning` `:398`. Nothing truncated or reordered.

**`SettingsScreen.kt:346`** — `switchTestTag = "ContactPaymentsToggle"`, still inside `if (state.isPaykitEnabled && state.isPubkyAuthenticated)` (`:339`). Nothing on the ungated settings path changed.

**Journey verified against the *merged* code, not the pre-merge trace.** The two merge commits brought a 214-line master delta into `PaymentRequestsScreen.kt` (#1217/#1186: card-level `onClick`, title now `displayContact.name`, `onReject`→`onDismiss`, subscriptions) and rewrote `AppViewModel` — which is exactly where steps 17-21 could have silently broken. They didn't:
- All 18 tags resolve at head, PascalCase and exact.
- Step 4's wording is what actually renders: `displayPath` strips the trailing slash (`PubkyAuthRequest.kt:54`), `rw` → `"READ, WRITE"` (`:57-63`), and `matchesWatchOnlyAccountCapabilities` (`:26`) pins the set to the two `bitkit/server` paths, so over-scope fails before the Authorize screen.
- Step 8 holds: profile creation routes to `PayContacts`, whose `continueToProfile` calls `setEnabled(true)` (`PayContactsViewModel:43`).
- Step 18 holds post-merge: system back → `SheetHost` `BackHandler` → `hideSheet()` → `clearActiveContactPaymentContext(retry = true)`, which defers only if `requestedPaymentRequestId == id` (null for auto-presented) or the id is in `automaticPendingRequests()` (which excludes presented ids). No re-presentation, and the bell still reads raw `_pendingRequests`, which stays populated because dismissal never rejects.
- Step 20 holds: subtitle is `subscriptions.nameFor ?: note ?: date` (`:502`), so "note when present" is exact.

**Parity with https://github.com/synonymdev/bitkit-ios/pull/720:** 30/30 actions, identical order and selector names; remaining divergences are platform-appropriate (`testTag` vs `id`, "Press Android back" vs "Tap back"). iOS touches only journey files because its master already carried most identifiers, whereas Android lacked `PubkyAuthAuthorize`, `PubkyAuthOK`, `PaymentRequestPay-<id>`, the hyphenated row tag and the `ContactPaymentsToggle` name. So the production edits here are instrumentation catch-up, and the asymmetry is expected rather than suspicious.

Gating: `SettingsStore.kt:42` defaults `isPaykitEnabled` false, written only by the Dev Settings toggle — dev/QA-facing today.

jvsena42 (dismissed):

ben-kaufman (dismissed):

jvsena42:
@ovitrif conflicts

ben-kaufman (approved):

PR #1222, opened 2026-09-02
fix(receive): handle additional receive liquidity edge cases

### LINKED ISSUE #1226: [Bug]: Receive liquidity flow edge cases
# Receive Liquidity Behavior

This document describes how the receive flow decides whether to show a normal Lightning invoice or route the user into CJIT liquidity setup.

## Cases

- Editing from Savings or Auto:
  - Editing the invoice amount only affects the Savings/onchain invoice.
  - If the edited amount cannot be received over Lightning, Auto falls back to the Savings tab and shows the onchain QR instead of routing to CJIT.
  - The edit flow does not create CJIT or route to CJIT amount entry.

- Lightning receive unavailable because there is no ready channel or inbound liquidity is `0`:
  - No Lightning invoice is created.
  - The normal QR remains Savings/onchain only.
  - The Spending tab shows CJIT onboarding.
  - Tapping receive spending routes to CJIT amount entry, or the CJIT geo-block screen when geo-blocked.
  - Editing the normal invoice amount only affects the Savings/onchain invoice; it does not create or route to CJIT.
  - When a channel already exists, later CJIT confirmation and learn-more screens use additional-liquidity copy.

- Ready channel, inbound liquidity greater than `0`, zero/variable amount:
  - A Lightning invoice is allowed.
  - A zero/variable Lightning invoice is allowed when inbound liquidity is greater than `0`, even though the sender could later choose an amount above the available inbound capacity.

- Ready channel, fixed amount less than or equal to inbound liquidity:
  - A normal BOLT11 invoice is created.
  - The unified QR includes Lightning.
  - The Spending tab shows the normal Lightning invoice.

- Ready channel, fixed amount greater than inbound liquidity but below CJIT minimum:
  - A normal Lightning invoice is not shown.
  - Editing from Spending routes to CJIT amount entry.
  - The user must choose at least the minimum CJIT amount.
  - Editing from Savings or Auto returns to the normal QR with Savings/onchain only.

- Ready channel, fixed amount greater than inbound liquidity and at or above CJIT minimum:
  - If editing from Spending and the amount can be backed by a CJIT channel without exceeding Blocktank's maximum channel size, the edit flow creates additional CJIT.
  - The user gets CJIT confirmation and then a CJIT Lightning invoice QR.
  - If editing from Spending and the amount is too large for CJIT, or the maximum cannot be calculated, the edit flow routes to CJIT amount entry.
  - The CJIT amount screen enforces the real maximum receivable amount, calculated from `invoiceSat + defaultLspBalance(invoiceSat) <= maxChannelSizeSat`.
  - Editing from Savings or Auto returns to the normal QR with Savings/onchain only.

- Geo-blocked and liquidity is needed:
  - The flow routes to the CJIT geo-block screen.
  - No CJIT invoice is created.

pwltr:
Fixes #1226

### Description

Ports the receive liquidity/CJIT behavior from bitkit-ios https://github.com/synonymdev/bitkit-ios/pull/711.

- Adds a pure `ReceiveLiquidityDecision` helper for Lightning invoice vs additional CJIT routing.
- Prevents Savings and Auto edits from routing into CJIT while still updating the shared receive amount.
- Prevents normal Lightning invoices when there is no ready channel, zero inbound liquidity, or a fixed amount exceeds inbound liquidity.
- Adds Spending-only additional CJIT routing for amounts above inbound liquidity, including below-min routing to amount entry and real max validation using Blocktank default LSP balance.
- Makes Auto fall back to Savings/on-chain QR when Lightning cannot receive the edited amount.
- Resets Receive sheet session state on close/reopen and defaults fresh sessions to Auto when available, otherwise Savings.
- Keeps direct Spending additional-CJIT creation from regenerating the normal receive invoice before CJIT confirmation navigation.
- Polishes Receive details spacing and copy card address text wrapping.

### Preview

https://github.com/user-attachments/assets/251e0fe6-7d4e-44c2-96c8-ee6242bbdfe9

### QA Notes

Manual checks:
- Savings edit above inbound returns to Savings/on-chain QR and does not route to CJIT.
- Auto edit above inbound falls back to Savings/on-chain QR.
- Spending edit below CJIT minimum routes to CJIT amount entry.
- Spending edit with valid CJIT amount creates additional CJIT and opens confirmation.
- Spending edit above real CJIT maximum routes to amount entry and shows max validation.
- Receive sheet close/reopen starts fresh on Auto when available, otherwise Savings.
- Auto details show both on-chain and Lightning copy cards with expected spacing.

Automated checks:
- `ReceiveLiquidityDecisionTest.kt`
- `ReceiveInvoiceUtilsTest.kt`
- `EditInvoiceVMTest.kt`


app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -409,6 +444,21 @@ class BlocktankRepo @Inject constructor(
         return@withContext getDefaultLspBalance(params)
     }
 
+    private suspend fun maxChannelSizeSat(): ULong? {
+        if (_blocktankState.value.info == null) {
+            refreshInfo()
+        }
+
+        return _blocktankState.value.info?.options?.maxChannelSizeSat?.takeIf { it > 0uL }
```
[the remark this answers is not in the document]
pwltr: Fixed in [3123f32](https://github.com/synonymdev/bitkit-android/pull/1222/commits/3123f32ca8148ff2c419d2f9ac1ef94cb1f1dea5)

ovitrif (dismissed):
### QA Notes

Pixel_8_Pro emulator (sdk_gphone16k_arm64), regtest via network Electrum/Blocktank.

- Savings edit above inbound stayed on the on-chain QR and did not route to CJIT.
- Auto edit above inbound fell back to Savings/on-chain QR.
- Spending edit below the CJIT minimum routed to CJIT amount entry.
- Spending edit with a valid additional-CJIT amount opened confirmation.
- Spending edit above the real CJIT maximum stayed on amount entry and clamped to max.
- Receive sheet close/reopen started fresh on Auto.
- Auto details showed both on-chain and Lightning copy cards.

Approve.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -272,6 +276,36 @@ class BlocktankRepo @Inject constructor(
...
+        runCatching {
+            val maxChannelSizeSat = freshMaxChannelSizeSat() ?: return@runCatching true
+            return@runCatching canCreateCjit(amountSats, maxChannelSizeSat)
+        }.onFailure {
+            Logger.error("Failed to check CJIT limit", it, context = TAG)
+        }
+    }
+
+    suspend fun maxCjitAmountSats(): Result<ULong?> = withContext(bgDispatcher) {
+        runCatching {
```
ovitrif: This block calls the suspend function `freshMaxChannelSizeSat()`, so `runSuspendCatching` is required to preserve coroutine cancellation, Please use runSuspendCatching here because this block calls suspend functions; runCatching can swallow CancellationException.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -272,6 +276,36 @@ class BlocktankRepo @Inject constructor(
...
+            val maxChannelSizeSat = freshMaxChannelSizeSat() ?: return@runCatching null
+            var lowerBound = 0uL
+            var upperBound = maxChannelSizeSat
+
+            while (lowerBound < upperBound) {
+                val candidate = lowerBound + (upperBound - lowerBound + 1uL) / 2uL
```
ovitrif: nit: `upperBound - lowerBound + 1uL` can overflow,, please use the `.safe()` extensions for this `ULong` arithmetic so the midpoint calculation cannot wrap at `ULong.MAX_VALUE`.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveCjitErrors.kt
```diff
@@ -0,0 +1,12 @@
+package to.bitkit.ui.screens.wallets.receive
```
ovitrif: nit: can be inlined in `BlocktankRepo`

Please move this classification into `BlocktankRepo` and normalize matching backend failures to `ServiceError.ChannelSizeExceedsMaximum` there. The UI would handle the typed error instead of matching service-error strings in a separate UI utility.

---

For example, normalize the raw SDK error inside `BlocktankRepo.createCjit()` before returning it:

```kotlin
runSuspendCatching {
    // Create CJIT…
}.fold(
    onSuccess = { Result.success(it) },
    onFailure = { Result.failure(it.toCjitError()) },
).onFailure {
    Logger.error("Failed to create CJIT", it, context = TAG)
}
```

Then define this in `BlocktankRepo.kt`:

```kotlin
private fun Throwable.toCjitError(): Throwable {
    if (this is ServiceError.ChannelSizeExceedsMaximum) return this

    val description = toString()
    return if (
        description.contains("Channel size is too big") ||
        description.contains("channelSizeExceedsMaximum") ||
        description.contains("maxChannelSizeSat") ||
        description.contains("channelSizeSat")
    ) {
        ServiceError.ChannelSizeExceedsMaximum()
    } else {
        this
    }
}
```

The UI can then use only:

```kotlin
error is ServiceError.ChannelSizeExceedsMaximum
```

This keeps SDK-string interpretation in the repository and removes `ReceiveCjitErrors.kt`.

piotr-iohk:
There are conflicts.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -869,7 +900,7 @@ private fun PreviewSmall() {
                 lightningState = LightningState(
                     nodeLifecycleState = NodeLifecycleState.Running,
                 ),
-                onClickEditInvoice = {},
+                onClickEditInvoice = { _ -> },
```
ovitrif: nit: the change from `{}` to `{ _ -> }` is not needed, it still works with empty block

ovitrif (commented):
Added some nits after manual review for covering potential edge cases and defaulting to existing patterns throughout the code.

Been db checking them with AI on the side.

pwltr:
All review comments addressed

piotr-iohk:
QA Android, regtest. Existing channel: inbound 766,164, spending 3,000.

Invoice at 766,164 is fine. Savings edit above inbound stayed on-chain. Auto edit above inbound fell back to Savings QR. Neither entered extra CJIT.

Spending 766,165 opens extra-CJIT amount (MINIMUM 3,000). Continue at 3,000 fails.

UI toast is the raw exception, not the new max copy:

`errorDetails=Failed to create CJIT entry: Client error: Bad request: Failed order channel size check for node 03a65053…de027dab. Node capacity is above our capacity limit.`

Same Blocktank node-cap rejection as synonymdev/bitkit-ios#711. Client max handling only covers per-channel `maxChannelSizeSat`, so extra CJIT is still offered and then fails.

Please don’t offer extra CJIT (or map this error) when the node is already at Blocktank’s capacity limit. Don’t toast `errorDetails` / node id.

Recording:

https://github.com/user-attachments/assets/664c07f5-3285-4901-8a5a-15306ced7b2d

Logs:
[logs-android.zip](https://github.com/user-attachments/files/31838122/logs-android.zip)

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -190,8 +216,8 @@ fun ReceiveQrScreen(
     }
 
     // Auto-switch to AUTO tab when it becomes available for the first time
-    LaunchedEffect(hasUsableChannels) {
-        if (initialTab == null && hasUsableChannels && visibleTabs.contains(ReceiveTab.AUTO)) {
+    LaunchedEffect(canCreateLightningInvoice, cjitInvoice) {
+        if (canCreateLightningInvoice && cjitInvoice.isNullOrEmpty() && visibleTabs.contains(ReceiveTab.AUTO)) {
```
jvsena42: **🟡 Auto-switch effect lost the `initialTab == null` guard; overrides the requested Trezor tab**

master had `if (initialTab == null && hasUsableChannels && ...)`. The rewrite drops `initialTab == null`, so this effect now runs on first composition and every re-entry, scrolling to AUTO even when the sheet was opened with `initialTab = TREZOR` (hardware wallet receive, or returning from a hardware invoice edit via `invoiceEditState.initialTab`). `defaultTab` already handles the initial choice; restore `initialTab == null` here. iOS keeps the equivalent `tab == nil` guard in `applyDefaultTabIfNeeded`.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -175,6 +193,14 @@ fun ReceiveQrScreen(
...
         if (selectedTab !in visibleTabs) {
             selectedTab = visibleTabs.first()
+            lazyListState.scrollToItem(0)
+        }
+    }
+
+    LaunchedEffect(canCreateLightningInvoice, cjitInvoice) {
+        if (!canCreateLightningInvoice && cjitInvoice.isNullOrEmpty()) {
+            selectedTab = ReceiveTab.SAVINGS
+            lazyListState.scrollToItem(0)
```
jvsena42: **⚪ Savings fallback scrolls to index 0, which is the Trezor tab when a hardware wallet is paired**

`selectedTab = SAVINGS` but `scrollToItem(0)`, and the `snapshotFlow { firstVisibleItemIndex }` collector below immediately overwrites `selectedTab` with `visibleTabs[0]`. With a hardware wallet paired, `visibleTabs[0]` is TREZOR, so the assignment on the line above is dead. No user-visible defect today (index 0 is only TREZOR when the sheet was opened for that hardware wallet, which also sets `initialTab = TREZOR`), but write it as `scrollToItem(visibleTabs.indexOf(ReceiveTab.SAVINGS))` so the two lines agree.

app/src/main/java/to/bitkit/ui/ContentView.kt
```diff
@@ -509,17 +510,20 @@ fun ContentView(
...
-                                hardwareWalletId = sheet.hardwareWalletId,
-                                walletState = walletState,
-                                isOffline = connectivityState != ConnectivityState.CONNECTED,
-                                navigateToExternalConnection = {
-                                    navController.navigateTo(ExternalConnection())
-                                    appViewModel.hideSheet()
-                                },
-                            )
+
+                            key(System.identityHashCode(sheet)) {
```
jvsena42: **⚪ Session reset is untested and keyed on identityHashCode**

iOS ships `ReceiveSheetSessionTests` for the per-presentation identity; Android has no test for this reset (no `identityHashCode` reference anywhere in `app/src/test` or `app/src/androidTest`). Add one. Optional robustness while you are here: give `Sheet.Receive` an explicit `presentationId` like iOS's `SheetConfiguration.presentationID` and key on that, so the reset does not depend on `System.identityHashCode` uniqueness.

app/src/main/res/values/strings.xml
```diff
@@ -1253,6 +1253,8 @@
     <string name="wallet__receive_bitcoin_invoice">Bitcoin invoice</string>
     <string name="wallet__receive_connect_additional">To receive more instant Bitcoin, Bitkit has to increase your liquidity. A &lt;accent&gt;{networkFee}&lt;/accent&gt; network fee and &lt;accent&gt;{serviceFee}&lt;/accent&gt; service provider fee will be deducted from the amount you specified.</string>
     <string name="wallet__receive_connect_initial">To set up your spending balance, a &lt;accent&gt;{networkFee}&lt;/accent&gt; network fee and &lt;accent&gt;{serviceFee}&lt;/accent&gt; service provider fee will be deducted.</string>
+    <string name="wallet__receive_cjit_error_max__description">The amount you can receive with additional liquidity is currently limited to ₿ {amount}.</string>
+    <string name="wallet__receive_cjit_error_max__title">Maximum exceeded</string>
```
jvsena42: **⚪ Max-exceeded toast copy differs from iOS**

iOS: "Receiving Capacity Maximum" / "The maximum you can receive to your spending balance right now is ₿ {amount}."; Android: "Maximum exceeded" / "The amount you can receive with additional liquidity is currently limited to ₿ {amount}." Same meaning, but the keys are shared and both repos sync from Transifex; pick one wording.

jvsena42 (commented):
Diffed against the iOS port (#711). `ReceiveLiquidityDecision` and its 12 tests are a 1:1 match, and the routing, CJIT max search and session reset are equivalent. Android is the better of the two in three places (refreshes Blocktank info before the max check and tests it, hides Auto while a CJIT invoice is shown, routes edit-flow failures to the amount screen).

The two items worth attention are the tab effects, where this PR changed behaviour that the iOS side did not. Three of these are replies on existing threads.

Separately, and not filed as a finding since it is a bitkit-core question rather than a defect in this PR: `get_default_lsp_balance` returns `max_channel_size_sat` above threshold 2 (~EUR 495), so `client + lsp > max` for every amount above that and the "real max" this PR surfaces is effectively capped there regardless of Blocktank's `maxChannelSizeSat`. `calc_default_lsp_balance` uses `max - client` in the same band. Worth confirming that cap is intended.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveCjitErrors.kt
jvsena42: **🟡 Substring match on "channelSizeSat" misclassifies the LSP's too-small rejection as ChannelSizeExceedsMaximum**

The `channelSizeSat` clause strictly subsumes the `maxChannelSizeSat` clause above it, so any LSP error naming that field - a *below-minimum* rejection included - maps to `ServiceError.ChannelSizeExceedsMaximum`; `EditInvoiceScreen` then suppresses the real error and `ReceiveAmountScreen` shows "Maximum exceeded". Match on the maximum wording only (e.g. `"bigger than the maximum"`). Fixing this also drops the condition below detekt's complexity threshold flagged on this line.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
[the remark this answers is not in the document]
jvsena42: **🔵 createCjit now fails whenever the Blocktank info refresh fails, even with cached info**

`freshMaxChannelSizeSat()` does `refreshInfo().getOrThrow()`, so a transient failure of the `/info` request aborts CJIT creation although `_blocktankState.value.info` is populated and the LSP `createCjit` call itself would succeed. master did not have this dependency and iOS only refreshes when `info == nil`. Refresh best-effort (`refreshInfo()` without `getOrThrow()`) and fall back to the cached `maxChannelSizeSat`.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
jvsena42: **⚪ Raw ULong subtraction; AGENTS.md requires USat**

AGENTS.md: "ALWAYS wrap ULong numbers with USat in arithmetic operations". `maxChannelSizeSat - amountSats` here and at line 470 is raw (guarded by the preceding check, so currently safe). Use `maxChannelSizeSat.safe() - amountSats.safe()` like the rest of this file.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -409,6 +450,42 @@ class BlocktankRepo @Inject constructor(
         return@withContext getDefaultLspBalance(params)
     }
 
+    private suspend fun freshMaxChannelSizeSat(): ULong? {
+        refreshInfo().getOrThrow()
```
ovitrif: `freshMaxChannelSizeSat()` calls `refreshInfo().getOrThrow()`, and `refreshInfo()` wraps the whole cache-then-server sequence in a single `runCatching`, so a failing `info(refresh = true)` request returns `Result.failure` even after the cached value has already been written into `_blocktankState`. Because `createCjit()` now calls `freshMaxChannelSizeSat()` before the LSP request, a momentary Blocktank `/info` outage aborts CJIT creation entirely, which is a regression from `master` where `createCjit` never depended on `/info` at all. Could we call `refreshInfo()` without `getOrThrow()` here so a failed refresh falls back to the cached `maxChannelSizeSat` and CJIT creation still reaches the LSP?

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -54,8 +54,10 @@ import to.bitkit.di.BgDispatcher
 import to.bitkit.env.Env
 import to.bitkit.ext.calculateRemoteBalance
 import to.bitkit.ext.nowTimestamp
+import to.bitkit.ext.runSuspendCatching
 import to.bitkit.models.BlocktankBackupV1
 import to.bitkit.models.EUR
+import to.bitkit.models.safe
```
ovitrif: This new `import to.bitkit.models.safe` sits above `import to.bitkit.models.msatCeilOf`, but `msatCeilOf` sorts first. `config/detekt/detekt.yml` enables `ImportOrdering` with layout `*,java.**,javax.**,kotlin.**,^`, so `./gradlew detekt` reports a violation on this file. Could we swap the two lines, or run `just format` to autocorrect it?

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -409,6 +450,42 @@ class BlocktankRepo @Inject constructor(
...
+
+    private fun Throwable.toCjitError(): Throwable {
+        if (this is ServiceError.ChannelSizeExceedsMaximum) return this
+
+        val description = toString()
+        return if (
+            description.contains("Channel size is too big") ||
+            description.contains("channelSizeExceedsMaximum") ||
+            description.contains("maxChannelSizeSat") ||
+            description.contains("channelSizeSat")
```
ovitrif: `description.contains("channelSizeSat")` is a strict superset of the `maxChannelSizeSat` clause above it, so any LSP error that merely names that field maps to `ServiceError.ChannelSizeExceedsMaximum` - a below-minimum channel rejection included. `EditInvoiceScreen` then deliberately suppresses the toast for that type and `ReceiveAmountScreen` shows "Maximum exceeded", so the user is told the wrong thing or nothing at all. Could we match on the maximum-specific wording only, for example `"bigger than the maximum"`, and drop this bare field-name clause?

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -266,12 +272,47 @@ class BlocktankRepo @Inject constructor(
...
+            return@runSuspendCatching cjitEntry
+        }.fold(
+            onSuccess = { Result.success(it) },
+            onFailure = { Result.failure(it.toCjitError()) },
+        ).onFailure {
             Logger.error("Failed to create CJIT", it, context = TAG)
         }
     }
 
+    suspend fun canCreateCjit(amountSats: ULong): Result<Boolean> = withContext(bgDispatcher) {
```
ovitrif: AGENTS.md says "ALWAYS remove unused code after refactors". This new public `canCreateCjit(amountSats)` has no production caller: lines 287 and 303 both invoke the private two-argument overload, and the only other reference anywhere is `BlocktankRepoTest.kt`, so the test exercises code nothing else uses. The shipped UI paths call `maxCjitAmountSats()` instead. Could we delete this overload together with its test, or wire it into a real caller if it is meant to replace one of the `maxCjitAmountSats()` uses?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceVM.kt
```diff
@@ -30,26 +36,55 @@ class EditInvoiceVM @Inject constructor(
...
+                inboundCapacitySats = walletRepo.inboundLiquiditySats(),
+                isGeoBlocked = isGeoBlocked,
+            )
+        ) {
+            return null
+        }
+
+        blocktankRepo.refreshMinCjitSats()
+        return blocktankRepo.maxCjitAmountSats().getOrElse {
+            Logger.warn("Failed to calculate max CJIT amount", it, context = TAG)
```
ovitrif: AGENTS.md says "NEVER duplicate error logging in `.onFailure {}` if the called method already logs the same error internally", and also "do not add logs to viewmodels". `BlocktankRepo.maxCjitAmountSats()` already logs `"Failed to calculate max CJIT amount"` with the same throwable before returning the failed `Result`, so this line writes a second entry for every failure and makes the receive-flow logs noisier. Could we drop the `Logger.warn` and just return `null` from the `getOrElse` block?

app/src/main/res/values/strings.xml
```diff
@@ -1253,6 +1253,8 @@
     <string name="wallet__receive_bitcoin_invoice">Bitcoin invoice</string>
     <string name="wallet__receive_connect_additional">To receive more instant Bitcoin, Bitkit has to increase your liquidity. A &lt;accent&gt;{networkFee}&lt;/accent&gt; network fee and &lt;accent&gt;{serviceFee}&lt;/accent&gt; service provider fee will be deducted from the amount you specified.</string>
     <string name="wallet__receive_connect_initial">To set up your spending balance, a &lt;accent&gt;{networkFee}&lt;/accent&gt; network fee and &lt;accent&gt;{serviceFee}&lt;/accent&gt; service provider fee will be deducted.</string>
+    <string name="wallet__receive_cjit_error_max__description">The amount you can receive with additional liquidity is currently limited to ₿ {amount}.</string>
```
ovitrif: AGENTS.md says "ALWAYS add new localizable string resources in alphabetical order in `strings.xml`". These two `wallet__receive_cjit_error_max__*` keys are inserted after `wallet__receive_connect_initial`, but `cjit` sorts before `connect`, so the surrounding block is no longer ordered. Could we move both entries up so they sit between `wallet__receive_bitcoin_invoice` and `wallet__receive_connect_additional`?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
```diff
@@ -579,7 +609,10 @@ private fun ReceiveDetailsView(
         shape = AppShapes.small,
         modifier = modifier
     ) {
-        Column {
+        Column(
+            verticalArrangement = Arrangement.spacedBy(32.dp),
+            modifier = Modifier.padding(32.dp),
```
ovitrif: AGENTS.md says "ALWAYS add trailing commas in multi-line declarations, EXCEPT after a `modifier = ...` last argument - never add a trailing comma there". `modifier = Modifier.padding(32.dp)` is the last argument of this `Column` call and carries a trailing comma, which is the one case the rule excludes and which every other composable call in this file follows. Could we drop the trailing comma on this line?

ovitrif (dismissed):
Two things regress behaviour that `master` had, both in code the rebase brought in, so I'd like them fixed before this goes in.

The bigger one is `BlocktankRepo.freshMaxChannelSizeSat()`. It calls `refreshInfo().getOrThrow()`, and `refreshInfo()` wraps the whole cache-then-server sequence in a single `runCatching`, so a failing `info(refresh = true)` request returns `Result.failure` even after the cached value was already written into `_blocktankState`. Since `createCjit()` now calls it before the LSP request, one flaky Blocktank `/info` call aborts CJIT creation entirely. On `master`, `createCjit` had no dependency on `/info` at all. The earlier thread on this was marked fixed, but `getOrThrow()` is still there on this head.

The second is the AUTO auto-switch effect in `ReceiveQrScreen`. It lost the `initialTab == null` guard that `master` had. A `LaunchedEffect` body runs on first composition, so it now scrolls to `ReceiveTab.AUTO` and sets `selectedTab = ReceiveTab.AUTO` whenever Auto is visible, undoing the effect above it that applied the requested tab. `ReceiveSheet` passes `initialTab = ReceiveTab.TREZOR` whenever a hardware wallet is attached or the user is coming back from a hardware invoice edit, so hardware receive lands on Auto instead of Trezor.

The rest are non-blocking. `./gradlew detekt` reports two violations on `BlocktankRepo.kt` on this head (`ComplexCondition` at line 478 and `ImportOrdering`), which is why the code-scanning alert is still open; the error-classification substring is broader than intended; and there are a few small rule and cleanup items noted inline.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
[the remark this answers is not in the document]
ovitrif: Reopening this: `getOrThrow()` is still on `BlocktankRepo.freshMaxChannelSizeSat()` at line 454 on `d9c603a`, and `createCjit()` calls it unconditionally before the LSP request, so a transient `/info` failure still aborts CJIT creation with the cached `info` sitting in `_blocktankState`. Could we drop `getOrThrow()` so the cached `maxChannelSizeSat` is used and creation still reaches the LSP?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveCjitErrors.kt
ovitrif: Reopening this: `description.contains("channelSizeSat")` at `BlocktankRepo.kt:481` still subsumes the `maxChannelSizeSat` clause above it, so a below-minimum LSP rejection is still rewritten to `ServiceError.ChannelSizeExceedsMaximum` and surfaced as "Maximum exceeded". Could we match the maximum-specific wording only and drop the bare field-name clause?

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
```diff
@@ -409,6 +450,42 @@
...
+
+    private fun Throwable.toCjitError(): Throwable {
+        if (this is ServiceError.ChannelSizeExceedsMaximum) return this
+
+        val description = toString()
+        return if (
+            description.contains("Channel size is too big") ||
+            description.contains("channelSizeExceedsMaximum") ||
+            description.contains("maxChannelSizeSat") ||
+            description.contains("channelSizeSat")
```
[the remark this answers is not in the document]
ovitrif: Still open on this head. `./gradlew detekt --rerun-tasks` reports `BlocktankRepo.kt:478:13: This condition is too complex (4)`, and there is no baseline under `config/detekt/` to absorb it. Could we extract the classification into a small named helper, or narrow it to the maximum-specific wording, so the condition drops below the threshold?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
ovitrif: Confirmed on `d9c603a`. The effect at `ReceiveQrScreen.kt:216` still has no `initialTab == null` guard, and `ReceiveSheet.kt:143` passes `initialTab = ReceiveTab.TREZOR` whenever `hardwareWalletId != null || isHardwareInvoice`, so the auto-switch overrides the requested Trezor tab on first composition. Could we restore that condition?

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
ovitrif: Still present. `selectedTab = ReceiveTab.SAVINGS` followed by `scrollToItem(0)` disagree whenever `hardwareWalletId != null`, because `visibleTabs` puts `ReceiveTab.TREZOR` first, and the `snapshotFlow` collector then overwrites `selectedTab` with `visibleTabs[0]`. Could we scroll to `visibleTabs.indexOf(ReceiveTab.SAVINGS).coerceAtLeast(0)` so the two lines agree?

piotr-iohk:
HW receive broke after rebasing onto master (Trezor tab). e2e `hardware_wallet` fails for the same reason (`hardware_wallet_2/3/4`: Timed out waiting for hardware receive address).

Home Receive, paired Trezor, Spending 0 / no LN:

- master → **Savings**
- this branch → **Trezor**

New Lightning-unavailable fallback does `scrollToItem(0)`. Tabs are now `[Trezor, Savings, Spending]`, so that lands on Trezor. Master never scrolled.

E2E helper reads the default-tab QR, taps Trezor, and waits for a different address. This branch already opened on Trezor, so the addresses match and it times out.

Home Receive should stay Savings (Auto if LN works). Trezor only when opening from the HW wallet screen.

Recording (master):


https://github.com/user-attachments/assets/033b3b08-186d-4fef-8762-29bfe32c4410



Recording (this branch):


https://github.com/user-attachments/assets/6453abb3-292b-41bd-8b6d-df0da51998f2

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveCjitErrors.kt
pwltr: Addressed locally. Removed `ReceiveCjitErrors.kt` and moved CJIT max-error normalization into `BlocktankRepo.createCjit()` via `toCjitError()`, so the UI now only checks `ServiceError.ChannelSizeExceedsMaximum`.

I also narrowed the string matching to maximum-specific errors and dropped the broad `channelSizeSat` match so below-minimum rejections are not misclassified as max-exceeded.

app/src/main/java/to/bitkit/repositories/BlocktankRepo.kt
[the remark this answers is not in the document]
pwltr: Addressed. `freshMaxChannelSizeSat()` now refreshes Blocktank info before calculating the CJIT max, so it does not rely on stale cached options.

The refresh is best-effort: if `/info` fails after cached info was loaded, we fall back to the cached `maxChannelSizeSat` instead of aborting CJIT creation before reaching the LSP call.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
pwltr: Addressed. The Savings fallback now scrolls to the actual Savings tab index with `visibleTabs.indexOf(ReceiveTab.SAVINGS).coerceAtLeast(0)` instead of hardcoding index `0`.

This keeps `selectedTab` and the visible pager item aligned when the Trezor tab is present.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveQrScreen.kt
pwltr: Addressed. Restored the `initialTab == null` guard on the Auto-switch effect, so the sheet only auto-selects `Auto` for default receive presentations.

When the sheet is opened with an explicit initial tab, like Trezor receive, that requested tab is preserved.

pwltr:
> HW receive broke after rebasing onto master (Trezor tab). e2e `hardware_wallet` fails for the same reason (`hardware_wallet_2/3/4`: Timed out waiting for hardware receive address).

Addressed. Home Receive no longer defaults to Trezor just because a single hardware wallet is paired; it defaults to Auto when Lightning is available, otherwise Savings.

Trezor is now only used as the initial tab when opening from the hardware wallet receive flow, or when returning from an edit that actually originated on the Trezor tab. I also fixed the Savings fallback to scroll to the actual Savings index instead of hardcoding `0`, so `[Trezor, Savings, Spending]` no longer lands on Trezor.

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -748,19 +747,17 @@ class WalletRepo @Inject constructor(
...
+    fun inboundLiquiditySats(): ULong {
+        return lightningRepo.lightningState.value.channels.calculateRemoteBalance()
+    }
 
-            return@runCatching (_walletState.value.bip21AmountSats ?: 0uL) >= inboundBalanceSats
-        }.onFailure {
-            Logger.error("shouldRequestAdditionalLiquidity error", it, context = TAG)
-        }
+    private fun canCreateLightningInvoice(amountSats: ULong?): Boolean {
+        val channels = lightningRepo.lightningState.value.channels
```
jvsena42: **Regression — this reads the cached channel list, but it's called to *decide* on the very event that invalidates that cache.**

Master used `lightningRepo.canReceive()` at both :325 and :339, which goes to `lightningService.canReceive()` → `node.listChannels()` (`LightningService.kt:1322`) and is always live. `lightningState.value.channels` is only refreshed asynchronously.

The close case is deterministic, not a race: `handleLdkEvent` launches `registerClosedChannel` (`LightningRepo.kt:737`), whose body never calls `syncState()`, and `AppViewModel.handleChannelClosed` → `walletRepo.syncBalances()` reads channels live but never writes `_lightningState.channels`. So when the only channel closes, the cache still holds it, `canCreateLightningInvoice` returns true, and the `!canCreate…` guard skips `setBolt11("")` at :339-342 — the Receive QR keeps offering a bolt11 that can no longer be paid until the sheet is reopened.

The open case is a genuine race but usually won: `handleLdkEvent` for `ChannelReady` is a fire-and-forget `scope.launch { refreshChannelCache(); syncState() }` (:732-735) on `Dispatchers.Default`, racing the `_nodeEventUpdates.emit` at :515, though the suspend `_eventHandlers` chain ahead of the emit biases it to land first.

Scope is narrower than it first looks — `PublicPaykitRepo` mints its own invoice via `lightningRepo.createInvoice` and never reads `walletState.bolt11`, and `ReceiveSheet.kt:100-103` calls `refreshReceiveState()` (syncState then refreshBip21) on every sheet open — so it only bites while the sheet is already open.

One line either way: read `lightningRepo.getChannels()` here, or `syncState()` at the top of `refreshBip21ForEvent` for channel events (that also freshens `ReceiveQrScreen`, which reads the same cache).

Separately: this swaps master's `isUsable` for `isChannelReady`, so a ready-but-peer-disconnected channel now counts as receivable. Intentional?

Regression test attached in the PR conversation.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveAmountScreen.kt
```diff
@@ -106,8 +154,13 @@ fun ReceiveAmountScreen(
                         )
                     )
                 }.onFailure { e ->
-                    app.toast(e)
                     Logger.error("Failed to create CJIT", e)
+                    if (e is ServiceError.ChannelSizeExceedsMaximum) {
+                        maxCjitAmountSats = runSuspendCatching { blocktank.maxCjitAmountSats() }.getOrNull()
+                        maxCjitAmountSats?.let { showMaxExceededToast(it) }
```
jvsena42: Nit: when the recompute returns null this shows nothing at all — the `else { app.toast(e) }` is skipped, so the user gets zero feedback and the button just re-enables.

Reachable with fiat rates unavailable: `maxCjitAmountSats()` needs `getSatsPerEur()` (`BlocktankRepo.kt:435-436`) so it's null on entry, `setMaxAmount(0)` installs no cap and `canContinue` (:127) treats null as unbounded, so `createCjit` throws from the `amountSats > maxChannelSizeSat` guard (:257-260) — which runs before any rate-dependent code — and the recompute is null again. Silent failure on every retry. Master toasted unconditionally.

`maxCjitAmountSats?.let { showMaxExceededToast(it) } ?: app.toast(e)`

jvsena42 (commented):
Traced the funds-critical parts and they hold up: `createCjit` still sends `invoiceSat = amountSats` with `channelSizeSat = amountSats + lspBalance`, the Confirm screen's `receiveAmountSats` is the same `action.amountSats` used for the invoice, the binary search in `maxCjitAmountSats()` is over a monotone predicate and is overflow-safe, `getInvoiceForTab` is called with the per-item `tab`, and the hardware address is only produced under `ReceiveTab.TREZOR`. I found no way to show a destination belonging to the wrong tab and no sat/msat or off-by-one error in the limits.

One regression and one nit inline.

jvsena42:
### Regression test — stale channel cache in `canCreateLightningInvoice`

Two tests that separate the cached source from the live one. Test-only, applies cleanly onto `fix/additional-cjit` (78fa7d470).

```
./gradlew :app:testDevDebugUnitTest --tests 'to.bitkit.repositories.WalletRepoTest'
```

**On head: 50 tests, 2 failed** (the two new ones; all 48 pre-existing pass).
```
refreshBip21ForEvent ChannelClosed should clear bolt11 when live channels are gone
  -> org.junit.ComparisonFailure: expected:<[]> but was:<[testInvoice]>

refreshBip21ForEvent ChannelReady should create invoice from live channels
  -> WantedButNotInvoked: lightningRepo.createInvoice(<any>, <any String>, <any int>);
     However, there were exactly 2 interactions with this mock: lightningRepo.ge...
```

The ChannelClosed one is the deterministic case: stale cache still holds the ready channel, live `getChannels()` is empty, so the `!canCreateLightningInvoice(...)` guard at :339 is skipped and the unpayable bolt11 survives in the BIP21/QR.

**Both pass** with the minimal fix at `WalletRepo.kt:755`:
```kotlin
val channels = lightningRepo.getChannels() ?: lightningRepo.lightningState.value.channels
```

⚠️ **One thing to know before you land the fix:** with it applied, 4 pre-existing tests fail — `updateBip21Invoice should create bolt11 when node can receive`, `refreshBip21ForEvent ChannelReady should update bolt11 and preserve amount`, `refreshBip21ForEvent ChannelClosed should not clear bolt11 when can still receive`, and `refreshBip21 should create a fresh invoice after PaymentReceived invalidates the old one`. They only stub the cached source, which is precisely why none of them caught this. The `?:` fallback doesn't rescue them because Mockito's default for a `List<ChannelDetails>?` getter is an empty list, not null. Adding `whenever(lightningRepo.getChannels()).thenReturn(channels)` beside each existing `lightningState` stub makes the class green at 50/50.

Those 4 stub updates are deliberately **not** in this diff — they'd be unused stubs on unfixed code, so they belong with the fix commit.

<details><summary>pr1222-regression.diff</summary>

```diff
diff --git a/app/src/test/java/to/bitkit/repositories/WalletRepoTest.kt b/app/src/test/java/to/bitkit/repositories/WalletRepoTest.kt
index c141eff37..84eccc614 100644
--- a/app/src/test/java/to/bitkit/repositories/WalletRepoTest.kt
+++ b/app/src/test/java/to/bitkit/repositories/WalletRepoTest.kt
@@ -588,6 +588,39 @@ class WalletRepoTest : BaseUnitTest() {
         assertEquals(INVOICE, sut.walletState.value.bolt11)
     }
 
+    @Test
+    fun `refreshBip21ForEvent ChannelClosed should clear bolt11 when live channels are gone`() = test {
+        sut.setBolt11(INVOICE)
+        // stale cache still holds the channel that just closed
+        whenever(lightningRepo.lightningState).thenReturn(MutableStateFlow(LightningState(channels = channels)))
+        whenever(lightningRepo.getChannels()).thenReturn(emptyList())
+
+        sut.refreshBip21ForEvent(
+            Event.ChannelClosed(
+                channelId = "testChannelId",
+                userChannelId = "testUserChannelId",
+                counterpartyNodeId = null,
+                reason = null,
+            )
+        )
+
+        assertEquals("", sut.walletState.value.bolt11)
+    }
+
+    @Test
+    fun `refreshBip21ForEvent ChannelReady should create invoice from live channels`() = test {
+        sut.setBip21AmountSats(SATS)
+        // stale cache does not know about the channel that just became ready
+        whenever(lightningRepo.lightningState).thenReturn(MutableStateFlow(LightningState()))
+        whenever(lightningRepo.getChannels()).thenReturn(channels)
+        whenever(lightningRepo.createInvoice(anyOrNull(), any(), any())).thenReturn(Result.success(INVOICE))
+
+        sut.refreshBip21ForEvent(channelReady)
+
+        verify(lightningRepo).createInvoice(anyOrNull(), any(), any())
+        assertEquals(INVOICE, sut.walletState.value.bolt11)
+    }
+
     @Test
     fun `refreshBip21ForEvent PaymentReceived should refresh address if used`() = test {
         whenever(cacheStore.data).thenReturn(flowOf(AppCacheData(onchainAddress = ADDRESS, bolt11 = INVOICE)))
```

</details>

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
pwltr: Addressed in [7fca106](https://github.com/synonymdev/bitkit-android/pull/1222/commits/7fca1061a2f93e40cf556c2e87c3e4e37a35c6b2)

jvsena42 (commented):
Re-reviewed at `3c21400`. No HIGH/MEDIUM — nothing blocking from me. One LOW nit inline, pre-existing, take it or leave it.

**Amount and fee honesty — clean.** `createCjit` sends `invoiceSat = amountSats` and `channelSizeSat = amountSats.safe() + lspBalance.safe()`; both confirm routes read `feeSat`/`invoice` off the same `IcJitEntry`, and `ReceiveConfirmScreen` displays `receiveAmountSats - feeSat`. CJIT entries are immutable and unpaid ones cost nothing, so a retry can't double-charge. Every CJIT still passes through `ReceiveConfirmScreen` (initial or `isAdditional`) before `cjitInvoice` is set — no path shows the LSP invoice without the fee screen first.

**Max-amount search — clean.** `maxCjitAmountSats()`'s binary search and `canFitChannelSize` are USat-guarded and monotone, and `freshMaxChannelSizeSat()` is now best-effort with a cached fallback, so a `/info` blip degrades instead of hard-failing.

**Stale-inbound — clean, and this was the one I most wanted to settle.** `currentChannels()` is live-first (`lightningRepo.getChannels()` → `node.listChannels()`, guarded by `isRunning`), `syncState()` runs before the Channel* decisions, and `calculateRemoteBalance()` filters `filterOpen` and then `isUsable` — so pending channels can't inflate displayed inbound and steer someone away from a CJIT they actually need.

**Also traced clean:** `updateBip21Invoice` clears bolt11 *before* `updateBip21Url`, so no stale `lightning=` param can survive; `getInvoiceForTab(AUTO)` strips lightning on fallback and `invoicesByTab` keys cover every input that can change it. `presentationId` only participates in `key()` — no production code compares `Sheet.Receive` by equality, so the session reset is sound. `createCjit`/`maxCjitAmountSats` use `runSuspendCatching`, and the effect `SharedFlow` drops with no subscriber, so a back-press mid-decision just discards the effect rather than stranding state. `toCjitError` only substring-matches into a typed error — no Blocktank string is rendered beyond the pre-existing `app.toast(e)`. Nothing seed-derived is touched, and nothing here is persisted, so there's no migration path from the shipped build to worry about.

**iOS #711 parity:** every finding from that side is either already handled here or structurally absent — pending-channel inbound (filtered by `isChannelReady && isUsable`), Auto hidden while CJIT is shown (`visibleTabs`), edit-flow failure routing (`navigateCjitAmount`), refresh-before-max (`freshMaxChannelSizeSat`), and `canCreateCjit`/max coverage (`BlocktankRepoTest`). The only iOS point with an Android analog not already on record is the in-flight guard, inline below — and it lands on the edit path here rather than the amount screen.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
```diff
@@ -85,49 +97,57 @@ fun EditInvoiceScreen(
...
+                        ReceiveAdditionalLiquidityAction.None -> {
+                            updateInvoice(receiveSats)
+                            onBack()
+                        }
+                        ReceiveAdditionalLiquidityAction.ChooseAmount -> {
+                            updateInvoice(receiveSats)
+                            navigateCjitAmount()
+                        }
+                        is ReceiveAdditionalLiquidityAction.CreateCjit -> {
+                            runSuspendCatching { blocktankVM.createCjit(action.amountSats) }.onSuccess { entry ->
```
jvsena42, line 131: Nit, optional, pre-existing on master — the Continue spinner drops before this call finishes.

`EditInvoiceVM.kt:56-57` emits `ApplyReceiveLiquidityAction(action)` and then immediately sets `_isLoading` false, and `editInvoiceEffect` is its own `viewModelScope.launch`, so `isLoading` clears regardless of what the collector is doing. By the time this line runs `refreshInfo` + `getDefaultLspBalance` + the LSP `createCjit` round trip, `PrimaryButton` is already back to `enabled && !isLoading` — so the user sees the spinner vanish while the network call is still in flight and the screen hasn't moved. `ReceiveAmountScreen.kt:252-253` guards the same call with `isCreatingInvoice`; the edit path doesn't.

A second tap is mostly harmless in practice — the collector is sequential and every branch navigates away, cancelling the `LaunchedEffect` — so this is about the misleading spinner, not a duplicate entry. `var isCreatingCjit by remember { mutableStateOf(false) }` around the call, passed as `isLoading = isLoading || isCreatingCjit`, mirrors the amount screen.

pwltr: Addressed in [10355fa](https://github.com/synonymdev/bitkit-android/pull/1222/commits/10355fa02a7ab5635401dcbc5d5e476efbfb0738)

jvsena42 (commented):
Re-reviewed at `10355fa`. The only change since `3c21400` is the `isCreatingCjit` guard in `EditInvoiceScreen`, and it's correct: the flag is set before the `createCjit` call, cleared after the whole `runSuspendCatching` block (so both the success and failure branches are covered), and OR'd into `isLoading` on the way to `EditInvoiceContent` — the Continue button now stays disabled and spinning for the duration of the LSP round trip. Cancellation is fine too: `runSuspendCatching` rethrows, but the composable is leaving at that point and the `remember` state goes with it.

No new findings. Still no HIGH/MEDIUM from me on this PR.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -344,17 +356,20 @@ fun ReceiveSheet(
...
     }
 
     fun beginHardwareEdit() {
         isHardwareInvoice = true
+        returnTab = ReceiveTab.TREZOR
     }
 
     fun initialTab(hardwareWalletId: String?): ReceiveTab? =
-        ReceiveTab.TREZOR.takeIf { hardwareWalletId != null || isHardwareInvoice }
+        returnTab ?: ReceiveTab.TREZOR.takeIf { hardwareWalletId != null }
```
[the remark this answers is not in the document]
pwltr: I did not change the returnTab thread. I think that one is a UX tradeoff, not a bug: after a user edits from a tab, preserving that tab for the rest of the current sheet presentation is deliberate. Closing/reopening still starts fresh because the receive sheet is keyed by presentationId.

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -318,11 +319,18 @@ class WalletRepo @Inject constructor(
         settledReceiveInvoice: SettledReceiveInvoice? = null,
         settledReceiveAddress: SettledReceiveAddress? = null,
     ) = withContext(bgDispatcher) {
+        when (event) {
+            is Event.ChannelReady,
+            is Event.ChannelClosed,
+            -> lightningRepo.syncState()
```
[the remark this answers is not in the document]
pwltr: Addressed in [07ec441](https://github.com/synonymdev/bitkit-android/pull/1222/commits/07ec441046b852739c95779dfc295de54e873ad4)

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -748,19 +755,25 @@ class WalletRepo @Inject constructor(
...
+        return ReceiveLiquidityDecision.canCreateLightningInvoice(
+            hasUsableChannels = usableChannels.isNotEmpty(),
+            inboundCapacitySats = usableChannels.calculateRemoteBalance(),
+            invoiceAmountSats = amountSats,
+        )
+    }
 
-            val inboundBalanceSats = channels.sumOf { msatFloorOf(it.inboundCapacityMsat) }
+    private fun currentChannels(): List<ChannelDetails> {
+        return lightningRepo.getChannels() ?: lightningRepo.lightningState.value.channels
```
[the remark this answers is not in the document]
pwltr: Addressed in [07ec441](https://github.com/synonymdev/bitkit-android/pull/1222/commits/07ec441046b852739c95779dfc295de54e873ad4)

jvsena42 (commented):
Re-reviewed the delta at `07ec44104`. **No HIGH/MEDIUM.**

The off-main move is right and it fixed a second thing at the same time. `inboundLiquiditySats()` is now `suspend` on `bgDispatcher`, so `currentUsableChannels().calculateRemoteBalance()` no longer runs on the caller's thread — and hoisting it to a single `inboundCapacitySats` local in `onClickContinue` removes the duplicate call that `needsCjitLimitsForAdditionalLiquidity` and `additionalLiquidityAction` were each making. Both now read the same snapshot, which also closes a small window where the two decisions could have been made against different channel state mid-evaluation.

`verify(walletRepo, times(1)).inboundLiquiditySats()` pins the dedup rather than just the happy path, so a future refactor that re-introduces the second read fails the test. The `syncState()` assertions added across the `WalletRepoTest` bip21 cases are the right shape too — they pin that the refresh path syncs before deciding, including on the negative branches.

Nothing else in the delta.

ovitrif (approved):
LGTM, review bot did the rest to confirm this is ready to merge from my side.

PR #1231, opened 2026-09-07
feat: add blockstream jade hardware wallet support

coreyphillips:
This PR:
1. Adds Blockstream Jade as a second hardware wallet vendor, over USB and Bluetooth
2. Generalises the hardware wallet layer so every device call is routed by the vendor of the paired entry
3. Releases an open Jade Bluetooth link while Bitkit is backgrounded and reconnects silently on return

Requires bitkit-core [0.5.16](https://github.com/synonymdev/bitkit-core/releases/tag/v0.5.16) ([synonymdev/bitkit-core#153](https://github.com/synonymdev/bitkit-core/pull/153)), which carries the Jade module and pins [`jade-client-rs`](https://github.com/synonymdev/jade-client-rs) at `d52ccd9`.

### Description

A Jade can now be paired from Connect Hardware over either transport, unlocked with its PIN, and
used exactly like a paired Trezor: watch-only balances, on-device receive address verification, and
on-device signing for both a normal send and a transfer to spending. The protocol, the pinserver
round trip and every deadline live in bitkit-core. This app supplies the byte transport over the
phone's radios and the UI that drives the flows.

The transport covers USB serial through a CP210x bridge on Jade v1 and native USB CDC on Jade Plus,
plus Bluetooth over the Nordic UART Service. Three USB device filter entries were added so Android
offers Bitkit when a Jade is plugged in.

Four things only a physical device revealed, each fixed here:

- The TX characteristic is indicate-only on this firmware, so the transport subscribes to
  indications when notify is absent instead of failing the connection.
- A Jade advertises under a new random Bluetooth address after every reboot or pairing reset, so a
  stored entry is recognised by name, which is Jade plus the last six hex digits of its efuse MAC,
  rather than by address.
- A link left open when the process dies wedges the device's single connection slot until it is
  power-cycled. Links are now closed when the activity finishes, and released after 30 seconds in
  the background so the same thing does not happen when Android kills a backgrounded Bitkit. Coming
  back to the foreground reconnects without a prompt.
- A bond that went stale across a re-pair used to stall the first write. It now gets a write budget
  wide enough to cover a re-pair, and a message telling the user to forget the Jade in Android's
  Bluetooth settings and pair again.

The vendor-neutral part is a refactor rather than new behaviour. The hardware wallet repository now
merges both vendors' discovery state, routes connect, verify and sign by the vendor stored on the
paired entry, and alternates which vendor gets the Bluetooth half of a scan so repeated searches
stay under Android's scan-rate limit. Watchers, transaction composition and broadcast are vendor
neutral already and stay where they are. Entries saved before this change carry no vendor and are
read as Trezor, so paired Trezors are untouched. Reconnect gets a longer deadline for a Jade,
because that reconnect may be waiting for a PIN to be entered on the device.

Session and identity hardening added during review:

- Cancelling pairing or receive verification explicitly cancels Jade work and closes the native transport in cleanup that survives caller cancellation.
- Every failed post-connect step closes both the Android transport and core session.
- A known USB reconnect checks each attached Jade until its efuse identity matches, but stops immediately if the expected Jade itself fails.
- The shared repository enforces one active vendor session and checks both vendor states before Bluetooth scans.
- Equal seeds on different vendors keep separate wallet identities, labels, watchers, and signing routes.
- Signing on a Jade checks that the connected session belongs to the wallet being spent from, as the Trezor path does.
- Finishing the activity releases the Jade transport, core session and cached connection together, off the main thread.
- The receive sheet closes a hardware session only if it used the device, and a USB attach for a vendor with nothing paired no longer drops the other vendor's session.
- The send sheet can be dismissed while it connects or waits for a Jade PIN; it stays locked while the device signs and while a broadcast is unresolved.

Jade-specific failures get their own copy: PIN entry, wrong PIN, an unreachable pinserver, a device
that is busy, firmware too old, a device that has no wallet yet, a network mismatch, and a PSBT the
device cannot hold.

Two gaps worth naming. The Jade illustration is a placeholder vector until design supplies the real
asset. Signet is not supported by Jade, so that combination throws rather than mapping to a network.

### QA Notes

Verified against a Jade v1 on firmware 1.0.41. There is no Jade emulator in `bitkit-docker`, so
these are all physical-device checks.

#### Manual Tests

- [x] **1.** Jade over USB → Connect Hardware → Search → Pair: PIN prompt shows on the device,
  unlock completes, accounts export and the wallet tile appears.
- [x] **2.** `regression:` USB → Send → pick the Jade source → sign on device → broadcast:
  transaction confirms.
- [x] **3.** Jade over Bluetooth → Connect Hardware → Search → Pair: pairs and unlocks.
- [x] **4.** Bluetooth → Receive → Hardware tab → Verify on Device: the address shown on the Jade
  matches the one in the app.
- [x] **5.** Bluetooth → Send → sign on device → broadcast: broadcast succeeded, txid
  `d955bc0c...`.
- [x] **6.** Bluetooth connected → background Bitkit for about 45 seconds → reopen: reconnects with
  no pairing prompt and no PIN re-entry.

#### Automated Checks

- Unit tests added: `JadeTransportTest.kt` covers USB driver selection, the CP210x and CDC open and
  close sequences, chunk sizing and read and write timeouts; `JadeRepoTest.kt` covers connect,
  unlock, replug and reconnect, recognising a Bluetooth Jade by name after its address changed, the
  background release and its USB counterpart, signing and address verification; `JadeServiceTest.kt`
  covers the `finalizePsbt` alias that used to recurse into itself; `HwUsbIdTest.kt` covers vendor
  detection from USB ids; `KnownDeviceTest.kt` covers vendor-aware entry matching, migration of
  pre-Jade entries, wallet identity and equal-seed vendor isolation; `HwErrorPresenterTest.kt` and `HwExceptionExtTest.kt` cover
  the Jade error copy and classification. The Bluetooth GATT paths themselves, including the
  indicate-only fallback, are not unit testable and were validated on hardware.
- Unit tests modified: `HwWalletRepoTest.kt`, `HwConnectViewModelTest.kt`, `HwSendViewModelTest.kt`,
  `HwReceiveViewModelTest.kt`, `TransferViewModelTest.kt`, `TrezorRepoTest.kt` and
  `ReceiveInvoiceUtilsTest.kt` move onto the vendor-neutral device state and the per-vendor routing.
- Local: `just compile`, `just test` (2626 tests, 0 failures) and `just lint` all pass.


app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -291,6 +294,9 @@ class MainActivity : FragmentActivity() {
 
     override fun onDestroy() {
         super.onDestroy()
+        // A Jade left with an open Bluetooth link when the process dies can refuse connections until
+        // it is power-cycled, so release every link when the activity is going away for good.
+        if (isFinishing) jadeTransport.closeAllConnections()
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 2dd3da77e. `MainActivity.onDestroy` now calls `HwWalletRepo.onActivityFinishing()`, which calls `JadeRepo.releaseAllConnections()`. It cancels the pending background release and transport reconnect jobs, then runs on the repository scope (off the main thread, and it outlives the activity) to close the transport links, call `jadeService.disconnect()` and clear `connected`. The transport closes first, because a device call still waiting on the link holds the core queue. Tests: `releasing all connections closes transport and core session and clears the cached link` (including that ordering), `foreground reconnect runs again after the connections were released`, and `activity finishing releases the jade connections`.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -504,35 +793,45 @@ class HwWalletRepo @Inject constructor(
...
-                if (it.isTrezorSessionFailure()) {
-                    transportDeviceIdOrNull(walletId)?.let { deviceId -> trezorRepo.disconnectStaleSession(deviceId) }
-                }
-                throw it
+            val serializedTx = when (vendorOf(walletId)) {
+                HwWalletVendor.TREZOR -> signTrezorFunding(walletId, funding)
+                HwWalletVendor.BLOCKSTREAM -> jadeRepo.signPsbt(funding.psbt).getOrElse {
+                    if (it.isHwSessionFailure()) disconnectStaleSession(walletId)
+                    throw it
+                }.serializedTx
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 2dd3da77e. The Blockstream branch now goes through `signJadeFunding`, which does the same identity check as `signTrezorFunding` before calling `signPsbt`. A connected session whose `walletId` differs throws `HwWalletMismatchError`. A session whose accounts could not be read must have a device id that belongs to the requested wallet's entries. With no session, core reports not connected and the existing reconnect path handles it, which is how Trezor behaves too. Tests cover a mismatched wallet, an unresolved session from another device (neither calls `signPsbt`), and the matching case.

jvsena42 (commented):
Reviewed at `ca919d98b` as a funds/signing change. **No HIGH, no MEDIUM.** Three LOW notes below, all on the lifecycle surface rather than the signing path — none loses funds, none wedges the app.

**The signing path holds up.** This was the thing I most wanted to break, so here is the trace, because the two vendors are protected by *different* mechanisms and that is worth writing down:

- The PSBT is composed app-side from the stored account xpub, so the change output is app-derived and never device-sourced. The only device-supplied input to compose is the master fingerprint, which affects key-origin metadata only.
- **Jade:** `signPsbt` feeds the device's return to `finalizePsbt(originalPsbt, signedPsbt)`. In bitkit-core v0.5.16 `src/modules/onchain/psbt.rs`, :55-59 rejects `original.unsigned_tx != signed.unsigned_tx`, :61-73 pins each input's previous output, then `finalize_mut` + `interpreter_check` verify the signatures against the pinned inputs. The device can contribute signatures and nothing else.
- **Trezor:** does *not* go through `finalizePsbt` — it broadcasts `serializedTx` directly. It is protected instead by trezor-connect-rs 0.4.0, which derives `expected_scripts` before signing and then runs `verify_signed_tx` on the device-returned bytes (`src/tx_verify.rs:23-73`, a port of @trezor/connect's `verifyTx`): output count, every output amount, every output scriptPubKey against independently derived expectations. Equivalent guarantee, different code path. Worth knowing if anyone later assumes `finalizePsbt` covers both.
- One caveat, **pre-existing and not this PR**: the Trezor path does not pin input outpoints, so a device could in principle substitute another UTXO the same seed controls. That matches upstream Trezor Connect.
- Every signing caller routes through `signFunding`/`broadcastFunding`. The only other `signTxFromPsbt`/`broadcastRawTx` user is the pre-existing dev TrezorScreen. Nothing broadcasts a device return without one of these checks.

Also verified clean: amount and address shown on `HwSendSignScreen` are the same values that build the `HwSendRequest`, and `miningFeeSats` comes from the same compose result that produced the PSBT — no recompute after display. A swapped Jade with a matching efuse MAC is caught downstream (different xpubs produce a new walletId, so `ensureConnected(oldWalletId)` throws; in the locked silent-reconnect path where xpubs aren't re-read, signing fails at `interpreter_check`). No xpub, PSBT or fingerprint reaches a `Logger` call. `runSuspendCatching` is used throughout; plain `runCatching` appears only around non-suspending calls and once with the correct explicit `CancellationException`/`TimeoutCancellationException` rethrow guard.

**Trezor regression surface** — the part I'd most expect a second-vendor PR to break. `TrezorRepo` changes are limited to vendor-scoped store reads/writes, the (null-for-Trezor) fingerprint passthrough, and helpers moved to `KnownDevice.kt` with identical logic plus a vendor equality check. `TrezorTransport` only extracts `requestUsbPermission` into `UsbPermissionRequester` with the same action, flags and timeout. `HwWalletStore.saveKnownDevices` keeps the other vendor's entries inside one `updateData`, so concurrent writes from both repos can't drop entries. `HwWalletId.derive`'s default stays `"trezor"`, so existing Trezor wallet ids are stable, and legacy entries deserialize with `vendor = TREZOR`. Bluetooth discovery now alternates vendors with `SCAN_INTERVAL` 2s → 4s, so Safe 7 discovery is slower — a deliberate trade against Android's scan-rate limit, not a defect.

Both greptile threads are genuinely fixed at head (`2dd3da77e`), with tests; I checked rather than taking the claim.

**Interaction with #1248** (`fix/receive-liquidity-parity`): no hidden semantic conflict, but expect a textual one. This PR renames `ReceiveTab.TREZOR` on the `onClickEditInvoice` line in `ReceiveQrScreen.kt` (~:383-390) while #1248 rewrites the two lines just below it, and similarly in `ReceiveSheet.kt` (~:156). #1248 adds no new `ReceiveTab.TREZOR` references, so once the conflict is resolved nothing compiles silently wrong. `ReceiveInvoiceEditStateTest.kt` hunks are disjoint. Whoever merges second should expect to resolve by hand rather than trusting a clean auto-merge.

`journeys/hardware-wallet/README.md` honestly scopes Jade as unit-test + manual-only, which is the right call given there's no Jade emulator.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
```diff
@@ -170,25 +175,36 @@ class HwReceiveViewModel @Inject constructor(
...
         verifyJob?.cancel()
         passphraseJob?.cancel()
         _uiState.update { it.copy(isPassphraseRequired = false) }
+        disconnect(walletId)
     }
 
-    private suspend fun handleVerifyFailure(error: Throwable) {
+    private fun disconnect(walletId: String?) {
+        walletId ?: return
+        viewModelScope.launch { hwWalletRepo.disconnectStaleSession(walletId) }
```
jvsena42, line 198: **LOW.** Merely *viewing* the Receive → Hardware tab now tears down a live Trezor session, because `walletId` is set by display rather than by use.

`ReceiveQrScreen.kt:256-259` calls `onLoadHardwareAddress` from a `LaunchedEffect` as soon as the tab is selected, and `loadAddress` sets `walletId` in the state. Closing the sheet then hits `DisposableEffect { onDispose(::cancel) }` (`ReceiveSheet.kt:103-104`) → `cancel()` → this `disconnect(walletId)` → `HwWalletRepo.disconnectStaleSession` → `TrezorRepo.disconnectStaleSession` → `trezorService.disconnect()` and `connected = null`. The device was never touched.

For a **passphrase** Trezor that's user-visible: `needsPassphrase` is `devices.any { it.passphraseProtected } && trezorRepo.state.value.connectedWalletId() != walletId`, so the next Send re-prompts for the passphrase although the session was live seconds earlier. `invalidateVerification()` at :186 also fires when the watcher rotates the receive address, so an incoming payment drops the session mid-sheet.

These three `disconnect(walletId)` calls came in with `dc6f266d7` "fix: harden jade sessions" and aren't vendor-scoped — master's `HwReceiveViewModel` has no `disconnect` at all.

**Why LOW rather than higher:** master already drops the Trezor session on every Send sheet close (`HwSendViewModel.kt:176-178`) and after transfers, so a session surviving across sheets was never the model — Receive was just the one sheet that didn't. Cost is one extra passphrase prompt, same prompt the user already gets after every Send. No fund risk, no stuck state.

**Fix**, mirroring what the Send sheet already does with `signingWalletId ?: return` — only disconnect when the device was actually engaged: set a flag in `verifyAddress()`/`submitPassphrase()` and make this a no-op otherwise, or gate on `vendorOf(walletId) == BLOCKSTREAM` if releasing the Jade's BLE link is the real goal. One boolean. Keep the timeout disconnect at :195-197 as-is.

If dropping the session on every Receive close is deliberate for both vendors, say so and I'll drop it — `cancel closes session` in `HwReceiveViewModelTest` currently enshrines it for all vendors.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -115,16 +135,107 @@ class HwWalletRepo @Inject constructor(
...
+            when {
+                jadeState.connected != null -> jadeRepo.disconnect()
+                jadeState.isConnecting || jadeState.isAutoReconnecting -> jadeRepo.cancelPendingConnection("")
+                else -> Result.success(Unit)
+            }
+        }
+        HwWalletVendor.BLOCKSTREAM -> {
+            val trezorState = trezorRepo.state.value
+            if (trezorState.connected != null || trezorState.isConnecting || trezorState.isAutoReconnecting) {
+                trezorRepo.cancelPairingCode()
```
jvsena42, line 207: **LOW.** `onTransportRestored` disconnects the other vendor *before* checking whether the target vendor has any known device — so a USB attach for a Jade that was never paired still kills a live Trezor session, including mid-signature.

The ordering: `AppViewModel.kt:5381` calls `hwWalletRepo.onTransportRestored(USB, vendor)` and only *then* checks `hasKnownDevice` at :5385. Inside, `disconnectOtherVendor(target)` runs unconditionally, reaching this block. Afterwards `jadeRepo.onTransportRestored` → `autoReconnect` → "No known devices" and nothing happens — but the Trezor is already gone.

This matters because the Jade USB IDs are necessarily generic: `usb_device_filter.xml:10` is `0x10C4:0xEA60`, the Silicon Labs CP210x default shared by ESP32 DevKits, Arduino clones and countless USB-serial dongles, and `:14` is `0x303A:0x1001`, Espressif's built-in USB-Serial/JTAG on every ESP32-S3/C3/C6. Jade genuinely uses these and Blockstream Green filters the same way, so the IDs aren't the fixable part — the unconditional disconnect is.

`signFunding` holds no `sessionMutex` (the mutex users are :140, :156, :217, :367, :393, :421, :445, :508), so a Trezor mid-signature can be disconnected. It's recoverable: `signTrezorFunding` fails, `isHwSessionFailure` triggers one `ensureConnected` retry, and `isBroadcastUnresolved` is never set before broadcast, so no wedge and no fund risk.

**Why LOW:** Android shows a chooser and the user must pick Bitkit for that device (or have ticked "use by default" previously). The realistic trigger is a Jade+Trezor user plugging the Jade in mid-Trezor-sign, which is arguably the intended "one vendor at a time" behaviour.

**Fix:** reorder the two lines in `AppViewModel.kt:5381-5385` so `hasKnownDevice` is checked first, or early-return from `onTransportRestored` when `loadKnownDevices().none { it.vendor == target }`. No new machinery. I wouldn't add an in-flight-sign guard — that needs a flag and there's no repro.

Related, as context rather than a separate ask: `JadeRepo.hasKnownUsbDevice` is true if *any* USB Jade is paired, and the candidate list is every scanned serial device that already has permission — so a foreign CP210x board can be handed to `jadeService.connect`, putting bitkit-core's CBOR parser on arbitrary device bytes. Fund-safe (identity is xpub-bound, mismatch throws `JadeIdentityMismatchError`), but a Rust panic across UniFFI aborts the process. Not worth acting on without a demonstrated panic.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -93,10 +105,18 @@ class HwWalletRepo @Inject constructor(
 
         /** Trezor v1 (2.4.0) tracks native SegWit accounts. */
         private val SUPPORTED_WATCHER_ADDRESS_TYPES = setOf(HwFundingAddressType.NATIVE_SEGWIT.settingsKey)
+
+        /** A Trezor reconnect is a session handshake; a Jade one may include entering the PIN on the device. */
+        private val TREZOR_RECONNECT_TIMEOUT = 30.seconds
+        private val JADE_RECONNECT_TIMEOUT = 5.minutes
```
jvsena42, line 112: **LOW.** A locked Jade can leave the Send sheet with no in-app exit for up to 5 minutes, against 30 seconds on the shipped Trezor path.

`HwSendViewModel.kt:226-231` wraps `ensureConnected` in `withTimeout(reconnectTimeout(walletId))`, which maps to this 5-minute constant for Jade and `TREZOR_RECONNECT_TIMEOUT = 30.seconds` otherwise. The Jade wait is a real block: `ensureConnected` → `unlockConnected()` suspends until the PIN is typed on the device or core's five-minute deadline expires.

Meanwhile every exit is closed. `ContentView.kt:473-474` computes `canDismissSheet` from `!isSigning && !isBroadcastUnresolved` and feeds it to the sheet host's `dismissEnabled`, blocking swipe and scrim; `HwSendSignScreen.kt:66` gates `onBackRequest` on the same, and both the `BackHandler` and the top-bar back use it. The only button on the screen starts signing. `cancel()` runs only on dispose, which those guards prevent. Since `sign` retries `ensureConnected` once on session failure, the theoretical ceiling is ~5m + 45s + 120s + 5m + 120s.

**This does not reintroduce the stuck-`isBroadcastUnresolved` wedge** — I checked specifically, since that one made every Send sheet undismissable until force-stop. `isBroadcastUnresolved = true` is still set only immediately before `broadcastFunding` and cleared on the first line of `handleFailure` and in `completeBroadcast`; the `cancel()` early return is byte-identical to master. The 5-minute wait happens in `ensureConnected`, before that flag can be set.

Escape hatches exist — enter the PIN, or unplug/power-cycle so `observeExternalDisconnects` fails the pending unlock and `handleFailure` clears `isSigning` — but neither is discoverable from a frozen sheet. The undismissable-while-signing design is pre-existing (master's `ContentView` guard is identical); this PR only lengthens the window for the new vendor.

**Fix**, two lines and no new control: gate `HwSendSignScreen.kt:66` and `ContentView.kt:474` on `isBroadcastUnresolved` alone, letting Back while `isSigning` run the existing `cancel()`, which already cancels `signingJob` and calls `disconnectStaleSession` (and `JadeRepo.cancelPendingConnection` closes the transport first, so the pending unlock unblocks). Cancelling mid-device-sign leaves the Jade on a confirm screen, same as Trezor today. Alternatively just lower the Jade budget on the Send path — `warmUp` pre-connects BLE, so a locked Jade at Send is the uncommon case.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/HwReceiveViewModel.kt
coreyphillips: Fixed in 65f2293fa. `HwReceiveViewModel` now tracks `engagedWalletId`, set only when `verifyAddress()` or `submitPassphrase()` runs. `cancel()`, `dismissPassphrase()` and `invalidateVerification()` all go through `disconnectEngaged()`, which does nothing if the sheet never used the device, so opening and closing the Hardware tab (or a watcher address rotation while only viewing) keeps a live Trezor session. The timeout disconnect is unchanged. `cancel closes session` is replaced by `cancel without using the device keeps its session`, and new tests cover a cancel after and during verification, plus an address change on a session the sheet never used.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
coreyphillips: Fixed in 65f2293fa. `HwWalletRepo.onTransportRestored` now returns before `disconnectOtherVendor` when `hwWalletStore.loadKnownDevices()` has no entry for the target vendor, so a generic CP210x or ESP32 attach with no Jade paired leaves the Trezor session alone. Tests: `usb attach for a vendor with nothing paired keeps the other vendor session` and `usb attach for a paired jade disconnects the trezor before reconnecting`. The check is per vendor, so with only a Bluetooth Jade paired, a foreign USB attach still goes through the existing Jade restore path. Agreed on leaving the CBOR-on-foreign-bytes point alone without a demonstrated panic.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
coreyphillips: Fixed in 65f2293fa, more narrowly than the suggestion, so Trezor stays undismissable mid-signature as it ships today. `HwSendUiState` has a new `isConnectingDevice`, true only while `ensureConnected` runs, and `canLeave = (!isSigning || isConnectingDevice) && !isBroadcastUnresolved`, used by both `ContentView` and `HwSendSignScreen`. Waiting for a PIN can now be left through the existing `cancel()`; compose, on-device signing and broadcast stay blocked. Two supporting changes: an attempt counter so a cancelled job that outlives `cancel()` cannot reset a newer attempt's state, and `JadeRepo.disconnectStaleSession` now closes the link before the core session, so the pending unlock fails immediately instead of holding the core queue until its deadline. Tests cover leaving while connecting, blocking while signing and while a broadcast is unresolved, a cancel during connect that never reaches compose, sign or broadcast, and the link-before-core ordering.

app/src/main/java/to/bitkit/repositories/JadeRepo.kt
```diff
@@ -0,0 +1,965 @@
...
+            if (!succeeded) cleanupFailedConnection(device.path)
+            if (connectingPath == device.path) connectingPath = null
+        }
+    }
+
+    private fun rejectUnusableDevice(version: JadeVersionInfo, expected: KnownDevice?) {
+        if (version.jadeState == JadeState.UNINIT) rejectDevice(HwDeviceUninitializedError())
+        val expectedHardwareId = expected?.jadeDeviceId
+        val hardwareId = version.efuseMac
+        if (expectedHardwareId != null && hardwareId != null && expectedHardwareId != hardwareId) {
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 7bc62d9f5. `rejectUnusableDevice` now drops the `hardwareId != null` condition, so a stored `jadeDeviceId` that does not match the reported `efuseMac` rejects the session whether the device reports one or not. The stored id is read through `takeIf { it.isNotBlank() }`, matching `deviceIdFor` and `advertisesAs`, so an entry that never carried an identity still connects. New test: `reconnecting a known jade rejects a device reporting no efuse mac` in `JadeRepoTest.kt`, with `knownUsb.jadeDeviceId` set and `efuseMac = null`, asserting the failure, the disconnect, and that accounts are never exported.

app/src/main/java/to/bitkit/repositories/JadeRepo.kt
```diff
@@ -0,0 +1,965 @@
...
+    private fun observeTransportRestored() {
+        jadeTransport.transportRestored.onEach { launchTransportReconnect(it) }.launchIn(scope)
+    }
+
+    private suspend fun awaitSetup() {
+        initialize().getOrThrow()
+        isSetup.await()
+    }
+
+    private suspend fun loadKnownDevices(): List<KnownDevice> = runCatching {
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 7bc62d9f5. It was an oversight, not a carve-out: `loadKnownDevices` now uses `runSuspendCatching`, so a cancellation while the store is read propagates instead of being logged as an error and returned as an empty list. The `runCatching` at the old `:790` (the documented `withTimeout` case) and the one wrapping the non-suspend `closeAllConnections()` are unchanged.

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
```diff
@@ -171,22 +288,101 @@ class HwWalletRepo @Inject constructor(
...
+        val found = trezor.getOrDefault(emptyList()).map { it.toHwNearbyDevice() } +
+            jade.getOrDefault(emptyList()).map { it.toHwNearbyDevice() }
+        Result.success(found)
+    }
+
+    private suspend fun hasOpenBluetoothSession(): Boolean {
+        if (jadeRepo.state.value.connected?.transport == JadeTransportKind.BLUETOOTH) return true
+        val trezorId = trezorRepo.state.value.connected?.id ?: return false
+        return devicesForDeviceId(trezorId).any { it.transportType == TransportType.BLUETOOTH } ||
+            trezorId.startsWith("ble:")
```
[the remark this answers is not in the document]
coreyphillips: Fixed in 7bc62d9f5. New `ext/HwDevicePath.kt` holds one `BLE_PATH_PREFIX` plus `String.isBlePath()`, `String.bleAddress()` and `blePath(address)`. Both file-level constants are gone, `JadeTransport.isBlePath`/`blePath` and `TrezorTransport.isBleDevice` are gone, and the bare literal in `HwWalletRepo.hasOpenBluetoothSession` is now `trezorId.isBlePath()`. `TrezorTransport` is folded in as you suggested, so its seven `"ble:${...address}"` constructions and its `removePrefix` now go through the same declaration. A search for `BLE_PATH_PREFIX` reaches every site.

jvsena42:
@coreyphillips conflict

app/src/main/java/to/bitkit/repositories/HwWalletRepo.kt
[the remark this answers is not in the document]
coreyphillips: Moved in 7e81da2f7 to follow your edit. The helpers now live in `models/HwDevicePath.kt` (package `to.bitkit.models`), next to `KnownDevice.kt` and `HwWalletId.kt`. `BLE_PATH_PREFIX`, `isBlePath()`, `bleAddress()` and `blePath()` are unchanged, and `HwWalletRepo`, `JadeRepo`, `JadeTransport` and `TrezorTransport` import them from there.

coreyphillips:
@jvsena42 Resolved in 44562df13, which merges current master. The only conflict was in `CustomTabRowWithSpacing.kt`: master moved the tab label into `TabLabel`, and this branch added `labelOverrides`. The resolution keeps master's layout and passes `labelOverrides[tab] ?: tab.uiText` as the label text. Compile, unit tests and detekt pass.

jvsena42 (commented):
Delta since my last pass (bf7745847, 7bc62d9f5, 7e81da2f7): one LOW observation inline, which I traced statically and haven't verified on hardware.

Checked:
- **Earlier threads.** ovi-reviewer's three threads are answered. `rejectUnusableDevice` now fails closed on a null efuse MAC while tolerating legacy blank ids, `loadKnownDevices` uses `runSuspendCatching`, and the BLE prefix is shared. My three earlier LOWs hold at head: `engagedWalletId` gating, the `onTransportRestored` early return, and `canLeave`.
- **`disconnectNoticeMutex`.** No deadlock: the notice path waits only on core `lifecycle`, and connect takes the mutex before `lifecycle`. The leftover race window needs a same-path reconnect within milliseconds of a GATT drop, and `retryAutoReconnect` waits at least 2s.
- **bitkit-core 0.5.14 → 0.5.17.** Adds the `Blockstream` vendor, catalog entries and the `finalize_psbt` sighash-type check (strictly safer, and Trezor doesn't use it). The backup migration change is rustfmt in tests only, and `jade:<hash>` ids survive it. No storage changes. The pinserver host is pinned to `jadepin.blockstream.com`.
- **Sign path.** Identity check → `check_signable` (ALL/DEFAULT sighash only, fingerprint must match) → `verify_signed_psbt` → core `finalize_psbt` with `interpreter_check`. A swapped or re-seeded Jade either fails the wallet-id check or produces signatures that fail interpretation. Change is derived app-side, and a retry reuses the cached `signedTx`.
- **Upgrade from Trezor-only state.** `vendor` defaults to TREZOR, `ignoreUnknownKeys` is set, and vendor-scoped saves keep the other vendor's rows.
- **Cancel/Back.** Leaving Send during the PIN wait cancels and closes the link before core, `NonCancellable`. Cancelling the connect sheet resets state. The Receive sheet only disconnects when a device was engaged.
- **Gating.** Not gated: Settings, the Home suggestion and USB attach are all reachable in release.
- **iOS twin, https://github.com/synonymdev/bitkit-ios/pull/765.** Reviewed clean. The USB case below doesn't apply there.

app/src/main/java/to/bitkit/repositories/JadeRepo.kt
```diff
@@ -0,0 +1,980 @@
...
+    private fun KnownDevice.advertisesAs(name: String?): Boolean {
+        val suffix = jadeDeviceId?.takeLast(BLE_NAME_SUFFIX_LENGTH)?.takeIf { it.isNotBlank() } ?: return false
+        return name?.endsWith(suffix, ignoreCase = true) == true
+    }
+
+    private fun KnownDevice.isSameDevice(device: JadeDeviceInfo): Boolean = when (device.transport) {
+        JadeTransportKind.BLUETOOTH -> path == device.path || advertisesAs(device.name)
+        // A plugged-in Jade cannot be told from a paired one before connecting, so a paired USB Jade
+        // claims every serial device; a second one is added through the Add button, which offers it anyway.
+        JadeTransportKind.SERIAL -> transportType == TransportType.USB
```
jvsena42: **LOW (observation, static trace only): a second USB Jade looks unpairable while any USB Jade is paired, which contradicts this comment.**

The SERIAL branch claims every plugged serial device for the paired USB entry, so `scan()` drops it from `nearbyDevices` (:196). Connect Hardware then falls back to `hasKnownDevice(it.id, advertisedName = it.name)` (`HwConnectViewModel.kt:333-335`). For a USB Jade, `id` is a path the device was never connected at and `name` is the USB product string. That string never ends with the efuse suffix, so `matches` and `advertisesAs` are both false. The OS attach route is closed as well: `hasKnownUsbDevice` returns true whenever any USB Jade is paired (`JadeRepo.kt:476`, `AppViewModel.kt:5390`), and the restore reconnect then rejects the second device with `JadeIdentityMismatchError`.

Scenario: Jade A is paired over USB. Plug in Jade B, then Settings → Hardware Wallets → Add → Continue: Searching keeps spinning. Accepting the Android USB chooser shows nothing. The workarounds are pairing B over Bluetooth or removing A first. No funds are involved.

Fix: in the SERIAL branch, treat a serial device as known only when a stored entry matches its path (or serial number when available), and let `hasKnownUsbDevice` require an exact `matches(deviceId)`. If you've confirmed Add does offer B with two devices, disregard this and fix the trace in the comment.

coreyphillips: Confirmed, the trace holds and the comment was wrong. Fixed in 187fd196f.

**Add flow.** `isSameDevice` now lets a paired USB entry claim a serial device only at the path it was last connected at, so Jade B is offered in Settings → Hardware Wallets → Add. A replugged Jade A at a new path is offered too, and pairing it again refreshes its existing entry rather than adding one (`a replugged jade refreshes its stored entry instead of adding one`). New tests in `JadeRepoTest.kt`: `scan offers a usb jade plugged in away from the paired entry's path`, `scan does not offer a usb jade at the paired entry's path` and `pairing a second usb jade keeps the paired one`.

**Attach route.** I kept `hasKnownUsbDevice` lenient on purpose. Before connecting there is no way to tell A from B, and A's path is renumbered on every plug, so an exact `matches(deviceId)` would open the Found sheet for the paired Jade each time it is plugged in through the chooser (over Home or Receive, since only Send and a few other sheets block it). So with A paired, attaching B still opens nothing, and B is paired from Add. The KDoc now says this, and `a usb attach at a new path counts as known while a usb jade is paired` pins it. Happy to switch if you'd rather take the replug prompt.

One related limit I left alone: with two USB Jades paired, the silent restore after an attach only tries the first USB entry, so the other one connects on first use instead. Send and Receive verification go through `ensureConnected`, which picks the right entry, so nothing is blocked.

jvsena42 (commented):
Delta since my last pass (`187fd196f`): **no findings.**

- **The USB thread's scenario is fixed.** The `SERIAL` branch of `isSameDevice` (`JadeRepo.kt:898`) now claims a serial device only at the stored path. With A paired and B plugged in, `scan()` keeps B, and Add pairs it with `expected = null`, B's own efuse id and B's own xpubs. A's row survives because `isReplacedBy` needs an equal id. The three new tests cover offered, not offered, and both-kept.
- **Replug.** A replugged A at a new path refreshes its own entry instead of adding a second one.
- **Attach route.** It is unchanged on purpose. The KDoc and a test pin the lenient `hasKnownUsbDevice`. The replug-prompt question stays on its thread.
- **Sign path.** Untouched. The `ensureConnected` walletId check and the `signJadeFunding` identity check still gate every Jade signature.
- **Cancel/Back with two Jades attached.** I walked Intro, Searching, Found (idle and waiting on the PIN) and Paired. Each one ends in `resetState` → `cancelPendingConnection`, with transport teardown before core under `NonCancellable`. A's stored row is only written after unlock and export, so cancelling B's connect can't touch it.
- **Bounds.** USB permission prompts are one per candidate. Scans and silent reconnects never prompt, and reconnect retries stay capped at 4.
- **Vendor mix-up.** Nearby Jade entries carry `BLOCKSTREAM`, the USB re-scan resolves by exact path, and the other vendor is disconnected under `sessionMutex` before connect.

jvsena42:
@coreyphillips conflicts

coreyphillips:
@jvsena42 Resolved in 92c3bbaf1, which merges current master. Two conflicts, both in the tab row:

- `CustomTabRowWithSpacing.kt`: master changed the default `selectedColor` to `Colors.White`, this branch added `labelOverrides`. Kept both.
- `ReceiveQrScreen.kt`: master dropped the now redundant `selectedColor = Colors.White` at the call site, this branch added `labelOverrides` on the same lines. Kept master's removal and our `labelOverrides`.

Compile, unit tests (2831) and detekt pass.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -1183,12 +1190,13 @@ class TransferViewModel @Inject constructor(
     @Suppress("ThrowsCount")
     private suspend fun ensureHardwareConnected(walletId: String) {
         runCatching {
-            withTimeout(HW_RECONNECT_TIMEOUT) {
+            // A Jade reconnect may include entering the PIN on the device, so the budget is per vendor.
+            withTimeout(hwWalletRepo.reconnectTimeout(walletId)) {
```
jvsena42: **The Send-sheet fix from `65f2293fa` was not carried to the transfer screen, so a locked Jade can trap the user for up to 5 minutes.** LOW.

This timeout is now `reconnectTimeout(walletId)`, which for Jade is `JADE_RECONNECT_TIMEOUT = 5.minutes`. `isSigning` is set before `prepareSignedHardwareFunding`, so `isBusy` stays true while `ensureConnected` waits for the PIN on the device. On `SpendingHwSignScreen` that removes system Back (`BackHandler(enabled = state.isBusy) {}`), the top-bar Back and the drawer icon at once, and `onCloseClick` only fires when `feeSat == 0`. A session failure retries `ensureHardwareConnected`, doubling the window.

Scenario: a paired Jade is powered on but locked, the user reaches the sign step, then changes their mind. Nothing in the app responds until the PIN is entered, the device is power-cycled, or five minutes pass. The Trezor path capped this at 30 s, and `HwSendUiState.canLeave` fixed exactly this for the Send sheet.

Fix: mirror the Send fix — an `isConnectingDevice` flag on `TransferToSpendingUiState`, set around `ensureHardwareConnected`, and gate Back and the drawer on `isBusy && !isConnectingDevice`. `cancelHardwareTransfer` already closes the link before the core session, so leaving releases the pending unlock.

jvsena42 (commented):
Only LOWs, both inline. Delta since `187fd196f` is three master merges with no PR-side commits; the conflict resolutions match what you described, and every earlier thread is still fixed at head.

Checked and clean, traced through core and jade-client-rs:
- **Signing.** `check_signable` rejects any sighash other than ALL/DEFAULT and requires a BIP32 origin matching the device fingerprint; `verify_signed_psbt` pins `unsigned_tx`, the input and output counts and the UTXOs, and requires a gained signature. `finalize_psbt` re-pins the previous outputs, fails when any input lacks a valid signature, and re-checks the sighash types on the finalized satisfactions. What the device shows is the transaction that gets broadcast, and fewer signatures than inputs means nothing is broadcast.
- **Change.** Composed app-side from the stored account xpub with key origin; the only device-sourced value is the fingerprint, and a wrong one makes Jade sign nothing.
- **Identity.** An entry is replaced only when the walletKey matches, a second device has a different id, and a walletId is adopted only on an exact xpub-set match. A null efuse fails closed. Equal seeds across vendors stay separate through `vendorWalletKey`.
- **Transport.** The BLE link is bonded, the PIN is entered on the device and the pinserver exchange is encrypted end to end inside jade-client-rs. No PSBT, xpub, fingerprint, PIN or signed transaction reaches a log. The two USB permission actions are distinct, so the pending intents do not collide.
- **Mid-flow failure.** The signed transaction is cached for both Send and Transfer, so a retry never re-signs, and the uniffi async calls let a timeout drop the Rust future while `cleanupFailedConnection` closes the link under `NonCancellable`.
- **Trezor regressions.** The changes there are vendor-scoped store access, the fingerprint passthrough and helpers moved to `KnownDevice.kt`. Legacy rows default to `TREZOR`, so derived ids are unchanged, and `ReceiveTab.TREZOR → HARDWARE` is a rename of an enum that is never persisted.

Design: no `### Design` section or Figma link in the body, and the Jade illustration is a declared placeholder, so `N/A — no design available.` is the right line. No new `*Screen.kt`, so `docs/screens-map.md` is untouched.

Both findings likely apply to https://github.com/synonymdev/bitkit-ios/pull/765 as well — the reconnect trigger exists over BLE there too — but it is conflicting, so I did not compare the code.

app/src/main/java/to/bitkit/repositories/JadeRepo.kt
```diff
@@ -0,0 +1,985 @@
...
+            var version = jadeService.connect(
+                device.transport,
+                device.path,
+                requestUsbPermission = requestUsbPermission
+            )
+            Logger.info(
+                "Connected Jade '${device.path}' firmware '${version.jadeVersion}' state '${version.jadeState}'",
+                context = TAG,
+            )
+            rejectUnusableDevice(version, expected)
```
jvsena42: **A Jade restored with a different seed silently adds a second wallet and keeps the old, unsignable one.** LOW.

`rejectUnusableDevice` compares only the efuse MAC, and the next lines export accounts and call `addOrUpdateKnownDevice` on every unlocked connect, including a silent `autoReconnect`. When the fetched xpubs share nothing with the stored entry, `previous` is null and a new entry with a new walletId is persisted. `isReplacedBy` then returns false, because the clause that drops a re-seeded device relies on the Trezor device id regenerating on wipe — a Jade's efuse MAC never changes, so it can never fire.

Scenario: pair a Jade on seed A, wipe it, restore seed B, unlock it, then background and foreground Bitkit. A second wallet appears with no pairing step, and the seed-A tile stays, still shows its balance and still hands out receive addresses, while signing fails closed at `ensureConnected`. Both tiles are named "Jade". A user who no longer holds seed A can be given a seed-A address from an indistinguishable tile. The same applies to Jade's TEMP restore state, which `isUnlocked()` treats as unlocked.

Fix: in `connectDevice`, when `expected != null` and the exported xpubs share nothing with `expected.xpubs`, reject before `addOrUpdateKnownDevice`, so a reconnect fails closed like the efuse check and pairing seed B stays an explicit Add.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
coreyphillips: Confirmed, fixed in 28113c190 by mirroring the Send fix.

- `TransferToSpendingUiState` gains `isConnectingDevice`, set around `ensureHardwareConnected` (both the first connect and the retry after a session failure), and a `canLeave` of `!isBusy || isConnectingDevice`.
- `SpendingHwSignScreen` gates system Back, the top-bar Back and the drawer icon on `canLeave`. Leaving disposes the screen, so `cancelHardwareTransfer` runs and closes the link before the core session, which releases the pending unlock.
- Signing and a pending broadcast still block leaving: `isConnectingDevice` is cleared before compose and sign, and `cancelHardwareTransfer` still returns early while a signed transaction is cached.
- Like `HwSendViewModel`, an attempt counter keeps a cancelled job's `finally` from resetting the state of a transfer started after it, which leaving mid-connect makes reachable.

New tests in `TransferViewModelTest.kt`: `hardware sign screen can be left while the device connects` and `hardware sign screen cannot be left while the device signs`.

app/src/main/java/to/bitkit/repositories/JadeRepo.kt
coreyphillips: Confirmed, fixed in 3c51337ed.

`connectDevice` now checks the exported xpubs against the expected entry before `addOrUpdateKnownDevice`. When a reconnect targets a stored entry that holds xpubs and none of them come back, it fails with `HwWalletMismatchError` and nothing is stored, so the old tile is not joined by a silent second wallet. This runs on every unlocked reconnect, including a silent `autoReconnect` of a Jade in the TEMP state. The failed connect goes through the usual cleanup, and `connectExpectedDevice` moves on to the next USB candidate as it does for an efuse mismatch. Pairing seed B stays an explicit Add, where `expected` is null. An entry with no stored xpubs is left alone, since there is nothing to compare against.

New test in `JadeRepoTest.kt`: `reconnecting a jade restored with another seed adds no wallet`.

jvsena42 (commented):
Delta since `1dd08c6e2` (`28113c190`, `3c51337ed`): **no findings.** Both LOWs are fixed.

- **Leaving while the device connects.** `isConnectingDevice` is set around `ensureHardwareConnected` only, and `canLeave = !isBusy || isConnectingDevice` now drives Back, the top bar and the drawer. The window is safe on the retry path too: `signHardwareFunding` only reconnects after a failed sign attempt, so `pendingHwFundingBroadcast` is still null there, and `cancelHardwareTransfer` proceeds normally. Once the device is asked to sign, or a broadcast is pending, `canLeave` is false again.
- **Attempt counter.** `hwTransferAttempt` stops a cancelled job's `finally` from resetting a newer attempt's state, and the same guard protects the `isConnectingDevice` reset.
- **Re-seeded Jade.** `rejectOtherWallet` runs before `addOrUpdateKnownDevice`, so a reconnect whose exported xpubs share nothing with the stored entry throws `HwWalletMismatchError` and persists nothing. The retry loop treats it like the identity mismatch, so it does not spin, and pairing the new seed stays an explicit Add.

app/src/main/java/to/bitkit/viewmodels/TransferViewModel.kt
```diff
@@ -1182,14 +1196,23 @@ class TransferViewModel @Inject constructor(
...
+        try {
+            runCatching {
+                // A Jade reconnect may include entering the PIN on the device, so the budget is per vendor.
+                withTimeout(hwWalletRepo.reconnectTimeout(walletId)) {
+                    hwWalletRepo.ensureConnected(walletId).getOrThrow()
+                }
+            }.getOrElse {
+                it.rethrowIfCancellation()
+                if (it.isHwUserCancellation()) throw it
+                throw HardwareReconnectError(it)
```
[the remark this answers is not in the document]
coreyphillips: Agreed, fixed in 8383cfe89.

A new `hardware__wallet_mismatch` string says the device holds a different wallet than the one paired, and suggests connecting the paired wallet or adding this one as a new hardware wallet. `handleHardwareTransferFailure` checks the cause chain for `HwWalletMismatchError` before the reconnect branch, so the wrapped reconnect failure and the sign-time session check both show it instead of the disconnected message.

The Send sheet had the same gap in another form: its fallback showed the error's hardcoded English message. `HwSendViewModel.handleFailure` now shows the same string.

New tests: `onTransferToSpendingHwConfirm shows wallet mismatch when the device holds another wallet` in `TransferViewModelTest.kt` and `a device holding another wallet shows the wallet mismatch` in `HwSendViewModelTest.kt`.

coreyphillips:
Resolved the conflict in 6c08cfea7, which merges current master. The only conflict was in `TransferViewModel.kt`: master added `confirmLeavingAmountSats` to `TransferToSpendingUiState` and this branch added `canLeave` in the same place. Kept both.

PR #1238, opened 2026-09-09
chore: add nix flake dev shell

icota:
### Description

This PR adds a Nix flake dev shell that provides the full toolchain needed to build the app without Android Studio:

1. JDK 17, Android SDK platform 36, build-tools 36.0.0, platform-tools, and NDK 28.1.13356709 (matching `ndk_ver` in `Justfile`), composed via nixpkgs `androidenv` with `flake.lock` pinning the nixpkgs revision
2. Exports `ANDROID_HOME`, `ANDROID_SDK_ROOT`, `ANDROID_NDK_ROOT`, and `JAVA_HOME` so `just compile` / `build` / `test` / `lint` work as-is inside `nix develop`
3. Works around a NixOS quirk: AGP downloads an unpatched aapt2 binary from Maven that cannot run on NixOS, so the shell hook points `android.aapt2FromMavenOverride` in `~/.gradle/gradle.properties` at the SDK's patched aapt2 (idempotent, refreshes stale store paths, leaves custom overrides alone)

Emulator and system images are excluded to keep the closure small (~2 GB with the NDK); the flags to enable them are documented in the flake. A sandboxed `nix build` of the APK is not provided since Gradle needs network access for Maven dependencies, which would require vendoring the whole dependency graph.

Note: building still requires `GITHUB_ACTOR`/`GITHUB_TOKEN` in `.env` for the `com.synonym:*` GitHub Packages repositories, same as any other environment.

### Preview

#### Manual Tests
- [ ] **1.** On a machine with Nix → `nix develop` → `just build`: devDebug APKs are produced in `app/build/outputs/bitkit/devDebug`.
#### Automated Checks
- Verified locally inside `nix develop` on NixOS: `just compile` and `just build` succeed, producing all four devDebug APKs.
- CI: standard compile, unit test, and detekt checks run by the PR bot; no Kotlin sources changed.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1788881743,
+        "narHash": "sha256-2V9GZGvPfrNzxFozhI9dcqV+c3QdA8YZrvAAzqEB+dI=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "d6524aaca2ff07876657ae2b323f24be4874944b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,107 @@
+{
+  description = "Bitkit Android dev shell (JDK 17 + Android SDK 36/NDK r28b via androidenv)";
+
+  # Provides the full toolchain to run: just compile / just build / just test / just lint
+  #
+  # Usage:
+  #   nix develop          # enter the shell
+  #   just compile         # or: ./gradlew assembleDevDebug
+  #
+  # Notes:
+  #   - First `nix develop` downloads ~2 GB of SDK/NDK; first gradle run
+  #     downloads the wrapper + all maven deps.
+  #   - Emulator/system images are excluded to keep the closure small.
+  #     For `just test android` without a physical device, flip the two
+  #     flags in androidenv below.
+  #   - On NixOS, adb access to a physical device needs udev rules:
+  #       services.udev.packages = [ pkgs.android-udev-rules ];
+  #   - A fully sandboxed `nix build` of the APK is not provided: gradle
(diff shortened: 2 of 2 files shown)


ovitrif:
@icota Awesome idea!
You may need some private keys from me so the builds actually use the test notifications firebase setup lane (they're off but build still works thanks to a dummy empty keys json file)

icota:
This could potentially fix https://github.com/synonymdev/bitkit-android/issues/953

ovitrif:
> This could potentially fix #953

To fully address that, we may need this for the rust repos too, or some `.tools-version` or `mise.toml`, but there was also a concern about reproducibility that was addressed in synonymdev/ldk-node#115 with the add to VCS of cargo lockfile.

Still, might need again to add the same change to all repos which build rust deps consumed by bitkit.

jvsena42 (commented):
Checked the things that actually go wrong with a flake — **clean, no findings.**

- **Pinning is correct.** One input (`github:NixOS/nixpkgs/nixos-unstable`), pinned in `flake.lock` to rev `d6524aaca2ff…` with a matching `narHash`, and the lock's `original` block matches the flake exactly. The floating `nixos-unstable` ref is only the update target — `nix develop` resolves through the lock. `cmdLineToolsVersion`/`platformToolsVersion = "latest"` resolve against the locked nixpkgs, so they're deterministic too. Nothing is fetched from an unpinned or non-content-addressed source at eval time.
- **Toolchain matches CI exactly**, so no "works on my machine" divergence: JDK 17 against `java-version: '17'` in all seven workflows; NDK `28.1.13356709` against `Justfile:6` `ndk_ver` and `NDK_VERSION` in both release workflows; `platformVersions = ["36"]` against `compileSdk`/`targetSdk = 36` (`app/build.gradle.kts:183`, `:188`). Gradle 9.5.0 via the checked-in wrapper, AGP 9.3.0 — JDK 17 satisfies both, and the repo pins no JVM elsewhere (no `jvmToolchain`, no `org.gradle.java.home`).
- **Nothing breaks non-nix contributors.** The repo root had no `flake.nix`/`shell.nix`/`.envrc`/`.tool-versions`/`mise.toml` before this, so there's no competing toolchain spec, and `.gitignore` has no pattern that would leave `flake.lock` untracked and let it drift. Nothing Gradle, detekt or `just` reads picks up either file.

The `~/.gradle/gradle.properties` `aapt2FromMavenOverride` write is the one rough edge and greptile already has it inline at `flake.nix:93`, including the effect on unrelated Gradle projects — not repeating it.

jvsena42 (approved):
utAck

Issue #736, opened 2026-01-27
Generic error message for Lightning payment failures hides actual cause


piotr-iohk:
## What happened?

When sending Lightning payments to Phoenix wallet, payments failed with the generic error "Your instant payment failed, please try again". The actual causes varied by payment size but were never surfaced to the user.

**Observed behavior:**
- Attempted multiple payments to Phoenix wallet (111 sats, 1k, 10k, 15k sats)
- All failed with generic error: "Your instant payment failed, please try again"
- Eventually succeeded with a 5k sat payment
- Error message gave no indication of the actual failure reason

**Expected behavior:**
The error message should:
1. Distinguish between routing failures vs recipient rejections
2. Indicate when retries might help vs when the payment amount is the issue
3. Provide actionable guidance based on the specific failure type

## Root Cause Analysis (from logs)

**Two distinct failure modes were observed:**

### Failure Mode 1: Small payments (111-1000 sats) - Phoenix JIT rejection
- Phoenix wallet rejected with `TemporaryNodeFailure(0x2002)`
- LDK marked route hint as failed, couldn't find alternative route
- Final error: `ROUTE_NOT_FOUND`

```
09:35:04 - Initiated sending 111000 msat
Onion Error[from 02bd5206...: TemporaryNodeFailure(0x2002)]
PaymentFailed { reason: Some(RouteNotFound) }

09:35:23 - Initiated sending 1000000 msat
Onion Error[from 02bd5206...: TemporaryNodeFailure(0x2002)]
PaymentFailed { reason: Some(RouteNotFound) }
```

### Failure Mode 2: Large payments (10k-15k sats) - Routing liquidity issues
- Various intermediate routing nodes rejected with `TemporaryChannelFailure(0x1007)`
- LDK exhausted all retry attempts across multiple routes
- Final error: `RETRIES_EXHAUSTED`

```
09:59:24 - Initiated sending 10000000 msat
Onion Error[from 027e9dd6...: TemporaryChannelFailure(0x1007)]
Onion Error[from 02baa725...: TemporaryChannelFailure(0x1007)]
Onion Error[from 028d1db1...: TemporaryChannelFailure(0x1007)]
Onion Error[from 030ef510...: TemporaryChannelFailure(0x1007)]
PaymentFailed { reason: Some(RetriesExhausted) }

10:02:38 - Initiated sending 15000000 msat
Onion Error[from 02dd3fcd...: TemporaryChannelFailure(0x1007)]
[...]

PR #1239, opened 2026-09-09
feat: add subscription requests

ben-kaufman:
### Description

This PR adds subscription proposals to contacts, building on the payer flow merged in #1186.

- Create a daily, weekly, monthly or yearly subscription with an amount, name, optional description and optional public Pubky icon.
- Reuse Payment Request amount entry, recipient rows and expiry controls; send to one saved, privately linked contact.
- Distinguish queued from sent and pending from accepted; show creator subscriber/received-payment counts and support pending/active deletion without editing.
- Validate the wire-size limit before icon upload/enqueue, downsample selected images, and aggregate valid billing-period proofs without creating payer prompts for creators.
- Keep the newly merged screen map in sync and add one user-facing changelog fragment.

Discover, autopay and renewal UI are intentionally excluded. Icon hosting is public by design. Base is `master`; no unmerged payer dependency remains.

iOS counterpart: https://github.com/synonymdev/bitkit-ios/pull/736.

### Design

- [Intro](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-239943&m=dev)
- [Overview](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-239663&m=dev)
- [Create Subscription](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-240132&m=dev)
- [Choose Recipient](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-240091&m=dev)
- [Sent Proposal](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?node-id=46920-240219&m=dev)

### Pubky file safety

This branch uses paykit-rs v0.1.0-rc54. Every displayed Pubky URI, including a descriptor's Pubky `src`, uses the bounded Kotlin fetch with a 1 MiB successful-body limit before bytes cross the FFI boundary. Existing fixed display sizes, the 15% memory-cache limit and Coil's bounded dedicated disk cache remain in place. Custom subscription icons and public icon hosting remain unchanged.

The 1 MiB value caps each successful encoded body. It is not a total-memory or decoded-pixel limit. Paykit rejects oversized successful bodies while reading and detects declared-length truncation. HTTP error bodies may still be buffered inside the Pubky client before Paykit regains control. The request timeout limits duration rather than that error-body memory, transport and current-chunk buffers add overhead, and a close-delimited body can still return partial bytes if the connection closes early. Shared-avatar handling and ambiguous proposal-publication outcomes remain outside this dependency update.

### Preview

Create → recipient → sent → overview using test wallets, silent at 4× speed. The first three seconds of the accelerated source clip were trimmed to exclude the system photo picker. All attached media was inspected and shows only Bitkit, including its keyboard; no device home screen or other app is included.

The walkthrough predates the final recipient corrections. The first screenshot below shows the final timer, Contacts spacing, Paste inset and 52dp field height.

https://github.com/user-attachments/assets/3853d647-d468-48b3-b338-16c5cd1f9973

<details>
<summary>Final recipient, sent confirmation and overview screenshots</summary>

Final recipient field, timer, Paste inset and Contacts spacing
Subscription request sent confirmation
Subscriptions overview after sending proposal

</details>

#### Manual Tests

- [ ] **1.** Subscriptions → Create Subscription → Amount: set a positive amount, name and frequency → Choose Recipient: only one saved/private-linked contact can be selected; expiry is configurable → Propose Subscription: confirmation shows recipient, amount and frequency.
- [ ] **2.** Create Subscription → upload custom icon → Choose Recipient → back: draft remains intact; the same public icon appears on the receiving device.
- [ ] **3.** Receiver → Review & Subscribe → accept and manually pay: creator shows one subscriber/one received payment attributed to the payer, without a payer payment prompt on the creator.
- [ ] **4.** Creator pending or accepted subscription → Delete → Swipe To Delete: request expires at the receiver and payment history remains available.
- [ ] **5.** Device offline → Propose Subscription: failure must not claim Sent; restore connectivity and retry the retained draft successfully.
- [ ] **6.** regression: Payment Requests → Create Payment Request: shared amount entry, contact selection and expiry remain functional.

Live regtest creation/delivery/acceptance/manual on-chain payments, custom icons and paid cancellation passed in both directions with iOS. Offline error/draft retention/retry and pending deletion also passed. Live transport-queued flushing, Lightning, mainnet and production push were not tested.

#### Automated Checks

- Full suite: 2,485 unit tests passed with zero failures/skips on the current source, including `ScreensMapTest.kt` and seven `PubkyImageFetcherTest.kt` cases.
- `PubkyImageFetcherTest.kt` verifies direct and descriptor-target fetches forward the 1 MiB limit and preserves coroutine cancellation. Native HTTP streaming behavior remains covered upstream in Paykit. `PaykitSubscriptionProposalTest.kt` covers UTF-8/wire-size boundaries. `PaykitPaymentRequestRepoSubscriptionTest.kt` covers creator terms, queued delivery, oversize rejection before icon upload/enqueue, pending cancellation and duplicate/off-schedule proofs. No automated coverage was removed.
- Nine Compose tests passed: `CreateSubscriptionScreenTest.kt`, `SubscriptionsScreenTest.kt` and existing `CreatePaymentRequestScreenTest.kt` regression coverage. Includes all four frequency tabs, no-Discover empty state, compact 520dp confirmation layouts, loading state, single recipient, expiry, and truthful sent/queued copy.
- Production and Android-test Kotlin compilation, rc54 native debug-symbol resolution and detekt completed with the dev flavor. Detekt retains its upstream advisory configuration. The cohesive shared subscription test fixture has a non-blocking LargeClass advisory, reviewed explicitly.
- Final local review found no remaining actionable issues. The final UI changes reuse the existing Payment Request picker and shared subscription card; no changed-UI detekt findings remain.


app/src/main/java/to/bitkit/ui/screens/subscriptions/SubscriptionsScreen.kt
```diff
@@ -153,19 +155,22 @@ internal fun SubscriptionsContent(
...
     paymentsContent: @Composable () -> Unit,
 ) {
-    val proposals = subscriptions.filter { it.isProposalVisible(now) }
-    val active = subscriptions.filter { it.isActive(now) }
-    val expired = subscriptions.filter { it.isExpired(now) && acceptedAt(it.id) != null }
-    val hasVisibleSubscriptions = proposals.isNotEmpty() || active.isNotEmpty() || expired.isNotEmpty()
+    val proposals = subscriptions.filter { it.isPayer && it.isProposalVisible(now) }
+    val active = subscriptions.filter { it.isPayer && it.isActive(now) }
+    val expired = subscriptions.filter { it.isPayer && it.isExpired(now) && acceptedAt(it.id) != null }
+    val created = subscriptions.filter { it.isCreatedVisible(now) }
```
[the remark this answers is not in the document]
ben-kaufman: Kept subscriptions with received payments in the Created list after deletion, so their payment history remains accessible. They show as expired, while canceled proposals without payments still disappear.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -419,6 +439,171 @@ class PaykitPaymentRequestRepo @Inject constructor(
...
+                mapOf(
+                    "note" to JsonPrimitive(name),
+                    "subscription" to JsonObject(subscriptionMetadata),
+                )
+            ).toString(),
+            recurrence = PaykitPaymentRequestRecurrenceTerms(
+                every = 1u,
+                unit = draft.frequency.rawValue,
+                startsAt = timestamp,
+                anchor = timestamp,
```
jvsena42, line 654: Low, disclosed, but worth a decision: anchoring the billing grid at proposal time lets the first billed period be arbitrarily short.

`startsAt = anchor = proposal time`, but the payer may accept any time up to `proposalExpiresAt` — and `PaymentRequestExpiration` goes up to 30 days. `periodsThrough()` bills the period *containing* `acceptedAt` in full, so with a `Day` plan and a `Week` expiry, accepting 6d23h later buys a period with an hour left, and the next full charge falls due an hour after that. Two consented payments within the hour.

It is disclosed — `SubscriptionsScreen.kt:703-712` shows "First billing period ends {date}. Each period is charged in full." — and every charge needs its own consent, which is why this is LOW rather than higher. But this PR is the first Bitkit code that *generates* such terms, so it's the right place to decide.

Either set `startsAt` to the proposal's expiry so the grid begins after the acceptance window, or cap the expiry options relative to the billing unit (hide `Month` for monthly, `Week`/`Month` for weekly). A payer-side minimum-first-period rule would also work but is a cross-platform behaviour change.

Same line-for-line on iOS (`PaykitPaymentRequestService.swift:648-653`) — worth deciding once for both.

app/src/main/java/to/bitkit/repositories/PaykitSubscription.kt
```diff
@@ -283,33 +330,63 @@ internal fun PaymentRequestRecord.toPaykitSubscription(): PaykitSubscription? {
...
+        return@runCatching PaykitSubscriptionMetadata(null, emptyList(), null)
     }
     val description = subscription["description"]?.jsonPrimitive?.contentOrNull?.clean(1024)
     val benefits = subscription["benefits"]?.jsonArray.orEmpty()
         .take(8)
         .mapNotNull { it.jsonPrimitive.contentOrNull?.clean(160) }
-    PaykitSubscriptionMetadata(description, benefits)
-}.getOrDefault(PaykitSubscriptionMetadata(null, emptyList()))
+    val iconUri = subscription["icon_uri"]?.jsonPrimitive?.contentOrNull
+        ?.clean(512)
```
jvsena42, line 386: Low: the only validation on the payee-supplied icon is the `pubky://` prefix, so it can point at any key's blob.

The URI goes to Coil via `SubscriptionAvatar` → `PubkyImage` and renders in place of the counterparty's avatar on the review sheet, list rows and detail screen. A proposer setting `icon_uri = pubky://<someone-else>/pub/bitkit.to/…avatar` puts that person's picture on the subscription, and the payer's device fetches from that homeserver on every render.

After the prefix check, require the URI's host segment to match `counterparty` (`PubkyPublicKeyFormat.matches(...)`) and fall back to `null` — default icon or contact avatar — otherwise.

Same check on iOS (`PaykitSubscription.swift`, also prefix-only).

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
```diff
@@ -419,6 +439,171 @@ class PaykitPaymentRequestRepo @Inject constructor(
...
+        val reservedIconUri = PaykitSubscriptionProposal.reservedIconUri.takeIf { draft.iconBytes != null }
+        val preflight = buildSubscriptionProposal(draft, name, description, reservedIconUri, endpoints, validationDate)
+        PaykitSubscriptionProposal.validate(preflight)
+        val iconUri = draft.iconBytes?.let {
+            paykitSdkService.uploadProfileAvatar(
+                bytes = SubscriptionIcon.compress(it),
+                contentType = "image/jpeg",
+                expectedIdentity = expectedIdentity,
+            )
+        }
```
jvsena42, line 565: Low: the public icon blob is uploaded before the proposal is committed, so a later failure leaves an orphaned world-readable file.

`uploadProfileAvatar` publishes under `/pub/bitkit.to/…` and runs *before* the post-upload `validateProposalExpiration`/`validate(proposal)` and before `proposePaymentRequest`. Any throw after this point leaves the blob on the homeserver, and each retry uploads another. Peer discovery succeeding but `proposePaymentRequest` failing on an identity check or transport is the ordinary way in.

It also holds `operationMutex` across a network upload, blocking refresh/dismiss/accept for the duration.

Reordering so the upload is the last fallible step before `proposePaymentRequest` fixes both, or delete the blob on failure (the SDK exposes `path`). Worth considering doing the upload outside the mutex and re-validating identity afterwards.

(For the record, `reservedIconUri` is sized against the *staging* namespace, which is longer than mainnet `bitkit.to`, so that preflight bound is conservative — not a bug.)

jvsena42 (commented):
Reviewed for fund draining specifically. **No critical or high findings.**

This adds the *creator/payee* side and re-tags the existing payer logic with `isPayer` guards. Every path that moves sats is untouched and still needs a fresh user action per period (`acceptSubscriptionAndStartPayment` → send flow; due periods → `openIncomingPaymentRequest` → send flow). Nothing auto-pays, nothing pays a cancelled or dismissed period, and the persisted-before-cleared dismissal ordering from the earlier fix is intact (`dismissSubscriptionPayment`, `:289-322`).

I specifically checked for the three bugs confirmed on the earlier subscriptions work — **none repeat**: creation uses a callback rather than a `StateFlow`, and the proof/dismissal code is unchanged, so neither the proof-kept-forever wedge nor the queue-cleared-before-persist ordering is reintroduced.

Also verified clean: amount shown vs paid is still enforced by `acceptsPaymentAmount`; cross-identity is guarded by `expectedIdentity`/`generation` under `operationMutex` with the SDK re-checking identity in `uploadProfileAvatar` and `proposePaymentRequest`, and `publishCreatedSubscription` gating on `isCurrentState`; payee records can't leak into one-off history (`toPaykitPaymentRequest:1188` rejects `recurrence != null`); the notification scheduler, monthly-cost, proposals and accept paths are all `isPayer`-filtered; `runSuspendCatching` throughout the new suspend paths; `creationMutex.tryLock` guards double-submit.

Three LOW notes inline, all dev/QA-facing today (`isPaykitEnabled` default false). Two of them also apply to the iOS twin (https://github.com/synonymdev/bitkit-ios/pull/736).

app/src/main/java/to/bitkit/repositories/PaykitSubscription.kt
ben-kaufman: The cross-key icon is possible, but checking `counterparty` directly would also reject the creator's own icon on created records, where `counterparty` is the payer. A top-level host check also would not constrain a descriptor's `src`. Leaving icon handling unchanged here for the coordinated follow-up.

app/src/main/java/to/bitkit/repositories/PaykitPaymentRequestRepo.kt
ben-kaufman: A failed proposal can leave a public upload behind, and the upload holds the shared operation lock. Moving it later does not make upload and proposal publication atomic, and deleting on a failed return needs care when publication may have succeeded. Leaving this for the recoverable-publication SDK follow-up documented in the description.

jvsena42:
conflicts

jvsena42 (commented):
Re-reviewed at `70629e09`. **No new findings.**

**Attribution:** the delta since my last pass (`29cecb8a`) is exactly one commit — `70629e095 chore: merge master into subscription requests` — touching 34 files, of which **30 are byte-identical to master** (#1216, #1235, #1236 landing) and 4 differ only on master's side. Net: this PR authored zero new lines since I last looked, so I re-verified the PR's own subscription code at head against the fund-safety list rather than reviewing a delta.

One merge *interaction* I checked because it's the kind of thing that slips through: master's #1216 made `MethodId.fromRawValue(value, network = Env.network)` network-aware, but `PaykitSubscription.kt:313` still calls the one-arg form. The default parameter makes it behaviourally identical in production, and creator-side `acceptedPaymentEndpointIdentifiers` uses `MethodId.rawValue` = `rawValueForNetwork(Env.network)`, so proposer and payer agree on identifiers. Not a finding, but worth a glance if the default ever changes.

**Fund safety — the reason this PR gets the scrutiny it does.** The creator-side role split only *narrows* the payer paths rather than adding new ones: `requestsThrough` returns empty for payee (`:240`), `accept()` requires `it == subscription && it.isPayer && isProposalActionable` and pins the displayed value (`repo:690`), the scheduler is gated on `isPayer` (`scheduler:70`), and `restoredAcceptances` is filtered on `isPayer` (`repo:767`). No auto-start charge path exists on the creator side. Period arithmetic holds: `every == 0u` rejected at `:296`, `every.toLong() * index` stays inside `Long`, and the `plusX` calls are wrapped in non-suspend `runCatching` — correct usage. No new `ULong` arithmetic was introduced.

Also clean: nothing seed-derived written, sent or logged; `uploadProfileAvatar(expectedIdentity)` and `proposePaymentRequest(expectedIdentity)` both check live identity, and `publishCreatedSubscription` gates on `isCurrentState(generation, expectedIdentity)`; inbound proposal fields are all bounded (`description` 1024, `benefits` 8×160, `note` 256, `icon_uri` 512); every suspend path uses `runSuspendCatching`, with `runCatching` only on non-suspend helpers; `dismissSubscriptionPayment` persists before mutating in-memory state.

**Recorded, not filed — the creator-side ledger counts payer-asserted proofs.** `PaykitSubscription.kt:345-354` reads only `billingPeriod` and `paymentEndpointIdentifier`; the SDK's `proof` payload is never inspected and nothing correlates against LDK or on-chain receipts, so a subscriber can make "Payments: N" and N `+amount` rows appear on the creator's detail screen without paying. Both platforms apply the same billing-grid filter before counting a proof (Android `periodsThrough` at `PaykitSubscription.kt:352`, iOS `recurrence.contains(billingPeriod)` at `PaykitSubscription.swift:547-557`), and Android derives the rows on the fly rather than persisting them. One Android-specific consequence worth knowing: `isCreatedVisible` (`:216`) includes `|| paidPeriods.isNotEmpty()`, so a single bogus proof pins a cancelled subscription in the Created list.

This is the same pre-existing Paykit-wide trust model I recorded on the iOS twin (synonymdev/bitkit-ios#736) rather than filing against it — the SDK does stateless correlation only and delegates settlement to the caller. Not this PR's defect; this is just the first Android code to render the number. If it's ever addressed, the narrow interim is to keep `paidPeriods.isNotEmpty()` out of the delete-visibility rule, or label those rows as reported rather than settled.

**Status of my open threads:** the short-first-billing-period one (`PaykitPaymentRequestRepo.kt:576`) is **still unanswered** — head still has `startsAt = timestamp, anchor = timestamp` with `timestamp` = proposal time. The icon-namespace and orphaned-icon-blob threads you replied to today are both deferred with reasons I accept; I've left a note on the first with input from the iOS side for whenever the coordinated follow-up happens.

Gating unchanged: dev/QA-facing today.

app/src/main/java/to/bitkit/repositories/PaykitSubscription.kt
jvsena42: Accepted — not reopening this. Your reason holds (a bare `counterparty` host check breaks creator-owned records where `counterparty` is the payer, and wouldn't constrain a descriptor's `src` either).

Input for the coordinated follow-up, since the iOS twin landed its half this week: synonymdev/bitkit-ios#736 fixed the *display* side rather than the URI side. It renders the saved contact's display name plus a locally-derived truncated public key on the review route, both resolved from `subscription.counterparty` and never from `metadata` — so the proposer-supplied icon can stay unconstrained while the sheet still carries an identity the proposer can't forge.

Relevant because the Android review sheet currently has no identity text at all: `SubscriptionReview` renders money, `SubscriptionProviderCard` (avatar + `note` + frequency), the period date and the swipe. I checked, and that's pre-existing rather than something this PR changed — `SubscriptionReview` is byte-identical to master. What this PR did change is the avatar itself (`SubscriptionsScreen.kt:776`, master's `PubkyContactAvatar(profile = contact, …)` → `SubscriptionAvatar`), so a non-null `icon_uri` now suppresses the one contact-derived cue that was there.

So the two halves compound: unconstrained icon + no identity text. Either half alone is much weaker. Whichever way the follow-up goes, worth deciding both together rather than just the URI check.

jvsena42 (dismissed):
Sent proposal sheet his not respecting top padding

Image

jvsena42:
## Cross-platform test with iOS #736

Android `902bb335f` (`devDebug`, Pixel 9 emulator) against iOS `d2f0585f` (iPhone 17 sim, Debug without `E2E_BUILD` → staging homeserver + staging regtest). Fresh identities: Android **AA**, iOS **BB**, saved as each other's contacts. Times are UTC.

**Verdict: the Android side behaved correctly everywhere I could test it. The test IDs don't match iOS, which I'm treating as blocking, and the top-padding issue from my earlier review is still there.**

### Results against the QA list

| # | Test | Result |
|---|---|---|
| 1 | Create & propose | ✅ Received from iOS: Review & Subscribe opened on its own about 11 s after iOS sent (₿5,000, Weekly, first billing period correct). ✅ Android → iOS: the Sent screen shows BB, ₿1,000, Daily. `deliveryStatus` was `Sent`, but iOS never showed it (iOS blocker, details on #736). |
| 2 | Draft kept + icon | ✅ Amount, Daily, name, description and the custom icon all survive Choose Recipient → Back. Couldn't check the icon on iOS because of the iOS blocker. |
| 3 | Accept & pay | ✅ Paying an iOS-created subscription: Review & Subscribe → on-chain confirm (₿141 fee) → Subscribed. The overview shows Active 1 and a monthly cost of ₿21,667, which is correct for ₿5,000/week. The payment proof was queued and reached iOS about 25 s later, credited to AA. |
| 4 | Creator cancels | ✅ iOS's delete reached Android in about 32 s: the status changed to Expired, Cancel disappeared, and the payment row (BB · −₿5,000) was kept under Expired. |
| 5 | Offline / oversized | Not tested |
| 6 | Payment request regression | ✅ Create Payment Request → BB → ₿1,000 → Send → Sent. It never appeared on iOS (iOS blocker). |
| 7 | Keyboard | Not applicable on the emulator run |

### 🔴 Blocking: IDs that don't match iOS

These have to match, per the shared journey vocabulary. AGENTS.md makes Android the reference, so most of the fixes may land on iOS (#736). Please settle each one with that PR so both land together.

| Element | Android | iOS |
|---|---|---|
| Frequency tabs | `Tab-day/week/month/year` (`CustomTabRowWithSpacing.kt:58`) | `Tab-daily/weekly/monthly/yearly` (also changes with the device language on iOS) |
| Subscription amount step | Reuses `PaymentRequestAmountContent` (`CreateSubscriptionScreen.kt:151`), so it exposes `PaymentRequestAmount`, `PaymentRequestAmountContinue`, `PaymentRequestNumberPad` | `SubscriptionAmount…` prefix |
| Recipient contact row | `SubscriptionContact<pubky>`, no separator (`CreatePaymentRequestScreen.kt:490`) | `SubscriptionContact-<pubky>` |

With the amount step reusing the `PaymentRequest*` tags, a journey can't tell the subscription amount step from the payment-request one on Android. Worth deciding on purpose.

### Still open from my earlier review: top padding
It still reproduces at `902bb335f`. The **Sent proposal** sheet's drag handle and "Sent" title are drawn inside the status bar. The **Create Subscription** sheet has the same problem: its title sits at status-bar height, right under the clock.

### Other findings
- **The first review step doesn't name the sender.** `SubscriptionReview` → `SubscriptionProviderCard(subscription, contact)` (`SubscriptionsScreen.kt:702`) shows only the contact's avatar ("B") next to the subscription name. The sender's name first appears on the on-chain confirm step. iOS fixed the equivalent in `123ccf51` (saved-contact name on review).
- **Link state for the iOS blocker, seen from Android.** After iOS's link to AA got stuck ("Encrypted Link Handshake is still in progress"), Android kept calling `restore_encrypted_link` for BB and never started a new handshake. Removing and re-adding BB on Android didn't change that either; the SDK link state is independent of the contact list. There's no in-app recovery on either side. Full details and a repro are on #736.

### Emulator note
The icon picker on the emulator is Google Photos' picker: tapping a tile only selects it, and it isn't applied until **Done** is tapped. The app correctly launches a single, image-only `PickVisualMedia` (`GetContent` is only the pre-Android-13 fallback), so this isn't a bug. It's just a trap when automating.

ben-kaufman:
Applied the standard sheet height to both the Create Subscription and Sent proposal states, so their titles keep the same top clearance as the other creation steps. Android's existing test tags remain unchanged as the shared journey reference. Sender identity and encrypted-link recovery remain separate from this layout fix.

app/src/main/java/to/bitkit/ui/screens/subscriptions/CreateSubscriptionScreen.kt
```diff
@@ -0,0 +1,558 @@
...
+        } else {
+            gallery.launch("image/*")
+        }
+    }
+
+    Column(
+        modifier = modifier
+            .fillMaxSize()
+            .gradientBackground()
+            .navigationBarsPadding()
```
jvsena42, line 245: Medium, gated: this chain never applies `imePadding()`, so with the keyboard up the "Choose Recipient" button — the only forward action on the screen — is completely hidden and nothing scrolls to recover it.

**Reproduced on device**, PR-head CI build (`bitkit-dev-debug-188` from `8f1e6f7eb`) on a Pixel 9 emulator, 1080x2424. Tapped the Description field, `mInputShown=true`, uiautomator bounds at that moment:

- `SubscriptionChooseRecipient` `[42,2172][1038,2319]` — keyboard top edge is at ~y1540, so the button is entirely underneath it
- `SubscriptionDescription` `[42,1317][1038,1554]` — clipped at the keyboard edge
- nothing moved: the sheet did not shift up at all

Scrolling cannot help, because the button at `:327` is a **sibling** of the weighted `LazyColumn` (which closes at `:326`), not a child of it. The user has to dismiss the keyboard with Back before the screen offers any way forward, which reads as a dead end.

**Why `adjustResize` doesn't save it:** the manifest does set `android:windowSoftInputMode="adjustResize|stateHidden"` (`AndroidManifest.xml:86`), but `MainActivity.kt:111` calls `enableAppEdgeToEdge()` -> `enableEdgeToEdge()` (`ui/utils/EdgeToEdge.kt:10`), which sets `decorFitsSystemWindows=false`. Under edge-to-edge the window no longer resizes for the IME and each screen must consume the inset itself. `sheetHeight()` sizes from `LocalWindowInfo.current.containerSize` (`SheetHeight.kt:64`), which doesn't shrink either. I checked `SheetHost.kt`, `ContentView.kt` and `TextInput.kt` — none of them supply IME insets.

**Fix — one line, matching an existing structural twin in this repo.** `HwPassphrasePromptSheet.kt` has a byte-for-byte identical chain with the line present:

```kotlin
modifier = modifier
    .fillMaxSize()
    .gradientBackground()
    .navigationBarsPadding()
    .imePadding()          // <- :126
    .testTag("HwTransferPassphraseSheet")
```

So add `.imePadding()` after `.navigationBarsPadding()` here at `:244`. The fixed height is not a problem: `modifier` at `:241` already carries `Modifier.sheetHeight()` as the outermost modifier (passed at `:150`), and Compose applies modifiers outside-in — the container stays at H while the content region shrinks to H minus navbar minus IME, and the `weight(1f)` `LazyColumn` absorbs it. (The broken ordering would be `Modifier.imePadding().sheetHeight()` at the call site, which is not what I'm suggesting.) `ReceiveSheet.kt:122-131` does the same thing the other way round — `sheetHeight()` outer, `fillMaxSize().imePadding()` inner.

**Severity and scope.** Introduced by this PR — `CreateSubscriptionScreen.kt` does not exist on master, so this ships with the new screen; it is not a regression of the padding commit. Not blocking, because `isPaykitEnabled` defaults false (`SettingsStore.kt:42`) and is only reachable via Dev Settings — so this affects opted-in users on released builds rather than everyone. MEDIUM rather than LOW only because the obscured control is the screen's sole forward action; I'd not argue hard against LOW, and the one-line fix makes the distinction academic.

jvsena42 (commented):
Re-reviewed at `8f1e6f7eb`. **Both of my blockers are resolved — I've dismissed the stale `CHANGES_REQUESTED`.** One new MEDIUM inline, gated so not blocking.

**Top padding: fixed, and I verified it on device** rather than from the diff. Installed the PR-head CI build (`bitkit-dev-debug-188` from `8f1e6f7eb`) on a Pixel 9 emulator: the Create Subscription drag handle and title now sit clear of both the status bar and the Subscriptions top bar. The fix is also *complete* — all four creation steps get the standard height now: Details (`:150`), Amount (`:165`, pre-existing), Recipient (`:384`, pre-existing), Sent (`:204`). `SubscriptionProposalSent` correctly takes `modifier: Modifier = Modifier` and applies it first in the chain (`:433`, `:436`), matching `CreateSubscriptionDetails` (`:224`, `:241`).

One honest gap: I could **not** reach the Sent proposal sheet on device this run. The recipient step showed `SubscriptionRecipientUnavailable` with Propose disabled — no privately-linked contact available, which is the stuck encrypted-link state I documented on https://github.com/synonymdev/bitkit-ios/pull/736. So the Sent sheet is verified by code reading only: identical `sheetHeight()` application, and the layout cannot clip because it uses `BoxWithConstraints(weight(1f))` -> `verticalScroll` + `heightIn(min = maxHeight)` (`:452-456`), with `fillMaxSize()` after a fixed `height()` being a no-op.

**Test IDs: resolved on the iOS side**, which is where the fixes belonged since AGENTS.md makes Android the reference. Verified at `1833d00a` against the tags I read off the live emulator:

| Element | Android (on device) | iOS at `1833d00a` |
|---|---|---|
| Frequency tabs | `Tab-day` / `-week` / `-month` / `-year` | `TabItem($0, accessibilityIdentifier: "Tab-\($0.rawValue)")` with `case day/week/month/year` — matches, and now locale-independent since it keys off `rawValue` rather than the localized `description` |
| Amount step | `PaymentRequestAmount`, `PaymentRequestAmountContinue`, `PaymentRequestAmountField`, `PaymentRequestNumberPad` | `testIdentifierPrefix: "PaymentRequest"` — matches |
| Recipient contact row | `SubscriptionContact<pubky>`, no separator | hyphen dropped — matches |

**The new finding closes row 7 of my own QA table**, which I had logged as "Keyboard — not applicable on the emulator run". It was never exercised; it is now, and there is a real gap. Inline at `:244`.

Checked and clean, at head:

- **Cancel and Back paths of the creation flow, enumerated.** Details -> system Back hides the sheet and only `remember` state is lost; nothing remote or local exists yet. Amount -> Back and Recipient -> Back both return to Details with the draft intact (`:160`, `:174`). While a propose is in flight, scrim, drag and system Back are all blocked by `ContentView.kt:502` `dismissEnabled = !isCreatingPaymentRequest && …` (honoured by `SheetHost` at `:182` and `:191`), and `creationMutex.tryLock` (`repo:528`) blocks double-submit. Sent -> OK/Back just hides. An identity change or generation mismatch mid-proposal gives `wasPublishedToActiveState=false` -> sheet hidden plus a "queued" toast, with no local record created for a foreign identity (`AppViewModel:5061-5071`, `repo:583`, `:677`). An in-progress icon load gates both Choose Recipient and Propose (`:330`, `:379`).
- **Creator cancel/delete is remote-first**: `repo:1069 operation(current)` then `:1072 applySubscriptionRecordLocked`, inside `runSuspendCatching`. A `CancellationException` landing between them leaves only a stale cache — the SDK record store, not local JSON, is the source of truth and is re-read by `synchronizeLocked` (`:919`). Delete goes through the `SubscriptionCancel` confirmation route (`SubscriptionsScreen.kt:653-661`). There is no "crash resurrects a cancelled subscription" path here.
- **Upgrade from v2.5.0.** `PaykitSubscription` is derived, not persisted; the only persisted local state is `subscriptionAcceptedAt` / `dismissedSubscriptionPaymentIds`, and the repo diff touches no persistence code. The new model fields all default (`role = Payer`, `iconUri = null`, `deliveryStatus = null`), and role comes from the SDK's `localRole` with `UNKNOWN`/null mapping to null — the record is dropped rather than mis-assigned a role. No SDK bump: paykit-android `0.1.0-rc51`, same as `v2.5.0`.
- **Fund safety, key material and trust boundaries are unchanged** since my `70629e09` pass — this commit is layout-only and introduces no new state, suspend calls or arithmetic.

I also corrected a line in my 2026-09-10 review: I'd written that Android was stricter than iOS on the billing-grid check. Both platforms apply the same filter — Android `periodsThrough` at `PaykitSubscription.kt:352`, iOS `recurrence.contains(billingPeriod)` at `PaykitSubscription.swift:547-557`.

app/src/main/java/to/bitkit/ui/screens/subscriptions/CreateSubscriptionScreen.kt
ben-kaufman: Added the standard IME inset to the Create Subscription details screen, so the Choose Recipient button stays above the keyboard.

jvsena42: Confirmed fixed on device. Installed the PR-head CI build (`b62e0d402`) on the Pixel 9 and repeated the exact measurement from this thread — with `mInputShown=true` on the Description field:

| | before (`8f1e6f7eb`) | now |
|---|---|---|
| `SubscriptionChooseRecipient` | `[42,2172][1038,2319]` — under the keyboard | `[42,1352][1038,1499]` — above it |
| `SubscriptionDescription` | `[42,1317][1038,1554]` — clipped at the edge | `[42,1115][1038,1352]` — fully visible |

Keyboard top is ~y1540 on this device. The placement is the one I suggested and the ordering is right: `sheetHeight()` stays outermost via the call site at `:151`, so the container keeps its fixed height and the `weight(1f)` `LazyColumn` absorbs the shrink. `navigationBarsPadding()` then `imePadding()` doesn't double-count either, since the outer consumes the nav inset and the inner applies only the excess.

One gap this didn't cover, which I've raised separately on `CreatePaymentRequestScreen.kt:415`: the **Recipient** step has the same exposure and still reproduces — `SubscriptionPropose` stays at `[42,2172][1038,2319]` with the keyboard up. Same one-line shape, different file.

jvsena42 (commented):
Re-reviewed at `b62e0d402`. **The keyboard fix works — verified on device — but it covers only the Details step.** The Recipient step has the same exposure and still reproduces. One MEDIUM inline, gated so not blocking.

**Details step: fixed.** Installed the PR-head CI build on a Pixel 9 and repeated the measurement from my earlier thread. With `mInputShown=true` on the Description field, `SubscriptionChooseRecipient` moved from `[42,2172][1038,2319]` (under the keyboard, top edge ~y1540) to `[42,1352][1038,1499]`, and `SubscriptionDescription` from `[42,1317][1038,1554]` to `[42,1115][1038,1352]` — fully visible instead of clipped. The commit is exactly two lines, the import at `:20` and `.imePadding()` at `:246`, placed inside the Column's chain as suggested.

**Recipient step: same bug, still there.** Same build, same device: tap `SubscriptionRecipientSearch`, keyboard opens, and `SubscriptionPropose` does not move — `[42,2172][1038,2319]` both before and after, entirely under the keyboard. Details inline on `CreatePaymentRequestScreen.kt:415`.

I checked the other two steps and they are genuinely not exposed, so this is the last one:

- **Amount** (`:153-167`) uses the custom `NumberPad` (`CreatePaymentRequestScreen.kt:185`); `NumberPadTextField.kt` imports nothing from `foundation.text` or `material3` and contains no `TextField`, so it never raises the IME.
- **Sent** (`:198-207`) has no text input, and its chain at `:437-444` is untouched.

Checked and clean:

- **The fix's ordering is correct and does not break the closed-keyboard case.** `modifier` is the head of the chain at `:242` and the call site at `:151` passes `Modifier.sheetHeight()`, so the effective order is `sheetHeight() -> fillMaxSize() -> … -> navigationBarsPadding() -> imePadding() -> padding()` — fixed height *outside* the IME padding, which is the safe direction. `imePadding()` is inset-driven, so it pads by zero when the keyboard is hidden; nothing is pushed off the sheet. `navigationBarsPadding()` then `imePadding()` does not double-count, because the outer consumes the nav inset and the inner applies only the excess. `SheetHost` (`BottomSheetScaffold`, `sheetPeekHeight = 0.dp`, no `windowInsets`) supplies nothing, so there is no third application. This is byte-identical to the in-repo twin `HwPassphrasePromptSheet.kt:121-127`.
- **No scope creep.** The diff since `8f1e6f7eb` is two files; the `.github/workflows/ci.yml` hunk is master's #1241 arriving through the merge commit, not the author's change (`git diff master refs/remotes/pr/1239 -- .github/workflows/ci.yml` is empty). The propose path, `creationMutex.tryLock` (`repo:528`), the `dismissEnabled = !isCreatingPaymentRequest` guard (`ContentView.kt:502`) and draft preservation on Back (`:161`, `:175`) are all zero-line diffs.
- **Conventions on the touched code**: `modifier` is the first optional parameter (`:225`, `:435`) and passed last with no trailing comma (`:151`, `:166`, `:205`, `:386`, `:556`).
- **CI is green** on this head.

No test was added, and I am not asking for one — nothing in the repo's androidTest asserts IME layout, and a Compose-rule test cannot drive a real IME. This stays a device check.

app/src/main/java/to/bitkit/ui/screens/paymentrequests/CreatePaymentRequestScreen.kt
```diff
@@ -406,10 +414,11 @@ internal fun PaymentRequestRecipientContent(
             .gradientBackground()
             .navigationBarsPadding()
```
jvsena42, line 416: Medium, gated: the Recipient step needs the same `.imePadding()` the Details step just got — "Propose Subscription" is entirely under the keyboard here.

**Reproduced on device**, same PR-head CI build (`b62e0d402`) and same Pixel 9 I used to confirm the Details fix. Tap `SubscriptionRecipientSearch`, keyboard opens (`mInputShown=true`), and `SubscriptionPropose` bounds are **unchanged**: `[42,2172][1038,2319]` before and after, with the keyboard top at ~y1540. The Details step's button moved to y1352 in the same session, so this is specifically the step the fix missed.

Selecting a contact does not get you out of it: `onSelected = { selectedTarget = it }` (`CreateSubscriptionScreen.kt:177`) only marks the row, and `PubkyContactRow` never clears focus or hides the keyboard — so the keyboard stays up and the contact rows below the keyboard edge stay unreachable, because the weighted `LazyColumn` viewport extends under the IME. Only the Back key recovers.

Structurally identical to the Details bug: this chain (`:411-418`) has `.fillMaxSize().gradientBackground().navigationBarsPadding().padding(horizontal = 16.dp).testTag(...)` and no `imePadding()` — zero hits for it in this file — while `TextInput` is an M3 `TextField` (`TextInput.kt:71`) that does raise the IME, and `footer()` at `:495` is a sibling after the weighted `LazyColumn` at `:460`.

**This is a regression introduced here, not pre-existing.** `PaymentRequestRecipientContent` does exist on master, but its only previous host was `ReceiveSheet.kt:122-131`, which wraps it in `Column.fillMaxSize().imePadding()` and so supplies the inset for it — and there its `onSelected` navigates immediately (`ReceiveSheet.kt:215-217`), so there was no footer button to hide. This PR is the first to host it directly under `SheetHost` (`ContentView.kt:559`), which supplies no insets, *and* the first to put a required action in the `footer` slot.

**Fix:** add `.imePadding()` after `.navigationBarsPadding()` on this line, mirroring `CreateSubscriptionScreen.kt:246`. Safe for the ReceiveSheet host too — its outer `imePadding()` at `ReceiveSheet.kt:130` already consumes the inset, so the inner one applies zero excess. (`Modifier.sheetHeight().imePadding()` at `CreateSubscriptionScreen.kt:386` would also work, but the shared-content line covers both hosts and matches the fix you just shipped.)

Not blocking: `isPaykitEnabled` defaults false (`SettingsStore.kt:42`) and is Dev-Settings-only, so this affects opted-in users on released builds.

ben-kaufman: Added the standard IME inset to the shared recipient screen, so Propose Subscription stays above the keyboard.

jvsena42: Confirmed fixed on device, and the shared-host regression I flagged as safe is actually safe — I checked rather than assumed.

Installed the PR-head CI build (`c0caca6a9`) on the Pixel 9. Subscription Recipient step, tapping `SubscriptionRecipientSearch` with `mInputShown=true`:

| | before (`b62e0d402`) | now |
|---|---|---|
| `SubscriptionPropose` | `[42,2172][1038,2319]` — under the keyboard | `[42,1352][1038,1499]` — above it |

**The nested-`imePadding()` case is clean.** This composable is also hosted inside `ReceiveSheet`, which already applies `.imePadding()` at `ReceiveSheet.kt:130`, so the new line sits inside an existing one. Measured it via Receive → Request from contact, keyboard up: `ReceiveScreen` bottom goes `2424` → `1541` and `PaymentRequestRecipient` goes `2361` → `1541` — i.e. the inner container lands exactly on the keyboard edge. Double-counting would have shrunk it twice, to roughly y657. `windowInsetsPadding` consumes the inset for descendants, so the inner call applies zero excess, which is what the bounds show.

That closes the keyboard issue across all four steps. Nothing further from me.

jvsena42 (commented):
Re-reviewed at `c0caca6a9`. **Fixed, verified on device, and the shared-host regression risk checked rather than assumed. Nothing further from me on this PR.**

The commit is the one-line `.imePadding()` at `CreatePaymentRequestScreen.kt:417` plus its import — exactly the shape I suggested, in the shared content composable so it covers both hosts.

**Measured on the PR-head CI build** (Pixel 9, `mInputShown=true`): `SubscriptionPropose` moved from `[42,2172][1038,2319]` to `[42,1352][1038,1499]`, i.e. from fully under the keyboard to above it. That was the last of the four creation steps — Details was fixed in the previous commit, and Amount and Sent have no IME exposure (the Amount step uses the custom `NumberPad`, and Sent has no text input).

**The nested-`imePadding()` case is clean.** Worth stating explicitly because I asserted it was safe when I proposed the fix, and an assertion is not evidence: `PaymentRequestRecipientContent` is also hosted inside `ReceiveSheet`, which already applies `.imePadding()` at `ReceiveSheet.kt:130`, so the new line now sits inside an existing one. Going Receive → Request from contact with the keyboard up, `ReceiveScreen` bottom goes `2424` → `1541` and `PaymentRequestRecipient` goes `2361` → `1541` — the inner container lands exactly on the keyboard edge. Double-counting would have shrunk it twice, to roughly y657. `windowInsetsPadding` consumes the inset for descendants, so the inner call contributes zero excess, and the bounds confirm it.

Also checked: the diff is that one file and two lines, so nothing else moved.

jvsena42 (dismissed):
✅ reviewed and cross tested with iOS PR

docs/pubky.md
```diff
@@ -37,10 +37,10 @@ The auth handshake uses a relay-based protocol:
 
 Wraps two FFI libraries:
 
-- **paykit-ffi** (`com.synonym:paykit-android`) — session management
-  - `paykitInitialize()`, `paykitImportSession()`, `paykitSignOut()`, `paykitForceSignOut()`
-- **bitkit-core** (`com.synonym:bitkit-core-android`) — auth relay, profile/contacts fetching, and file fetching
-  - `startPubkyAuth()`, `completePubkyAuth()`, `cancelPubkyAuth()`, `fetchPubkyProfile()`, `fetchPubkyContacts()`, `fetchPubkyFile()`
+- **paykit-ffi** (`com.synonym:paykit-android`) — session management and bounded file fetching
+  - `paykitInitialize()`, `paykitImportSession()`, `paykitSignOut()`, `paykitForceSignOut()`, `fetchPubkyFileBounded()`
+- **bitkit-core** (`com.synonym:bitkit-core-android`) — auth relay and profile/contacts fetching
```
jvsena42: Low: this bullet is stale in a way that matters for exactly the next person who touches the bound.

The commit correctly moved `fetchPubkyFile()` -> `fetchPubkyFileBounded()` up to the paykit-ffi bullet, but left bitkit-core credited with "auth relay and profile/contacts fetching" plus five function names the app does not call. Checked two ways: `git grep` for `startPubkyAuth|completePubkyAuth|cancelPubkyAuth|fetchPubkyContacts` across `app/src/main` returns nothing, and the only `com.synonym.bitkitcore` import in `PaykitSdkService.kt` is `mnemonicToSeed` (`:4`). Profile fetching is on the paykit handle — `handle().fetchPubkyProfile(publicKey)` at `PaykitSdkService.kt:416`.

Why it is worth fixing in this commit rather than later: this section is the first place someone tightening the byte cap will look, and it currently implies a second, separate fetch path living in bitkit-core. They will either waste time looking for it or — worse — conclude there is an unbounded path still to be closed. There is not; the enumeration is clean, every counterparty-image byte goes through the one fetcher.

Narrowest fix: drop the Pubky function names from the bitkit-core bullet (or the bullet), and list `fetchPubkyProfile()` / `resolveContactProfile()` under paykit-ffi where they actually live.

jvsena42 (commented):
Re-reviewed at `7cef55a35`. **No HIGH or MEDIUM.** One LOW doc note inline. The bound is correct and the coverage is complete — traces below.

**It throws, it does not truncate** — the question I most wanted settled, because a silently truncated blob would be worse than the unbounded fetch it replaces. From the SDK at tag `v0.1.0-rc54` (`paykit-sdk/src/net.rs:10-36`), `read_bounded_body` rejects on `Content-Length > max_bytes` before reading anything, then checks `body.len() + chunk.len() > max_bytes` *before* appending each chunk — so the buffer is complete or absent, never partial, and the response is dropped. Oversize surfaces as `PaykitSdkError::Protocol`; a body that ends short of its declared length is a `Transport` error, not a success (`net.rs:138-149`). The effective cap is `min(max_bytes, 5 MiB)`, so a caller cannot raise it past the SDK ceiling.

I also checked the Kotlin binding rather than inferring it: pulling the rc54 AAR from GitHub Packages and running `javap` gives `fetchPubkyFileBounded-z13BHRw(String, long, Continuation<? super byte[]>) throws PaykitException, CancellationException`. So `PaykitSdkService.kt:376` genuinely sees an exception on oversize, which means the `?: throw AppError("Pubky file not found")` now only fires on a real 404 — it is no longer reachable for an oversized file.

**Coverage is complete: every counterparty-image byte goes through one fetcher.** `PaykitSdkService.fetchFile` <- `PubkyService.fetchFile` <- `PubkyImageFetcher` is the only path that fetches Pubky bytes, and `git grep` for the unbounded `fetchPubkyFile\b` at head returns no callers. All 20-odd `PubkyImage` / `PubkyContactAvatar` / `SubscriptionAvatar` call sites render through Coil's singleton `ImageLoader` (`App.kt:44`, `ImageModule.kt:29`), so they are all bounded by construction. The two raw `AsyncImage` uses (`EditProfileScreen:232`, `CreateSubscriptionScreen:313`) load the user's own picker `content://` URI, not counterparty data. SDK-internal profile resolution fetches JSON only, under a separate 64 KiB cap, and never dereferences `image_uri`. No hole.

**Failure degrades to the placeholder.** The throw reaches Coil's `EngineInterceptor`, which catches non-`CancellationException` throwables and returns `ErrorResult`, so `PubkyImage.kt:66` sets `ImageState.Error` and draws the gray `ic_user_square`. Same for a descriptor pointing at another descriptor, or for non-image bytes. A counterparty who publishes an oversized subscription icon gets the default glyph, not a broken screen.

**The descriptor indirection is depth-1 and totals at most 2 MiB.** `resolveImageData` (`:36-45`) parses once and returns the second fetch's bytes without re-parsing, so there is no loop; both legs pass `PUBKY_IMAGE_MAX_BYTES`. The namespace question stays deferred per your earlier thread and I am not reopening it.

**The `runCatching` -> `runSuspendCatching` change is real, and I proved the new test earns its place** rather than assuming it. In a throwaway worktree at this head with the rc54 AAR: head passes 7/7, and with the wrapper reverted to `runCatching`, `fetch should preserve cancellation when following file descriptor` **fails** — `Expected an exception of class java.util.concurrent.CancellationException to be thrown, but was completed successfully with the result: SourceFetchResult(...)`. That is exactly the bug: the old code swallowed cancellation and handed Coil the descriptor JSON as though it were an image. Worth noting Coil itself rethrows `CancellationException`, so the old wrapper was also defeating Coil's own cancellation. Working tree left as found, worktree removed.

**On the rc51 -> rc54 bump landing here too:** this PR needs nothing from rc54 beyond `fetchPubkyFileBounded`. I checked the one compatibility item — `PrivateStreamIntakeReport.receive_batch_id` becoming optional — and `git grep` for `receiveBatchId|PrivateStreamIntakeReport` across `app/src/main` finds no usages, so nothing reads it. The pin is textually identical to the one on https://github.com/synonymdev/bitkit-android/pull/1255, and `git merge-tree --write-tree` on the two heads merges clean, so the duplicate bump is not a conflict — whichever lands second just no-ops that line.

One small thing for the PR description rather than the code: it credits the bounded API to rc53, but there is no `v0.1.0-rc53` tag in `pubky/paykit-rs` (tags go rc50, rc51, rc52, rc54). The API is in rc54; the "bump to rc53" commit sits inside the rc54 range. Only worth mentioning so nobody goes looking for a tag that isn't there.

Parity with the iOS twin (https://github.com/synonymdev/bitkit-ios/pull/736, `3f0e83f19`): same 1 MiB limit, same throw-not-truncate failure mode, same 2/2 call-site coverage, same single-hop descriptor following, same ≤2 MiB per image, same rc54 pin. Two pre-existing divergences, neither introduced here — iOS rejects a cross-pubkey `src` and Android does not (that is the deferred namespace item), and iOS ignores cancellation where this commit now propagates it, so Android is the better-behaved side on that one. iOS also added explicit decode/pixel/cache caps; Android gets the equivalent from Coil's `maxBitmapSize` plus layout-driven downsampling, so I am not filing it.

**Gating:** `PubkyImageFetcher` is reachable only behind `isPaykitEnabled` — Profile and contacts routes sit inside `PaykitRouteGuard` (`ContentView.kt:1458-1468`), Home's avatar button is `showProfileButton = isPaykitEnabled` (`HomeScreen.kt:279`), activity contact avatars come from a view model that emits an empty contact list when disabled (`:91-93`), the send flow gates at `AppViewModel.kt:1937`, and the auth-approval sheet's registrar combines the flag (`PubkyAuthHandlerRegistrar.kt:41`). So this affects opted-in users on released builds, with no non-Paykit surface rendering it.

Also checked and clean: the constant is documented with `/** */` (`:20`); the `ULong` is passed through with no arithmetic, so no `USat` wrapper is needed; no `runCatching` around a suspend call remains in the file; `docs/pubky.md:118-123` describes the bound accurately, including the honest caveat that HTTP *error* bodies are still buffered unbounded inside the Pubky client upstream. No `*Screen.kt` was added or removed, so `docs/screens-map.md` needs nothing.

**Not verified:** end-to-end behaviour with a real >1 MiB icon served from a homeserver. I verified the SDK error path from source, the Kotlin binding from the AAR, and the UI fallback from code, but did not exercise the three together on a device. The e2e CI lanes were still pending when I looked; build, detekt and lint were green, which is what confirms the new binding compiles.

jvsena42 (approved):

jvsena42:
@ben-kaufman you have some commit with unverified signature

docs/pubky.md
ben-kaufman: Updated the Pubky integration docs to put Ring auth, profile/contact resolution, and bounded file fetching under paykit-ffi. bitkit-core is now listed only for the receiver noise-key seed conversion.

PR #1245, opened 2026-09-10
fix: skip push registration in e2e

### LINKED ISSUE #1244: fix: skip push registration in e2e
## Summary

The app registers every install for push notifications with the Bitkit notification server, including E2E builds whose device is wiped minutes later. Each wiped install leaves a dead FCM token that the LSP keeps pushing to.

## Scope

- Cross-platform: same behavior on Android and iOS, one issue per repo (counterpart linked below).
- Requested change: E2E builds skip the network registration with the notification server and log one line, keyed on the existing `E2E` build flag. Every other build keeps today's behavior.
- Out of scope: how the notification server handles dead tokens (synonymdev/bitkit-notification-server#19), and Blocktank's retry loop.

## Observed on staging

On 2026-09-10 the Blocktank staging ln2 watcher logged 210 `Failed to notify peer … Notification reject by Apple/Google` errors for 5 node ids, all E2E wallets. `WakeToTimeoutWatcher` retries an offline peer with a pending HTLC every 5 minutes for up to 14 days, and a provider reject does not stop it. No E2E spec exercises push notifications, so registration from those builds serves nothing.

## Why E2E builds can skip it

No E2E test uses push notifications: `grep -ri notification bitkit-e2e-tests/test/specs` finds nothing. The suite drives the app in the foreground, and receiving a payment while backgrounded is never asserted, so a registration from an E2E build is never consumed. The only effect of registering is the dead token the wiped device leaves behind, which Blocktank then retries against every 5 minutes for up to 14 days.

## Mechanism

- `LspNotificationsService.registerDevice` calls bitkit-core `registerDevice` against `Env.blocktankNotificationApiUrl` unconditionally and caches the token in the keychain.
- Callers: `LightningRepo.registerForNotifications` on node start, `FcmService.onNewToken`, and the DevSettings manual action.
- `E2E=true` is already set on exactly the builds that matter: the E2E APK steps in `.github/workflows/e2e.yml` and `e2e_migration.yml`, and `just e2e`. `ci.yml` and `ui-tests` never launch the app on a device; release builds never set it.

## Fix

- `LightningRepo.registerForNotifications` returns early with a single `Logger.info` line when `Env.isE2eTest` is set. All three callers go through it, so they skip together. FCM token retrieval, the keychain cache, and receive-side handling in `FcmService` are not reached and stay untouched.
- The Dev Settings manual action toasts `Skipped push registration` in E2E builds.
- No new flag, no Gradle or workflow change.

## Verification

- `E2E=true` build → create wallet → node start logs the skip line and no `Registering device for notifications…`.
- Default build → node start proceeds to registration as before.
- Existing lint and unit tests pass.
- Staging: after the E2E APK carries the change, no new node ids appear in the `WakeToTimeout` reject lines.

Counterpart: synonymdev/bitkit-ios#739.

ovitrif:
Closes #1244

This PR skips push notification registration in E2E builds so those wallets stop registering short-lived FCM tokens with the Bitkit notification server.

### Description

- `LightningRepo.registerForNotifications` returns early with one info log when `Env.isE2eTest` is set. It is the single path for node start, `FcmService.onNewToken`, and the DevSettings action, so all three skip together. The FCM token fetch and keychain cache are simply not reached; receive-side handling is untouched.
- The Dev Settings "Register For LSP Notifications" action toasts `Skipped push registration` in E2E builds.
- `E2E=true` is already used in the three E2E APK steps in `e2e.yml` and `e2e_migration.yml`, also in plus `just e2e`.

Limited scope change, nothing relevant of the happy flow is affected because the build flag is not exercised for those builds.

Context: [Blocktank staging thread](https://synonymworkspace.slack.com/archives/C03UWHGKUNP/p1789031337183729); server-side pruning of dead tokens is synonymdev/bitkit-notification-server#19.

Follow-up: iOS counterpart synonymdev/bitkit-ios#739.

### Design

N/A — no UI changes.

### Preview

#### Manual Tests

- [ ] **1.** `just e2e assembleDevDebug` → install → create wallet → node start: log shows `Skipped push registration in E2E build` and no `Registering device for notifications…` line.
- [ ] **2.** `regression:` `just build` → create wallet → node start: `registerForNotifications` proceeds to the FCM token fetch as before.
- [ ] **2.** `just e2e assembleDevDebug` → Settings → Dev Settings → Register For LSP Notifications: toast `Skipped push registration`.
#### Automated Checks

- Verified on a regtest emulator with the earlier flag-based build of this branch: the skip path logged on node start; the default build entered the registration path. The guard is the same, only the condition changed to `Env.isE2eTest`.
- No unit test: `Env.isE2eTest` is a compile-time constant, so the skip path is exercised by any `E2E=true` build, including every CI E2E run.
- `just build` and `just lint` pass on the current head.
- No changelog fragment: no user-facing change.


app/src/test/java/to/bitkit/repositories/LightningRepoTest.kt
```diff
@@ -959,6 +958,29 @@ class LightningRepoTest : BaseUnitTest() {
...
+
+    @Test
+    fun `registerForNotifications should skip device registration when push registration is disabled`() = test {
+        startNodeForTesting()
+
+        val result = sut.registerForNotifications(token = "fcm-token", isPushRegistrationEnabled = false)
+
+        assertTrue(result.isSuccess)
+        verifyBlocking(lspNotificationsService, never()) { registerDevice(any()) }
+    }
```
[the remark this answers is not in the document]
ovitrif: Test removed: the flag is now a plain BuildConfig constant with no injection seam, so the off path is verified by the emulator build documented in QA Notes.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -1670,7 +1670,13 @@ class LightningRepo @Inject constructor(
 
     fun separateTrustedChannels(channels: List<ChannelDetails>) = lightningService.separateTrustedChannels(channels)
 
+    @Suppress("KotlinConstantConditions")
     suspend fun registerForNotifications(token: String? = null) = executeWhenNodeRunning("registerForNotifications") {
+        if (Env.isE2eTest) {
+            Logger.info("Skipped push registration in E2E build", context = TAG)
+            return@executeWhenNodeRunning Result.success(Unit)
```
ben-kaufman, line 1677: Non-blocking nit: with a running node in an E2E build, `DevSettingsViewModel` still shows "Registered for notifications" after this returns success. Could the manual action show that registration was skipped instead?

ben-kaufman (dismissed):

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
ovitrif: Addressed in ab24334

ovitrif:
The reason for this PR is to reduce the load of push notifications tokens registrations (issue flagged by @dzdidi in thread about staging backend overload by bitkit ci).

The issue is also related to missing cleanup in our notification server, an issue flagged separately in that repo -- cleanup can be done on trigger of deterministic signal returned by the specific platform's notifications delivery server when trying to access it, the returned payload includes a certain error code which our backend can use to ack those devices are never gonna be available anymore in there, so they can be purged.

E2E tests never use those anyways but @piotr-iohk can db check to confirm.

ben-kaufman (approved):

jvsena42 (commented):
Reviewed as one question: does this change behaviour for anyone outside an E2E build? **No.** `Env.isE2eTest` is a compile-time constant, not a runtime toggle, and the release workflows never set it.

| Step | Location | Content |
|---|---|---|
| Gate | `LightningRepo.kt:1675` | `if (Env.isE2eTest) {` |
| Constant | `Env.kt:21` | `const val isE2eTest = BuildConfig.E2E` |
| BuildConfig field | `app/build.gradle.kts:323` | `buildConfigFields.put("E2E", e2eEnv.booleanField())`, inside `androidComponents { onVariants { … } }` — the same value is stamped into every variant, and there is no per-flavor override in `productFlavors` (`:201-222`) |
| Source | `app/build.gradle.kts:77` | `val e2eEnv = envFlag("E2E", default = false)` |
| Parser | `app/build.gradle.kts:63-67` | true only when the `E2E` env var is literally `"true"` at Gradle time |

`E2E=true` is exported in exactly four places — `Justfile:119`, `Justfile:200`, `.github/workflows/e2e.yml:76,126`, `.github/workflows/e2e_migration.yml:56`. `release.yml:70-79` and `release-internal.yml:69-78` pass only `GPR_*`, `GITHUB_TOKEN` and the keystore vars before `assembleMainnetRelease bundleMainnetRelease`, and the workflow-level `env:` is `TERM`/`FORCE_COLOR`/`NDK_VERSION`. So `BuildConfig.E2E = false` in mainnet and the new branch is dead code there — which is exactly why `@Suppress("KotlinConstantConditions")` is needed at `:1673`.

Also checked and clean:

- **No runtime toggle, no stuck state.** `DevSettingsViewModel.kt:173-175` only reads the same constant to pick a toast title; it writes nothing. There is no persisted "registration disabled" flag that a real user could get wedged into.
- **The skip does not poison retry.** It returns `Result.success(Unit)` *before* the `runCatching`, so `Keychain.Key.PUSH_NOTIFICATION_TOKEN` — written in `LspNotificationsService.kt:63` from inside `registerDevice()` — is never written on this path. All three callers are fire-and-forget or toast-only (`LightningRepo.kt:416`, `FcmService.kt:158`, `DevSettingsViewModel.kt:171`), so nothing caches "registered" and stops retrying.
- **Skipped is distinguishable from failed**: `:1676` info, `:1687` debug (already registered), `:1693` error — three separate messages.
- **The wake path is untouched.** `WakeNodeWorker`, `LightningNodeService` and `NotifyPaymentReceived` are not in the diff, so in non-E2E builds the registration that feeds them runs exactly as before. This was the failure mode worth ruling out: a real user whose node never wakes for an incoming payment.

One caveat for the record, not a finding and not introduced here — because it is an env var rather than a flavor constant, `E2E=true ./gradlew assembleMainnetRelease` run locally would ship the skip. That is already true of every other `Env.isE2eTest` consumer (`Env.kt:147`, `QrScanningScreen.kt:342`, `BackupSettingsScreen.kt:87`).

PR #1246, opened 2026-09-10
chore: bump version 2.5.0

piotr-iohk:
Release 2.5.0.

### Description

- Bump `versionCode` to 189 and `versionName` to 2.5.0
- Collect changelog fragments from `changelog.d/next/` into `CHANGELOG.md` under `[2.5.0] - 2026-09-10`

### Design

N/A — no UI changes.

### Preview

N/A

### QA Notes

N/A — version bump and changelog only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,52 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.5.0] - 2026-09-10
+
+### Added
+- Added receiver-specific Paykit contact support for Bitkit wallet and server integrations. #1066
+- Bitkit now creates, names, backs up, and manages separate watch-only Bitcoin accounts and securely delivers signed setup claims to Paykit servers. #1084
+- Incoming Paykit payment requests can now be reviewed and approved through the existing payment flow. #1098
+- Developers running their own build can now open supported screens directly with `bitkit://screen/...` links while dev mode is enabled. #1119
+- Passphrase-protected (hidden) Trezor wallets can now be paired from the connect flow, each appearing as its own watch-only balance with its own label, activity and removal, and asking for its passphrase again when a transfer needs signing. #1142
+- Incoming Paykit payment requests now stay discoverable until handled, and users can send new private requests to connected contacts from the invoice flow. #1172
+- The name you give a hardware wallet is now included in your backup and comes back when you pair the device again, and removing a hardware wallet asks first whether to keep its name and tags in your backup. #1173
+- Payments made from incoming private payment requests now send a payment proof back to the requester. #1178
+- Bitkit can now review, manage, and pay recurring payment requests from Paykit contacts, with on-chain fees shown before confirming the first payment and unresolved payments protected from accidental retries. #1186
+- Added on-chain send support for paired Trezor wallets, including transaction approval on the device. #1187
+- Added a Trezor Receive tab with fast watch-only addresses and on-device verification. #1189
+
+### Changed
+- Added support for tags and transaction details to hardware wallet activity. #1044
+- Updated Pubky profiles, contacts, and Paykit settings to match the latest designs and simplified contact payments. #1097
+- Send now keeps the funding-source button, fee label, and swipe-to-confirm track stable while switching sources, continuing from amount, or refreshing fees. #1195
+
+### Fixed
+- Receive requests now refresh reliably after a payment is completed. #1086
+- The Show Details button on the Receive screen is now available while the Lightning node is still starting up, so you can view and copy your on-chain address right away. #1090
+- Savings transfers to spending now use a faster on-chain fee rate, reserve mining fees when sizing max transfers, and only drain the wallet when leftover change would be dust or a retry can still fully fund the order. #1091
+- Fixed profile deletion being blocked when private Paykit contact cleanup could not complete. #1108
+- Activity timestamps now follow the device's 12/24-hour time setting instead of always using a 12-hour clock. #1111
+- App Status and Backup timestamps now follow the device language instead of always using the US date format. #1124
+- Fixed Paykit Server authorization to require both public and private payment capabilities. #1131
+- Restored the transfer label on hardware wallet transactions after removing and re-adding a device. #1133
+- Improved Lightning send failure recovery with clearer messages and a retry action that refreshes payment routing. #1140
+- Fixed delayed private payment requests after adding a Paykit contact or returning to the app. #1141
+- Tags you add to hardware wallet activity are now included in your backup, and older backups are upgraded automatically the first time you restore them. #1163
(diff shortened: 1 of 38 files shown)


jvsena42 (commented):
Release mechanics check out. **No findings — safe to ship from my side.**

**Version:** `app/build.gradle.kts:189-190` → `versionCode = 189`, `versionName = "2.5.0"`. Previous release (`git cat-file -p "v2.4.1^{commit}:app/build.gradle.kts"`) was 188 / 2.4.1, so monotonic +1. I also checked the ABI-split logic, since the APK/AAB output code was restructured between 2.4.1 and 2.5.0 (`applicationVariants.all` → `androidComponents.onVariants`) — neither form applies an ABI offset to `versionCode`, both only rename output files, so the Play-uploaded AAB goes 188 → 189 with no regression.

**Changelog consumption — all 36 fragments verified individually, not spot-checked:**
- 36 fragments deleted (the 37th path is `.gitkeep`, correctly retained), 36 bullets in the new section. 1:1.
- Each fragment's exact text appears **exactly once**, under a heading matching its filename category. Zero dropped, zero duplicated, zero miscategorised. Category counts match filenames: added 11, changed 3, fixed 19, security 3.
- Issue-ID sets identical between fragment filenames and the `#NNNN` suffixes (sorted `diff` is empty).
- **Released sections untouched** — everything from `## [2.4.1]` down is byte-identical apart from the link-reference block, which correctly repoints `[Unreleased]` and adds the `[2.5.0]` compare link. That's the expected mechanical output, not a manual edit.
- No `changelog.d/hotfix/` fragments swept in (that dir holds only `.gitkeep` at both base and head).

**Stray changes:** none. `git diff --name-status` is exactly `M CHANGELOG.md`, `M app/build.gradle.kts`, and 36 `D changelog.d/next/*.md`.

**Semver:** 2.5.0 is right. Even discounting the Paykit and dev-only entries, Trezor on-chain send (#1187), the Trezor Receive tab (#1189), hidden/passphrase Trezor wallets (#1142) and HW-name backup (#1173) are reachable new features, so a minor is justified rather than a patch.

ovitrif (dismissed):
utAck

jvsena42 (dismissed):

ovitrif (approved):

piotr-iohk:
Resolved conflicts, please review @ovitrif @jvsena42.

CHANGELOG.md
```diff
@@ -7,6 +7,53 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added on-chain send support for paired Trezor wallets, including transaction approval on the device. #1187
+- Added a Trezor Receive tab with fast watch-only addresses and on-device verification. #1189
+
+### Changed
+- Added support for tags and transaction details to hardware wallet activity. #1044
+- Updated Pubky profiles, contacts, and Paykit settings to match the latest designs and simplified contact payments. #1097
+- Send now keeps the funding-source button, fee label, and swipe-to-confirm track stable while switching sources, continuing from amount, or refreshing fees. #1195
+
+### Fixed
+- Resetting the wallet no longer leaves backups broken after a failed setup, and no longer races an in-flight Lightning start. #1266
```
jvsena42, line 31: Anchoring here because the code is not in this PR's diff: #1266 landed on master as `21c62aefa` and is cherry-picked into this branch as `81207664c`. **This is a note for a follow-up against master, not a change request on the release back-merge.**

**A wipe landing mid-setup orphans the VSS gate.** LOW on its own. `setup()` captures `gate = isSetup`, and `reset()` cancels that deferred and installs a fresh one under a different lock (`synchronized(this)` vs `setupMutex`). When the network call returns, `gate.complete(Unit)` no-ops on the cancelled deferred, the `isSetup === gate` swap is skipped, and `setup()` returns success with the live deferred never completed. The `scope.launch` at `BackupRepo.kt:172` is untracked, so `stopObservingBackups()` never cancels it, and `WipeWalletUseCase.stopNode()` calls `reset()` first. Nothing user-visible follows, because every real caller runs `setup()` again before awaiting, and a new wallet re-runs it anyway.

**The same window has a consequence worth more attention.** The in-flight setup already read the store id before the network call, `reset()` clears the `VssStoreIdProvider` cache, and then the success path launches `vssBackupClientLdk.setup()`, which re-derives the store id from the keychain. At that point `keychain.wipe()` has not run yet (it comes after `lightningRepo.stop()` and `cleanupRemote()`), so the **old** wallet's mnemonic is still there and the old store id is cached again. `clearCache()` runs only from `reset()`, so the entry survives until the process dies. Create or restore a wallet in that same process and `LightningService` builds the new node against the old wallet's VSS store, with the LDK deletes pointing there too. Whether the server binds a store id to the LNURL-auth identity is unverified here: if it does, the node fails to start until a restart; if it doesn't, the new wallet's LDK state is written under the old store id. This is pre-existing, not introduced by #1266.

One line in both clients closes both: `check(gate.complete(Unit))` instead of `gate.complete(Unit)`, so a reset mid-setup becomes a failure and the LDK setup launch never fires. Clearing the store-id cache on wallet create/restore would harden it further.

jvsena42 (commented):
**No blocker.** The bump is clean and `v2.5.0` is already tagged and published, so the shipped artifact is the tag, not this head. One advisory inline about code that came in via #1266, for a follow-up against master.

Checked and clean:
- **Version.** `versionCode = 190`, `versionName = "2.5.0"`, monotonic over 188. 189 was the first bump commit, superseded. No other file carries a version string.
- **Tag vs notes.** `v2.5.0` does not contain post-cut master, so the 2.5.0 section describes exactly what shipped.
- **Changelog.** 36 fragments consumed into 36 bullets with matching issue ids, released sections untouched apart from the link block, and the 37 fragments still in `changelog.d/next/` all arrived after the cut, so they belong to 2.6.0. Nothing dropped. The `#1266` bullet was hand-added rather than collected from a fragment; the result is right, but it bypassed the script.
- **Carry-over.** The diff against master is only the changelog, the gradle bump and the deleted fragments.
- **Cherry-pick.** `81207664c` and master's `21c62aefa` differ only in hunk offsets.
- **No config drift.** No secret, signing, endpoint or flavor change. I also diffed `v2.4.1..v2.5.0` on `app/build.gradle.kts`: the `androidComponents.onVariants` restructure keeps every default, including `GEO=true` and `FEATURE_PAYKIT_UI_DISABLED=false`. No flag flipped on in this release.

CHANGELOG.md
jvsena42: Opened a issue for this https://github.com/synonymdev/bitkit-android/issues/1331

ovitrif (approved):
ack - I trust [reviewer bot on this](https://github.com/synonymdev/bitkit-android/pull/1246#pullrequestreview-5289947262), it does much more than I can do wrt. matching prs before and after merge / rebase / conflicts resolution

edit: scanned manually ✅

PR #1248, opened 2026-09-10
fix: align receive liquidity

pwltr:
### Description

Follow-up from https://github.com/synonymdev/bitkit-ios/pull/711#discussion_r3969484202

This PR:
1. Aligns Receive Lightning invoice eligibility with the shared receive-liquidity spec by using ready-channel inbound capacity for normal invoice creation, Auto availability, and Spending CJIT routing.
2. Clears receive-session CJIT invoice/details state when editing or starting a fresh Receive session so immutable CJIT invoices are not reused after amount changes, including CJIT QR edits that can return to a normal BOLT11 invoice when the new amount fits inbound liquidity.
3. Validates CJIT quotes before confirmation UI and maps node-wide Blocktank capacity failures to a separate user-facing error instead of the per-channel maximum message.
4. Updates the platform-agnostic receive-liquidity spec and focused unit coverage for the receive decision matrix, CJIT quote validation, CJIT session state, and Blocktank error classification.

### Design

Mapped frames in `docs/screens-map.md`:
- `EditInvoiceScreen.kt`: Send (Contact) (Lightning) › Edit invoice populated
- `ReceiveAmountScreen.kt`: Receive CJIT Onboarding › CJIT Enter amount
- `ReceiveConfirmScreen.kt`: Send (Enter Manually) › CJIT Fee to connect
- `ReceiveQrScreen.kt`: Receive Lightning Active › Receive Auto (Unified QR)

### Preview

https://github.com/user-attachments/assets/9c3f1772-825f-4931-8dd2-f2ffe0b87dfb

#### Manual Tests

- [ ] **1.** Receive → Spending CJIT QR → Edit Invoice with an amount that fits ready inbound liquidity → QR: returns to Spending, shows the newly generated normal LN invoice, and does not show the previous CJIT invoice.
- [ ] **2.** Receive → Spending CJIT QR → Edit Invoice with an amount that still needs additional liquidity → Confirm → QR: shows a fresh Spending-only CJIT invoice and does not expose the old CJIT invoice.
- [ ] **3.** Receive → create or view CJIT QR → close sheet → reopen Receive: starts from fresh Receive state with no CJIT invoice/details preserved.
- [ ] **4.** `regression:` ready channel with inbound > 0 but peer not yet usable → Receive: normal LN invoice/Auto availability follows ready inbound capacity.

#### Automated Checks

- Unit tests added: `CjitQuoteValidatorTest.kt` covers invalid fee/net quote cases and a valid quote.
- Unit tests added: `CjitEntryDetailsTest.kt` covers confirmation-entry mapping and quote rejection before confirmation UI.
- Unit tests updated: `ReceiveInvoiceEditStateTest.kt` covers clearing stale CJIT invoice state on edit and fresh CJIT creation.
- Unit tests updated: `ReceiveLiquidityDecisionTest.kt` and `WalletRepoTest.kt` cover ready-channel receive eligibility and ready inbound liquidity.
- Unit tests updated: `BlocktankRepoTest.kt` covers node capacity error classification separately from per-channel max-size errors.
- Local checks:
  - `./gradlew testDevDebugUnitTest --tests to.bitkit.models.ReceiveLiquidityDecisionTest --tests to.bitkit.models.CjitQuoteValidatorTest --tests to.bitkit.ui.screens.wallets.receive.ReceiveInvoiceEditStateTest --tests to.bitkit.ui.screens.wallets.receive.CjitEntryDetailsTest --tests to.bitkit.repositories.WalletRepoTest --tests to.bitkit.repositories.BlocktankRepoTest`


app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -150,11 +152,13 @@ fun ReceiveSheet(
                         onClickEditInvoice = {
                             editInvoiceSourceTab = it
                             invoiceEditState.beginSoftwareEdit(it)
+                            cjitSessionState.beginReceiveEdit()
```
jvsena42: **Regression — tapping Edit on a CJIT QR and then Back loses the invoice the user was just shown.**

`beginReceiveEdit()` (`= clear()`, `:432-434`, nulling both `cjitInvoice` and `entryDetails`) runs here at edit **start**, unconditionally, before navigating — and at `:161` for the Trezor tab. Back is just `popBackStack()` (`:363`) and restores nothing. So a user who creates a CJIT, shows the QR to a payer, taps Edit to check something, and taps Back returns to a QR with `cjitInvoice == null`.

Nothing re-derives it. `cjitInvoice` is sheet-level `remember` state and `ReceiveRoute.QR` is a `data object` with no argument; `entryDetails` was cleared too; `BlocktankRepo.cjitEntries` does keep the entry, but only `ChannelDetailScreen` reads it, and only once a channel exists. The Spending tab renders `cjitInvoice ?: bolt11` (`ReceiveQrScreen.kt:690`), so with no ready channel the tab flips to Savings and Spending shows CJIT onboarding; with one, it shows the bolt11. Either way it looks as though no CJIT was ever made, with no toast, and the only recovery is re-running the whole quote flow — which may come back with different fees.

To be clear about impact: **no sats are at risk.** The old LSP invoice is still payable and would still open a channel. What's lost is the ability to re-show or copy an invoice the user may already have handed out.

**It's a regression from both Edit entry points.** On master, the QR-view Edit went to `onClickReceiveCjit` (master `ReceiveQrScreen.kt:~386-390`) and Back from the Amount screen left `cjitInvoice` intact — that branch is removed here (head `:384-388`). The details-card Edit is identical on master and head, and master's sheet handler had no clear, so Back restored the CJIT there too. (Master's details-card path *on completion* is the stale-invoice bug this PR correctly fixed — the fix just moved the clear too early.) It's also the one place this diverges from the iOS twin on a PR titled "align": `ReceiveQr.swift:504` pushes the edit *over* the CJIT QR and `ReceiveEdit.swift:273-287` only replaces it in `finishWithQr`/`finishWithRoute`, so Back on iOS keeps the invoice — and `ReceiveEditTests.swift` tests exactly that.

Reachable by every user: both the Edit under the QR (`ReceiveQrScreen.kt:515-531`) and the details-card Edit are shown on Spending while a CJIT is displayed.

**Fix — clear on completion, not on start.** Drop the `beginReceiveEdit()` calls here and at `:161`, and clear inside the completing callback at `:364`:

```kotlin
updateInvoice = { cjitSessionState.clear(); wallet.updateBip21Invoice(it) }
```

I walked every completing path in `EditInvoiceScreen.kt:123-146` to check this doesn't reopen the stale-invoice bug:
- **`None`** — clears, then `onBack`, so no stale CJIT is shown. The "edit down with a ready channel" bug stays closed.
- **`ChooseAmount`** — clears, then opens Amount; a new CJIT or a Back both land on a cleared QR.
- **`CreateCjit` success** — no `updateInvoice`, but `navigateReceiveConfirm` → `onCjitCreated` already nulls `cjitInvoice` (`:436-439`) before Confirm.
- **`CreateCjit` failure / `GeoBlocked`** — nothing clears, so Back returns to the old, still-valid CJIT at its original amount. That isn't stale, because the edit was never applied. (Clearing inside `navigateCjitAmount` too would match iOS on the failure path; harmless either way.)
- **Trezor on-chain edit** — doesn't call `updateInvoice`, so the CJIT is kept, matching iOS.

Optionally clear only when `editInvoiceSourceTab == SPENDING` for exact iOS parity — as written, a completed Savings edit also clears the CJIT, which is no worse than head.

One test needs changing with it: `receive CJIT session clears stale invoice when editing starts` in `ReceiveInvoiceEditStateTest.kt` currently pins the start-of-edit clear — i.e. it asserts the regression. Replace it with create → confirm → begin edit → back without applying → invoice still present, plus a case that a *completed* edit does clear.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -374,7 +378,7 @@ fun ReceiveSheet(
                             navController.navigateTo(ReceiveRoute.PaymentRequestRecipient)
                         },
                         navigateReceiveConfirm = { entry ->
```
jvsena42, line 384: Low, pre-existing, but it sits right on the invariant this PR's docs now state: after an *additional*-CJIT edit, the Savings tab keeps the old amount.

Invoice at 20k, a ready channel with 50k inbound, edit from Spending to 100k → `CreateCjit(100k)` → confirm → QR. Spending shows the 100k CJIT correctly (the amount is in the bolt11), but `walletState.bip21AmountSats` is still 20k: the `CreateCjit` branch deliberately skips `updateInvoice`, and nothing refreshes bip21 afterwards (`ReceiveQrScreen.kt` has no `refreshBip21`/`updateBip21`, and `ReceiveSheet.kt:119-122` runs once per sheet). Swipe to Savings and the on-chain QR encodes `amount=0.0002`, not the 100k the user just approved. The auto-switch to Spending (`:243-251`) makes it one swipe away rather than the default view, which is why it's LOW.

iOS avoids it by setting `wallet.invoiceAmountSats = amountSats` before branching (`ReceiveEdit.swift:177`) and calling `refreshBip21()` on every QR appear (`ReceiveQr.swift:229-235`). The minimal Android version is to set the bip21 amount here before navigating: `wallet.setBip21AmountSats(entry.receiveAmountSats.toULong())`. `docs/receive-liquidity.md` now says the normal invoice "must not regenerate before creating CJIT" — true, but that leaves the *after* case unaddressed.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
```diff
@@ -129,19 +131,12 @@ fun EditInvoiceScreen(
...
-                                    CjitEntryDetails(
-                                        networkFeeSat = entry.networkFeeSat.toLong(),
-                                        serviceFeeSat = entry.serviceFeeSat.toLong(),
-                                        channelSizeSat = entry.channelSizeSat.toLong(),
-                                        feeSat = entry.feeSat.toLong(),
-                                        receiveAmountSats = action.amountSats.toLong(),
-                                        invoice = entry.invoice.request,
-                                    )
-                                )
+                                navigateReceiveConfirm(CjitEntryDetails.from(entry, action.amountSats).getOrThrow())
```
jvsena42: Low, latent: this `getOrThrow()` sits inside `.onSuccess {}` — outside `runSuspendCatching {}` — so a failure would propagate out of `collect` and take down the `LaunchedEffect`, unlike `ReceiveAmountScreen.kt:145-146` where the same call is inside the catching block.

Not reachable today: `BlocktankRepo.createCjit:283-287` already ran the identical validator on identical inputs, so `from` is deterministic-success here. Worth moving inside the block anyway, for symmetry with `ReceiveAmountScreen` and so the next change to either validator can't turn it into a crash.

jvsena42 (dismissed):
Funds-focused review. **No sats are at risk anywhere in this PR**, and three of the four iOS-parity gaps I'd recorded for Android are closed:

- **`validateCjitEntry` twin — closed.** `CjitQuoteValidator` (`models/CjitQuoteValidator.kt:17-24`) rejects `feeSat >= invoiceSat` *before* subtracting, then `channelSizeSat < invoiceSat - feeSat`, all in `ULong` — the guard makes the subtraction underflow-proof. Placement is what matters and it's right: it runs inside `BlocktankRepo.createCjit` (`:283-287`) right after the only `coreService.blocktank.createCjit` call in the app, and both UI callers go through `BlocktankViewModel.createCjit` → repo. `CjitEntryDetails.from` (`ReceiveConfirmScreen.kt:208-223`) is the only production constructor of the confirm model, so the old signed-`Long` path at `:83` can no longer render a negative "you will receive".
- **Stale CJIT after editing, ready-channel half — closed**, by a different mechanism than iOS: clearing eagerly rather than replacing the route on completion. That eagerness is the problem inline.
- **Stale channel cache in `canCreateLightningInvoice` — closed, not reintroduced.** `currentChannels()` (`WalletRepo.kt:771-773`) reads live `getChannels()` when the node is running; this PR only swaps `isUsable` → `isChannelReady` (`:776`), and `calculateRemoteBalance()` already filtered on `isChannelReady`, so the liquidity number is consistent.

Also checked clean: the hardware address is only emitted under `ReceiveTab.TREZOR` and CJIT only under `SPENDING` (`ReceiveInvoiceUtils.kt:40-52`), with AUTO hidden while a CJIT invoice exists — no wrong-destination display. All new arithmetic is `ULong` behind the underflow guard. Session reset per presentation is guaranteed by `key(receiveSheetPresentationKey(sheet))` plus `LaunchedEffect(startRoute) { clear() }`.

**Still open at LOW — zero-inbound first-channel half.** With no channels, Edit on the initial CJIT QR → Continue returns `None` via `(inboundCapacitySats ?: 0uL) == 0uL` (`ReceiveLiquidityDecision.kt:45`), so the CJIT is cleared, the typed amount only reaches the bip21, and `ReceiveQrScreen.kt:211-217` flips to Savings. "Receive on Spending" then opens `ReceiveAmountScreen` empty — it uses a fresh `hiltViewModel()` `AmountInputViewModel` whose only `setSats` calls are the max clamp and the min button. Same ~3-line fix as I suggested on iOS: prefill from `walletState.bip21AmountSats` in a `LaunchedEffect`, since both the `None` and `ChooseAmount` branches already persist the typed amount there. (Not in the diff, so noting it here.)

Requesting changes for one regression — inline. It's in core receive rather than Paykit, so it isn't gated.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/EditInvoiceScreen.kt
pwltr: Agreed. Moved the `CjitEntryDetails.from(...).getOrThrow()` call inside `runSuspendCatching`, matching the `ReceiveAmountScreen` path so future validation changes are handled as typed failures instead of escaping the collector.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Fixed, and your suggested direction exposed one extra detail: updating `bip21AmountSats` alone was not enough because Savings renders the prebuilt BIP21 URI.

I added a wallet-layer path that updates the shared amount and rebuilds the on-chain BIP21 URI without creating/replacing a Lightning invoice. Added `WalletRepoTest` coverage to assert the new amount is encoded and no `lightning=` param is added.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Good catch. Updated this so CJIT state is no longer cleared when edit starts. Back from the edit screen now preserves the existing CJIT invoice, and stale state is cleared only when an edit is actually applied.

I also updated `ReceiveInvoiceEditStateTest` to cover both cases: cancelled edit keeps the invoice, completed edit clears it.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -418,6 +429,28 @@ fun ReceiveSheet(
...
 
+@Stable
+internal class ReceiveCjitSessionState {
+    var cjitInvoice by mutableStateOf<String?>(null)
+        private set
+    var entryDetails by mutableStateOf<CjitEntryDetails?>(null)
+        private set
+
+    fun onCjitCreated(entry: CjitEntryDetails) {
+        cjitInvoice = null
```
jvsena42: Low, and a both-platforms follow-up rather than something to fix here alone: **Confirm→Back after an Edit-driven CJIT still loses the CJIT the user already showed, and now leaves Savings at the declined amount.**

A CJIT QR at 20k is shown to a payer → Edit → 100k → Continue → `CreateCjit` → `navigateReceiveConfirm` (`:384-390`) runs `updateOnchainBip21Amount(100k)`, then `onCjitCreated` nulls `cjitInvoice` here — both *before* the user confirms. They see the fee, decline, Back → Back. The QR comes back with `cjitInvoice == null`: `ReceiveRoute.QR` is a `data object`, `entryDetails` now holds the declined quote, and `cjitEntries` is only read by `ChannelDetailScreen`. `bip21AmountSats` is 100k, above ready inbound, so Spending shows CJIT onboarding and the tab flips to Savings, which encodes 100k. The Amount-screen path (`onCjitCreated` at `:268`, reached when creation fails) nulls it the same way.

It's a regression against Android master, where neither call touched `cjitInvoice` — it was only *replaced* at Confirm (master `:287`, `:302`). The two halves came from different commits: the null from `f604acdf2`, hidden behind A until `79dd4c3bf` removed A; and the Savings-at-100k half from `79dd4c3bf`, which followed my own suggestion on `:384` to set the bip21 amount before navigating. My last walkthrough also cited this null as the guard keeping the `CreateCjit` path clean — true for the success case, but I didn't walk the decline.

**Why this isn't blocking: iOS does exactly the same thing.** For `CreateCjit`, `ReceiveEdit.swift:244-246` calls `finishWithRoute(.cjitConfirm)` → `replaceEditedQrRoute`, which removes the stale CJIT QR at *creation*, not at Confirm — and `testReplacingEditedQrRouteCanNavigateToFreshCjitFlow` (`ReceiveEditTests.swift:40-50`) pins that. iOS also sets `invoiceAmountSats` before branching (`:177`), so its Savings shows the new amount after a decline too. A was blocking partly *because* Android diverged from iOS; blocking an alignment PR in order to make it diverge would be backwards. It's also narrower than A — `CreateCjit` needs ready inbound that's non-zero but below the new amount (`ReceiveLiquidityDecision.kt:43-45`), so zero-channel first-CJIT users can't reach it — and no sats are at risk, since the old invoice stays payable.

If you want it fixed, do it on both platforms: replace at Confirm, not at creation. Drop `cjitInvoice = null` here (`onCjitConfirmed` overwrites it anyway) and move `updateOnchainBip21Amount` into the `ConfirmIncreaseInbound` `onContinue` (`:306`). Two caveats: `onContinue` isn't a suspend function, so call `onCjitConfirmed` and navigate synchronously and launch the bip21 update separately; and `ConfirmIncreaseInbound` is also reached from the Amount screen, so gate or accept that.

**Related, introduced by `79dd4c3bf`:** wrapping `navigateReceiveConfirm` in `scope.launch` means `isCreatingCjit = false` (`EditInvoiceScreen.kt:147`) runs *before* navigation, so a second tap during the DataStore write re-runs `createCjit` and leaves an orphan LSP quote, overwriting `entryDetails`. `navigateTo` is single-top so there's no duplicate screen, and no sats are at risk. Making the navigation synchronous again — which the fix above does — removes it.

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
```diff
@@ -630,6 +630,19 @@ class WalletRepo @Inject constructor(
 
     fun setBip21AmountSats(amount: ULong?) = _walletState.update { it.copy(bip21AmountSats = amount) }
 
+    suspend fun updateOnchainBip21Amount(amountSats: ULong?): Result<Unit> = withContext(bgDispatcher) {
```
jvsena42, line 633: Low, latent: this moves `bip21AmountSats` but leaves `walletState.bolt11` at the old amount.

`updateBip21Invoice` clears `bolt11` when the amount can't be covered (`:755-756`); this rebuilds `bip21` without `lightning=` but never clears `bolt11`. So after a `CreateCjit` edit the wallet holds `bolt11` at 20k and `bip21AmountSats` at 100k.

It's masked today: `CreateCjit` only fires when the amount exceeds ready inbound, so `canCreateLightningInvoice` is false and `getInvoiceForTab` hides the stale bolt11, and a `ChannelReady` regenerates it. It would surface only if ready inbound grew *without* a `ChannelReady` (say, after an outbound payment) while the QR sits with `cjitInvoice == null` — i.e. after the Confirm→Back above — at which point Spending would show a payable 20k bolt11 under a 100k Savings request.

A `setBolt11("")` in here matches `updateBip21Invoice`'s cannot-cover branch, and the existing test's `never().createInvoice` assert still holds.

app/src/test/java/to/bitkit/ui/screens/wallets/receive/ReceiveInvoiceEditStateTest.kt
```diff
@@ -40,4 +40,64 @@ class ReceiveInvoiceEditStateTest {
 
         assertNull(state.initialTab(hardwareWalletId = null))
     }
+
+    @Test
+    fun `receive CJIT session keeps invoice when edit is cancelled`() {
```
jvsena42, line 45: These two tests would still pass with the A fix reverted, so the regression isn't actually pinned.

The fix lives in composable lambda wiring (`ReceiveSheet.kt:155-164`, `:365-368`) that this unit test can't reach. `keeps invoice when edit is cancelled` never simulates an edit — it's create → confirm → assert, effectively the same as `exposes confirmed fresh invoice` (`:84-93`). `clears stale invoice when edit is applied` calls `clear()` directly. Re-adding `beginReceiveEdit()` and its two call sites would leave both green.

Not blocking — the state class is thin and there's no Compose harness for `ReceiveSheet` — but worth knowing so nobody reads these as coverage for the Edit→Back behaviour.

jvsena42 (commented):
**Finding A is fixed** — `79dd4c3bf` drops the start-of-edit clear (`beginReceiveEdit()` is gone from `:155-164` and deleted) and clears on completion in `updateInvoice` (`:365-368`). Lifting my `REQUEST_CHANGES`.

I walked every path in `EditInvoiceScreen.kt:116-154` to check the stale-invoice bug this PR originally fixed stays closed:

| Path | Edit applied? | CJIT |
|---|---|---|
| `None` | yes | cleared via `updateInvoice` → the "edit down with a ready channel" bug stays closed |
| `ChooseAmount` | yes | cleared, then Amount |
| `CreateCjit` success | on Confirm | `onCjitConfirmed` sets the new invoice |
| `CreateCjit` failure / `GeoBlocked` | no | old, still-valid CJIT kept — correct |
| Trezor on-chain edit | on-chain only | kept — correct |
| **Back from Edit without applying** | no | **kept — the bug is fixed** |

Per-presentation scoping still holds: `cjitSessionState` is still `remember`ed inside `ReceiveSheet` (`:81`) under `key(receiveSheetPresentationKey(sheet))`, with a fresh `UUID` per `Sheet.Receive`, and no CJIT state moved into `WalletRepo`/`WalletViewModel`. The only new repo writes are `bip21AmountSats`/`bip21`, which every sheet open resets via `refreshReceiveState()`, so nothing leaks into the next presentation.

**B is addressed** — `updateOnchainBip21Amount` keeps the Savings QR on the approved amount. **C is addressed** — `getOrThrow()` now sits inside `runSuspendCatching` (`EditInvoiceScreen.kt:133-135`). The zero-inbound prefill is still open, but I'd marked it optional.

Three LOW notes inline, none blocking.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Fixed in dff7f789b. Android now preserves the currently displayed CJIT QR until the replacement quote is actually confirmed: onCjitCreated only stores the pending quote details, and the BIP21 amount update moved to Confirm onContinue. Back from confirmation returns to the old QR/amount; confirming replaces it with the fresh CJIT. Navigation is synchronous again as well, so the duplicate-tap window from the previous scope.launch wrapper is gone.

app/src/main/java/to/bitkit/repositories/WalletRepo.kt
pwltr: Fixed in dff7f789b. updateOnchainBip21Amount now clears bolt11 before rebuilding the on-chain-only BIP21 URI, matching the cannot-cover branch in updateBip21Invoice. WalletRepoTest now seeds an old bolt11 and asserts it is cleared while also asserting no new Lightning invoice is created.

app/src/test/java/to/bitkit/ui/screens/wallets/receive/ReceiveInvoiceEditStateTest.kt
pwltr: Acknowledged. I did not add a Compose harness for ReceiveSheet in this PR, so the test remains state-level coverage rather than direct wiring coverage for the edit callbacks. The behavior was manually verified in the sheet flow, and the state test was updated only to pin the session invariant that old CJIT stays visible until fresh CJIT confirmation.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -279,12 +294,12 @@ fun ReceiveSheet(
...
                 }
                 composableWithDefaultTransitions<ReceiveRoute.Confirm> {
-                    cjitEntryDetails.value?.let { entryDetails ->
+                    cjitSessionState.entryDetails?.let { entryDetails ->
                         ReceiveConfirmScreen(
                             entry = entryDetails,
                             onLearnMore = { navController.navigateTo(ReceiveRoute.Liquidity) },
                             onContinue = { invoice ->
-                                cjitInvoice.value = invoice
+                                cjitSessionState.onCjitConfirmed(invoice)
```
jvsena42, line 302: Medium: this route is missing the `updateOnchainBip21Amount` call its sibling has two blocks below, and the new `resetEditInvoiceAmount()` helper turns that gap into a destroyed CJIT.

Compare `ConfirmIncreaseInbound.onContinue` at `:317`:
```kotlin
wallet.updateOnchainBip21Amount(entryDetails.receiveAmountSats.toULong())
cjitSessionState.onCjitConfirmed(invoice)
```
against this one, which goes straight to `onCjitConfirmed`. So after the *first* CJIT, `bip21AmountSats` is still null.

**Reproduced on device** (Pixel 9, PR-head CI build `dff7f789b`), regtest wallet with no channels:

1. Receive -> Spending -> Receive Lightning funds -> Min (₿2,000) -> Continue -> Confirm ("YOU WILL RECEIVE ₿593") -> Continue. CJIT QR shown on the Spending tab.
2. Tap **Edit** — the amount field reads **₿0 / $0.00**, not 2,000. `resetEditInvoiceAmount()` (`:121-128`) reads `walletState.bip21AmountSats`, which this route never set, so it takes the `clearInput()` branch.
3. Tap **Show QR** without touching the amount — the intent being to add a note or a tag.
4. `onClickContinue(SPENDING, 0, …)` -> `needsInboundLiquidity(0, 0)` is true (`ReceiveLiquidityDecision.kt:71-73`) -> `:45` returns `None` -> `updateInvoice(0)` -> `clear()` (`:377`). **The CJIT is gone and the sheet silently switches to the Savings tab showing an on-chain address.**

No sats are at risk — the LSP invoice stays payable for anyone who already scanned it, and an on-chain payment still credits the user. What makes this MEDIUM rather than LOW is that the trigger is completely benign (open Edit to add a note), and the outcome silently changes the receive destination the user explicitly chose, with no indication. A user who shows the new QR believing they are still receiving on Spending gets an on-chain payment and no channel.

**On scope, because this matters for whether it is fair to raise here:** the *symptom* predates this PR — the edit view model was 0 on this route before `dff7f789b` too, since it is cleared once per sheet at `:79` and the Amount screen uses its own `hiltViewModel()` instance. But this commit added `resetEditInvoiceAmount()` specifically to prefill the edit amount, and wired it to both Edit entry points; this is the one route that does not feed it. The fix is one line, in code this commit already edits:

```kotlin
onContinue = { invoice ->
    wallet.updateOnchainBip21Amount(entryDetails.receiveAmountSats.toULong())
    cjitSessionState.onCjitConfirmed(invoice)
```

That makes the prefill show 2,000, makes the Savings QR encode the same amount the quote was priced for, and stops step 4 from collapsing to `None`.

If you would rather take this as a separate PR since the symptom is pre-existing, say so and I will drop the block — that is a reasonable call and I would not argue it. iOS has the same gap (`ReceiveCjitConfirmation.swift` never sets `invoiceAmountSats`, while `ReceiveEdit.swift:160-161` prefills from it), so it is worth fixing on both sides either way.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -73,13 +74,16 @@ fun ReceiveSheet(
...
     val navController = rememberNavController()
     val rootRoute = startRoute.rootRoute()
+    val currencies = LocalCurrencies.current
 
     LaunchedEffect(Unit) { editInvoiceAmountViewModel.clearInput() }
-    LaunchedEffect(startRoute) { navController.navigateToReceiveStart(startRoute) }
-
-    val cjitInvoice = remember { mutableStateOf<String?>(null) }
-    val cjitEntryDetails = remember { mutableStateOf<CjitEntryDetails?>(null) }
+    val cjitSessionState = remember { ReceiveCjitSessionState() }
```
jvsena42, line 80: Low, pre-existing: `cjitSessionState` is `remember` rather than `rememberSaveable`, so an activity recreation drops a displayed CJIT while the sheet itself survives.

**Reproduced on device** (same build): with a confirmed CJIT QR showing on the Spending tab, toggling system dark mode (`adb shell cmd uimode night yes`) recreates `MainActivity`. `AppViewModel._currentSheet` is view-model-scoped so the Receive sheet reappears — but `cjitSessionState` is fresh and `LaunchedEffect(startRoute)` at `:84` clears it. The sheet came back on **Savings** with an on-chain QR; the CJIT was gone.

Not introduced here — master already held `cjitInvoice` in a plain `remember { mutableStateOf<String?>(null) }`. And it is narrower than it sounds: `MainActivity` is locked to portrait, so rotation cannot trigger it; it takes a uiMode, locale, or font-scale change, or a low-memory kill. The LSP invoice also stays payable.

Raising it because it is the same failure the rest of this PR is about — a CJIT the user is actively showing to a payer disappearing underneath them — and because both fields are strings, so `rememberSaveable` with a small `Saver` would close it. Entirely optional, and reasonable to defer.

jvsena42 (changes requested):
Re-reviewed at `dff7f789b`, and I re-derived the whole path table from the code rather than leaning on my previous sign-off — the commit changes exactly the mechanism I signed off on last time, and the skill I work from is explicit that a safety net a prior review praised is an unverified claim until stressed again.

**The regression I flagged is fixed.** `Confirm -> Back` now leaves everything alone: `onCjitCreated` (`:447-449`) no longer nulls `cjitInvoice`, `navigateReceiveConfirm` (`:396-397`) is synchronous again and no longer touches bip21, and the bip21 write moved into `ConfirmIncreaseInbound.onContinue` (`:317`). So declining a replacement quote leaves the old CJIT intact and the QR recomposes onto Spending showing it.

Re-derived table at head:

| Path | Edit applied | Existing CJIT | Payer-visible QR stale? |
|---|---|---|---|
| `None` | yes | cleared | no — re-derived from bip21 at the new amount |
| `ChooseAmount` | yes | cleared, then Amount | no |
| `CreateCjit` -> Confirm | at Confirm only | kept until `:318`, then replaced | no — fee, "you will receive" and `entry.invoice` all come from the same `entryDetails` object |
| **`CreateCjit` -> Confirm -> Back** | **no** | **kept** | **no — the fix** |
| `CreateCjit` failure | no | kept | no |
| `GeoBlocked` | no | kept | no |
| Trezor on-chain edit | on-chain only | kept | no |
| **Back from Edit unapplied** | **no** | **kept** | no — `resetEditInvoiceAmount()` only writes to the edit view model, never to wallet state |
| Sheet dismissed at any point | no | lost with the presentation | by design; and dismissing at Confirm now leaves `bip21AmountSats`/`bolt11` untouched |
| System Back at Confirm | no | kept | no |
| Config change mid-flow | n/a | **lost** | see the second inline note |

No completing or cancelling path replaces or destroys a displayed CJIT before the user confirms the replacement. Nothing moved CJIT state into `WalletRepo`/`WalletViewModel` — `remember` at `:80` under `key(receiveSheetPresentationKey(sheet))` still scopes it per presentation, and the one new repo write (`setBolt11("")`) is regenerated by `refreshBip21()` on every sheet open.

**Two findings inline, both reproduced on a Pixel 9 against the PR-head CI build rather than argued from code.** One is MEDIUM and is why I am requesting changes; the other is LOW and optional. Both are pre-existing symptoms rather than regressions — I have said so explicitly on each, and on the MEDIUM I have offered to drop the block if you would rather take it separately.

Also checked and clean:

- **Approve-time display vs encoded invoice.** `ReceiveConfirmScreen.kt:82-93` computes fees and "you will receive" from `entry`, and `:110` passes `entry.invoice`; `:316-322` reads `receiveAmountSats` off the same `let`-bound object, so the screen and the lambda recompose together and cannot disagree.
- **Savings vs Spending after an additional CJIT.** `:317` sets Savings to the exact amount the quote was priced for and `WalletRepo.kt:636` clears the stale-amount `bolt11`; `getInvoiceForTab` prefers `cjitInvoice` and AUTO is hidden while one exists, so no tab can show the old-amount bolt11.
- **Arithmetic.** No new signed-`Long` subtraction. `CjitEntryDetails.from` is still the only production constructor and still runs `CjitQuoteValidator.validate` before the `Long` fields exist.
- **Partial failure.** `CreateCjit` failure touches neither `cjitInvoice` nor `entryDetails`.
- **Double-tap.** Navigation is synchronous again, which closes the window I flagged last round. `_isLoading` is set on `Main.immediate` before the first suspension, `PrimaryButton` is `enabled && !isLoading` with a 500 ms debounce. The residual — a tap during the 300 ms exit transition starting a second `createCjit` — has its `LaunchedEffect` cancelled when Edit leaves composition, so nothing reaches `onCjitCreated`; worst case is an orphan quote at the LSP.
- **Fire-and-forget `updateOnchainBip21Amount`.** `WalletViewModel.kt:522` is now `viewModelScope.launch`. The only concurrent writer is `refreshBip21()` on a new presentation, which needs dismiss-and-reopen inside the DataStore write window; `SCREEN_TRANSITION_DELAY` in `showSheet` widens that further. Not reachable by a human.
- **`setBolt11("")` vs sheet auto-close.** `observeReceiveSheetInvoice` (`AppViewModel.kt:550`) returns early on an empty bolt11, so `receiveSheetContext` keeps the previous value and auto-close behaviour is unchanged.
- **Conventions.** `runSuspendCatching` kept at `WalletRepo.kt:634`; `Logger.error(..., context = ...)`; no `runBlocking`, no `!!`, no `System.currentTimeMillis()`.
- `ReceiveInvoiceEditStateTest.kt:70-86` now pins keep-until-confirm at the state level.

Gating: core receive, no feature flag, so severity is assigned without any gating discount.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Fixed in `a2a1edf74`. First-CJIT confirmation already syncs the amount, and unchanged edits of an already-confirmed CJIT now preserve the active CJIT invoice instead of clearing the session through the generic `None` path. Added state coverage for matching confirmed CJIT amounts.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Good catch. I’m going to leave this out of this PR because it’s a pre-existing lifecycle/persistence limitation rather than part of the receive-liquidity regression being fixed here.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -418,6 +440,31 @@ fun ReceiveSheet(
...
+    fun onCjitCreated(entry: CjitEntryDetails) {
+        entryDetails = entry
+    }
+
+    fun onCjitConfirmed(invoice: String) {
+        cjitInvoice = invoice
+    }
+
+    fun hasConfirmedInvoiceForAmount(amountSats: ULong?): Boolean {
+        return cjitInvoice != null && entryDetails?.receiveAmountSats?.toULong() == amountSats
```
jvsena42: **`hasConfirmedInvoiceForAmount` compares against the *pending* quote, not the confirmed one — so a declined replacement quote makes a later Savings edit silently no-op.**

`entryDetails` is written by `onCjitCreated` (:451) and never cleared on Back; `cjitInvoice` is written by `onCjitConfirmed` (:455). After a declined quote the two are desynced, and this line reads the wrong one.

Trace — ready channel, 50k inbound:

1. Edit from Spending → 80k → CreateCjit → Confirm. :318 sets bip21 to 80k, :319 sets `cjitInvoice = inv80k`. `entryDetails = E80k`.
2. Edit from Spending again → 120k → CreateCjit (120k > 50k) → `navigateReceiveConfirm` (:398) sets `entryDetails = E120k` → user declines the fee, Back. Back is `popBackStack()` (:325), so nothing is restored.
3. State: `cjitInvoice = inv80k`, `entryDetails = E120k`, `bip21AmountSats = 80k`.
4. User swipes to Savings — the only non-CJIT tab, since `canCreateLightningInvoice` is false at 80k so AUTO is hidden — taps Edit, types 120k to take it on-chain instead.
5. `ReceiveLiquidityDecision.kt:42` returns `None` for any non-SPENDING source, so `EditInvoiceScreen.kt:124` calls `updateInvoice(120k)` → :378 `hasConfirmedInvoiceForAmount(120k)`: `cjitInvoice != null` is true, `E120k.receiveAmountSats == 120k` is true → **the whole branch is skipped** → :125 `onBack()`.

QR returns on Savings with `walletState.bip21` still encoding 80k. The user typed 120k, got no change, and no toast — the only trace is `amount=0.0008` inside the copy-address URI text. A payer scanning it sends 80k.

Before `a2a1edf74` this path cleared the CJIT and set 120k, so it's a regression from the guard itself rather than from keep-until-confirm.

**Fix** — snapshot the amount at confirm time, no call-site changes needed:

```kotlin
private var confirmedAmountSats: ULong? = null

fun onCjitConfirmed(invoice: String) {
    cjitInvoice = invoice
    confirmedAmountSats = entryDetails?.receiveAmountSats?.toULong()
}

fun hasConfirmedInvoiceForAmount(amountSats: ULong?): Boolean =
    cjitInvoice != null && confirmedAmountSats == amountSats
```

`entryDetails` is the quote being confirmed at both :303 and :319, so the snapshot is always the right one.

`ReceiveInvoiceEditStateTest` :100-114 only exercises a single quote, which is why it passes. Worth adding: confirm 1_000 → `onCjitCreated` a 2_000 quote → assert `hasConfirmedInvoiceForAmount(2_000uL)` is false and `(1_000uL)` is still true.

While you're here — `docs/receive-liquidity.md:46` and `:58` ("Editing from Savings or Auto returns to the normal QR with Savings/onchain only") are no longer unconditionally true after this guard, and nothing in the doc mentions that an unchanged edit preserves the CJIT. One bullet covers it: re-applying an edit at the confirmed CJIT amount keeps the current CJIT; only a changed amount clears or replaces it. (`:52-53` is still accurate — the replacement does happen before the updated result is shown.)

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
```diff
@@ -357,7 +374,12 @@ fun ReceiveSheet(
                         lightningState = lightningState,
                         sourceTab = editInvoiceSourceTab,
                         onBack = { navController.popBackStack() },
-                        updateInvoice = wallet::updateBip21Invoice,
+                        updateInvoice = {
+                            if (!cjitSessionState.hasConfirmedInvoiceForAmount(it)) {
+                                cjitSessionState.clear()
+                                wallet.updateBip21Invoice(it)
```
jvsena42, line 380: **A note typed during an unchanged-amount edit never reaches the Savings BIP21 URI.**

`updateBip21Invoice` is the only thing on this path that rebuilds the URI, and it now sits inside the `if`. `onDescriptionUpdate = wallet::updateBip21Description` (:384) → `WalletRepo.setBip21Description` (:647) only does `_walletState.update { it.copy(bip21Description = ...) }`.

So: CJIT confirmed at amount A → Edit → type a note → Show QR with the amount unchanged → `None` → short-circuit → `walletState.bip21` untouched. Savings QR and copy text both read `walletState.bip21` (`ReceiveQrScreen.kt:369-372`, `ReceiveInvoiceUtils.kt:30-33`), so `message=` is missing. The note reappears if Edit is reopened, and lands on the next rebuild (ChannelReady, another edit), so it looks saved.

This is the exact trigger from my earlier :303 comment — "open Edit to add a note or a tag". Tags are fine (`addTagToSelected` persists immediately); only the description is affected. The Trezor tab reads `bip21Description` directly so it's unaffected too.

**Fix** — one line:

```kotlin
updateInvoice = {
    if (!cjitSessionState.hasConfirmedInvoiceForAmount(it)) {
        cjitSessionState.clear()
        wallet.updateBip21Invoice(it)
    } else {
        wallet.updateOnchainBip21Amount(it)
    }
},
```

Same call Confirm already makes for this amount; it re-encodes `bip21Description` and creates no Lightning invoice. Its `setBolt11("")` is a no-op here — `bolt11` is already empty in every CJIT-displayed state.

LOW. Take it or document it.

jvsena42 (changes requested):
Re-reviewed at `a2a1edf74`. The two new commits are `f1f088ad` (the `updateOnchainBip21Amount` call I asked for on :303) and `a2a1edf74` (the unchanged-edit guard). `f1f088ad` is correct. The guard in `a2a1edf74` has a bug inside it — one MEDIUM below, plus a LOW on the same branch.

Neither finding reverses anything: keep-until-confirm (`dff7f789b`) and the :303 call stay exactly as they are. The MEDIUM is a defect *in* the new guard, not an argument against having one.

**Checked and clean** (exits walked at head, and what keeps each correct):

| Exit | Result |
|---|---|
| QR → Edit → system/gesture Back | `onBack = popBackStack` (:376); `resetEditInvoiceAmount()` writes only to the edit VM. CJIT and bip21 untouched |
| Edit → CreateCjit → Confirm → Back | `onCjitCreated` (:451) only sets `entryDetails`; `navigateReceiveConfirm` touches no wallet state. Displayed CJIT and Savings amount unchanged |
| Edit → CreateCjit failure / GeoBlocked | `navigateCjitAmount` / `navigateGeoBlock`; nothing cleared; both Backs are plain pops |
| Confirm → Continue (both routes) | fee, net and `entry.invoice` all read off the same `let`-bound `entryDetails`; `popUpTo(QR)` inclusive makes a double-tap inert |
| Liquidity / LiquidityAdditional → Back or Continue | both `popBackStack`; `entryDetails` retained |
| Sheet swipe-dismiss → reopen | fresh `presentationId` → `key(...)` recreates the sheet → fresh `ReceiveCjitSessionState` + `LaunchedEffect(startRoute) { clear() }` + `clearBip21State`. No CJIT state lives in WalletRepo/WalletViewModel |
| Backgrounding without death | a `ChannelReady` regenerates `bolt11`, but Spending still prefers `cjitInvoice` and AUTO stays hidden while one exists |

Also verified: `CjitQuoteValidator` runs twice (in `createCjit` and again in `CjitEntryDetails.from`) before any `Long` conversion, and net receive is guarded by `feeSat < invoiceSat`; `ReceiveCjitErrorPresenter` renders only string resources, no server text; `refreshCjitEntries` still has the correct `withTimeout` + explicit `CancellationException`/`TimeoutCancellationException` guard while `createCjit` uses `runSuspendCatching`. `BlocktankRepo`, `WalletRepo`, the validator and the decision table are unchanged since `dff7f789b` and still look right.

Process death mid-session is still out of scope — you deferred it earlier and I'm not reopening it.

---

Non-blocking, separate from the findings: `### Design` says `N/A — no design available.`, but all four screens this PR touches are mapped to real frames in `docs/screens-map.md` — `EditInvoiceScreen.kt`, `ReceiveAmountScreen.kt`, `ReceiveConfirmScreen.kt`, `ReceiveQrScreen.kt` (lines 192-197), none of them `todo` or `n/a`. Link the relevant frames and I'll diff the new error states against them next pass. Doesn't block anything.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Fixed in `69a076b47`. `ReceiveCjitSessionState` now snapshots `confirmedAmountSats` when the CJIT invoice is confirmed, and `hasConfirmedInvoiceForAmount` compares against that confirmed snapshot instead of the latest pending `entryDetails`.

I also added coverage for the declined-replacement shape: confirm 1,000 sats, create a pending 2,000 sats quote, then assert 1,000 still matches and 2,000 does not.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
pwltr: Fixed in `69a076b47`. The unchanged-confirmed-CJIT edit path now preserves the active CJIT invoice but still calls `updateOnchainBip21Amount`, so the on-chain BIP21 URI is rebuilt with the current note/metadata without creating a normal Lightning invoice.

app/src/main/java/to/bitkit/ui/screens/wallets/receive/ReceiveSheet.kt
jvsena42: Fixed. `confirmedAmountSats` is snapshotted at :459 from the same `entryDetails` both Confirm routes write their bip21 amount from, and `clear()` nulls it alongside `cjitInvoice` at :469.

Walked my own repro against the new head: after the declined 120k quote the state is `cjitInvoice = inv80k`, `entryDetails = E120k`, `confirmedAmountSats = 80k`, and the Savings edit to 120k now evaluates `80k == 120k` -> false -> takes the clearing branch and rebuilds bip21 at 120k.

I also went after the mirror bug, since a guard fixed in review is where the next one lands. `hasConfirmedInvoiceForAmount` returning false for the genuinely confirmed amount would destroy a live CJIT, so I traced every writer of `bip21AmountSats` that could make the Edit prefill diverge from the snapshot: `setSats` stores sats verbatim and `switchUnit` only rewrites the display text, so no fiat rounding drift on a prefilled untouched amount; the only in-sheet trigger of `refreshBip21`/`clearBip21State` is flipping the notification switch on the Confirm/Liquidity screen, which happens before Confirm's own amount write. Inbound liquidity growing mid-session no longer matters either, since the guard is now pure amount equality. Nothing diverges.

The test pins it. I reverted `ReceiveSheet.kt` to `a2a1edf74` while keeping the new test: `receive CJIT session matches confirmed amount when pending quote differs` fails at :126. At head, 10/10 pass. Throwaway worktree outside the repo, nothing committed.

The `else { updateOnchainBip21Amount(it) }` branch is right too — it rebuilds with `bip21Description`, so the note reaches the URI, and its `setBolt11("")` is a no-op in every CJIT-displayed state reachable without an LDK event. One bonus I hadn't anticipated: `ChooseAmount` at the confirmed amount now lands here as well and keeps the CJIT while opening the Amount screen, instead of the previous silent no-op.

Docs line reads correctly against :378-383. Clearing my block.

jvsena42 (approved):
`69a076b47` fixes the MEDIUM and the LOW. Clearing my `REQUEST_CHANGES` — details on the two threads.

The repro is closed at the guard itself: `confirmedAmountSats` is snapshotted in `onCjitConfirmed` from the same `entryDetails` both Confirm routes write their bip21 amount from, and `clear()` nulls it alongside `cjitInvoice`. A declined replacement quote now leaves `entryDetails` desynced harmlessly, because the guard no longer reads it.

I spent most of this pass on the mirror bug rather than the one I reported, since a guard fixed in response to review is exactly where the next defect lands — that is how the original got in. `hasConfirmedInvoiceForAmount` returning *false* for the genuinely confirmed amount would destroy a CJIT invoice the user has already shown a payer, so I traced every writer of `bip21AmountSats` that could make the Edit prefill diverge from the snapshot: `setSats` stores sats verbatim and `switchUnit` only rewrites display text (no fiat rounding drift on a prefilled untouched amount), and the only in-sheet trigger of `refreshBip21`/`clearBip21State` is the notification switch on Confirm/Liquidity, which fires before Confirm's own amount write. Inbound liquidity growing mid-session is no longer relevant either, now that the guard is pure amount equality. Nothing diverges.

Exits re-walked at the new head, since the control flow changed shape: Edit toolbar/system/gesture Back; Confirm → Back after a declined replacement quote, and the mirror where the confirmed amount is re-applied afterwards; an Amount-created quote declined twice back to QR; Liquidity and LiquidityAdditional Back and Continue; CreateCjit failure and GeoBlocked; system Back on the QR root; swipe and scrim dismiss then reopen (fresh `presentationId` → fresh session state → `refreshBip21` reset); config change (all three fields are `remember`ed so they die together — no partial state, and the CJIT loss there is the limitation you explicitly deferred, not reopening it); process death; backgrounding without death.

The test is load-bearing: reverting `ReceiveSheet.kt` to `a2a1edf74` while keeping the new test fails `receive CJIT session matches confirmed amount when pending quote differs`; at head 10/10 pass. Throwaway worktree outside the repo, nothing committed.

Docs line matches the code. One note for whoever merges: https://github.com/synonymdev/bitkit-android/pull/1231 renames `ReceiveTab.TREZOR` on lines immediately adjacent to the ones this PR rewrites in `ReceiveQrScreen.kt` and `ReceiveSheet.kt`, so expect a textual conflict — no semantic conflict, and `ReceiveInvoiceEditStateTest.kt` hunks are disjoint.

PR #1224, opened 2026-09-02
feat: support Pubky signup

ben-kaufman:
### Description

- Accept app-authorized `pubkyring://signup` and auth-bearing `pubkyauth://signup` requests, plus direct `pubkyauth://direct_signup` and parameter-only legacy `pubkyauth://signup`, through the normal scanner and deep-link flow.
- Register the wallet-derived Pubky identity with the requested Homeserver and optional signup token.
- Require explicit approval and the existing PIN/biometric check for every signup format; show the requested homeserver before registration.
- Restore locally owned identities by signing in on their existing homeserver, keeping credentials retryable after failures instead of registering at Homegate.
- Activate the returned Paykit session and continue into the existing profile setup flow.
- Reuse the approval sheet’s loading state while signup completes, report Already signed in when a local identity exists, and keep payment-only scanner state intact when rejecting Pubky requests.

This PR is stacked on #1200 and uses its Paykit rc51 authorization model. Ordinary Pubky App sign-in must use that grant-auth model; compatibility with the older sign-in request is intentionally outside this signup PR.

### Preview

Not included; this reuses the existing scanner, authorization approval sheet, loading treatment, and profile setup UI.

### QA Notes

1. With no Pubky identity, scan a signup QR from staging Pubky App; approve it, confirm the website continues, and complete profile setup.
2. With no Pubky identity, scan either direct signup format; confirm the homeserver is visible, Cancel performs no registration, and Authorize requires local authentication before profile setup opens.
3. Scan a signup request while signed in and confirm the Already signed in state.
4. Confirm payment-only scanners reject Pubky requests without clearing the current payment flow.
5. With the wallet locked, open a signup deep link; confirm nothing registers or opens until unlocking, then the approval sheet appears.
6. After signup on a non-default homeserver, make session recovery fail, restore connectivity, and retry Create/Restore Profile. Confirm it keeps the same homeserver. Repeat with a failed profile save and confirm retry never calls signup.

Validation:
- Full `testDevDebugUnitTest`: 2,346 tests passed on the restacked branch.
- `compileDevDebugKotlin` and `detekt` passed.
- Regression coverage includes locked signup deep links, consent/local-auth ordering, real SDK activation and partial-persistence cleanup, stale pending state, and one-time profile navigation.
- `PubkyRepoTest.kt`: 97 tests passed, including existing-key recovery, secure-storage errors, sign-in/profile failure retries, cancellation, and no-key signup. Detekt reports no changed-file findings.

PR #1253, opened 2026-09-11
ci: add e2e-staging workflow for nightly staging tests

### LINKED ISSUE #221: ci: split local merge-gate E2E from staging nightly/release
## Goal

Split Bitkit E2E CI so:

1. **Local docker/regtest** is the PR merge gate (self-contained app/UI signal).
2. **Staging (Blocktank/pubky)** runs nightly, on `release-*` PRs, and via `workflow_dispatch`.
3. The required GitHub check **`e2e-status` keeps working** (same job name, still reported on every non-draft PR).

This is an implementation ticket for an agent. Work spans this repo plus `bitkit-ios` and `bitkit-android`. Do not merge `skip_local` on the existing workflow.

### Infra must not red every PR

On 10 Sep 2026, Blocktank staging (`api.stag0.blocktank.to`) stopped marking transfer-to-spending orders paid (`state: created`, `paidSat: 0`) after an ln2-watcher crash and a ~2k unpaid-invoice backlog. Same class of stag0 failure as Feb 23 and Aug 8 (RabbitMQ / watcher). That reded `e2e-status` on **all** bitkit-android and bitkit-ios PRs even when the app was fine.

Staging shards that complete a real LSP path (`@multi_address_2`, transfer-to-spending) belong off the merge gate so stag0 stays usable for the real LSP path. Migration workflows are already nightly — leave them.

### Local is not actually self-contained today

`BACKEND=local` only switches Electrum/mining to docker. Both apps still hardcode Blocktank to stag0 on regtest:

- iOS `Bitkit/Constants/Env.swift` → `blocktankBaseUrl` default `https://api.stag0.blocktank.to`
- Android `app/src/main/java/to/bitkit/env/Env.kt` → same

Local `@transfer_1` / `@transfer_max` (and HW transfer) **create Blocktank orders**, then pay a deposit address on the **docker chain**. Those orders never go paid. Hardware wallet already documents this:

```ts
// test/specs/hardware-wallet.e2e.ts (@hardware_wallet_3)
if (getBackend() === 'regtest') {
  // We only check balance on staging regtest backend, local does not have Blocktank.
  await expectSpendingBalance(0, { condition: 'gt', timeout: 60_000 });
}
```

That unpaid-order leak is a plausible source of the 2k backlog, not just “three staging transfers per PR.” Moving staging off PRs without skipping these local cases only solves half the incident.

Related, not blocking: #34 (Blocktank in local docker) would make local truly self-contained later. Do not wait on it.

### Better signal for agentic PR volume

A lot of recent Bitkit work is agent-driven: many small PRs, parallel branches, retries on red CI. That only works if the **merge gate is a product signal**, not a shared-infra weather report.

- Staging-on-PR couples every agent PR to stag0 + GitHub-hosted Macs. One watcher crash makes every agent treat infra as an app bug, re-push, and pile more unpaid orders.
- Local-on-PR (docker Electrum + local LND) is the right gate for “did this change break the app?”
- Staging nightly / release / dispatch is the right gate for “does the real LSP/pubky path still work?”

Keep `e2e-status` as that local signal so agents and humans can trust a green merge box.

## Current shape

| Workflow | Repo | Trigger | What it runs |
|---|---|---|---|
| `e2e-tests.yml` | bitkit-ios | PR + dispatch | Local (self-hosted Mac Mini, 1 shard) **and** staging (`macos-latest`: `@multi_address_2`, `@pubky`) |
| `e2e.yml` | bitkit-android | PR + dispatch | Local (`ubuntu-latest` × 7) **and** staging (`@multi_address_2`, `@pubky`, `@hardware_wallet`) |
| `e2e_migration.yml` | both | cron 02:00/03:00 UTC + dispatch | Unchanged — leave it |
| `e2e-status` job | both app workflows | every non-draft PR | ANDs **local and staging**; this name is the required ruleset check |

Default-branch rulesets:

- iOS: required `e2e-status`, `Run Integration Tests`, `Run Tests`
- Android: required `e2e-status`, `build`

GitHub matches the **check name** `e2e-status`. Renaming it, not emitting it on some PRs, or leaving staging in `needs:` after those jobs stop running, will block merges.

Do **not** merge https://github.com/synonymdev/bitkit-ios/pull/744 (`skip_local` dispatch input). After this split, `gh workflow run e2e-staging` is the staging poke and never queues the Mini. Close #744 when the new workflow exists.

## Target shape — separate workflows

**Keep** existing `e2e-tests.yml` / `e2e.yml` as the local merge gate.

**Add** `e2e-staging.yml` on each app repo (nightly + dispatch + release PRs).

**Leave** `e2e_migration.yml` alone. Stagger staging nightly **after** migration (suggest ~04:00 UTC) so stag0 is not hit by restore wallets and LSP transfers in the same hour.

| Trigger | Local docker | Staging matrix | Status check |
|---|---|---|---|
| PR → master (feature) | Yes | No | `e2e-status` = local only |
| PR head or base `release-*` | Yes | Yes | `e2e-status` + `e2e-staging` (see skip-success below) |
| Nightly cron ~04:00 UTC on default branch | No | Yes | Slack on failure; not a PR check |
| `workflow_dispatch` on **local** workflow (`e2e.yml` / `e2e-tests.yml`) | Yes | No | Informational (does not satisfy the PR `e2e-status` check) |
| `workflow_dispatch` on **staging** workflow (`e2e-staging.yml`) | No | Yes | Informational + Slack `#bitkit-staging-nightly` |
| Migration crons | Unchanged | Unchanged | Unchanged |

Staging matrix (move as-is, then align HW):

- `@multi_address_2`
- `@pubky` (paykit/profile — pubky infra, not Blocktank, but still not a merge gate)
- Android `@hardware_wallet` full path (channel assert). iOS HW currently runs on the Mini and skips spending-balance; after the split, put **HW transfer completion** on staging for both platforms; keep connect/receive/send-onchain on local where Trezor emulator works.

## Hard constraint: do not break `e2e-status`

In the **existing local workflow**, keep a job whose `name:` is exactly `e2e-status`.

It must still run on every non-draft PR, including the path-filter skip path that currently succeeds with “no code changes.”

Change the job, not the name:

```yaml
e2e-status:
  if: always() && github.event.pull_request.draft == false
  name: e2e-status
  needs: [detect-changes, e2e-tests-local]   # drop e2e-tests-staging
  # verify only needs.e2e-tests-local.result == success when code == true
```

If staging stays in `needs:` after those jobs are removed or skipped on feature PRs, `needs.*.result` is `skipped` and the aggregator fails — that is the usual way this check breaks.

Do **not** add `e2e-staging` as a **required** ruleset context unless that workflow also runs on **every** PR and skip-succeeds on feature PRs. A missing required check blocks merge the same way a red one does.

Recommended skip-success pattern for a new `e2e-staging` job (required only if we decide to enforce staging on release PRs via GitHub):

- Feature PRs: job runs, exits 0, prints “skipped — not a release PR.”
- `release-*` head or base: run the real staging matrix; fail the job if shards fail.

v1: implement that for `release-*` only. A `needs-staging-e2e` label is v2 (`pull_request` types must include `labeled`). Dispatch is enough for LSP/pubky feature PRs in v1.

Do **not** implement “green nightly within N hours of cut” as a GitHub required check. Nightly is master SHA, not the release SHA.

## Implementation order

GitHub constraint: **`workflow_dispatch` only exists for workflow files already on the default branch.** A brand-new `e2e-staging.yml` cannot be “Run workflow”’d until it is merged. It **can** run on the PR that introduces it if that file has `on: pull_request` (same-repo PRs; the workflow from the PR head is used).

Do **not** ship skip-success on the first staging PR or that PR will no-op and you will not have tested the matrix.

Stacked PRs, per app repo (`bitkit-ios`, `bitkit-android` in parallel):

### PR A — add `e2e-staging.yml` only (merge first)

- New file: `schedule` + `workflow_dispatch` (`e2e_branch` only) + `pull_request`.
- **Always run the real staging matrix on this PR** (no release-* skip yet). Existing `e2e.yml` / `e2e-tests.yml` stays unchanged, so the merge gate is untouched. This PR will run staging twice (old job + new workflow); that is expected and is the test.
- Staging grep includes `@transfer_1|@transfer_max` plus current `@multi_address_2`, `@pubky`, and Android `@hardware_wallet` as specified above.
- Slack to `#bitkit-staging-nightly` via `secrets.SLACK_WEBHOOK_URL_STAGING` (already set).
- After merge: `gh workflow run e2e-staging.yml --ref <branch>` works for pokes.

### PR B — stacked on A (or on master after A merges) — switch the merge gate

- Strip `build-staging` / `e2e-tests-staging` from the **existing** local workflow.
- `e2e-status` needs local only (same job `name:`).
- On `e2e-staging.yml`, add skip-success for feature PRs; real run on `release-*` head/base, schedule, and dispatch. Feature PRs must still **emit** the `e2e-staging` check (green skip) so a later required context cannot go missing. Do not add it to the ruleset in v1 unless product wants it required on release PRs only via that skip-success job.
- Do not add `suite=` / `skip_local` / combined `full` mode.

### This repo — skip local LSP orders, with or just after A

- Spec-level skip when `BACKEND=local` for `@transfer_1` / `@transfer_max`. Safe on staging (`BACKEND=regtest` still runs them).
- **Do not merge this skip before PR A is live**, or those tests disappear from CI until staging exists.
- Keep `@transfer_2` on local grep; stop grepping `@transfer` as a blob.

### Then

- Close ios#744 without merge (dispatch `e2e-staging.yml` instead).
- Confirm one scheduled or dispatched run posted to `#bitkit-staging-nightly`.

### Later (out of v1)

- `needs-staging-e2e` label
- iOS local sharding (one Mini: sharding does not shorten a single PR)
- GitHub-hosted iOS sim + remote Linux docker (separate track)
- #34 local Blocktank in docker

## Acceptance criteria

- [ ] Feature PRs: required `e2e-status` still appears and goes green when local passes (or path-filter skip). Staging infra down does **not** red that check.
- [ ] `e2e-status` job `name:` unchanged. Ruleset does not need a rename.
- [ ] Staging lives in a **new** workflow file, not a `skip_local` flag on the old one.
- [ ] `BACKEND=local` no longer creates unpaid Blocktank orders. `@transfer_1` / `@transfer_max` run on staging (`e2e-staging.yml` shard); `@transfer_2` stays local-only.
- [ ] Nightly staging runs on master; posts to `#bitkit-staging-nightly` via `secrets.SLACK_WEBHOOK_URL_STAGING` (already set on ios + android).
- [ ] `workflow_dispatch` on the new workflow can poke staging without queuing iOS local Mini.
- [ ] `release-*` PRs still get a staging signal (skip-success job or explicit required check — implement option A).
- [ ] Migration crons unchanged.
- [ ] ios#744 closed without merge.
- [ ] AGENTS.md / README in this repo updated for which tags are local vs staging.

## Context

- Slack (10 Sep, stag0 down, Denys volume discussion): https://synonymworkspace.slack.com/archives/C03UWHGKUNP/p1789031337183729
- ln2 follow-up: https://github.com/synonymdev/blocktank-lsp-ln2/issues/277
- Do not merge: https://github.com/synonymdev/bitkit-ios/pull/744
- Local Blocktank (follow-up, not this ticket): #34

## Agent notes

- Implement in this order: PR A (`e2e-staging.yml` on ios + android, self-test via pull_request) → merge A → spec skip in this repo → PR B (strip staging from local workflow, skip-success on feature PRs) → close ios#744.
- Touch app repos in PRs of their own; do not try to change ios/android files from this repo.
- Do not edit GitHub rulesets unless `e2e-status` would stop reporting; the intended change is job internals, not the required context name.
- Do not add “Made with Cursor” or similar attribution in PRs/issues.
- English, conventional commits (`ci:`, `test:`). No changelog fragments unless a workflow change is user-facing (it is not).

piotr-iohk:
Fixes synonymdev/bitkit-e2e-tests#221 (PR A — android)

### Description

Adds a new workflow `.github/workflows/e2e-staging.yml` that runs staging E2E tests:

- **Triggers**:
  - `schedule`: runs daily at 04:00 UTC (after migration cron at 02:00)
  - `workflow_dispatch`: manual dispatch with inputs:
    - `e2e_branch` — branch of bitkit-e2e-tests to use
    - `post_to_slack` — checkbox to post summary to Slack (off by default)
  - **No PR trigger** — nightly + dispatch only

- **Staging shards**:
  - `@multi_address_2`
  - `@pubky`
  - `@hardware_wallet`
  - `@transfer_max`
  - ~~`@transfer_1`~~ — temporarily omitted; will return in PR B / e2e-tests fix

- **Build**: Uses `BACKEND=regtest` with `TREZOR_BRIDGE: true` matching existing staging build config
- **Slack**: Posts summary via shared `scripts/slack_summary.py e2e-staging` from `bitkit-e2e-tests` repo (merged in synonymdev/bitkit-e2e-tests#223) to `#bitkit-staging-nightly`:
  - On **schedule**: always posts (success, failure, or cancelled)
  - On **workflow_dispatch**: posts only when `post_to_slack` checkbox is checked
- **Concurrency**: group is `workflow-ref-event_name` with `cancel-in-progress: true` — schedule and dispatch cannot cancel each other; back-to-back same-trigger runs still cancel in progress
- **Timeout**: `timeout-minutes: 90` on `e2e-tests-staging` job (matches iOS)
- **Test coverage**: Added `e2e-staging.yml` to `ARTIFACT_CONSUMERS` in `BuildOutputContractTest.kt`

The existing `e2e.yml` workflow remains unchanged — the `e2e-status` merge gate is unaffected.

### Design

N/A — no UI changes.

### Preview

N/A

#### Manual Tests
N/A

#### Automated Checks
- Unit tests modified: `BuildOutputContractTest.kt` now includes `e2e-staging.yml` in `ARTIFACT_CONSUMERS`.
- Workflow concurrency (after merge): schedule and dispatch share a ref but not a group, so a master dispatch cannot cancel the nightly.

**To dispatch after merge:**

```sh
gh workflow run e2e-staging.yml -f e2e_branch=main
# or with Slack summary:
gh workflow run e2e-staging.yml -f e2e_branch=main -f post_to_slack=true
```

The workflow runs nightly at 04:00 UTC or via manual dispatch.

<div><a href="https://cursor.com/agents/bc-33edabbc-f060-4072-9641-d599d313d3c9?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-33edabbc-f060-4072-9641-d599d313d3c9&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -234,12 +234,17 @@ jobs:
         uses: jwalton/gh-docker-logs@v2
 
   notify-failure:
-    if: always() && github.event_name != 'pull_request' && needs.e2e-tests-staging.result == 'failure'
-    needs: [e2e-tests-staging]
+    if: >-
+      always() &&
+      github.event_name != 'pull_request' &&
+      (needs.build-staging.result == 'failure' ||
+       needs.e2e-branch.result == 'failure' ||
+       needs.e2e-tests-staging.result == 'failure')
+    needs: [build-staging, e2e-branch, e2e-tests-staging]
     runs-on: ubuntu-latest
     steps:
       - name: Notify Slack on failure
-        uses: slackapi/slack-github-action@v2.0.0
+        uses: slackapi/slack-github-action@e598089eaef53883a2d9b325e044899548518a03
         with:
           webhook: ${{ secrets.SLACK_WEBHOOK_URL_STAGING }}
           webhook-type: incoming-webhook
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -1,16 +1,14 @@
 name: E2E Staging
 
 on:
+  schedule:
+    - cron: "0 4 * * *"
   workflow_dispatch:
     inputs:
       e2e_branch:
         description: "Branch of synonymdev/bitkit-e2e-tests to use (main | default-feature-branch | custom branch name)"
         required: false
         default: "default-feature-branch"
-  schedule:
-    - cron: "0 4 * * *"
-  pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
 
 env:
   TERM: xterm-256color
@@ -22,7 +20,6 @@ concurrency:
 
 jobs:
   build-staging:
-    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
 
     steps:
@@ -74,14 +71,12 @@ jobs:
           path: app/build/outputs/bitkit/devDebug/bitkit_e2e.apk
 
   e2e-branch:
-    if: github.event.pull_request.draft == false
     uses: synonymdev/bitkit-e2e-tests/.github/workflows/determine-e2e-branch.yml@main
     with:
       app_branch: ${{ github.head_ref || github.ref_name }}
       e2e_branch_input: ${{ github.event.inputs.e2e_branch || 'default-feature-branch' }}
 
   e2e-tests-staging:
-    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     needs: [build-staging, e2e-branch]
 
@@ -92,7 +87,6 @@ jobs:
           - { name: multi_address_2_regtest, grep: "@multi_address_2" }
           - { name: pubky_paykit, grep: "@pubky" }
           - { name: hardware_wallet, grep: "@hardware_wallet" }
-          - { name: transfer_staging, grep: "@transfer_1|@transfer_max" }
 
     name: e2e-tests-staging - ${{ matrix.shard.name }}
 
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -1,0 +1,252 @@
+name: E2E Staging
+
+on:
+  workflow_dispatch:
+    inputs:
+      e2e_branch:
+        description: "Branch of synonymdev/bitkit-e2e-tests to use (main | default-feature-branch | custom branch name)"
+        required: false
+        default: "default-feature-branch"
+  schedule:
+    - cron: "0 4 * * *"
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+env:
+  TERM: xterm-256color
+  FORCE_COLOR: 1
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-staging:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Java
+        uses: actions/setup-java@v6
+        with:
+          java-version: "17"
+          distribution: "temurin"
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v6
+
+      - name: Decode google-services.json
+        run: |
+          set -euo pipefail
+          if [ -z "${GOOGLE_SERVICES_JSON_BASE64:-}" ]; then
+            echo "GOOGLE_SERVICES_JSON_BASE64 is empty; using checked-in placeholder."
+            exit 0
+          fi
+          mkdir -p app/src/debug
+          printf '%s' "$GOOGLE_SERVICES_JSON_BASE64" | base64 -d > app/src/debug/google-services.json
+        env:
+          GOOGLE_SERVICES_JSON_BASE64: ${{ secrets.GOOGLE_SERVICES_JSON_BASE64 }}
+
+      - name: Build debug app (regtest)
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CHATWOOT_API: ${{ secrets.CHATWOOT_API }}
+          E2E: true
+          E2E_BACKEND: network
+          GEO: false
+          TREZOR_BRIDGE: true
+          TREZOR_BRIDGE_URL: http://10.0.2.2:21325
+        run: ./gradlew assembleDevDebug
+
+      - name: Rename APK
+        run: |
+          apk=$(find app/build/outputs/bitkit/devDebug -name 'bitkit-dev-debug-*-universal.apk')
+          mv "$apk" app/build/outputs/bitkit/devDebug/bitkit_e2e.apk
+
+      - name: Upload APK (regtest)
+        uses: actions/upload-artifact@v7
+        with:
+          name: bitkit-e2e-apk-staging_${{ github.run_number }}
+          path: app/build/outputs/bitkit/devDebug/bitkit_e2e.apk
+
+  e2e-branch:
+    if: github.event.pull_request.draft == false
+    uses: synonymdev/bitkit-e2e-tests/.github/workflows/determine-e2e-branch.yml@main
+    with:
+      app_branch: ${{ github.head_ref || github.ref_name }}
+      e2e_branch_input: ${{ github.event.inputs.e2e_branch || 'default-feature-branch' }}
+
+  e2e-tests-staging:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    needs: [build-staging, e2e-branch]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        shard:
+          - { name: multi_address_2_regtest, grep: "@multi_address_2" }
+          - { name: pubky_paykit, grep: "@pubky" }
+          - { name: hardware_wallet, grep: "@hardware_wallet" }
+          - { name: transfer_staging, grep: "@transfer_1|@transfer_max" }
+
+    name: e2e-tests-staging - ${{ matrix.shard.name }}
+
+    steps:
+      - name: Show selected E2E branch
+        env:
+          E2E_BRANCH: ${{ needs.e2e-branch.outputs.branch }}
+        run: echo $E2E_BRANCH
+
+      - name: Clone E2E tests
+        uses: actions/checkout@v7
+        with:
+          repository: synonymdev/bitkit-e2e-tests
+          path: bitkit-e2e-tests
+          ref: ${{ needs.e2e-branch.outputs.branch }}
+
+      - name: Enable KVM
+        run: |
+          echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+          sudo udevadm control --reload-rules
+          sudo udevadm trigger --name-match=kvm
+
+      - name: AVD cache
+        uses: actions/cache@v6
+        id: avd-cache
+        with:
+          path: |
+            ~/.android/avd/*
+            ~/.android/adb*
+          key: avd-33-x86_64-pixel_6
+
+      - name: Create AVD and generate snapshot for caching
+        if: steps.avd-cache.outputs.cache-hit != 'true'
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          profile: pixel_6
+          api-level: 33
+          arch: x86_64
+          avd-name: Pixel_6
+          force-avd-creation: false
+          emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-front none
+          disable-animations: false
+          script: echo "Generated AVD snapshot for caching."
+
+      - name: Download APK (regtest)
+        uses: actions/download-artifact@v8
+        with:
+          name: bitkit-e2e-apk-staging_${{ github.run_number }}
+          path: bitkit-e2e-tests/aut
+
+      - name: List APK directory contents
+        run: ls -l bitkit-e2e-tests/aut
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: 22
+
+      - name: Cache npm cache
+        uses: actions/cache@v6
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Install dependencies
+        working-directory: bitkit-e2e-tests
+        run: npm ci
+
+      - name: Pull Trezor emulator image
+        if: matrix.shard.name == 'hardware_wallet'
+        working-directory: bitkit-e2e-tests
+        run: |
+          cd docker
+          docker compose --profile trezor pull --quiet trezor-user-env
+
+      - name: Run E2E Tests 1 (${{ matrix.shard.name }})
+        continue-on-error: true
+        id: test1
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          profile: pixel_6
+          api-level: 33
+          arch: x86_64
+          avd-name: Pixel_6
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-front none
+          script: cd bitkit-e2e-tests && ./ci_run_android.sh --mochaOpts.grep "${{ matrix.shard.grep }}"
+        env:
+          BACKEND: regtest
+          RECORD_VIDEO: true
+          ATTEMPT: 1
+
+      - name: Run E2E Tests 2 (${{ matrix.shard.name }})
+        continue-on-error: true
+        id: test2
+        if: steps.test1.outcome != 'success'
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          profile: pixel_6
+          api-level: 33
+          arch: x86_64
+          avd-name: Pixel_6
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-front none
+          script: cd bitkit-e2e-tests && ./ci_run_android.sh --mochaOpts.grep "${{ matrix.shard.grep }}"
+        env:
+          BACKEND: regtest
+          RECORD_VIDEO: true
+          ATTEMPT: 2
+
+      - name: Run E2E Tests 3 (${{ matrix.shard.name }})
+        id: test3
+        if: steps.test1.outcome != 'success' && steps.test2.outcome != 'success'
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          profile: pixel_6
+          api-level: 33
+          arch: x86_64
+          avd-name: Pixel_6
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-front none
+          script: cd bitkit-e2e-tests && ./ci_run_android.sh --mochaOpts.grep "${{ matrix.shard.grep }}"
+        env:
+          BACKEND: regtest
+          RECORD_VIDEO: true
+          ATTEMPT: 3
+
+      - name: Upload E2E Artifacts (${{ matrix.shard.name }})
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: e2e-artifacts-staging_${{ matrix.shard.name }}_${{ github.run_number }}
+          path: bitkit-e2e-tests/artifacts/
+
+      - name: Dump docker logs on failure (${{ matrix.shard.name }})
+        if: failure()
+        uses: jwalton/gh-docker-logs@v2
+
+  notify-failure:
+    if: always() && github.event_name != 'pull_request' && needs.e2e-tests-staging.result == 'failure'
+    needs: [e2e-tests-staging]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Slack on failure
+        uses: slackapi/slack-github-action@v2.0.0
+        with:
+          webhook: ${{ secrets.SLACK_WEBHOOK_URL_STAGING }}
+          webhook-type: incoming-webhook
+          payload: |
+            text: "❌ E2E Staging nightly failed"
+            blocks:
+              - type: "section"
+                text:
+                  type: "mrkdwn"
+                  text: "❌ *E2E Staging nightly failed*\n<${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View workflow run>"
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -87,6 +87,7 @@ jobs:
           - { name: multi_address_2_regtest, grep: "@multi_address_2" }
           - { name: pubky_paykit, grep: "@pubky" }
           - { name: hardware_wallet, grep: "@hardware_wallet" }
+          - { name: transfer_max, grep: "@transfer_max" }
 
     name: e2e-tests-staging - ${{ matrix.shard.name }}
 
--- a/.github/workflows/e2e-staging.yml
+++ b/.github/workflows/e2e-staging.yml
@@ -9,6 +9,11 @@ on:
         description: "Branch of synonymdev/bitkit-e2e-tests to use (main | default-feature-branch | custom branch name)"
         required: false
         default: "default-feature-branch"
+      notify_slack:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,252 @@
...
+          name: e2e-artifacts-staging_${{ matrix.shard.name }}_${{ github.run_number }}
+          path: bitkit-e2e-tests/artifacts/
+
+      - name: Dump docker logs on failure (${{ matrix.shard.name }})
+        if: failure()
+        uses: jwalton/gh-docker-logs@v2
+
+  notify-failure:
+    if: always() && github.event_name != 'pull_request' && needs.e2e-tests-staging.result == 'failure'
+    needs: [e2e-tests-staging]
```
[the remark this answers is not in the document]
piotr-iohk: Addressed: `notify-failure` now watches `build-staging`, `e2e-branch`, and `e2e-tests-staging` on schedule/dispatch (still skipped on pull_request).

.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,252 @@
...
+  workflow_dispatch:
+    inputs:
+      e2e_branch:
+        description: "Branch of synonymdev/bitkit-e2e-tests to use (main | default-feature-branch | custom branch name)"
+        required: false
+        default: "default-feature-branch"
+  schedule:
+    - cron: "0 4 * * *"
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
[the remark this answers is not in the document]
piotr-iohk: Superseded: `pull_request` trigger removed entirely. `e2e-staging.yml` is schedule + `workflow_dispatch` only — no more double-run with `e2e.yml` on PRs.

.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,258 @@
...
+        required: false
+        default: false
+
+env:
+  TERM: xterm-256color
+  FORCE_COLOR: 1
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
jvsena42, line 24: Low: on the default branch this concurrency group collapses to a single key, so a manual dispatch can silently kill the running nightly with no alert.

In the sibling `e2e.yml` this group is per-PR (`github.ref` is the PR ref), so cancel-on-new-push is right. Here both triggers land on the default branch — `schedule` always runs on `refs/heads/master`, and `workflow_dispatch` defaults to `master` in the UI — so every run shares the group `E2E Staging-refs/heads/master`.

Trigger: the 04:00 UTC nightly is still running (4 shards x up to 3 attempts) when someone clicks Run workflow on `master`. The nightly is cancelled, `needs.e2e-tests-staging.result` becomes `cancelled`, and `notify-failure` (lines 241-243) only fires on `== 'failure'` — so no Slack message goes out and that night produces no staging signal at all.

The discriminator has to go in the **group key**, because `cancel-in-progress` is the incoming run's knob and the incoming run is the one that needs restraining:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event_name }}
  cancel-in-progress: true
```

That puts schedule and dispatch in disjoint groups, so a dispatch can no longer reach the nightly's group, while back-to-back dispatches still cancel each other. It also matches the flat `cancel-in-progress: true` every other workflow in this repo already uses. Fixing the notification gate so a cancelled run still reports is worth doing as well, and is the half that actually removes the "no staging signal" outcome.

.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,258 @@
...
+          name: bitkit-e2e-apk-staging_${{ github.run_number }}
+          path: app/build/outputs/bitkit/devDebug/bitkit_e2e.apk
+
+  e2e-branch:
+    uses: synonymdev/bitkit-e2e-tests/.github/workflows/determine-e2e-branch.yml@main
+    with:
+      app_branch: ${{ github.head_ref || github.ref_name }}
+      e2e_branch_input: ${{ github.event.inputs.e2e_branch || 'default-feature-branch' }}
+
+  e2e-tests-staging:
```
jvsena42, line 84: Low: this job has no `timeout-minutes`, while the iOS twin caps the same job at 90 (https://github.com/synonymdev/bitkit-ios/pull/747 line 122).

`reactivecircus/android-emulator-runner` imposes none either, so an emulator that boots but never gets past a WDIO wait hangs attempt 1 — a known failure mode for the `hardware_wallet` shard, which depends on the Trezor emulator image pulled at line 171. Unattended overnight, the job burns to GitHub's 6-hour ceiling on each of the 4 matrix legs before failing. On PRs (`e2e.yml`, same gap) a human notices within minutes; on a nightly nobody does, and the Slack alert is delayed by up to 6h instead of 90min.

`timeout-minutes: 90` under line 85.

jvsena42 (commented):
Two LOW observations inline, neither blocking. The staging targeting is right and the matrix does fail the run — traces below.

Cross-repo consistency with the iOS twin https://github.com/synonymdev/bitkit-ios/pull/747, which I reviewed in the same pass:

- **The Slack action is pinned to a different release on each side.** Here, `e598089…` (line 248) is v2.0.0; iOS line 269 pins `b0fa283…` = v2.1.0. Both are SHA-pinned and both parse the YAML `payload:` form, so nothing breaks — but the pins were clearly meant to match, and one of the two should move.
- **The Slack message bodies diverge.** This one carries a title plus the run link (lines 253-258); iOS also includes workflow, run number, branch and `github.event_name`. Both post to the same `SLACK_WEBHOOK_URL_STAGING`, so the channel gets two differently-shaped alerts and the Android one will not say which trigger fired it.
- `notify-failure` conditions are logically identical on both sides (lines 237-243 here vs 257-263 there, just reordered). Shard sets differ only by `hardware_wallet`, which is Android-only and expected.

Checked and clean:

- **Staging targeting is correct.** The build (lines 55-65) sets `E2E_BACKEND: network` on `assembleDevDebug`, and every test attempt sets `BACKEND: regtest` (lines 186, 204, 221) — which per the e2e repo's README means the Blocktank API over the internet, not the docker stack. Matching build to runner backend is what that README requires, and this workflow correctly omits `e2e.yml`'s `Run regtest setup` / `Wait for electrum` docker steps.
- **All four shard tags resolve to real tests** on `bitkit-e2e-tests@main`: `@multi_address_2` (`multiaddress.e2e.ts:116`, also `@regtest_only` — consistent with `BACKEND: regtest`), `@pubky` (`paykit.e2e.ts:138`, `pubky-profile.e2e.ts:47`), `@hardware_wallet` (`hardware-wallet.e2e.ts:33`), `@transfer_max` (`transfer.e2e.ts:303`). No shard silently matches zero tests and passes green.
- **The matrix really does fail the run.** `continue-on-error: true` is on attempts 1 and 2 only (lines 174, 191); attempt 3 (line 208) has none, so a shard failing all three retries fails its job, and `fail-fast: false` still leaves `e2e-tests-staging.result == 'failure'`. The absence of an `e2e-status`-style aggregator is fine here — that job exists to give PRs a stable required-check name.
- **E2E branch resolution on a cron run**: `github.event.inputs.e2e_branch` is null on `schedule`, so line 82 falls through to `default-feature-branch`; `determine-e2e-branch.yml` probes `refs/heads/master` in `bitkit-e2e-tests`, which does not exist, and falls back to `main` as intended.
- **No injection surface or secret exposure.** No `pull_request_target`. The only `${{ }}` values reaching a `run:` shell are `matrix.shard.grep` (workflow-literal) and `needs.e2e-branch.outputs.branch`, which is piped through an `env:` var at line 102 first. Secrets appear only in `env:` blocks and action inputs. `GOOGLE_SERVICES_JSON_BASE64` decoding (lines 44-53) uses `set -euo pipefail` and an env indirection, same as the sibling.
- **The schedule will fire** — the workflow lands on `master`, the repo's default branch.
- **Artifact names do not collide** with `e2e.yml` (`bitkit-e2e-apk-staging_*` vs `-regtest_*`), and upload/download share the same `github.run_number`.

.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,262 @@
...
+      (github.event_name == 'schedule' ||
+       (github.event_name == 'workflow_dispatch' && inputs.post_to_slack == true))
+    needs: [build-staging, e2e-branch, e2e-tests-staging]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout E2E tests
+        uses: actions/checkout@v7
+        with:
+          repository: synonymdev/bitkit-e2e-tests
+          ref: ${{ needs.e2e-branch.outputs.branch }}
```
jvsena42, line 249: Low: this checks out the *resolved e2e feature branch*, but `scripts/slack_summary.py` exists only on `main` — so on a feature-branch dispatch the report step dies and no Slack message goes out at all.

I enumerated it rather than assuming: of the 9 branches in `synonymdev/bitkit-e2e-tests`, only `main` carries `scripts/slack_summary.py`. It landed there today in `6fbd12e`, so every branch cut before that lacks it — including `codex/pubky-ring-signup`, which exists in the e2e repo *and* is a live app-side branch right now (https://github.com/synonymdev/bitkit-android/pull/1224 and https://github.com/synonymdev/bitkit-ios/pull/724).

Scenario: you dispatch E2E Staging from `codex/pubky-ring-signup` with `post_to_slack: true`. `determine-e2e-branch` resolves `default-feature-branch` to the same-named e2e branch, this line checks it out, and the next step dies with `python3: can't open file '.../scripts/slack_summary.py'`. Red `slack-report` job, zero Slack output — the exact reporting loss this job exists to prevent.

The scheduled nightly is safe today, which is why this is LOW: on `schedule`, `github.ref_name` is `master`, no `master` branch exists in the e2e repo, so it falls back to `main`. The window also self-heals as branches rebase past `6fbd12e`.

Pin this one checkout to `ref: main`. The script is CI infrastructure, not test code — pinning it also stops a feature branch from shipping a broken `slack_summary.py` into the nightly report. The test-job checkout must keep tracking the resolved branch; that one *is* test code.

jvsena42 (commented):
Re-reviewed at `2bcb3695c`. **Both of my notes are addressed and the timeout fix is right. The concurrency one needs a different shape — my original suggestion was misdirected, and I've corrected that comment.** Three LOW notes, none blocking.

**`timeout-minutes: 90` — correct.** Line 87, on `e2e-tests-staging` (not the build job), so it applies per matrix leg, and it matches iOS line 122. I sanity-checked the value against real data rather than eyeballing it: recent green `e2e.yml` runs (34816482602, 34810970657, 34792240593) give per-leg wall times of 10-24 min, so a worst-case 3-attempt leg lands around 75 min, inside the cap.

**Concurrency — see the corrected comment on line 24.** Short version: `cancel-in-progress` is the *incoming* run's knob, and the incoming run is the one that needs restraining, so conditioning it on `github.event_name` cannot protect a run already in flight — the discriminator has to be in the group key. That was my error last round, not yours; I gave you that expression.

The good news is that the half that actually mattered is fixed. The old gate was `needs.e2e-tests-staging.result == 'failure'`, which never fires on a cancellation — confirmed empirically in cancelled run `34628569879`, where `notify-failure` concluded `cancelled`. The new `slack-report` job runs on `always()` with `STAGING_RESULT` computed as success-only-if-all-three-succeeded, so a cancelled nightly now *does* post `failure`, and `BUILD_RESULT: cancelled` / `E2E_TESTS_RESULT: cancelled` in the payload make the message self-diagnosing. I verified `always()` survives concurrency cancellation using the structurally identical `e2e-status` job in `e2e.yml`: in cancelled runs `34485911442` and `34482039396` it ran and reported `failure` while its upstreams were `cancelled`. So the silent-nightly outcome is genuinely closed, which is why the residual is LOW.

Worth knowing: `e2e_migration.yml` in both repos has the same ref-only group with both `schedule` and `workflow_dispatch`, so the same latent issue predates this PR. If you change the shape here, it's worth doing all four. And the mechanism is demonstrably live on `master` — in iOS `e2e_migration.yml` on 2026-09-10, run `34531572432` was cancelled 21 seconds after `34531814945` was created, both `workflow_dispatch` on `refs/heads/master`.

**The `slack_summary.py` refactor is a genuine improvement** and I have no objection to it: the env-var contract matches exactly (all nine names the script reads are supplied, no orphans in either direction), and it strictly *reduces* injection surface — every `${{ }}` now lands in an `env:` block or an action input, and line 252 is a literal command. It also removed a third-party action from both repos. Two notes on it inline, both non-blocking.

**Convergence with iOS https://github.com/synonymdev/bitkit-ios/pull/747 — both of last round's divergences are resolved.** The differing Slack action pins are moot (`slackapi/slack-github-action` is gone from both files entirely), and the differently-shaped payloads into the same channel are gone too — both now render through the one shared script, differing only by the platform title. The `timeout-minutes` gap is closed. The notify condition and concurrency block are now byte-identical across the two repos, which unfortunately also means the concurrency note applies verbatim to both.

Re-verified clean at head: all four shard tags still resolve; `continue-on-error` on attempts 1-2 only with attempt 3 bare (175, 192, 209-211), so the matrix can still fail the run; `E2E_BACKEND: network` (61) and `BACKEND: regtest` (187, 205, 222) unchanged; `GOOGLE_SERVICES_JSON_BASE64` decode still `set -euo pipefail` + env indirection (45, 51); artifact names still collision-free vs `e2e.yml`. The `BuildOutputContractTest.kt` addition is correct — the file contains `app/build/outputs/bitkit/` at 69, 70 and 76 and neither forbidden AGP-native path.

One unrelated thing so it doesn't cost you time: the red `build` check on this head is **not yours**. It is `QuickPayRepoTest > reconcile during live dispatched op does not steal completion`, a `Pending`-vs-`Success` race — and master commit `50b2c4cab` ("test: synchronize quickpay dispatch race") was already an attempt to fix that same race. #1224 and #1239 share this exact merge-base (`57dea75e7`) and their `build` checks passed. A re-run should clear it.

.github/workflows/e2e-staging.yml
```diff
@@ -0,0 +1,262 @@
...
+
+      - name: Post Slack summary
+        run: python3 scripts/slack_summary.py e2e-staging
```
jvsena42, line 252: Low, upstream: the shared script fails open, so a broken webhook leaves the nightly green and silent.

In `synonymdev/bitkit-e2e-tests@main:scripts/slack_summary.py`, `post_to_slack` returns `False` on a non-200 or a `URLError`, and returns `True` while printing to stderr when `SLACK_WEBHOOK_URL` is empty — but `main()` discards the value:

```python
        post_to_slack(message)
        return 0
```

Scenario: the `#bitkit-staging-nightly` incoming webhook is revoked or rotated (routine when a Slack app is reinstalled). At 04:00 a shard fails, this step runs, Slack answers `404 invalid_token`, the script prints one stderr line and exits 0, and the job goes green. The nightly silently stops reporting, and now that it posts on success too, "no message" is indistinguishable from "nothing ran". The `slackapi/slack-github-action` this replaced surfaced that as a red step.

Not this PR's defect and not blocking it — but both platforms now depend on that script, so it is worth a one-liner upstream: `return 0 if post_to_slack(message) else 1`.

I checked the adjacent risk and it is not live: `SLACK_WEBHOOK_URL_STAGING` does exist as a repo secret, so the empty-webhook silent-skip path isn't reachable today.

piotr-iohk:
Suggest we merge with the remaining Lows and enhance later.

This is a new schedule + workflow_dispatch flow — hard to fully exercise Slack/nightly behaviour until it lands on master. The Low notes (Slack script checkout on feature e2e branches; fail-open webhook) are valid follow-ups, but they should not block landing PR A.

Happy to follow up in a small PR after we have a real scheduled/dispatch run to look at.

ovitrif (approved):

.github/workflows/e2e-staging.yml
ovitrif: Low
Resolution adopted: we should iterate afterwards if needed, suggested by piotr in this [comment](https://github.com/synonymdev/bitkit-android/pull/1253#issuecomment-5668228615)

resolving

.github/workflows/e2e-staging.yml
ovitrif: Resolution: intentional
resolving

Issue #1254, opened 2026-09-14
[Bug]: Reset then restore picks the RN backup after a failed VSS lookup and drops tags / HW names


piotr-iohk:
### What happened?

On a **migrated mainnet** wallet (RN remote backup still on the server), Settings → Reset wallet → restore the same seed came back **without activity tags and without hardware-wallet names**, and the wiped local METADATA / ACTIVITY were then uploaded over the native VSS envelopes.

**Corrected root cause** (supersedes the original description, which blamed a "mnemonic read too soon" race). The full logcat shows the mnemonic read succeeded; three native defects chained instead. All times UTC (device log; CET = +2):

1. **Reset ran while a node start was in flight** (→ #1257). The app had been backgrounded at 08:21:57; the node stop hung until 08:26:29; the foreground `start()` then began building the node at 08:26:30. Reset was tapped at 08:26:49 while that build was still running. At 08:26:50.918 the node started, lifecycle went `Running`, and `BackupsViewModel.observeAndSyncBackups` → `BackupRepo.startObservingBackups()` → `VssBackupClient.setupWithRetry()` ran against the already wiped keychain: `VSS client setup error [MnemonicNotAvailableException]` (08:26:50.944, retries at 51.9 / 53.9 / 56.9).
2. **That first failure poisons `VssBackupClient`** (→ #1256): `isSetup.completeExceptionally()` is never undone. When the restore's `getLatestBackupTime()` called `setup()` at 08:26:57.724 it read the freshly saved mnemonic fine and logged `VSS client setup with server` (57.848) — but `getObject()` → `isSetup.await()` re-threw the stale exception 1 ms later: `Failed to get VSS backup timestamp: MnemonicNotAvailableException` (57.849).
3. **A failed lookup is treated as "no native backup"**: `restoreFromMostRecentBackup()` maps `vssTimestamp == null` to "use RN" because an RN timestamp exists → `RN migration completed` (08:27:23.9) with RN-era metadata (`tags: 0`, no HW names). `Full restore starting` never ran.
[...]

PR #1266, opened 2026-09-15
fix: vss client recovery and reset race

### LINKED ISSUE #1256: [Bug]: VssBackupClient never recovers after a failed setup(): later setup reports success but every VSS call throws the stale error
### What happened?

`VssBackupClient.setup()` completes its `isSetup: CompletableDeferred<Unit>` exceptionally on the first failure and never undoes that. A later `setup()` call re-runs the setup (the `isCompleted && !isCancelled` guard falls through because an exceptionally completed deferred reports `isCancelled == true`), loads the mnemonic, builds the VSS client and logs `VSS client setup with server` — but `isSetup.complete(Unit)` on an already-completed deferred is a no-op. From then on every `putObject` / `getObject` / `listKeys` / `deleteObject` hits `isSetup.await()` and throws the **original** exception, for the rest of the process lifetime or until `reset()` (only called from `WipeWalletUseCase`).

Observed on mainnet during 2.5.0 QA (the incident behind #1254): `BackupRepo.startObservingBackups()` → `setupWithRetry()` ran during a wallet wipe and failed once with `MnemonicNotAvailableException` (08:26:50.944 UTC). Seven seconds later the restore's `getLatestBackupTime()` called `setup()`, which read the freshly saved mnemonic fine and logged `VSS client setup with server` (08:26:57.848) — and 1 ms later `getObject()` failed: `Failed to get VSS backup timestamp: MnemonicNotAvailableException` (08:26:57.849). That `null` made the restore picker choose the RN remote backup instead of the native VSS backup. The later `VSS client setup succeeded on attempt 5` (08:27:01.124) was equally hollow.

Consequences:

- Any setup failure silently disables VSS backups until the app is restarted. Realistic triggers are keychain-side: the mnemonic not being present yet (wipe racing a node start, #1257) or a keychain load/decrypt error (cf. #897); `vssNewClientWithLnurlAuth` itself does no network I/O (the LNURL-auth JWT is fetched lazily on the first request), so a network hiccup at setup is not one of them. `setupWithRetry` returns immediately on non-mnemonic errors, so nothing logs "exhausted"; `triggerBackup()` then dies on `isSetup.await()` (outside its `runCatching`), which only surfaces as `Backup job cancelled for: '<category>'`.
- Restore source selection is wrong (#1254): a poisoned client makes `getLatestBackupTime()` return `null`, which the picker reads as "no native backup".

Verified with a JVM unit test against the real class (FFI facade `com.synonym.vssclient.Vss_rust_client_ffiKt` static-mocked with Mockito 5 so `setup()` can succeed off-device), 4/4 as predicted:

1. control — fresh client: `setup()` → `getObject()` OK
2. `setup()` fails (mnemonic `null`) → mnemonic present → `setup()` returns `Result.success` → `getObject()` **throws** `MnemonicNotAvailableException`
3. same via `setupWithRetry()` reporting "succeeded on attempt N" → `putObject()` **throws**
4. `reset()` → `setup()` → `getObject()` OK

The test was not committed; it can be added to `VssBackupClientTest`.

### Expected behavior

A `setup()` that returns success must leave the client usable. On failure, either leave the deferred incomplete and fail only that call, or replace it with a fresh `CompletableDeferred` before the next attempt. `putObject` / `getObject` must never re-throw a stale setup error as if it were current, and a failed setup must not turn into a silent, process-long backup outage.

### Steps to Reproduce

Unit level (deterministic):

1. Mock `Keychain.loadString(BIP39_MNEMONIC)` → `null`; call `setup()` → failure with `MnemonicNotAvailableException`.
2. Mock the mnemonic present (static-mock `vssNewClientWithLnurlAuth` / `vssGet` to return defaults); call `setup()` → `Result.success`.
3. Call `getObject("METADATA")` → throws `MnemonicNotAvailableException` instead of returning a result.

Device level: make the first `setup()` of a process fail (e.g. tap Reset wallet while a node start is in flight, see #1257), then observe that no VSS backup succeeds until the app is killed and relaunched.

Same device session as #1254 — see the log excerpt attached there: [2026-09-14-reset-restore-loses-tags-logs.zip](https://github.com/user-attachments/files/32189950/2026-09-14-reset-restore-loses-tags-logs.zip). The lines below are from the full logcat of that session (UTC):

```
08:26:49.266  VSS client reset                                     (WipeWalletUseCase → BackupRepo.reset)
08:26:50.918  Node started                                         (pre-wipe start finishing mid-wipe)
08:26:50.922  Start observing backup statuses and data store changes
08:26:50.944  VSS client setup error [MnemonicNotAvailableException='Mnemonic not available']   ← isSetup poisoned
08:26:50.947  VSS client setup deferred, retrying in 1000ms (attempt 1/10)
08:26:57.549  Saved value for key 'BIP39_MNEMONIC'
08:26:57.724  VSS client setting up…
08:26:57.848  VSS client setup with server: 'https://bitkit.to/vss_rs_auth'       ← reported success
08:26:57.849  Failed to get VSS backup timestamp: MnemonicNotAvailableException   ← stale error from isSetup.await()
08:27:01.122  VSS client setup with server: 'https://bitkit.to/vss_rs_auth'
08:27:01.124  VSS client setup succeeded on attempt 5                             ← also hollow
```

### Bitkit Version

2.5.0 (mainnet `to.bitkit`, `release-2.5.0` @ `eb77c8953`). Not a 2.5.0 regression: poisoning on a missing mnemonic was introduced on 2026-01-21 by `b3b7e5a6f` (PR #678 "fix: mnemonic not found crash"), which changed the missing-mnemonic path from `return@runCatching false` (no effect on the deferred; a later attempt could still succeed) to `throw MnemonicNotAvailableException()` inside the `runCatching` whose `onFailure` calls `isSetup.completeExceptionally()`. `setupWithRetry()` was written for the old semantics and kept retrying a client that can no longer recover. The same one-shot handling for other setup errors predates that (`e573b844d`, 2025-07-31). Identical in v2.4.1.

### Device / OS

Samsung Galaxy S22 (SM-S901B), Android — mainnet. Also reproducible on the JVM (unit test).

### Reproducibility

Always (once the first `setup()` in a process has failed)

### Additional context

- Not RN-triggered. Pure native defect in `app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt`: `setup()` L37-73 (guard L40, `complete` L66, `completeExceptionally` L70), `setupWithRetry()` L82-108, `reset()` L110-117, `isSetup.await()` in `putObject` L123 / `getObject` L138 / `listKeys` L156 / `deleteObject` L168 / `deleteAllKeys` L184.
- Callers: `BackupRepo.startObservingBackups()` L141-164 (`setupWithRetry`), `BackupRepo.getLatestBackupTime()` L734-750 (swallows every failure into `null`), `BackupRepo.triggerBackup()` L459-488.
- Related: #1254 (incident where this selected the RN backup), #1257 (the trigger in that incident), #678 (introduced the poisoning), #691 (concurrent `setup()` guard added in the same code).
- Why it went unnoticed: `VssBackupClientTest` (added in #678) covers repeated failure but never fail-then-succeed, because a successful `setup()` needs the native FFI and nobody static-mocked it; in production the first `setup()` of a process normally succeeds (mnemonic present, network up), so the poisoned path only runs after a failure, and its symptom — backups silently stopped until the next app restart — looks like transient network trouble.
- Environment: mainnet, wallet migrated from RN. Lightning channels not affected (ldk-node uses its own VSS store via `VssBackupClientLdk`); the damage is app-level backups (METADATA, ACTIVITY, SETTINGS, WIDGETS, WALLET, BLOCKTANK) and restore-source selection.
- iOS: worth checking whether the Swift `VssBackupClient` has the same one-shot state (not verified).

### LINKED ISSUE #1257: [Bug]: Wipe wallet races an in-flight node start: backup observers run on a wiped keychain and the following restore gets stuck on initializing
### What happened?

`WalletViewModel.wipeWallet()` → `WipeWalletUseCase` neither waits for nor cancels a `LightningRepo.start()` that is already in flight. When Reset wallet is tapped while the node is being built, the wipe and the start interleave and produce three separate failures. Seen on mainnet during 2.5.0 QA (same session as #1254), all times UTC:

- 08:21:57 app backgrounded → node stop begins; the stop hangs until 08:26:29 (4.5 min).
- App foregrounded meanwhile; its `start()` waited on `lifecycleMutex` and began building the node at 08:26:30 (`Building node…`, build ≈ 20 s).
- 08:26:49.266 user taps Reset → `backupRepo.reset()`, then `keychain.wipe()` (`Reset keychain encryption key and deleted all '4' entries` 08:26:50.956).
- 08:26:50.918 **`Node started`** — the pre-wipe start finishes in the middle of the wipe. Lifecycle goes `Running` → `BackupsViewModel.observeAndSyncBackups` → `BackupRepo.startObservingBackups()` → `VssBackupClient.setupWithRetry()` runs against the already wiped keychain → `VSS client setup error [MnemonicNotAvailableException]` (08:26:50.944, retries 51.9 / 53.9 / 56.9). This is what poisons `VssBackupClient` (#1256) and ultimately sent the restore down the RN path in #1254.
- 08:26:51.096 `wipeStorage called, stopping node first` → `stop()` blocks on `lifecycleMutex`, which `start()` holds until its initial sync completes (08:27:00.4).
- The UI had already switched to onboarding (`resetWalletState()` runs before `lightningRepo.wipeStorage()`), so the user restored the seed at 08:26:57.549 while the wipe was still pending.
- `ContentView` re-entered → `ON_START` → `walletViewModel.start()` **returned early**: `isStarting` was still `true` from the pre-wipe start (it cleared at ≈08:27:01). Nothing started the node afterwards. The user sat on the initializing screen from `RN migration completed` (08:27:23.9) until force-killing the app at 08:36:16 (`ActivityManager: Killing 1936:to.bitkit … remove task`).
- 08:27:01.477 `LDK storage wiped` — **4 s after the new wallet's mnemonic had been saved**.

### Expected behavior

- Wipe must not run concurrently with a node start: cancel the in-flight start or await it, and keep the UI on the wipe until `wipeStorage()` has completed.
- `BackupRepo.startObservingBackups()` must be a no-op while `_isWiping` is set; backup observers must never call `setup()` when no mnemonic is present.
- A restore-triggered `start()` must not be dropped because `isStarting` is still `true` from a previous, unrelated start; re-check once that start finishes.
- `lightningRepo.wipeStorage()` must never execute after a new mnemonic has been saved.

### Steps to Reproduce

1. Mainnet (or regtest) wallet with channels / payment history so the background stop takes long.
2. Background the app for a few minutes, foreground it, and while `Building node…` is in the log go to Settings → Reset wallet.
3. Immediately restore any seed.
4. Observe: `VSS client setup error` right after `Node started`; `Stopped observing backup statuses` at restore; `LDK storage wiped` after `Saved value for key 'BIP39_MNEMONIC'`; no `Starting node…` after the restore — the initializing screen never finishes until the app is killed.

Same device session as #1254 — see the log excerpt attached there: [2026-09-14-reset-restore-loses-tags-logs.zip](https://github.com/user-attachments/files/32189950/2026-09-14-reset-restore-loses-tags-logs.zip). The lines below are from the full logcat of that session (UTC):

```
08:21:57.300  Stopping node…                                   (background stop; hangs)
08:26:29.187  Node stopped
08:26:30.088  Building node…                                   (foreground start, queued behind the stop)
08:26:49.266  VSS client reset                                 (WipeWalletUseCase begins)
08:26:50.918  Node started                                     ← pre-wipe start finishes mid-wipe
08:26:50.922  Start observing backup statuses and data store changes
08:26:50.944  VSS client setup error [MnemonicNotAvailableException]
08:26:50.956  Reset keychain encryption key and deleted all '4' entries
08:26:51.096  wipeStorage called, stopping node first          ← waits on lifecycleMutex
08:26:57.443  Stopped observing backup statuses and data store changes   (restore → Initializing)
08:26:57.549  Saved value for key 'BIP39_MNEMONIC'             ← new wallet restored
08:27:00.515  Stopping node…
08:27:01.458  Node stopped
08:27:01.477  LDK storage wiped                                ← after the new mnemonic was saved
08:27:23.904  RN migration completed, marked for post-migration sync
              (no Starting node… until 08:36:14; process killed 08:36:16 by "remove task")
```

### Bitkit Version

2.5.0 (mainnet `to.bitkit`, `release-2.5.0` @ `eb77c8953`). Same flow in v2.4.1 — not a 2.5.0 regression.

### Device / OS

Samsung Galaxy S22 (SM-S901B), Android — mainnet.

### Reproducibility

Sometimes (<50%)

### Additional context

- Not RN-triggered. Native lifecycle race; the RN restore in this session is incidental.
- Code (`release-2.5.0`): `WipeWalletUseCase.kt` L47-90 (`backupRepo.reset()` L55, `keychain.wipe()` L64, `resetWalletState()` L77, `lightningRepo.wipeStorage()` L81); `LightningRepo.wipeStorage()` L811-827 → `stop()` L616-643 (blocks on `lifecycleMutex`); `LightningRepo.start()` L320+ holds `lifecycleMutex` through setup, start and the initial `sync()`; `WalletViewModel.start()` L278-306 (`isStarting` guard L283); `BackupsViewModel.observeAndSyncBackups()` L49-58; `BackupRepo.startObservingBackups()` L141-164 (no `_isWiping` check).
- Related: #1254 (incident), #1256 (consequence), #1252 (node start/stop recovery; the 4.5-minute hung background stop that set this up is the same stop path).
- Why it went unnoticed: the race window is the node build/start (2–25 s on every foreground, see #1252), and Reset is normally tapped long after the app has settled. Here the window lined up only because a hung background stop (4.5 min) deferred the build to the moment Reset was tapped. The visible symptom — stuck on initializing, force-kill, works after relaunch — reads as a one-off glitch, and no e2e covers Reset during node startup.
- Environment: mainnet, wallet migrated from RN, 3 open channels. Channels were intact after relaunch (loaded from ldk-node's VSS store); the LDK-dir wipe after the new mnemonic did not cause loss here because the node never started in that process, but the ordering is unsafe.

ovitrif:
Fixes #1256
Fixes #1257

This PR:
1. Makes the VSS backup client recover after a failed setup instead of staying broken until the app restarts
2. Stops the node before the keychain is wiped on Reset, so a wipe no longer races an in-flight node start

### Description

- Resets the setup state when setup fails, so a later successful setup actually makes the client usable. Previously the first failure was latched: every following setup reported success without doing anything, and every VSS call threw the stale error. Setup also captures its own gate, so a reset that swaps the gate mid-setup can no longer be completed by the old wallet's client.
- Restructures the wallet wipe into three phases. First it resets the backup clients and stops the node; this is the only step that can fail the reset, and it fails before anything is destroyed, so an in-flight node start is awaited under the lifecycle mutex instead of raced. Then it runs the remote Paykit and Pubky cleanup, which needs the keychain session, as best-effort. Finally it wipes local state, with the LDK storage wipe as the last abort point: if the node directory cannot be removed, nothing else is destroyed and the reset fails with the wallet intact. Every later local step is logged on failure and never aborts the rest; if the keychain wipe itself fails, the remaining steps still run, the reset reports a failure, and the wallet-exists state is re-read from the keychain so the user stays in the wallet where Reset can be retried. A reset cannot leave a wallet that still looks alive with its node data gone, and the LDK directory is never wiped after a new mnemonic was saved.
- Skips starting the backup observers while a wipe is in progress, so they never call VSS setup without a mnemonic, and restarts them after a failed node stop when the node is still running.
- Rejects a second Reset while one is in flight, and shows the Reset button in a loading state with back navigation and the Backup button disabled until the wipe completes.
- Makes `LightningRepo.stop()` really stop a node object that a failed start left alive instead of short-circuiting on the Stopped state, so Reset during the start retry window tears the node down before any cleanup rather than failing later at the storage wipe. The storage wipe now holds the lifecycle lock across the stop and the directory removal, and any node start arriving during the wipe is refused under that lock instead of queued, so nothing can rebuild the old seed's node before the keychain is gone.
- Applies the same setup recovery to the LDK VSS client used by Reset network graph and Reset pathfinding scores.
- Motivation: one failed setup (network or auth hiccup at first node start, or the wipe race from [#1254](https://github.com/synonymdev/bitkit-android/issues/1254)) used to silently kill all VSS backups for the session and could send a restore to the RN backup.

#### Out of Scope

- `WalletViewModel.start()`: a restore-triggered start dropped by a stale `isStarting` flag ([#1257](https://github.com/synonymdev/bitkit-android/issues/1257), remaining part).
- `BackupRepo` restore picker: falling back to the RN backup when the VSS lookup returns null ([#1254](https://github.com/synonymdev/bitkit-android/issues/1254)).
- Surfacing setup failures in the UI or adding retries beyond the existing setup-with-retry path.
- Cancelling an in-flight node start from Reset; the wipe waits for it as before, now visibly.

### Design

`ResetAndRestoreScreen` gains a loading state on the Reset button while wiping. N/A — no design available.

### Preview

N/A

#### Journeys
N/A — no backup/restore journey exists yet.
#### Manual Tests
- [x] **1.** Onboarding → New Wallet: node starts and all backup categories upload (`Backup succeeded` in logs).
- [x] **2.** Settings → Security → Reset and Restore → Reset → New Wallet: logs show backup reset, node stopped, remote cleanup, LDK storage wiped, then keychain wiped; the new wallet's node starts and backups upload.
- [x] **3.** Reset and Restore → Reset → Yes, Reset: Reset button shows a spinner, Backup button and back are disabled until onboarding shows.

The failed-setup path itself could not be reproduced on-device: setup does no network I/O and the app skips node start while offline, so the only real trigger is the wipe race from [#1254](https://github.com/synonymdev/bitkit-android/issues/1254). That path is proven by the unit tests only.
#### Automated Checks
- Unit tests added: cover a successful setup and setup-with-retry after a failed attempt leaving the client usable in `VssBackupClientTest.kt`, the same recovery in `VssBackupClientLdkTest.kt`, and observers being skipped while wiping in `BackupRepoTest.kt`.
- Unit tests added: cover the stop of a node left alive by a failed start, a start being excluded while the storage wipe holds the lifecycle lock, and a start being refused while wiping, in `LightningRepoTest.kt`; wallet existence re-read after an incomplete wipe in `WalletRepoTest.kt`.
- Unit tests added and modified: assert the phase order, the re-entrancy guard, completion despite later local step failures, the reported failure when the keychain wipe fails, the no-wipe path when the LDK storage wipe fails, and the no-wipe plus observer restart path when the node stop fails in `WipeWalletUseCaseTest.kt`.
- Local: `just compile`, `just test`, and `just lint` pass.
- On-device (Pixel 10 emulator, dev build): Reset → new wallet logs the phase order (backup reset, node stopped, remote cleanup, LDK storage wiped, then keychain wiped), the new wallet's node starts, VSS setup succeeds on attempt 1 and every backup category uploads. Reset shows the loading state on the Reset button with Backup and back disabled until onboarding appears.


app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
```diff
@@ -31,15 +31,14 @@ class VssBackupClient @Inject constructor(
...
     private var isSetup = CompletableDeferred<Unit>()
     private val setupMutex = Mutex()
 
     suspend fun setup(walletIndex: Int = 0): Result<Unit> = withContext(ioDispatcher) {
         setupMutex.withLock {
             runCatching {
-                if (isSetup.isCompleted && !isSetup.isCancelled) {
-                    runCatching { isSetup.await() }.onSuccess { return@runCatching }
-                }
+                if (isSetup.isCompleted) return@runCatching
```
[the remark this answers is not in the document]
ovitrif: Good catch, the guard was not part of the fix. Restored it as `isCompleted && !isCancelled` in e1c2a83; a failed setup now replaces the deferred anyway, so the recovery path is unaffected.

piotr-iohk (commented):
QA review from the diff and thread only — not run on device.

1. **[gap]** `VssBackupClientLdk.setup()` still does `isSetup.completeExceptionally(it)` without replacing the deferred — the same latch fixed here in `VssBackupClient`. Callers are `clearNetworkGraph` and `resetPathfindingScores` (`LightningRepo.kt:665`, `:1904`): one failed LDK-client setup makes every later "Reset network graph" / "Reset pathfinding scores" fail with the stale error and restart the node until the app is restarted. Same two-line fix; arguably in #1256's scope.

2. **[+1 greptile]** The partial-wipe window (LDK deleted, keychain kept) is real but narrow — I checked the steps between `wipeStorage` and `keychain.wipe()` and only `privatePaykitAddressReservationRepo.clear()` and `keychain.wipe()` itself can throw. Still, if it happens the app presents a wallet whose channel state is gone. Consider calling `resetWalletState()` regardless of failure once `wipeStorage` has succeeded, or treating the remaining local wipes as non-fatal.

3. **[gap]** Reset is now a long blocking operation with no UI feedback. `ResetAndRestoreScreen` never clears `showDialog` on confirm, so the dialog stays up until `resetWalletState()`, which now runs after `stop()` — the full node build + initial sync, or minutes in the #1257 stop-hang case. `rememberDebouncedClick` only debounces the tap; nothing stops a second `wipeWallet()` mid-flight, and `WipeWalletUseCase` has no re-entrancy guard, so the second run's `finally { setWiping(false) }` can clear the flag while the first is still wiping. Before this PR the onboarding switch was near-immediate so the window was tiny. A `wiping` flag on the dialog/button or a mutex in the use case would close it.

4. **[nit]** If `stop()` fails, `wipeStorage().getOrThrow()` aborts after `backupRepo.reset()` already stopped the observers and reset the VSS clients. Wallet survives, but backups are silently off until the next `Running` transition (app restart). Pre-existing ordering, new abort point — worth restarting observers on that failure path.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
```diff
@@ -53,6 +53,7 @@ class WipeWalletUseCase @Inject constructor(
         return try {
             runSuspendCatching {
                 backupRepo.reset()
+                lightningRepo.wipeStorage(walletIndex).getOrThrow()
```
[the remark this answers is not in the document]
ovitrif: Agreed. Fixed in 2b5431e by wiping the keychain right after the LDK storage, so any later cleanup failure leaves the same state as before this PR (no keychain, no node data) instead of a wallet without its node data.

ovitrif:
> [Original review](https://github.com/synonymdev/bitkit-android/pull/1266#pullrequestreview-5210285398)
>
> `VssBackupClientLdk.setup()` still does `isSetup.completeExceptionally(it)` without replacing the deferred
>
> The partial-wipe window (LDK deleted, keychain kept)
>
> no UI feedback
>
> backups are silently off

@piotr-iohk Thanks, all four addressed in 2b5431e:

1. `VssBackupClientLdk.setup()` now replaces the deferred on failure, same as `VssBackupClient`; covered by `VssBackupClientLdkTest.kt`.
2. `keychain.wipe()` moved right after `wipeStorage()`, so a later cleanup failure leaves the pre-PR state (no keychain, no node data) rather than a wallet without its node data.
3. `WipeWalletUseCase` takes a `tryLock` mutex and returns `WipeAlreadyInProgress` for a second call. The screen dismisses the dialog on confirm and, while `BackupRepo.isWiping`, shows the Reset button loading, disables Backup, and blocks back navigation.
4. On wipe failure the use case restarts the backup observers when the node is still Running.

ovitrif:
An agent is running and will check the Test 3 checkbox if it succeeds.

piotr-iohk (dismissed):
Re-reviewed `2b5431e` from the diff — not run on device. Items 1, 3 and the Greptile P1 look good; two new issues from the fix-up:

1. **[blocker]** `keychain.wipe()` now runs before `removePublishedEndpointsForCleanup`, `removeBitkitPaymentEndpoints` and `closeAndClear`. All three go through `PaykitSdkService`, whose `PaykitSdkSessionProvider.loadSessionAccess()` reads `PAYKIT_SESSION` from the keychain on every call and returns `null` once it is gone (`liveSessionAccess` is only returned when its secret matches the keychain value). So `syncPublicEndpoints(emptyList())` / `removePaykitReceiverMarker()` run without a session, fail, and are swallowed — and the `contactSharingCleanupPending` flag is then erased by `settingsStore.reset()`. Result: the wiped wallet's public endpoints and receiver marker stay published after Reset; on master they were removed. Every step between `wipeStorage` and the old `keychain.wipe()` position returns `Result` or swallows (only `privatePaykitAddressReservationRepo.clear()` can throw), so moving `keychain.wipe()` back after `pubkyRepo.wipeLocalState()` keeps Greptile's window closed in practice; wrap `clear()` if you want it airtight.

2. **[gap]** The observer restart in `onFailure` is dead code: `backupRepo.startObservingBackups()` runs while `_isWiping` is still `true` (cleared in `finally`), so the guard added in this PR returns early with "Skipped observing backups while wiping". The unit test only passes because `BackupRepo` is a mock. Call `setWiping(false)` before the restart, or move the restart after the `finally`.

ovitrif:
> [Original review](https://github.com/synonymdev/bitkit-android/pull/1266#pullrequestreview-5211241612)
>
> `keychain.wipe()` now runs before `removePublishedEndpointsForCleanup`, `removeBitkitPaymentEndpoints` and `closeAndClear`.
>
> The observer restart in `onFailure` is dead code

@piotr-iohk Both addressed in 47187b3ac:

1. `keychain.wipe()` is back after `pubkyRepo.wipeLocalState()`, so the Paykit and Pubky cleanup still see the session. `privatePaykitAddressReservationRepo.clear()` is wrapped and logged, so nothing between `wipeStorage()` and the keychain wipe can abort anymore; covered by a new case in `WipeWalletUseCaseTest.kt`.
2. The observer restart now runs after the `finally` that clears `_isWiping`, so the guard no longer short-circuits it. The test asserts `setWiping(false)` precedes `startObservingBackups()`.

ovitrif:
Handed off to `pr-babysit` skill which watches over reviews and CI. Warning: first run of this skill.

ovitrif:
> [Original review](https://github.com/synonymdev/bitkit-android/pull/1266#pullrequestreview-5211241612)
>
> two new issues from the fix-up:

@piotr-iohk Both addressed in 43449af, which replaces the previous fix-up rather than moving lines around again. The wipe is now phased:

1. Reset backup clients and `lightningRepo.stop()`. This is the only step that can fail the reset, and it fails before anything is destroyed. On failure the observers restart after the wiping flag is cleared, so the guard no longer short-circuits it (your point 2).
2. Remote Paykit and Pubky cleanup, best-effort, while the keychain session still exists (your point 1).
3. Local wipe: LDK storage, reservations, Pubky local state, keychain, FCM token, Core, Room, stores, repo state. Every step is wrapped and logged and never aborts the rest, so the partial-wipe window is closed by construction rather than by ordering.

`WipeWalletUseCaseTest.kt` covers the phase order, completion despite local step failures, and the no-wipe plus observer restart path when the stop fails. Verified on the emulator: the log shows reset, node stopped, remote cleanup, LDK wiped, keychain wiped, then onboarding.

app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
```diff
@@ -31,15 +31,14 @@ class VssBackupClient @Inject constructor(
...
     private var isSetup = CompletableDeferred<Unit>()
     private val setupMutex = Mutex()
 
     suspend fun setup(walletIndex: Int = 0): Result<Unit> = withContext(ioDispatcher) {
         setupMutex.withLock {
             runCatching {
-                if (isSetup.isCompleted && !isSetup.isCancelled) {
-                    runCatching { isSetup.await() }.onSuccess { return@runCatching }
-                }
+                if (isSetup.isCompleted && !isSetup.isCancelled) return@runCatching
```
jvsena42: **LOW / hardening.** `setup()` reads `isSetup` as a field at every use rather than capturing it, so a `setup()` that is mid-flight when `reset()` swaps the deferred completes the *replacement*.

The real line is :65 (`isSetup.complete(Unit)`), which isn't in the diff — anchoring here since this is the guard the PR rewrote.

`reset()` (:110-117) does `synchronized(this) { isSetup.cancel(); isSetup = CompletableDeferred() }` and does **not** take `setupMutex`, so it doesn't wait for an in-flight `setup()`. If one is between :43 (mnemonic read) and :65 when that happens, :65 completes the fresh deferred. Then the next wallet's `setup()` short-circuits at this line and every VSS call uses the Rust client built for the *old* wallet — `vss_new_client_with_lnurl_auth` stores into a global and `vssStore`/`vssGet` take no client handle, so the stale client really is reachable. `putObject`/`getObject` await the field (:123, :138), i.e. the already-completed replacement, so it returns immediately and silently: the new wallet's backups land in the old store encrypted with the old key, and on the restore path `getLatestBackupTime()` → `setup().getOrThrow()` (BackupRepo.kt:742) reads the old wallet's store into a wallet restored from a different seed. It persists until process death.

**Why it's LOW rather than what that description sounds like:** I couldn't construct a realistic trigger. `setup()` does no network I/O, so the window is milliseconds of key derivation. The `_isWiping` guard you added at :144 already keeps `startObservingBackups` from launching a fresh `setup()` once a wipe starts, `setupWithRetry` only re-enters on `MnemonicNotAvailableException`, and the other callers are user-driven. Identical code is on v2.4.1, so this is not a regression — the PR only rewrote the guard, added `@Volatile`, and added the `onFailure` reset.

**Fix**, if you want it closed while you're in here — capture the gate under the mutex:

```kotlin
setupMutex.withLock {
    val gate = isSetup
    if (gate.isCompleted && !gate.isCancelled) return@runCatching
    ...
    gate.complete(Unit)
}
```

and in `onFailure`: `gate.completeExceptionally(it); if (isSetup === gate) isSetup = CompletableDeferred()`. `complete()` on the cancelled old gate is a no-op, so the replacement stays incomplete and the next `setup()` rebuilds the global. Three lines here and the same three in `VssBackupClientLdk.kt` (:49, :66). Equally fine to leave it — it's theoretical hardening.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
```diff
@@ -44,52 +46,72 @@ class WipeWalletUseCase @Inject constructor(
...
+    }
+
+    private suspend fun cleanupRemote() {
+        step("remove Paykit published endpoints") { privatePaykitRepo.get().removePublishedEndpointsForCleanup(TAG) }
+        step("remove Bitkit payment endpoints") { pubkyRepo.removeBitkitPaymentEndpoints() }
+        step("close Paykit SDK") { privatePaykitRepo.get().closeAndClear() }
+    }
+
+    private suspend fun wipeLocal(walletIndex: Int, resetWalletState: () -> Unit) {
+        step("wipe LDK storage") { lightningRepo.wipeStorage(walletIndex) }
```
jvsena42: **LOW.** `stopNode()` can succeed *vacuously* while the node object is still alive, and `step()` then swallows the `wipeStorage` failure — so the keychain and stores are destroyed with the old seed's LDK directory left behind.

The PR body says an in-flight node start is "awaited under the lifecycle mutex instead of raced", and "closed by construction". That holds when the start *succeeds*. It doesn't when the start fails into the retry path:

1. `LightningRepo.start()`: `lightningService.setup()` succeeds (node assigned, LightningService.kt:209), `lightningService.start()` throws. The `getOrElse` at :418 reverts state to `initialLifecycleState` — `Stopped` (:427) — **without nulling the node**, releases `lifecycleMutex`, then `delay(2.seconds)` (:439) and retries (:440).
2. Reset confirmed inside that window: `stop()` waits on `lifecycleMutex`, gets it, sees `isStoppedOrStopping()` (LightningRepo.kt:618-621) and returns success without touching the node.
3. This line: `wipeStorage` → `stop()` succeeds vacuously again → `LightningService.wipeStorage` :506 throws `NodeStillRunning` → `step()` logs a warning and continues.
4. :91 `keychain.wipe()`, then core, DB, settings. `onSuccess()` runs, onboarding shows.
5. 2s later the queued retry `start()` runs. `node != null` so setup is skipped, and the **old seed's** node starts on the retained old directory. Create a new wallet and `start()` sees "LDK node already running" — the new seed's wallet drives the old seed's node. Next process start builds the new seed on the old LDK dir and fails.

Trigger B is the same ending via `awaitNodeRelease()` throwing `NodeReleaseTimeout` after 90s when `node.destroy()` wedges.

**Pre-existing, and head is arguably better:** on v2.4.1 `keychain.wipe()` ran unconditionally *before* `wipeStorage(...).getOrThrow()`, with the same vacuous early return, so the same trigger left the same on-disk state — but it surfaced an error and dropped the user on a dead wallet screen instead of onboarding. The PR didn't touch `LightningRepo`/`LightningService` at all; the vacuous gate is entirely theirs. What changed is that the failure is now only a `Logger.warn`.

**Fix**, if you want the first local step to be the gate it reads as — nothing local is destroyed before it, so it's the one safe place to abort:

```kotlin
lightningRepo.wipeStorage(walletIndex).onFailure { return Result.failure(it) }
```

with `wipeLocal` returning `Result` and :59 becoming `stopNode().mapCatching { cleanupRemote(); wipeLocal(...).getOrThrow(); onSuccess() }`. ~4 lines, no new abstractions. It does **not** reverse greptile's P1 (that was aborting *after* LDK deletion) and is compatible with piotr's "treat the remaining local wipes as non-fatal".

One collateral: it flips `invoke should complete the wipe when lightningRepo wipeStorage fails` (WipeWalletUseCaseTest.kt:288), which encodes the never-abort choice. If that choice is deliberate, leave the code and soften the PR body — "closed by construction" is what I'd push back on, not the design.

jvsena42 (commented):
No blockers. Two LOW observations below, both **pre-existing on v2.4.1** and neither introduced by this PR — I'm raising them only because they sit inside the window this PR's title claims to close, and in one case the PR body's wording is stronger than the code.

Neither is a reversal: greptile's guard, piotr's items 1-4 and the phased-wipe design in `43449af` all stay as landed.

**Checked and clean.**

Exits from `ResetAndRestoreScreen`: system Back while wiping is swallowed by `BackHandler(enabled = isWiping) {}` (:76); top-bar back is hidden via `onBackClick = null` (:81); dialog dismiss only flips `showDialog`; Confirm launches on the activity-scoped `WalletViewModel`, so navigating away via the still-enabled drawer icon doesn't cancel the wipe; backgrounding pauses collection but the wipe continues on `viewModelScope`, and `ON_STOP`'s `stopDebounced()` is a no-op against an already-stopped node; on failure `isWiping` clears, a toast fires, and the wallet is intact.

Ordering and lifecycle: remote cleanup doesn't depend on the LDK node, so moving it after `stop()` is safe, and it runs before `keychain.wipe()` so the Paykit session is still present — the ordering piotr asked for. `stop()` can't fail merely because the node never started (`LightningService.stop()` returns early on `node == null`), so Forgot-PIN before node start and `ErrorStarting` still reset. Observer restart on `stop()` failure runs after `setWiping(false)`, so the new `_isWiping` guard in `startObservingBackups` doesn't short-circuit it.

Concurrency: `wipeMutex.tryLock` + `finally` unlock is correct, and a second concurrent wipe gets `WipeAlreadyInProgress`. `step()` uses `runSuspendCatching`, so a genuine `CancellationException` propagates out of the wipe. Cancelling during `stopNode`/`cleanupRemote` leaves the wallet intact — `stop()` is `NonCancellable` inside its lock and nothing destructive has run. An `ON_START` node restart mid-wipe is self-healing: `wipeStorage`'s own `stop()` serialises on `lifecycleMutex`.

Key material: nothing in the touched code logs a mnemonic or passphrase; `VssBackupClient` logs only the VSS/LNURL URLs. `Keychain.wipe()` clears the DataStore and resets the keystore key, and the PR adds no seed-derived artifact that survives it — except the LDK-directory case below. Remote VSS state is intentionally kept for restore; unchanged.

Upgrade: no persisted format changed, and a v2.4.1 VSS backup is read by unchanged `BackupRepo` code.

I also checked the claim that `setup()` does no network I/O and it holds — `VssClient::new_with_lnurl_auth` only derives xprivs and builds the header provider; the JWT is fetched lazily on first request. That's what makes finding 1 below a millisecond-wide window rather than a 30-second one.

app/src/main/java/to/bitkit/data/backup/VssBackupClient.kt
ovitrif: Applied in ee283d43f: `setup()` captures `gate = isSetup` under the mutex, completes that gate, and on failure only replaces `isSetup` when it still is that gate. Same change in `VssBackupClientLdk`.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
ovitrif: Agreed, and taken in ee283d43f: `wipeLocal()` returns a `Result` and the LDK storage wipe is its abort point, since nothing local is destroyed before it. Later local steps stay best-effort. The test now asserts the no-wipe path when `wipeStorage` fails, and the PR body no longer claims the window is closed by construction alone.

jvsena42: `ee283d43f` does what I asked, and I walked trigger A end to end at the new head to confirm: the vacuous `stop()` still returns success, `wipeStorage` still throws `NodeStillRunning`, and `wipeLocal` now returns before :92 `keychain.wipe()`. Nothing local is destroyed. The flipped test is load-bearing — I reverted the three main-code files to `43449afd6` while keeping the new tests and got exactly one failure, `invoke should fail before wiping local state when LDK storage wipe fails`, and it asserts `verify(keychain, never()).wipe()` rather than just a Result.

One consequence of my own request, which I should have called out when I made it. The abort lands *after* `cleanupRemote()`, so it is now the first exit that leaves the wallet alive with remote state already torn down. At the previous head the only fatal step was `stopNode()`, before any cleanup, so "keychain intact + remote torn down" was unreachable except by process death in that window.

What the user is left in on trigger A: mnemonic, passphrase, `PAYKIT_SESSION`, `PUBKY_SECRET_KEY`, LDK dir, Room and settings all intact — but private endpoints unpublished with `contactSharingCleanupPending(false)` so nothing retries, public endpoints synced to empty, the receiver marker removed, `PAYKIT_SDK_STATE` deleted and the cache store reset. The wallet is briefly "Paykit-dark": payers resolving its endpoints get nothing until the next foreground cycle republishes them (`AppViewModel.refreshPublicPaykitEndpoints` / `refreshPrivatePaykitEndpoints`), and the SDK handle rebuilds lazily from the surviving session. Backup observers stay stopped until the next Running transition, which is the same shape as the `stop()`-fails case piotr already accepted.

This is still the better trade — a recoverable half-torn remote beats a destroyed keychain with the old seed's LDK directory left behind — so I am not asking you to change it. Two things worth knowing:

Reset is not permanently blocked in the realistic trigger. 2s after the failed start the queued retry runs, state leaves `Stopped`, and the next `stop()` really stops and nulls the node, so a second Reset succeeds. The one contrived exception is a recovery-mode deeplink inside that 2s window, which makes the retry return `RecoveryModeError` before touching state and leaves every Recovery-screen wipe aborting until relaunch.

If you want the half-torn state unreachable rather than merely recoverable, the root cause is one condition: `LightningRepo.stop()` :618 treats `Stopped` as "nothing to stop" while a failed start leaves `lightningService.node` alive. Requiring `lightningService.node == null` there turns the vacuous success into a real stop and trigger A never reaches `cleanupRemote()`. That is in `LightningRepo`, which this PR deliberately does not touch, so deferring it to the #1256 follow-up is a fine call.

Fix #1 also checks out: the gate is captured once under `setupMutex` and used at the guard, at `complete`, and in `onFailure` in both clients, and `isSetup === gate` keeps a stale setup from clobbering a fresh deferred. I traced the `setupWithRetry` path too — an exceptionally-completed gate is `isCancelled`, so the guard does not short-circuit and the next `setup()` gets a fresh one. No poisoned gate. Neither client has a test covering the race in either direction; just an observation, not a request.

jvsena42 (commented):
`ee283d43f` closes both LOW observations from my last pass. No blockers.

I verified the wipe fix by reverting the three main-code files to `43449afd6` while keeping the new tests: exactly one failure, `invoke should fail before wiping local state when LDK storage wipe fails`. So the test is load-bearing rather than decorative, and it asserts `verify(keychain, never()).wipe()` and `verify(db, never()).clearAllTables()` rather than just a Result. At head, `WipeWalletUseCaseTest` 11/11, `VssBackupClientTest` 6/6, `VssBackupClientLdkTest` 1/1 all pass. Run in a throwaway worktree outside the repo; nothing committed, working tree untouched.

One consequence of the change I asked for is worth your attention before merge — the abort now lands after `cleanupRemote()`, so a recoverable "wallet alive, Paykit endpoints unpublished" state is newly reachable. Details and the recovery path are in my reply on the `WipeWalletUseCase.kt:88` thread. I'm not asking you to change it; it's a better trade than the leftover-LDK-directory outcome it replaces.

Also checked on the new control flow:

- `runSuspendCatching` at :59 replacing `stopNode().map {}` keeps cancellation propagating, and `finally` still clears `isWiping` and unlocks. A side benefit: the non-`step` calls in `wipeLocal` (the `resetState()`s, `resetWalletState()`, `onSuccess()`) now surface as a Result failure and a toast instead of an uncaught exception in `viewModelScope`.
- `keychain.wipe()` is unreachable on the abort path by construction.
- A second concurrent wipe still returns `WipeAlreadyInProgress`.
- The new abort path leaves the user on ResetAndRestore with Back and Reset re-enabled and a toast — it doesn't wedge `isWiping`.
- Re-walked all seven exits from `ResetAndRestoreScreen` (the file didn't change, but the wipe's control flow did): Back while wiping, top-bar back, dialog dismiss, Confirm, drawer navigation mid-wipe, backgrounding, process death before and after the keychain step. All still correct.
- No seed-derived material reaches a log or a toast on any new path — the new lines add no `Logger` calls, and `NodeStillRunning`'s message is a fixed string.

On the VSS gate: the capture is right in both clients, and `isSetup === gate` stops a stale setup clobbering a fresh deferred. I traced the `setupWithRetry` path specifically for a poisoned-gate regression and there isn't one — an exceptionally-completed deferred reports `isCancelled`, so the guard doesn't short-circuit and the next `setup()` gets a fresh gate. Neither client has a test covering that race in either direction; noting it, not asking for one.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
ovitrif: Thanks for walking trigger A through. Took the root fix in ed82850af rather than leaving the half-torn state reachable: `LightningRepo.stop()` now only short-circuits when the state is stopped and `lightningService.node == null`, so a node left alive by a failed start gets a real stop before `cleanupRemote()` runs. Covered by two new cases in `LightningRepoTest.kt`.

jvsena42: `ed82850af` closes trigger A properly. At :618 `Stopped && node != null` is now false, so `stop()` takes the real path and `LightningService.stop()` handles the never-started node cleanly — `listenerJob` is null because `start()` threw before assigning it, `node.stop()` throws `NotRunning` which is swallowed, and `node = null`. `stopNode()` returns success with the node genuinely gone, so `wipeStorage` no longer throws `NodeStillRunning` and the half-torn "wallet alive, remote already cleaned up" state is unreachable via that trigger rather than merely less likely.

I checked the thing I was most worried about — that widening into `LightningRepo` would regress the shipped `stop()` path — and it does not. "Stopped with a live node" is bounded to the 2s retry window plus a recovery-mode edge, because the only writer of a non-null node is `setup()` under `start()`'s mutex, and every other initial state (`Initializing`, `ErrorStarting`) already took the real path before this commit. `stopDebounced`'s 5s delay outlives the 2s retry, so a normal background cycle is unchanged. `LightningService.stop()` cannot realistically throw (`NonCancellable`, `node.stop()` inside `runSuspendCatching`, `releaseHandle` catches destroy failures), so no caller inherits a failure it never used to get. The `node` read is safe: `@Volatile` at LightningService.kt:171, and writes and reads are both under `lifecycleMutex` anyway. In recovery mode the new behaviour is strictly better — that state previously made every Recovery-screen wipe abort.

The tests pin it. Reverting line 618 alone fails exactly `stop tears down a node object left alive by a failed start` with `WantedButNotInvoked: lightningService.stop()`; the second case passes either way and pins the no-op branch. 115/115 at head.

One refinement, LOW, and it is a consequence of the option I offered rather than a new problem. Now that the retry no longer finds a retained node, it does a full `setup()` — and `LightningRepo.wipeStorage` runs `lightningService.wipeStorage` at :815 *outside* the mutex that `stop()` released at :642. So if the 2s retry lands in that gap (most plausibly while `awaitNodeRelease()` is waiting on a wedged `free_node`, which the comment at :471 says takes tens of seconds after a failed start), the retry can take the mutex, rebuild from VSS, and bring up an old-seed node while the wipe deletes the directory and clears the keychain around it. The wallet ends up on onboarding with the old seed's node running behind it.

Genuinely LOW: it needs a start that fails into the retry path from a `Stopped` initial state, a Reset confirmed and `cleanupRemote()` completed inside ~2s of a failure the UI never surfaces, and either the destroy wedge or a millisecond-exact landing. In the likelier ordering the retry holds the mutex through its rebuild, `wipeStorage`'s `stop()` waits, and the wipe then succeeds correctly — just slower. The underlying race is not on v2.4.1: there `keychain.wipe()` ran before `wipeStorage`, so a `start()` in that gap failed `build` with `MnemonicNotFound`. It became reachable when a9e08f82f moved the keychain wipe after the node stop, and this commit adds a second trigger to it.

If you want it closed here, the narrow version is to hold `lifecycleMutex` across both halves of `wipeStorage`: split `stop()` into a `stopLocked()` and have `wipeStorage` do `lifecycleMutex.withLock { stopLocked().mapCatching { lightningService.wipeStorage(walletIndex); … } }`. Both `setup()` and `wipeStorage` already gate on `awaitNodeRelease()`, so nothing else moves. Equally defensible to leave it for the #1256 follow-up, given it predates this PR — your call, and not a blocker either way.

jvsena42 (commented):
`ed82850af` is correct. No blockers, nothing new to raise as its own thread — one LOW refinement is on the `WipeWalletUseCase.kt:88` thread, and it's a consequence of the change I asked for rather than a defect you introduced.

This commit reaches into `LightningRepo`, which the PR had deliberately left alone, so I reviewed it as a change to shipped lifecycle code rather than as a one-line follow-up. The question that mattered was whether making the early return conditional regresses `stop()` on the hot path. It doesn't:

- **"Stopped with a live node" is not a routine state.** The only writer of a non-null node is `setup()` under `start()`'s mutex, and it's reachable only when `initialLifecycleState` was `Stopped` — bounded to the 2s retry window, plus a recovery-mode edge. `Initializing` and `ErrorStarting` already took the real path before this commit, so a real stop against a set-up-but-unstarted node was already exercised (`restartWithElectrumServer` failure → `restartWithPreviousConfig`).
- **`stopDebounced`'s 5s delay outlives the 2s retry**, so a normal background cycle sees no new behaviour. In recovery mode the change is strictly better: `start()` returns early without touching state, so `Stopped + live node` used to persist and make every Recovery-screen wipe abort.
- **No caller inherits a new failure.** `LightningService.stop()` can't realistically throw — it's `NonCancellable`, `node.stop()` is inside `runSuspendCatching`, and `releaseHandle` catches `destroy()` failures. I walked all ten `stop()` call sites anyway; the ones that ignore the result (`LightningNodeService` ×2, `WakeNodeWorker`, `onProceedWithoutRestore`) get the intended outcome, and the ones that handle failure already did.
- **The `node` read is safely published**: `@Volatile` at `LightningService.kt:171`, and both the write at :209 and the read at :618 are under `lifecycleMutex` regardless.
- **The fall-through emits `Stopping` → `Stopped` and replaces the whole `LightningState`**, which the early return didn't. Every real stop already does this and observers tolerate it; the retry re-derives `isGeoBlocked`. No probe-cache leak, since a node that never ran has emitted no events.

Tests are load-bearing: reverting line 618 alone fails exactly `stop tears down a node object left alive by a failed start` with `WantedButNotInvoked: lightningService.stop()`. The second case passes in both states, which is right — it pins the no-op branch. 115/115 at head. Throwaway worktree outside the repo, nothing committed, main checkout untouched.

Also confirmed the previous commit's work is byte-identical — the captured VSS setup gate and the fatal LDK-wipe step are untouched, and `git diff --stat ee283d43f ed82850af` is just these two files.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
ovitrif: Closed here in effa097: `stop()` delegates to a private `stopLocked()`, and `wipeStorage()` holds `lifecycleMutex` across `stopLocked()` and `lightningService.wipeStorage()`, so a start cannot take the lock between the stop and the directory removal. Pinned by a new `LightningRepoTest.kt` case that blocks the wipe inside the service call and asserts a concurrent `start()` does not reach the node until the wipe releases; it fails without the lock.

piotr-iohk (commented):
Re-reviewed `43449af`..`effa097` from the diff — not run on device. Both items from my last pass are fixed, and the phased design plus the `stop()`/lock hardening from the jvsena42 threads look right. One consequence of the phases, fine to take as a follow-up:

1. **[gap]** `keychain.wipe()` is now a best-effort `step()`. If it throws, `wipeLocal()` still returns success, `resetWalletState()` and `onSuccess()` run, and `setWalletExistsState()` sees the mnemonic still there → `walletExists = true`. The user lands back in a wallet whose LDK dir, Core data, Room and settings are gone but seed and PIN remain, with only a `warn` log; before the rewrite this aborted with a toast. Rare (DataStore/Keystore write failure) and recoverable (LDK state rebuilds from VSS, a second Reset retries), so not a merge blocker — but either make `keychain.wipe()` a second abort point (fail the `Result` so the toast fires) or toast when any local step fails. Past `wipeStorage`, "success" should mean onboarding.

ovitrif:
> [Original review](https://github.com/synonymdev/bitkit-android/pull/1266#pullrequestreview-5221045706)
>
> either make `keychain.wipe()` a second abort point (fail the `Result` so the toast fires) or toast when any local step fails.

@piotr-iohk Taken in 38c9df844, your first option: `keychain.wipe()` is now the second abort point. The remaining local steps still run so the stores stay consistent, then `wipeLocal()` returns `WipeIncomplete` ("please reset again"), `onSuccess()` is skipped and the toast fires. Past `wipeStorage`, success now always means onboarding. Covered by a new case in `WipeWalletUseCaseTest.kt`.

piotr-iohk (dismissed):
QA from the diff plus emulator (emulator-5554, CI APK 38c9df8). Happy-path Reset and New Wallet backups pass; Pubky cleanup ran with a session before the keychain wipe.

1. **[nit]** If `keychain.wipe()` throws, `resetWalletState()` still sends the user to onboarding while the mnemonic is on the keychain, and the toast says “please reset again” from a screen that no longer has Reset. Rare, recoverable. Not a merge blocker.

On device:
- Reset while the node was Running landed on TOS with the expected order (observers stopped → VSS reset → node stopped → remote cleanup → LDK wiped → keychain wiped). New Wallet then set up VSS on attempt 1 and uploaded METADATA / WIDGETS / ACTIVITY. No `MnemonicNotAvailableException`.
- Re-tested on a wallet with Pubky + a contact: private Paykit cleanup warned `PrivateUnavailable` (best-effort); public cleanup then issued `delete_payment_endpoint` (`btc-regtest-p2wpkh`) and `delete_paykit_receiver_marker` **before** keychain wipe. No `identity_error` / `no Pubky session`. Did not confirm from a second device that the old identity is unresolvable.
- Reset-during-`Building node…` and the double-tap guard were not reproduced (wipe ~0.5s without Paykit; ~6s with).

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
```diff
@@ -44,52 +46,76 @@ class WipeWalletUseCase @Inject constructor(
...
+        step("reset settings") { settingsStore.reset() }
+        step("reset cache") { cacheStore.reset() }
+        step("clear watch-only accounts") { watchOnlyAccountRepo.clear() }
+        step("reset widgets") { widgetsStore.reset() }
+        blocktankRepo.resetState()
+        activityRepo.resetState()
+        hwWalletRepo.resetState()
+        resetWalletState()
+        step("mark migration checked") { migrationService.markMigrationChecked() }
+        return if (keychainWiped) Result.success(Unit) else Result.failure(WipeIncomplete())
```
jvsena42, line 107: **LOW.** If `keychain.wipe()` fails, the user lands on onboarding with the seed still in the keychain. The toast says "please reset again", but onboarding has no Reset, and Create and Restore both fail until the app is relaunched.

Trace:
1. `keychain.edit { it.clear() }` throws, for example a DataStore `IOException` (`Keychain.kt:165`). `keychainWiped = false` (:92) and the remaining steps run.
2. `resetWalletState()` (:103) is `WalletRepo.resetState()`, which replaces the state with `WalletState()`, whose default is `walletExists = false` (`WalletRepo.kt:807`). `onSuccess` (`setWalletExistsState`) is skipped here, so nothing sets it back.
3. `MainActivity.kt:144`: `!walletExists && !isRecoveryMode` shows `OnboardingNav`. `wipeStorage` already cleared recovery mode.
4. Create wallet runs `keychain.saveString(BIP39_MNEMONIC, …)` (`WalletRepo.kt:440`), and `Keychain.save` throws `FailedToSaveAlreadyExists` because the old mnemonic is still stored (`Keychain.kt:121`). Restore fails the same way at `WalletRepo.kt:459`. Every button leads to an error toast.
5. After a relaunch `walletExists()` reads the keychain again and the user is back in the partly wiped wallet, where a second Reset works.

This regressed in this commit. At `ed82850af` the same failure left the user in the wallet, which is what piotr flagged. On v2.4.1 a keychain throw aborted everything before `resetWalletState()`. piotr asked for the failure to show a toast, not to land on onboarding.

LOW because a failed keychain DataStore write is rare and a relaunch recovers.

**Fix**, one line in `WalletRepo.wipeWallet` (`WalletRepo.kt:475-481`, not in the diff): `.onFailure { if (it is WipeIncomplete) setWalletExistsState() }`. It reads the keychain again, so the result is correct in both sub-cases. If `clear()` failed, the mnemonic still exists: the user stays in the wallet and "reset again" is something they can do. If only `resetEncryptionKey()` failed, the entries are gone and onboarding is the right place. `assertFalse(onSetWalletExistsStateCalled)` in the new test stays green, because the fix lives in the repo.

jvsena42 (commented):
Reviewed `effa097b7` and `38c9df844`. No HIGH/MEDIUM. Two LOWs: one inline below, and one as a reply on the `WipeWalletUseCase.kt:88` thread, because it refines the lock I asked for there.

**Checked and clean**

- **`stop()` / `stopLocked()` split.** The body is equivalent: `return Result.success(Unit)` replaces `return@withLock`, and `runCatching`/`NonCancellable`/`onFailure` are unchanged. `wipeStorage` is still `withContext(bgDispatcher)`, so `stopLocked` runs on the same dispatcher.
- **No re-entrant lock.** Nothing inside `wipeStorage`'s hold acquires `lifecycleMutex` again. `LightningService.stop()` → `listenerJob?.cancelAndJoin()` cancels any handler waiting on the mutex, since `Mutex.lock()` is cancellable. `releaseHandle`/`awaitNodeRelease` join an `ioDispatcher` job that only calls `destroy()`. `clearProbeOutcomes` and `setRecoveryMode` are map and StateFlow writes. All six `stop()` call sites enter through the public `stop()`.
- **Hold duration.** The lock is now also held across `awaitNodeRelease()` (≤90s) and the delete. The callers that can block are `start()`, `stopDebounced`, `restartWithPreviousConfig` and `WakeNodeWorker`. The worker's 2-minute `withTimeout` covers the deliver signal, not `start`, so a long hold delays it without failing it. Every entry point is on `bgDispatcher`, so there is no main-thread wait.
- **Second Reset after `WipeIncomplete`.** `deleteRecursively()` on a missing directory returns true, `awaitNodeRelease` is a no-op on a null job, and `cleanupRemote` without a session is swallowed by `step()`. A retry from inside the wallet works.
- **Cancellation.** `step()` still uses `runSuspendCatching`. A cancellation that `stopLocked().mapCatching` wraps inside `wipeStorage` is rethrown by `getOrThrow()` inside the outer `runSuspendCatching`. Nothing local is destroyed before `deleteRecursively`.
- **`Keychain.wipe()` partial failure.** `edit { clear() }` is atomic. If it succeeds and `resetEncryptionKey()` throws, the entries are gone and the toast is misleading but harmless.
- **Callers of the new failure.** `WalletRepo.wipeWallet` is the only caller of the use case. `WalletViewModel`, `RecoveryViewModel` and ForgotPin show the failure as a toast; `DevSettingsViewModel` ignores it. Only the use case calls `lightningRepo.wipeStorage`.
- **Tests.** `wipeStorage holds the lifecycle lock…` fails without the lock. `invoke should fail after wiping the rest when keychain wipe fails` pins the `WipeIncomplete` result. See the inline comment for the landing it also pins.
- **Logs.** No new log line touches seed material, and the `WipeIncomplete` message is a fixed string.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
jvsena42: **LOW, refinement of effa097.** The lock does what it says: nothing can rebuild the node while the directory is being deleted. It queues a `start()` though, it does not refuse one. The queued start runs as soon as the lock is released, and the keychain is still intact at that point.

Trace at head:
1. A `start()` arrives while `wipeStorage` holds `lifecycleMutex` (`LightningRepo.kt:815`). Possible sources: ContentView ON_START (`walletExists` stays true until `resetWalletState()` at `WipeWalletUseCase.kt:103`, and `WalletViewModel.start()` only checks `walletExists`/`isStarting`), `WakeNodeWorker.kt:85`, `LightningNodeService.kt:96`, or the 2s retry. None of these paths looks at `isWiping`. It queues at `LightningRepo.kt:343`.
2. `wipeStorage` deletes the directory and releases the lock at :827. The queued start sees `Stopped` with `node == null`, so it runs `setup()`, and `LightningService.build` reads `BIP39_MNEMONIC` (:277).
3. Meanwhile the use case runs `privatePaykitAddressReservationRepo.clear()` (:90), then `pubkyRepo.wipeLocalState()` (:91), and only then `keychain.wipe()` (:92). If `build` reads the mnemonic first, `buildWithVssStore` rebuilds the old seed from VSS and the node reaches `Running`. The wipe then finishes and onboarding shows with that node still running.
4. Create wallet: `setInitNodeLifecycleState()` sets `Initializing`, so `start()` skips setup because `node != null`, hits "LDK node already running" (:373), and the new wallet drives the old seed's node until process death. An address or invoice shown in that session belongs to the old seed.

The last assertion in `wipeStorage holds the lifecycle lock so a start cannot rebuild the node mid-wipe` (`LightningRepoTest.kt:893`, `times(2)`) pins this: the queued start builds and starts once the wipe releases.

This corrects my earlier comment, which I have edited. The race is not on v2.4.1. There `keychain.wipe()` ran before `wipeStorage`, so a start in this gap failed with `MnemonicNotFound`. The reorder in a9e08f82f opened it, not effa097.

LOW because the window is short. Phase 1 already stopped the node, so the `stopLocked()` inside `wipeStorage` is a no-op. The lock is held for `awaitNodeRelease()` plus the delete, which takes milliseconds unless `free_node` is wedged (up to 90s, most likely after a failed start). On top of that the build has to win a race of tens of milliseconds against :90-92. The outcome is bad when it does land, which is why I'm raising it.

**Fix**, if you want it closed: refuse the start rather than queue it. Add a wipe flag to `LightningRepo`, set and cleared next to `backupRepo.setWiping` (`WipeWalletUseCase.kt:57`, :66), and check it *inside* `lifecycleMutex.withLock` in `start()`, returning a failure before `setup()`. The recovery-mode check at :330 sits outside the lock, so it would not catch a start that is already queued. One check covers ON_START, the worker, the service and the retry. The test's final `times(2)` becomes `times(1)`. Leaving it for the #1256 follow-up is also reasonable.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
ovitrif: Taken in 8acaa6759: `WalletRepo.wipeWallet` re-reads wallet existence on `WipeIncomplete`, so a failed keychain clear keeps the user in the wallet where Reset can be retried, and a failed key reset alone still lands on onboarding. Covered in `WalletRepoTest.kt`.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
ovitrif: Closed in 8acaa6759: `LightningRepo` gains a wiping flag, set and cleared by the use case next to the backup flag, and `start()` checks it inside `lifecycleMutex` before `setup()`, returning `WipeInProgressError`. ON_START, the worker, the service and the retry all go through that check. Pinned by `start is refused while a wipe is in progress` in `LightningRepoTest.kt`; the lock test keeps `times(2)` since it exercises the lock without the flag.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -341,6 +344,7 @@ class LightningRepo @Inject constructor(
         var initialLifecycleState: NodeLifecycleState
 
         val result = lifecycleMutex.withLock {
+            if (isWiping) return@withLock Result.failure(WipeInProgressError())
```
jvsena42, line 347: **LOW.** A start refused during a wipe shows up as an error toast.

`WalletViewModel.startNode` (`WalletViewModel.kt:343-347`, not in the diff) logs every failure with `Logger.error("Node startup error")` and sends it to `ToastEventBus`, skipping only `RecoveryModeError`. If the user backgrounds and reopens the app while the Reset spinner is up, ContentView ON_START calls `walletViewModel.start()`, because `walletExists` stays true until `resetWalletState()`. The start is refused here, and the user gets an "Error / Wallet wipe in progress, refusing node start" toast in the middle of a Reset that is working as intended. The spinner can stay up for tens of seconds (`node.stop()` deadline plus remote cleanup), so this is easy to hit.

**Fix:** `if (it !is RecoveryModeError && it !is WipeInProgressError)` at `WalletViewModel.kt:345`. Logging it at debug rather than error would also keep it out of error logs.

jvsena42 (commented):
Reviewed `8acaa6759`. No HIGH/MEDIUM, one LOW inline. Both fixes from my last pass hold.

**Checked and clean**

- **Wipe flag.** It is checked inside `lifecycleMutex` (`LightningRepo.kt:347`) before `initialLifecycleState` is read and before `Starting` is emitted. A refused start leaves the lifecycle state alone and schedules no retry, and a start that queued during `wipeStorage`'s hold is now refused when it takes the lock. The flag is set right after `wipeMutex.tryLock()` and cleared in `finally`, so a failed or cancelled wipe cannot leave it stuck. A start after `finally` either finds an empty keychain (`MnemonicNotFound`) or, on `WipeIncomplete`, a wallet the user is intentionally kept in.
- **Callers of the refused start.** `WakeNodeWorker.kt:85` ignores the `Result` and waits on the deliver signal as before. `LightningNodeService.setupService` only acts on success. `AppViewModel.completeRNRemoteBackupRestore` logs the error. In recovery mode, `RecoveryModeError` still wins because it is checked first.
- **`WipeIncomplete` landing.** `WalletRepo.wipeWallet` reads `keychain.exists(BIP39_MNEMONIC)` again after `resetState()`. A failed `clear()` keeps the user in the wallet with Reset available; a failed `resetEncryptionKey()` alone still lands on onboarding, where Create works. `WalletRepoTest.kt` pins the first case.
- **Tests.** `start is refused while a wipe is in progress` checks both the failure type and that `lightningService.start` is never called. The lock test keeping `times(2)` is correct, because it never sets the flag.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
ovitrif: Taken in bf677b369: `WalletViewModel.startNode` now treats `WipeInProgressError` like `RecoveryModeError`, logging at debug and sending no toast.

jvsena42 (commented):
Reviewed `bf677b369`. No findings, and the toast LOW from my last pass is fixed.

**Checked and clean**

- `WalletViewModel.startNode` (:344-351) returns early for both `RecoveryModeError` and `WipeInProgressError`, logging at debug with no toast. Every other failure still logs at error and shows a toast, same as before.
- The only change in behaviour for `RecoveryModeError` is that it is now logged at debug instead of error. Recovery mode is a deliberate user state, so that fits.
- `Logger.debug` has no `Throwable` parameter (`Logger.kt:81`), so writing the message into the log text is the only way to keep the reason.
- No other start caller surfaces `WipeInProgressError` to the user: `WakeNodeWorker` ignores the `Result`, `LightningNodeService.setupService` only acts on success, and `AppViewModel.completeRNRemoteBackupRestore` only logs.
- No existing test asserts on the `RecoveryModeError` toast branch, so nothing depends on the old logging.

At `bf677b369` I have nothing further on this PR. CI was still pending when I checked.

jvsena42 (approved):
tAck

piotr-iohk (approved):
QA from the diff plus emulator (emulator-5554, CI APK bf677b36). No new findings. The keychain-incomplete and start-during-wipe follow-ups look right in the diff.

On device:
- Reset while the node was Running landed on TOS with the expected order (observers stopped → VSS reset → node stopped → remote cleanup → LDK wiped → keychain wiped). New Wallet then set up VSS on attempt 1 and uploaded METADATA / ACTIVITY. No `MnemonicNotAvailableException`.
- Background + resume during Reset did not hit `WipeInProgress` — wipe was ~0.6s on this empty wallet. That path is unit-covered only.
- Did not re-run the Pubky endpoint delete (last pass on `38c9df8`).

Issue #1275, opened 2026-09-15
Backlog triage: unsolved issues ranked by agent-handoff safety


jvsena42:
## Context

A triage pass compared all 124 then-open issues against `master` (`cc1023ac`), reading the current code and the fixing commits, and driving a Pixel_9 emulator where the flow was observable. 13 issues turned out to be already fixed or stale and were closed; regression pins for those landed in #1274.

This issue collects what is **left**, split by how safely it can be handed to a coding agent running with little or no supervision. Each row names the exact file and the shape of the change, so a task can be cut from it directly. Nothing here is a new bug report — every item links an existing issue.

The point of the split is that "small diff" and "safe for an agent" are not the same thing. Tier A items have one obvious correct implementation. Tier B items are still bounded but have a failure mode worth a human look. Everything in the last section should **not** go to an unsupervised agent, and the reason is stated.

---

## Tier A — mechanical, one obvious implementation

Self-contained, unit-testable, no product or design decision left open. Each mirrors a pattern that already exists in the repo.

| Issue | Change | Where |
| - | - | - |
| #809 | Transfer-to-savings success returns to the Spending details screen instead of Home. Replace `popBackStack<Routes.TransferRoot>(inclusive = true)` with `navigateToHome()` — the sibling external-node flow already does this (`a13842243`), and iOS calls `navigation.reset()`. | `ContentView.kt:940-947` |
| #1110 | Channel details hardcodes 24h time. `DatePattern.CHANNEL_DETAILS = "MMM d, yyyy, HH:mm"`; route it through the existing `UiDateStyle.pattern(is24Hour)` + `LocalIs24HourFormat` machinery that the activity list already uses (the #677 fix). | `ChannelDetailScreen.kt:588-604`, `ext/DateTime.kt:264` |
| #1042 | A successful retry keeps the failed attempt's amount. The existing-row branch copies only `updatedAt`/`status`/`contact`; add `value`, `fee`, `preimage`, `message`, which are already on `payment`/`kind`. | `CoreService.kt:748-753` |
| #1167 | Three Core restore slices share one `runCatching`, so one failure silently drops the rest. Split into three `runSuspendCatching` blocks with per-slice `Logger.error` and an aggregate failure result. | `ActivityRepo.kt:985-989` |
[...]

PR #1282, opened 2026-09-16
fix: refresh lightning activity on retry

### LINKED ISSUE #1042: [Bug]: LN activity keeps failed attempt amount after successful retry (same invoice)
## What happened?

After a **failed** Lightning send (e.g. 300,000 sats, `ROUTE_NOT_FOUND`), the activity correctly shows **Failed** with that amount.

Retrying to the **same address / invoice** (same payment hash) with a **different amount** succeeds, but the activity updates to **Sent** while still showing the **first (failed) amount** (e.g. 300k instead of the amount that actually went out).

Logs from repro session show one payment hash `3b8b626f…`:

- 09:16:17 — initiated send **300,000** sats → `PaymentFailed` (`ROUTE_NOT_FOUND`)
- 09:16:33 — same hash → `PaymentSuccessful` at **22** sats (`22000` msat)

Screen recording of the UI symptom to be attached.

## Expected behavior

One activity row per payment hash (LDK model). After a failed attempt and a successful retry on the same invoice, the activity should reflect the **final succeeded payment**: correct **Sent** status and **actual amount** (and fee / preimage), matching iOS behavior.

Showing both failed and succeeded rows for the same payment hash would double-count mentally; overwrite with correct final fields is preferred.

## Steps to Reproduce

1. Regtest (or staging) wallet with spending balance and an open channel.
2. Send a Lightning payment large enough to fail routing (e.g. **300,000** sats to a reachable but under-capacity peer, or force `ROUTE_NOT_FOUND`).
3. Confirm activity shows **Failed** with that amount.
4. Retry to the **same** Lightning Address / invoice flow (same payment hash — typical for amountless invoices or “send again” to same LNURL).
5. Enter a **smaller** amount that succeeds.
6. Open Activity → payment shows **Sent** but amount is still the **failed attempt** amount.

- Logs zip (attach manually): [logs.zip](https://github.com/user-attachments/files/29286729/logs.zip)
  - Key lines: `PaymentFailed` then `PaymentSuccessful` for hash `3b8b626f8cdc85e5a4dea06c4431137a1d357ee434e43051cacbfff2639743c2` (300k fail → 22 sats success)
- Screen recording: 

https://github.com/user-attachments/assets/1506473d-93c1-4be0-bc4c-42b2894a1f91

## Bitkit Version

2.3.0 (182)

## Device / OS

Pixel 6 emulator (`Pixel_6` AVD), Android 13 (API 33)

## Reproducibility

Often (>50%) — depends on provoking a failed outbound payment then retrying same invoice with different amount; confirmed in one logged session.

## Additional context

- **Environment:** Regtest / local docker stack
- **Similar issues:** [#790](https://github.com/synonymdev/bitkit-android/issues/790) (max send failures) — different symptom; no duplicate found for activity amount on retry.
- **iOS note:** iOS rebuilds full `LightningActivity` from `payment.amountSats` on each sync (`CoreService.processLightningPayment`). Android only updates `status` / `updatedAt` / `contact` on existing rows in `processBolt11` - **Android-only** display bug.
- **Root cause (code review):** `CoreService.processBolt11` — when updating existing `Activity.Lightning`, `copy()` updates `status` but not `value`, `fee`, `preimage`, or `message`. New activities set `value = payment.amountSats`; updates do not.
- **Suggested fix:** On update, refresh `value`, `fee`, `preimage`, `message` from `PaymentDetails` like iOS `processLightningPayment`.
- **Related file:** `app/src/main/java/to/bitkit/services/CoreService.kt` (`processBolt11`, ~521–526)

jvsena42:
Fixes #1042
Related to #1275

This PR fixes Lightning activity rows keeping the failed attempt's amount after a retry of the same invoice succeeds.

### Description

- Refreshes the amount, fee and preimage of an existing Lightning activity when LDK reports an update for its payment hash, so a retry with a different amount shows what was actually sent. Before, only the status, update time and contact were refreshed.
- Keeps the stored amount, fee and preimage when the update reports none. A pending retry has no fee yet, so it shows the earlier fee until the payment succeeds.
- Keeps the stored message on every update. LDK reports a description-hash invoice's hash as its description, so copying it would overwrite an LNURL comment.
- Keeps one activity row per payment hash, overwritten with the final attempt, which matches iOS.

#### Out of Scope

- Activity sync: rows created for the first time still take LDK's description as the message, which for description-hash invoices is the hex hash. This is unchanged from master.
- Activity sync: the guard that decides whether an existing row is updated is unchanged and has no dedicated test.
- Activity type, timestamp, invoice, seen/created times and contact attribution of an existing row are not changed.
- iOS: no change needed. The issue says iOS already shows the final attempt.

### Design

N/A — no UI changes.

### Preview

N/A. The activity screens do not change. The fix changes which amount is stored, and the repro needs a second node to issue an amountless invoice plus a forced asynchronous route failure, so no recording was made.

#### Manual Tests

- [ ] **1a.** Amountless LN invoice from another node → Send 300,000 sats so routing fails (`ROUTE_NOT_FOUND`) → Activity: row shows Failed with 300,000 sats.
  - [ ] **1b.** Pay the same invoice again with a small amount (e.g. 22 sats) → Activity: the same row shows Sent with 22 sats and the fee of the successful attempt, and no second row appears.
- [ ] **2.** `regression:` Send to an LNURL-pay address with a comment → Activity Detail after it succeeds: the comment is still shown.
- [ ] **3.** `regression:` Receive an LN payment → Activity: amount and status are correct.

These steps were not run on a device. The fix is covered by unit tests only, and the call from the activity sync into the new update step was checked by code review.

#### Automated Checks

- Unit tests added in `CoreServiceTest.kt`: a failed 300,000 sat row retried as a succeeded 22 sat payment takes the new amount, fee (1,999 msat rounds down to 1 sat), preimage, status, update time and contact; untouched fields stay the same even when the retry reports a different invoice; a missing fee, amount or preimage keeps the stored value; the stored message survives a description hash and a different description.
- `just test file CoreServiceTest` is rejected by an existing guard in `app/build.gradle.kts`, so the class was run with `just test file to.bitkit.services.CoreServiceTest`.
- Local runs of `just compile`, `just test` and `just lint` passed.


jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `da30e5580`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Traced `LightningActivity.withPaymentUpdate` (app/src/main/java/to/bitkit/services/CoreService.kt:334-346) and its only caller `processBolt11` (CoreService.kt:769-775): the fields now overwritten on an existing row are value, fee and preimage, each guarded with `?:` so a null LDK value keeps the stored one; status/updatedAt/contact behave as on master and message/invoice/timestamp/createdAt/seenAt are untouched. Enumerated every writer of Lightning rows at the PR head: LDK sync (this path), `setContact`/`clearContact` (copy contact + updatedAt=now, so the pre-existing guard at CoreService.kt:756 skips the next sync), `markAsSeen` (seenAt only), gift `SuccessWithoutLiquidity` (id is the funding txid/order id, never a payment hash, so sync cannot match it; the with-liquidity gift path inserts no row and relies on LDK), CJIT insert (funding txid id) and RN migration upserts (ldk-node's payment store is not populated by migrated channel monitors, so no hash overlap). No writer stores an LNURL comment or any value/fee/preimage that LDK's `PaymentDetails` could degrade. Checked received payments (inbound PENDING skipped; SUCCEEDED carries the claimed amount and preimage, fee null keeps 0), pending->failed (amount unchanged, fee/preimage null keep stored), a retry that fails synchronously (ldk-node records a FAILED row with the new amount, guard passes on newer timestamp, value updated to the new attempt), and zero-amount invoices (outbound amount always set by ldk-node at send; inbound amount set on claim). Rounding is consistent with row creation (`amountSats` ceil, `msatFloorOf` fee). `ServiceQueue.CORE` is a single-thread dispatcher, so the only interleaving is the pre-existing stale-list status regression between a full sync and `handlePaymentEvent`, which this PR does not worsen and the next sync repairs. The linked issue's retry path (FAILED 300k -> PENDING/SUCCEEDED 22 sats via `handlePaymentEvent` or full sync) now lands the final attempt's amount, fee and preimage on the same row.

jvsena42 (commented):
✅ LGTM

ovitrif (approved):
utAck

piotr-iohk:
QA reviewed on 8c250db4002438ce255d2af4a935eb0f9712a8ec.

Retried a failed Lightning send with a smaller amount on the same invoice. The activity row now shows the new amount; the old failed amount is gone.

No findings.

Checked and clean
- Retry updates amount, fee, and preimage on the existing row and leaves the payment message alone.
- Unit tests cover that retry path.

QA LGTM

piotr-iohk (approved):

PR #1284, opened 2026-09-16
fix: stop pin result toggle overlapping label

### LINKED ISSUE #638: [Bug]:  Random toggle in screen after confirming PIN change

jvsena42:
Fixes #638
Related to #1275

This PR fixes the "Also require for payments" switch on the PIN setup result sheet drawing over its label when the label is long.

### Description

- Fixes the switch overlapping the label on the "Wallet Secured" sheet, because the label had no width limit and ran under the switch in longer translations such as Spanish ("También se requiere para los pagos") or at larger font scales. The label now takes the remaining width and wraps, and the switch keeps its natural size to the right.
- Adds a preview of the result sheet in Spanish at 1.3x font scale so the long-label case stays visible in Android Studio.
- Updates the screen map row for the result screen to the "Security › Create passcode" frame, since the previous "PIN Change Succes" frame belongs to the change PIN flow.

#### Out of Scope

- `ui/settings/pin/PinBiometricsScreen.kt`: the biometrics toggle row uses the same label and switch layout. No current translation overlaps there at default font scale, so it is left unchanged.
- `ui/settings/pin/PinResultScreen.kt`: the previews still use a plain themed surface instead of the in-sheet preview wrapper, matching the existing previews on this screen.
- iOS: the same toggle exists on the iOS security success screen; checking it for the same overlap is a separate follow-up in `bitkit-ios`.

### Design

[Security › Create passcode](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-147248)

### Preview

| Figma | Before (es, 1.3x) | After (es, 1.3x) |
| --- | --- | --- |
| Figma Security Create passcode | Before es 1.3x | After es 1.3x |

| After (es, 1.3x, toggled on) | After (en, 1.0x) |
| --- | --- |
| After es 1.3x toggled on | After en 1.0x |

The Figma frame shows the biometrics variant of the body text and the switch on; the device has no biometrics enrolled and the switch defaults to off.

#### Manual Tests

- [x] **1a.** App language Spanish, font scale 1.3x → Settings → Security → PIN Code → Enable PIN → set and confirm PIN → skip biometrics → Pin Result: "También se requiere para los pagos" wraps to two lines and the switch sits fully to its right.
  - [x] **1b.** tap the toggle row: switch turns on; tap again: switch turns off.
- [x] **2.** `regression:` English, default font scale → same flow → Pin Result: label stays on one line with the switch at the right edge, aligned with the OK button.
- [x] **3.** `regression:` Pin Result → toggle on → OK → Settings → Security: PIN for payments shows enabled.

Journey covering the above, committed at `journeys/security/pin-result-long-label.xml`
(with its rows in `journeys/README.md`). Inline for review:

<details>
<summary>pin-result-long-label.xml</summary>

```diff
diff --git a/journeys/security/pin-result-long-label.xml b/journeys/security/pin-result-long-label.xml
new file mode 100644
index 000000000..73c0ae9d1
--- /dev/null
+++ b/journeys/security/pin-result-long-label.xml
@@ -0,0 +1,49 @@
+<journey name="pin result toggle wraps long label">
+  <description>
+    On the "Wallet Secured" result sheet shown after creating a PIN, the "Also require for
+    payments" switch sits to the right of its label and never draws over it when the label is long.
+    Spanish at 1.3x font scale makes the label ("También se requiere para los pagos") wider than the
+    row, so it must wrap onto a second line instead of running under the switch. The row (tag
+    "ToggleBioForPayments") still toggles the PIN-for-payments setting.
+
+    Precondition: onboarded dev wallet with the PIN disabled (Settings ▸ Security ▸ PIN Code shows
+    "Disabled"), no biometrics enrolled on the device, and PIN for payments off. Start on the wallet
+    home screen. The journey enables a PIN you choose and disables it again at the end, so remember
+    it; never enter a wrong PIN more than once.
+
+    With no biometrics enrolled the biometrics step shows the "not available" variant, whose Skip
+    button carries tag "SkipButton"; on a device with biometrics the screen shows a toggle and
+    "ContinueButton" instead and the route differs.
+
+    Enabling the PIN pops the PIN Code screen back to Settings ▸ Security while the result sheet is
+    still up, so after OK the runner is on Settings ▸ Security and the cleanup has to re-enter PIN
+    Code before "Disable PIN" exists. Disabling pops back the same way, so the last check reads the
+    PIN Code row on Settings ▸ Security.
+
+    Assert the overlap from `adb shell uiautomator dump` bounds: the Switch has no node of its own
+    in `android layout`, so compare the label's right edge against the row's. On iOS, compare the
+    label and switch frames from `snapshot-ui` instead.
+  </description>
+  <actions>
+    <action>Run `adb shell cmd locale set-app-locales to.bitkit.dev --locales es`</action>
+    <action>Run `adb shell settings put system font_scale 1.3`</action>
+    <action>Tap the menu button (tag "HeaderMenu") and tap Settings (tag "DrawerSettings")</action>
+    <action>Tap the Security tab (tag "Tab-security")</action>
+    <action>Tap PIN Code (tag "PINCode")</action>
+    <action>Tap Enable PIN (tag "EnablePin")</action>
+    <action>Enter a 4-digit PIN with the number pad (tags "N0"–"N9")</action>
+    <action>Re-enter the same 4-digit PIN to confirm it</action>
+    <action>On the biometrics screen, tap Skip (tag "SkipButton")</action>
+    <action>Verify the result sheet titled "Cartera asegurada" is visible with the row (tag "ToggleBioForPayments") and the OK button (tag "OK")</action>
+    <action>Verify from a screenshot that "También se requiere para los pagos" wraps onto two lines and the switch is fully to its right, inside the sheet's horizontal padding and aligned with the right edge of the OK button</action>
+    <action>Verify from `adb shell uiautomator dump` that the label's right bound is less than the "ToggleBioForPayments" row's right bound minus the switch width</action>
+    <action>Tap the "ToggleBioForPayments" row and verify the switch turns on</action>
+    <action>Tap the "ToggleBioForPayments" row again and verify the switch turns off</action>
+    <action>Tap OK (tag "OK") and verify the screen underneath is Settings ▸ Security, not PIN Code</action>
+    <action>Tap PIN Code (tag "PINCode")</action>
+    <action>Tap Disable PIN (tag "DisablePin") and enter the PIN chosen above on the disable sheet (tag "DisablePIN")</action>
+    <action>Verify the PIN Code row (tag "PINCode") shows "Desactivado"</action>
+    <action>Run `adb shell cmd locale set-app-locales to.bitkit.dev --locales ''`</action>
+    <action>Run `adb shell settings put system font_scale 1.0`</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- No unit tests added: the change is layout only, with no logic changes.
- `ScreensMapTest.kt` passes with the updated screen map row.
- Ran `just compile`, `just test` and `just lint` locally; all passed.


jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `d03ddb345`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Read PinResultScreen.kt in full at the PR head. The toggle row now gives the label `Modifier.weight(1f)` with an 8dp spacer before the Switch, so the label is measured against the remaining width after the switch and wraps instead of running under it; the row keeps `fillMaxWidth`, `clickableAlpha` and the `ToggleBioForPayments` tag, and the toggle callback, BackHandler and OK path in `PinResultScreen`/`PinSheet` are unchanged. Dropping `Arrangement.SpaceBetween` is safe because the weighted label pushes the switch to the right edge on its own. The docs/screens-map.md row is consistent: Figma node 46920:147248 on the handoff file is a top-level frame named "Create passcode" containing the "Also require for payments" row with the switch, the check visual and the OK button, which is this screen, and the file name still matches the `*Screen.kt` regex `ScreensMapTest` checks. The changelog fragment is a single fixed entry under changelog.d/next.

jvsena42 (commented):
✅ LGTM

journeys/security/pin-result-long-label.xml
```diff
@@ -0,0 +1,43 @@
...
+    <action>Enter a 4-digit PIN with the number pad (tags "N0"–"N9")</action>
+    <action>Re-enter the same 4-digit PIN to confirm it</action>
+    <action>On the biometrics screen, tap Skip (tag "SkipButton")</action>
+    <action>Verify the result sheet titled "Cartera asegurada" is visible with the row (tag "ToggleBioForPayments") and the OK button (tag "OK")</action>
+    <action>Verify from a screenshot that "También se requiere para los pagos" wraps onto two lines and the switch is fully to its right, inside the sheet's horizontal padding and aligned with the right edge of the OK button</action>
+    <action>Verify from `adb shell uiautomator dump` that the label's right bound is less than the "ToggleBioForPayments" row's right bound minus the switch width</action>
+    <action>Tap the "ToggleBioForPayments" row and verify the switch turns on</action>
+    <action>Tap the "ToggleBioForPayments" row again and verify the switch turns off</action>
+    <action>Tap OK (tag "OK")</action>
+    <action>Tap Disable PIN (tag "DisablePin") and enter the PIN chosen above on the disable sheet (tag "DisablePIN")</action>
```
jvsena42, line 44: **MEDIUM, journey, non-blocking** — **After OK the runner is on Settings › Security, not PIN Code, so the "Tap Disable PIN" step cannot be performed and the cleanup steps never run.**

`PinConfirmScreen` calls `app.addPin(pin)` when the confirmation matches. `editPin` then sets `isPinEnabled = true` (AppViewModel.kt:4876). `PinManagementScreen` stays composed under the PIN sheet with `initialPinState = remember { isPinEnabled }` (false), and its `LaunchedEffect(isPinEnabled)` calls `navController.popBackStack()` when the value changes (PinManagementScreen.kt:46-51). So by the time the result sheet appears, the screen underneath has already returned to Settings. Test 3 in the QA run shows the same landing: after OK the runner was on "Settings > Security" with PIN Code enabled. `DisablePin` only exists in `PinManagementScreen` (line 131), so step 38 finds nothing. The README says to stop rather than improvise in this case. The run ends as a false fail and leaves the emulator with a PIN set, Spanish app locale and 1.3x font scale, because steps 40-41 never run.

Fix: add `<action>Tap PIN Code (tag "PINCode")</action>` between "Tap OK" and "Tap Disable PIN". Step 39 is correct as written: the second visit remembers `true` and pops back to Settings when the PIN is removed.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `2e374c9cc`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: MEDIUM (journey, non-blocking) — Journey taps DisablePin on a screen that has already popped

Checked: Round 4 (8a52de683, new journey; the finding was re-checked against PinManagementScreen.kt at head): Reviewed the one new commit since d03ddb345: journeys/security/pin-result-long-label.xml plus its suite and cross-platform rows in journeys/README.md. Every tag the journey names exists at head: HeaderMenu, DrawerSettings, Tab-security (from the SettingsTab enum through CustomTabRowWithSpacing), PINCode, EnablePin, the N0-N9 number pad keys, SkipButton in the biometrics-unavailable view, ToggleBioForPayments and OK in PinResultScreen, and DisablePin and DisablePIN. The Spanish strings the journey asserts match values-es ("Cartera asegurada", "También se requiere para los pagos", "Desactivado"). The bounds check (label right edge below row right edge minus switch width) holds with the weighted label and would fail if the unweighted label came back, and the numbers in the QA run confirm it. The journey is Android-only, and its cross-platform row records it as not ported.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `744cdbc54`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 744cdbc54: the journey MEDIUM is fixed. After OK the runner now verifies it is on Settings ▸ Security and taps PIN Code (tag "PINCode") before Disable PIN, which matches the `LaunchedEffect(isPinEnabled)` pop in PinManagementScreen.kt, and the final check reads the PIN Code row where the disable pop lands. The cleanup steps now run.

journeys/security/pin-result-long-label.xml
jvsena42: Verified. `PinManagementScreen.kt:46-52` remembers `isPinEnabled` on entry and pops back as soon as it changes, and `editPin` sets it to true (`AppViewModel.kt:4875-4877`), so the screen under the result sheet is already Settings > Security by the time OK is tapped. `DisablePin` only exists on that screen (`PinManagementScreen.kt:131`), so the old step 38 had nothing to tap and the locale and font-scale cleanup never ran.

Added `Tap PIN Code (tag "PINCode")` between OK and Disable PIN, made the OK step assert the Settings > Security landing, and noted the pop-back behaviour in the description so the last check reading the `PINCode` row (`SettingsScreen.kt:440-452`) is clearly on Settings > Security too. Step 39 was already correct and is unchanged.

Commit 744cdbc54; the PR body diff is back in step with the file.

ovitrif (approved):
utAck

PR #1287, opened 2026-09-16
fix: show specific electrum error toasts

### LINKED ISSUE #1177: Electrum server errors show a generic message, discarding the probe's specific cause
### What happened?

`ElectrumProbeService` produces precise, actionable failure reasons, but `ElectrumConfigScreen` throws them away and always shows the same generic toast.

The failure branch ignores `result.exceptionOrNull()` entirely:

```kotlin
// app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigScreen.kt:70-77
} else {
    app.toast(
        type = Toast.ToastType.WARNING,
        title = context.getString(R.string.settings__es__server_error),
        description = context.getString(R.string.settings__es__server_error_description),
        testTag = "ElectrumErrorToast",
    )
}
```

`ElectrumConfigViewModel` does carry the real cause through (`connectionResult = Result.failure(e)`), so the information is available at the UI layer — it is just never read.

Every rejection therefore renders as:

> **Electrum Connection Failed**
> Bitkit could not establish a connection to Electrum.

even though `ElectrumProbeError` already distinguishes four cases, two of which are directly actionable by the user:

| Error | Message the probe already builds |
|---|---|
| `NetworkMismatch` | wrong network, with expected vs actual genesis hash |
| `ProtocolMismatch` | failed TLS handshake, "check the protocol" |
| `Unreachable` | could not reach the server |
| `NotElectrum` | host answered, but does not speak Electrum |

The two worth surfacing are `NetworkMismatch` (user pointed a testnet/regtest wallet at a mainnet server, or vice versa) and `ProtocolMismatch` (user picked TLS on a plain-TCP port, or the reverse). Today both look identical to "the server is down", so there is nothing to act on.

### Expected behavior

The toast should reflect the probe's verdict — at minimum a distinct, localized message for `NetworkMismatch` and `ProtocolMismatch`, falling back to the current generic string for `Unreachable` / `NotElectrum` / anything unrecognized.

### Steps to Reproduce

1. Run a non-mainnet build (dev/regtest or tnet).
2. Settings → Advanced → Electrum Server.
3. Enter `electrum.blockstream.info`, port `50002`, protocol TLS (a healthy **mainnet** server).
4. Tap **Connect To Host**.
5. A generic "Bitkit could not establish a connection to Electrum." toast appears — despite the connection succeeding and the server being rejected purely for being on the wrong network.

The same generic toast appears for an unreachable host (`10.255.255.1:9999`) and for a non-Electrum host (`tcp://google.com:443`).

Verified on-device; the log has the precise reason the user never sees:

```
11:40:53.705 INFO  [LightningRepo.kt:734] Changing ldk-node electrum server to: 'ssl://electrum.blockstream.info:50002'
11:40:55.241 WARN  [LightningRepo.kt:737] Rejected electrum server 'ssl://electrum.blockstream.info:50002'
  [NetworkMismatch='Rejected electrum server 'ssl://electrum.blockstream.info:50002' on wrong network,
   expected '0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206'
   got '000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f'']
```

UI at that moment (via layout dump):

```
Electrum Connection Failed
Bitkit could not establish a connection to Electrum.
```

### Bitkit Version

2.4.1 (build 188). `ElectrumConfigScreen.kt` and `ElectrumConfigViewModel.kt` are byte-identical on `master` and on `release-2.4.1`, so this affects both.

### Device / OS

Android emulator, `sdk_gphone16k_arm64` (16 KB page size), dev/regtest flavor.

### Reproducibility

Always

### Additional context

- Behaviour is unchanged from before `ElectrumProbeService` was introduced — the probe added the good diagnostics, the UI just never started using them. So this is a follow-up to the probe work rather than a regression.
- Nothing is persisted on rejection and the node is correctly never torn down, so the impact is purely that the user cannot tell *why* their server was refused.
- Scope is small and self-contained: branch on the `ElectrumProbeError` subclass in the `LaunchedEffect` failure branch and add two string resources.

jvsena42:
Fixes #1177
Related to #1275

This PR shows the specific cause when connecting to a custom Electrum server fails, instead of the same generic message for every error.

### Description

- Shows a network-specific message when the server is on a different Bitcoin network, so the user knows to pick a server for the network Bitkit is using.
- Shows a protocol-specific message when the TLS handshake fails against a peer that does not speak TLS, so the user knows to check that TCP or TLS matches the server port.
- Shows a certificate-specific message when the handshake fails on certificate validation instead, so a self-signed Fulcrum or electrs on its SSL port is not told to check a protocol and port that are already correct.
- Asks JSSE to verify the certificate's hostname during the probe handshake. A raw `SSLSocket` validates the chain but not the name, so a server reached by IP or alias that presents a CA-valid certificate for a different name used to probe clean and was only rejected afterwards by the node's own Electrum client, which validates the domain — the user got the generic message after a full node restart. It now fails in the probe and shows the certificate message.
- Keeps the generic "could not establish a connection" message for unreachable hosts, non-Electrum servers and any other failure.
- Moves the success and error toasts from the screen into the view model so the toast can use the probe's error instead of a plain success/failure result. Toast titles and test tags (`ElectrumUpdatedToast`, `ElectrumErrorToast`) are unchanged, so the e2e settings tests still match.
- Returns `Result.failure` from `LightningRepo.restartWithElectrumServer` when the settings write after a successful restart fails, so that failure reaches the toast instead of escaping the caller's coroutine.
- Takes the probe's `SSLSocketFactory` as a dependency instead of reading `SSLSocketFactory.getDefault()` inside it. Production still gets the platform default through Hilt; the change is so the TLS tests can hand the probe a factory of their own rather than swapping the JVM's default SSL configuration underneath it.

#### Out of Scope

- Settings › Advanced › Electrum Server: choosing TCP on a TLS-only port still shows the generic message, because the probe reports that case as "not Electrum" rather than a protocol mismatch.
- The new certificate cases are not in the journey below: they need a self-hosted Electrum server with a self-signed certificate, or one whose certificate names another host, which the journey's preconditions do not set up.
- Settings › Advanced › Rapid Gossip Sync: still uses the old screen-side result toast pattern, and `restartWithRgsServer` still writes its setting unguarded; not part of the issue.
- Locales: only the English strings are added; translations follow through the usual localization flow.
- iOS: parity for the specific network and protocol messages on manual connect is a possible follow-up in `bitkit-ios`, along with porting the new journey below.

### Design

- [Settings > Advanced › Advanced > Electrum server](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-150682). No toast frame exists in the design, so the toast copy and styling cannot be compared against one.

### Preview

| Figma | Before: wrong network | After: wrong network |
| --- | --- | --- |
| Figma Electrum server screen | Before: generic toast for wrong network | After: network mismatch toast |

| After: TLS on plain TCP port | After: unreachable host |
| --- | --- |
| After: protocol mismatch toast | After: generic toast for unreachable host |

Captured on a dev (regtest) build connected to the staging Electrum server. The wrong-network case uses `electrum.blockstream.info:50002` over TLS, the protocol case uses `electrum.blockstream.info:50001` over TLS, and the unreachable case uses `10.255.255.1:9999`. TLS is selected before the port is typed, because selecting a protocol rewrites a port that is still a default one.

#### Manual Tests

- [x] **1.** Settings → Advanced → Electrum Server → select TLS, then enter `electrum.blockstream.info` / `50002` (selecting a protocol rewrites a default port, so set it first) → Connect To Host: warning toast "This server is on a different Bitcoin network. Choose a server for the network Bitkit is using."
- [x] **2.** Electrum Server → keeping TLS selected, enter `electrum.blockstream.info` / `50001` → Connect To Host: warning toast "Secure connection failed. Check that the protocol (TCP or TLS) matches the server port."
- [x] **3.** Electrum Server → enter `10.255.255.1` / `9999` → Connect To Host: after ~5s, warning toast "Bitkit could not establish a connection to Electrum."
- [x] **4.** `regression:` After each failure above: "Currently connected to" still shows the previous server, and back → Electrum Server row is unchanged.
- [x] **5.** `regression:` Electrum Server → enter a valid server for the current network → Connect To Host: success toast "Electrum Server Updated" and the connected server updates.
- [x] **6.** Needs a self-hosted Electrum with a self-signed certificate: Electrum Server → enter that host / its SSL port, TLS → Connect To Host: warning toast "This server's certificate is not trusted. Use a server with a trusted certificate, or connect to its TCP port." Before this PR, and before the last commit on this branch, that case showed the protocol message instead.
- [x] **7.** Needs an Electrum server reachable by IP whose TLS certificate is CA-valid but issued for another name: Electrum Server → enter that IP / its SSL port, TLS → Connect To Host: the same certificate toast, now without a node restart first.

The steps above are also written as a journey, committed in this PR as
`journeys/settings/electrum-server-error-toasts.xml`, that passes on this branch on a Pixel 9
emulator. Its row is added to the suites table and the cross-platform table in `journeys/README.md`.

<details>
<summary>journeys/settings/electrum-server-error-toasts.xml</summary>

```diff
diff --git a/journeys/settings/electrum-server-error-toasts.xml b/journeys/settings/electrum-server-error-toasts.xml
new file mode 100644
index 000000000..eb3431ba3
--- /dev/null
+++ b/journeys/settings/electrum-server-error-toasts.xml
@@ -0,0 +1,40 @@
+<journey name="electrum server error toasts">
+  <description>
+    Verifies that a rejected custom Electrum server shows a warning toast whose description names the
+    cause: a wrong-network server and a TLS handshake against a plain TCP port get their own messages,
+    and an unreachable host keeps the generic one. The server is probed before the node stops, so a
+    rejected server is never persisted and the wallet stays on its current server.
+
+    Precondition: onboarded dev wallet on the default staging regtest server (Settings shows Electrum
+    Server "Auto"). Start on the wallet home screen. Toasts never appear in `android layout` and last
+    about 2s, so assert them from a screen recording or a screenshot taken 2-6s after tapping connect.
+    Type hosts in short chunks and verify the field text; `adb shell input text` drops characters.
+    Do not tap "Reset To Default" unless the starting server was the default. Tapping a protocol row
+    rewrites the port whenever it is empty or still one of 51002/50002/51001/50001, to that
+    protocol's default for the build's network (60002 for TLS on the regtest dev build), so always
+    set the protocol before typing the port and never re-tap a protocol row afterwards.
+  </description>
+  <actions>
+    <action>Tap the menu icon (testTag "HeaderMenu")</action>
+    <action>Tap Settings (testTag "DrawerSettings")</action>
+    <action>Tap the Advanced tab (testTag "Tab-advanced")</action>
+    <action>Tap Electrum Server (testTag "ElectrumConfig")</action>
+    <action>Verify "Currently connected to" is visible and note the connected server (testTag "Connected")</action>
+    <action>Select "TLS" under protocol (testTag "ElectrumProtocol")</action>
+    <action>Replace the host (testTag "HostInput") with "electrum.blockstream.info" and the port (testTag "PortInput") with "50002", then verify the port field still reads "50002"</action>
+    <action>Hide the keyboard and tap Connect To Host (testTag "ConnectToHost")</action>
+    <action>Verify a warning toast "Electrum Connection Failed" appears with the description "This server is on a different Bitcoin network. Choose a server for the network Bitkit is using."</action>
+    <action>Verify the connected server (testTag "Connected") is unchanged</action>
+    <action>Replace the port (testTag "PortInput") with "50001" without tapping a protocol row, keeping "TLS" selected, and verify the port field still reads "50001"</action>
+    <action>Hide the keyboard and tap Connect To Host (testTag "ConnectToHost")</action>
+    <action>Verify a warning toast "Electrum Connection Failed" appears with the description "Secure connection failed. Check that the protocol (TCP or TLS) matches the server port."</action>
+    <action>Replace the host (testTag "HostInput") with "10.255.255.1" and the port (testTag "PortInput") with "9999"</action>
+    <action>Hide the keyboard and tap Connect To Host (testTag "ConnectToHost")</action>
+    <action>Verify a warning toast "Electrum Connection Failed" appears with the description "Bitkit could not establish a connection to Electrum." after the ~5s connect timeout</action>
+    <action>Verify the connected server (testTag "Connected") is unchanged</action>
+    <action>Tap Reset To Default (testTag "ResetToDefault")</action>
+    <action>Verify the host and port fields show the default server, which is the connected one, and Connect To Host (testTag "ConnectToHost") is disabled</action>
+    <action>Tap back (testTag "NavigationBack")</action>
+    <action>Verify the Electrum Server row (testTag "ElectrumConfig") still shows "Auto"</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `ElectrumConfigViewModelTest.kt` covers the toast type, title, description, test tag and loading state for network mismatch, protocol mismatch, untrusted certificate, unreachable, not-Electrum, other failures and success.
- `ElectrumProbeServiceTest.kt` pins how a failed TLS handshake is classified: a certificate path failure or an unverified peer becomes `UntrustedCertificate`, while a handshake timeout or a non-TLS reply stays `ProtocolMismatch`. It also probes a real TLS Electrum server whose certificate is trusted but issued for another name and asserts `UntrustedCertificate`, plus the paired case of a certificate that does name the host, which must still succeed. Reverting the hostname-verification line locally fails the first of those two and leaves the second passing. Both TLS tests build their own `SSLContext` from the JDK's own JSSE provider and pass its socket factory to the probe, so they no longer read or write any JVM-wide SSL state: Robolectric installs Conscrypt as the JVM's first security provider and never removes it, which made the paired success case fail on Linux CI while passing on macOS.
- `LightningRepoTest.kt` pins that a failed settings write after a successful restart comes back as `Result.failure` instead of throwing out of the returned result.
- Ran locally: `just compile`, `just test`, `just lint`.


app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigViewModel.kt
```diff
@@ -150,33 +152,33 @@ class ElectrumConfigViewModel @Inject constructor(
...
-                    it.copy(
-                        isLoading = false,
-                        connectionResult = Result.failure(e),
+            val serverUrl = ElectrumServer.fromUserInput(
+                host = currentState.host,
+                port = port,
+                protocol = protocol,
+            ).toString()
+
+            lightningRepo.restartWithElectrumServer(serverUrl)
```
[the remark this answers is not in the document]
jvsena42: Verified the path: `restartWithElectrumServer` only leaks an exception through the unguarded `settingsStore.update` after a successful restart — every other step (`validateElectrumServer`, `waitForNodeToStop`, `stop`, `start`) already returns a Result. Rather than re-wrap a Result-returning call in the view model, I fixed the contract at the repository in a008aae1: the settings write is now inside `runSuspendCatching` and a write failure comes back as `Result.failure`, so the view model's `onFailure` clears `isLoading` and shows the error toast as before. `LightningRepoTest.kt` gained a regression test pinning that.

app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigViewModel.kt
```diff
@@ -357,6 +355,12 @@ data class ElectrumConfigUiState(
...
     val protocol: ElectrumProtocol? = null,
     val isLoading: Boolean = false,
-    val connectionResult: Result<Unit>? = null,
     val hasEdited: Boolean = false,
 )
+
+@StringRes
+private fun Throwable.toServerErrorDescriptionRes(): Int = when (this) {
+    is ElectrumProbeError.NetworkMismatch -> R.string.settings__es__server_error_network
+    is ElectrumProbeError.ProtocolMismatch -> R.string.settings__es__server_error_protocol
```
jvsena42, line 364: **LOW** — **A self-signed or otherwise untrusted certificate is reported as a protocol mismatch.**

`openSocket` maps every exception from `startHandshake()` to `ElectrumProbeError.ProtocolMismatch` (ElectrumProbeService.kt:112-121), including `SSLHandshakeException` caused by `CertPathValidatorException` or a hostname mismatch. This line turns that into "Secure connection failed. Check that the protocol (TCP or TLS) matches the server port." Scenario: a self-hosted Fulcrum/electrs on `host:50002` with its default self-signed cert, TLS selected. Protocol and port already match, but the toast tells the user to change them, so they look in the wrong place. Before this PR the same case showed the generic message.

Fix: in `openSocket`, keep `ProtocolMismatch` for handshake timeouts and non-TLS replies (`SocketTimeoutException`, `SSLException` that is not `SSLHandshakeException` caused by a cert/peer-verification error). Map certificate-validation failures to the generic branch or to a separate certificate message.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `a008aae12`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Certificate failures on a correct TLS port get the "check TCP or TLS" message

Checked: Traced the error types from `ElectrumProbeService.probe` through `LightningRepo.validateElectrumServer` and `restartWithElectrumServer` into the view model: `NetworkMismatch` and `ProtocolMismatch` are thrown unwrapped from inside `runSuspendCatching` (the inner `runCatching` at ElectrumProbeService.kt:91-93 re-throws existing `ElectrumProbeError`s as-is), and `restartWithElectrumServer` returns the probe failure without wrapping, so the `when` in `toServerErrorDescriptionRes` sees the concrete subclass. `restartWithElectrumServer` has a single caller (`ElectrumConfigViewModel.connectToServer`), and on master that caller's `runCatching` already turned a thrown settings write into an error toast, so the new `Result.failure` return keeps the user-facing outcome identical while fixing the stuck-loading path greptile flagged; `runSuspendCatching` still propagates cancellation. `ElectrumServer.fromUserInput` cannot throw, so dropping the view model's `runCatching` leaves no unguarded throw before the repository call. Toasts go through `ToastEventBus`, collected by the activity-scoped `AppViewModel`, and the success path's settings write re-seeds host/port through `observeState` as before.

app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigViewModel.kt
jvsena42: Confirmed on the branch: openSocket wrapped every startHandshake failure in ProtocolMismatch, so an SSLHandshakeException from certificate-path validation or a hostname mismatch produced the "check that the protocol (TCP or TLS) matches the server port" toast even though both were already right.

Fixed in 57b4e9415. The handshake failure is now classified before it becomes a probe error: a CertificateException, CertPathValidatorException or SSLPeerUnverifiedException anywhere in the cause chain maps to a new ElectrumProbeError.UntrustedCertificate, everything else (handshake timeout, a plain-TCP server answering a TLS handshake) stays ProtocolMismatch. UntrustedCertificate gets its own string, settings__es__server_error_certificate: "This server's certificate is not trusted. Use a server with a trusted certificate, or connect to its TCP port." — actionable for a self-hosted Fulcrum/electrs on 50002 with a self-signed cert.

Tests: ElectrumProbeServiceTest covers the four classification cases, ElectrumConfigViewModelTest covers the new toast, and the existing "probe rejects TLS against a plain tcp server" test still lands on ProtocolMismatch. just compile, just test and just lint pass locally. PR description and QA notes updated.

jvsena42: **LOW (pre-existing)** — **The hostname-mismatch half of the original LOW is not covered: `startHandshake()` never checks the hostname, so `SSLPeerUnverifiedException` is never raised here.**

`openSocket` (ElectrumProbeService.kt:116-119) uses `SSLSocketFactory.getDefault()` and calls `startHandshake()` with no `endpointIdentificationAlgorithm` and no `HostnameVerifier`. Conscrypt validates the chain but not the name on a raw `SSLSocket`. Scenario: TLS selected, `host` is an IP or alias, and the server presents a CA-valid cert for `electrum.example.com`. The probe succeeds, `restartWithElectrumServer` proceeds to stop and start the node, the Rust electrum client (domain validation on by default) rejects the name, and the user gets the generic toast after a full node restart, the path the probe exists to avoid. The `toTlsProbeError reports an unverified peer` test builds the exception by hand, so it passes without this path being reachable. Pre-existing probe behaviour, not a regression of this commit.

Fix: before `ssl.startHandshake()` set `ssl.sslParameters = ssl.sslParameters.apply { endpointIdentificationAlgorithm = "HTTPS" }` (minSdk 28 supports it). Conscrypt then fails the handshake with a `CertificateException` cause, which the new classifier already maps to `UntrustedCertificate`.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `57b4e9415`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 57b4e9415: the certificate case is now classified separately. One pre-existing LOW (the probe skips hostname verification) is posted as a reply on the earlier thread. Re-reviewed at 57b4e9415. `toTlsProbeError` walks the handshake failure's cause chain (bounded at 8, cycle-safe) and maps `CertificateException`, `CertPathValidatorException` and `SSLPeerUnverifiedException` to the new `UntrustedCertificate`. On Conscrypt an untrusted or self-signed chain arrives as `SSLHandshakeException` -> `CertificateException` -> `CertPathValidatorException`, and an expired cert adds `CertificateExpiredException` (a `CertificateException` subclass), so both now get the certificate toast. The genuine mismatch cases keep `ProtocolMismatch`: a plaintext server answering the ClientHello fails with an `SSLException`/`SSLProtocolException` that has no certificate cause, and a handshake timeout is a `SocketTimeoutException`. TLS against a TCP port never reaches the handshake (TCP returns the plain socket at ElectrumProbeService.kt:111), so it still ends as `NotElectrum` as before. `settings__es__server_error_certificate` sits alphabetically before `_description` in strings.xml and is resolved through `toServerErrorDescriptionRes`, not hardcoded. The new probe tests exercise the new classifier and the view-model test fails on the old `when` (it would fall to the generic string). The rest of the PR (repository `runSuspendCatching` around the settings write, toast mapping for network/protocol) is unchanged by this commit.

app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigViewModel.kt
jvsena42: Confirmed and fixed in 5291fe216.

I verified the claim before changing anything, with a standalone JVM harness: a self-signed cert for `wrong.example` served on 127.0.0.1, with that cert as the only trusted anchor so the chain is valid and only the name can fail. A raw `SSLSocket` reports `endpointIdentificationAlgorithm = null` and the handshake **succeeds** — no name checking, exactly as described. Setting it to `"HTTPS"` fails the handshake with `SSLHandshakeException: No subject alternative names matching IP address 127.0.0.1 found`, caused by `java.security.cert.CertificateException`, so `isCertificateFailure()` returns true and it lands in `UntrustedCertificate`, not `ProtocolMismatch`. minSdk is 28 and `SSLParameters.setEndpointIdentificationAlgorithm` is API 24+.

The change is the two lines you suggested in `openSocket`, before `startHandshake()`.

For the test I did not assert on the parameter — that would have the same weakness as the hand-built-exception test you flagged. Instead `ElectrumProbeServiceTest` now runs the probe against a real TLS electrum server whose certificate is trusted but issued for another name, and asserts `UntrustedCertificate`. I ran it as a negative control with the fix reverted: `probe rejects a certificate issued for another host` fails there, because the probe returns success. A paired test, `probe accepts a certificate issued for the host it connected to` (SAN `IP:127.0.0.1`), passes both with and without the fix.

On whether this can newly reject a server the app would otherwise accept: the Rust electrum client has domain validation on by default and rejects the same certificate anyway, so the probe is only failing earlier and with a specific toast instead of after a full node restart with the generic one. The paired test above is the evidence that a correctly named certificate is unaffected.

No local or regtest setup is affected. The only IP-addressed server in the repo is the local e2e electrum, `ElectrumServers.REGTEST.LOCAL = tcp://${Env.e2eLocalHost}:60001` (default `10.0.2.2`), which is plain TCP — `openSocket` returns before the TLS branch. The other defaults are all `ssl://` with real hostnames (`bitkit.to`, `electrs.bitkit.stag0.blocktank.to`, `electrum.blockstream.info`). There is no docker electrum on a TLS port in this repo.

One caveat worth recording: unit tests run against JVM JSSE, while the app runs on Conscrypt. Both honour `endpointIdentificationAlgorithm`, but the assertion above is JSSE-verified, not Conscrypt-verified.

`just compile`, `just test` and `just lint` all pass.

jvsena42 (commented):
✅ Self contained change

jvsena42:
The CI failure of `ElectrumProbeServiceTest > probe accepts a certificate issued for the host it connected to` was the test harness, not the production change. Fixed in f9ce27047.

Cause. Robolectric installs Conscrypt as the JVM's first security provider and never removes it, and unit tests share one JVM, so every test class that runs after a Robolectric one resolves `SSLContext.getInstance("TLS")`, `TrustManagerFactory` and friends to Conscrypt. The TLS harness built its fake Electrum server from that provider, so on CI the server socket was a Conscrypt one. A Conscrypt server socket cannot finish a handshake on a current JDK — it reflects into `java.net.InetAddress`, which the module system refuses without `--add-opens` — so the server dropped the connection right after the client handshake and the probe read EOF, reporting `NotElectrum`. The wrong-name test still passed because the client rejects the certificate name before the server ever gets that far, which is why only the positive case failed.

It passes locally because Conscrypt ships no darwin-aarch64 native in `conscrypt-openjdk-uber` 2.5.2, so Robolectric's install throws `UnsatisfiedLinkError` on an Apple Silicon machine and SunJSSE stays in charge. Test class order mattered too: on a clean build the classes run roughly alphabetically and this one lands at position 90, but locally an incremental build had hoisted it to position 1, ahead of every Robolectric class.

Reproduced by running the compiled test classes on the Gradle test classpath under Linux x86_64: `LightningNodeServiceTest` followed by `ElectrumProbeServiceTest` fails the positive case, the same pair passes on macOS, and `ElectrumProbeServiceTest` alone passes on both.

Fix. The TLS tests now build one `SSLContext` pinned to the JDK's own JSSE provider, holding both the key managers that serve the certificate and the trust managers that trust it, and use it for both ends. `ElectrumProbeService` takes its `SSLSocketFactory` as a constructor dependency — Hilt provides the platform default, so production behaviour is unchanged — and the tests hand it the factory from that context. `SSLContext.setDefault` and its teardown restore are gone, so the tests neither read nor write JVM-wide SSL state.

The production change is unchanged and still correct: hostname verification behaved identically under SunJSSE and Conscrypt, on JDK 17 and 21, on macOS and Linux. Negative control rerun: with the `endpointIdentificationAlgorithm` line removed, `probe rejects a certificate issued for another host` fails and `probe accepts a certificate issued for the host it connected to` passes, so the pair still discriminates.

Local: `just compile`, `just test` (2608 passing, 0 failures), `just lint` all green.

journeys/settings/electrum-server-error-toasts.xml
```diff
@@ -0,0 +1,37 @@
...
+    Do not tap "Reset To Default" unless the starting server was the default.
+  </description>
+  <actions>
+    <action>Tap the menu icon (testTag "HeaderMenu")</action>
+    <action>Tap Settings (testTag "DrawerSettings")</action>
+    <action>Tap the Advanced tab (testTag "Tab-advanced")</action>
+    <action>Tap Electrum Server (testTag "ElectrumConfig")</action>
+    <action>Verify "Currently connected to" is visible and note the connected server (testTag "Connected")</action>
+    <action>Replace the host (testTag "HostInput") with "electrum.blockstream.info" and the port (testTag "PortInput") with "50002"</action>
+    <action>Select "TLS" under protocol (testTag "ElectrumProtocol")</action>
```
jvsena42, line 23: **LOW** — **Tapping the TLS row after entering port 50002 changes the port field to 60002 on a dev build, so the wrong-network step does not run against the server the journey describes.**

`ElectrumConfigViewModel.setProtocol` (ElectrumConfigViewModel.kt:115-130) replaces the port whenever it is in `defaultElectrumPorts` (`51002, 50002, 51001, 50001`) with `protocol.getDefaultPort()`, which on regtest is `60002` for SSL (ElectrumServer.kt:106-111). The row's `onClick` fires even though the dev default (`ssl://electrs.bitkit.stag0.blocktank.to:9999`) already has TLS selected. Line 20 types `50002`, line 21 taps TLS, so at line 22 the probe runs against `ssl://electrum.blockstream.info:60002`. The network toast still appears because that port is Blockstream's testnet TLS endpoint, so the run passes, but a runner that verifies the field text as the description tells it to sees `60002` where the journey says `50002` and reports a stale step; the PR's manual test 1 and preview describe `50002` for the same reason.

Fix: swap lines 20 and 21 (select TLS first, then type host and port; `setPort` does not touch the protocol), or state in the step that the port is rewritten to `60002` and that this is the intended target.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `2e36ed64b`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Selecting TLS after typing port 50002 silently rewrites the port to 60002, so the journey's stated state does not hold

Checked: Follow-up on 5291fe216 (hostname verification), f9ce27047 (tests) and 3c03cfc3a (journey): the earlier hostname LOW is fixed. Re-reviewed the three own commits since 57b4e9415. Hostname fix (5291fe216): `endpointIdentificationAlgorithm = "HTTPS"` is applied through the get/set SSLParameters round trip at ElectrumProbeService.kt:125, before `startHandshake()` at :126, on the socket built with `createSocket(plain, server.host, port, true)`, so the name checked is the user-typed host. On Android this reaches Conscrypt's `TrustManagerImpl.checkTrusted(chain, authType, session, parameters)` via `RootTrustManager -> NetworkSecurityTrustManager -> getTrustedChainForServer(certs, authType, socket)` (the app has no network security config), which throws `CertificateException` on a name mismatch; that is wrapped as `SSLHandshakeException` and `isCertificateFailure()` (:242-253) maps it to `UntrustedCertificate` -> `settings__es__server_error_certificate`, so a hostname mismatch shows the certificate toast, as the PR body states. Self-signed certs are unchanged: chain validation still fails with a `CertPathValidatorException`/`CertificateException` cause, same class as before this commit; hostname verification never turns a previously-rejected chain into an accepted one. IP hosts: the node's client is ldk-node 0.7.0-rc.66 -> electrum-client 0.24 with `use-rustls-ring` and `ConfigBuilder::new()` defaults (validate_domain = true, ldk-node/src/chain/electrum.rs:565-568), which passes the URL's host to `ServerName::try_from` (IP literal -> `IpAddress`, requires an iPAddress SAN; DNS name -> dNSName SAN, no CN fallback). Conscrypt's `OkHostnameVerifier` has the same IP-SAN/DNS-SAN split and no CN fallback since API 28 (minSdk), so a server the probe now rejects is one the node would reject after the restart; the paired positive test (IP SAN `127.0.0.1`) shows a correctly named certificate still passes. `.onion` hosts over TLS never reach the TLS branch: `plain.connect` fails DNS resolution and reports `Unreachable` (:109-113), unchanged. Provider decoupling (f9ce27047): `HttpModule.provideSslSocketFactory` returns `SSLSocketFactory.getDefault()`, the same factory the probe called inline before, and `ElectrumProbeService` is the only consumer of the binding. Tests: `tlsContext` pins SunJSSE for the context, key managers and trust managers and hands its socket factory to the probe, so no JVM-wide SSL state is read or written; the negative test discriminates because the only difference from the passing positive test is the SAN, so it can only fail through name checking. CI `testDevDebugUnitTest` passed at head 2e36ed64b. Journey: every testTag it names exists at head (`HeaderMenu`, `DrawerSettings`, `Tab-advanced` via `Tab-${tab.name.lowercase()}`, `ElectrumConfig`, `Connected`, `HostInput`, `PortInput`, `ElectrumProtocol` on the TLS row, `ConnectToHost`, `ResetToDefault`, `NavigationBack`), the Auto label logic and the disabled-Connect assertion match `AdvancedSettingsViewModel.electrumHost` and `enabled = !isLoading && hasEdited || !isConnected`, the suites and cross-platform tables are updated, and no bitkit-ios twin exists.

journeys/settings/electrum-server-error-toasts.xml
jvsena42: Confirmed on the branch head. `ElectrumConfigViewModel.setProtocol` (app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigViewModel.kt:115-130) replaces the port with `protocol.getDefaultPort()` whenever it is empty or in `defaultElectrumPorts` (line 59: 51002/50002/51001/50001), and `getDefaultPort` returns 60002 for SSL on REGTEST (app/src/main/java/to/bitkit/models/ElectrumServer.kt:103-111). The TLS row fires `onChangeProtocol` unconditionally (app/src/main/java/to/bitkit/ui/settings/advanced/ElectrumConfigScreen.kt:161-166), and the dev default is `ssl://electrs.bitkit.stag0.blocktank.to:9999` (app/src/main/java/to/bitkit/env/Env.kt:289), so TLS is already selected and the tap only served to rewrite 50002 to 60002.

Fixed by selecting TLS before typing the host and port, since `setPort` (ElectrumConfigViewModel.kt:107-112) does not touch the protocol, and port 9999 from the default server is not a default port so the earlier tap rewrites nothing. Added a verification of the port field text after typing, and spelled the rewrite rule out in the description.

Checked the rest of the journey for the same trap: the 50001 step is the only other one at risk, since 50001 is also in `defaultElectrumPorts` — it never taps a protocol row, and the step now says so explicitly. The 9999 steps are unaffected. Also brought manual tests 1 and 2 and the preview note in the PR body into the same order. Commit 8f00d51b2.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `8f00d51b2`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up: the earlier journey finding is fixed. Follow-up on 8f00d51b2: the earlier port-rewrite LOW is fixed. `ElectrumConfigViewModel.setProtocol` (ElectrumConfigViewModel.kt:115-130) replaces the port with `protocol.getDefaultPort()` whenever it is empty or in `defaultElectrumPorts` (51002/50002/51001/50001), and `getDefaultPort()` returns 60002 for TLS on REGTEST (ElectrumServer.kt:110), so the old order rewrote 50002. The journey now selects TLS before typing host and port, verifies the field still reads 50002, and changes the port to 50001 without re-tapping a protocol row; `setPort` only trims, so both typed values survive. A failed connect writes no settings, so `observeState` does not re-seed the fields between steps. The description records the rewrite rule accurately, including the network-dependent default. No later step taps a protocol row.

ovitrif (approved):
utAck

ovitrif:
@jvsena42 conflicts

ovitrif (dismissed):

jvsena42:
Both points are covered centrally in https://github.com/synonymdev/bitkit-android/issues/1275#issuecomment-5732780955: the conflict is the shared `journeys/README.md` row that every PR in this batch touches, and a merge of current master is running across all of them; the red e2e shards failed in setup (snapshot load, `adb` exit 1, app never reaching Home), not in an assertion.

I will confirm the result on this PR once the merge lands and CI re-runs on the new head.

journeys/README.md
```diff
@@ -129,6 +129,7 @@ fixtures, push notifications) live in each suite's README.
 | [receive](receive) | 1 | Receive sheet tab selection; needs a spending channel, no README |
 | [security](security) | 1 | PIN result sheet layout at a long locale and font scale; no README |
 | [shop](shop) | 1 | Shop Discover category titles and web view handoff; needs Bitrefill reachable; no README |
```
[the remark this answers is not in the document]
jvsena42: Right, and fixed in a1cd640d5: the `settings` row now sits above `shop`. My conflict resolution had appended master's rows after this branch's instead of merging the table in order. The same mistake had hit the suite table on eight other open PRs from this batch, so I fixed those too and checked each result against its `journeys/` folder: one row per suite, alphabetical, and each count matching the folder's `.xml` files.

ovitrif (approved):
utAck

Issue #1125, opened 2026-07-31
fix: recover ldk restart after background timeout


ovitrif:
### What happened?

Backgrounding Bitkit about one second after tapping **Restart** can make the restart's `start()` call stall in Electrum fee-rate estimation until its approximately 15-second deadline. The timeout leaves the node lifecycle in `ErrorStarting`, shows a failure toast, and requires the next foreground resume to trigger recovery.

Unrelated Ktor requests continue succeeding during the stall. The next fee-rate cache update completes in approximately 680 ms.

### Expected behavior

The restart flow should recover automatically from a transient fee-rate estimation timeout and return the node to `Running` without another foreground cycle or manual action.

### Steps to Reproduce

1. Open **Settings → Advanced → Dev Settings → LDK**.
2. Tap **Restart**.
3. Background Bitkit approximately one second later.
4. Wait for the restart attempt to finish.
5. Foreground Bitkit and inspect the node state.

### Logs / Screenshots / Recordings

```text
Starting up LDK Node …
Updating fee rate estimates timed out: deadline has elapsed
Failed to start node during restart
LDK Node error: Feerate estimation update timeout.
Fee rate cache update finished in 682ms.
```

[Joao's review evidence](https://github.com/synonymdev/bitkit-android/pull/1122#pullrequestreview-4830791946) includes the full run details and the known-failing `restart-while-backgrounded.xml` journey.

### Bitkit Version

2.4.0 development build from #1122.

### Device / OS

Pixel-class arm64 emulator, Android API 36.

### Reproducibility

Always — 3/3 with `com.synonym:ldk-node-android:0.7.0-rc.62` and 2/2 with `0.7.0-rc.58`.

### Additional context

`restartNode()` calls `start(shouldRetry = false)`. A transient start failure therefore commits `ErrorStarting` immediately, while the next foreground-triggered start can complete the same fee-rate work successfully.

### Acceptance criteria

- Backgrounding Bitkit during **Restart** does not leave the node in `ErrorStarting`.
- A transient fee-rate estimation timeout triggers bounded automatic recovery.
- Successful recovery returns the node to `Running` without another foreground cycle or manual restart.
- Recovered transient failures do not show a terminal restart-failure toast.
- Automated coverage reproduces the background-during-restart sequence and verifies node usability after recovery.

PR #1294, opened 2026-09-16
fix: retry node start on restart

jvsena42:
Fixes #1125
Related to #1275

This PR makes the node restart retry a transient start failure instead of stopping at ErrorStarting.

### Description

- Fixes a restart giving up after one transient start error (e.g. "Feerate estimation update timeout"): the restart now uses the same bounded retry as a normal start, 2 attempts 2s apart.
- Makes the restart retry yield to a background stop: if the app was backgrounded during the failed attempt, the retry is skipped, the node goes to Stopped instead of ErrorStarting, and the next foreground starts it. If the app returns to the foreground before the retry, the retry runs.
- Stops the restart retry from cancelling a background stop, so it cannot start the node while the app is in the background.
- Hides the "Failed to restart node" toast on LDK Debug when the restart yielded to a background stop. Real failures still show it.
- Keeps the retry alive when the caller's scope dies: LDK Debug's Restart now goes through `LightningRepo.restartNodeDetached`, which runs the restart on the repo scope the way `stopDebounced` already does. Leaving the screen inside the 2s retry window no longer cancels the retry and leaves the node Stopped with a non-null service node. The toast is still collected on `viewModelScope`, so a live screen reports the outcome unchanged.
- Adds a changelog fragment for the fix.

#### Out of Scope

- `LightningRepo.restartNode`: a restart that stays backgrounded while the first attempt fails does not recover on its own. It yields to the stop, ends Stopped, and recovers on the next foreground. The issue's "back to Running without another foreground cycle" criterion is not met in that case, to avoid starting the node in the background.
- `LightningRepo.restartNode` graph-reset leg (known gap from review): when the required peers are missing from the network graph, the restart stops the node and calls the public start, which now retries (it did not before) and cancels a pending background stop. A backgrounded restart that also needs a graph reset and hits a transient error can start the node in the background and show the failure toast. Deferred to a follow-up.
- `LightningRepo.start`: the normal app-launch start retry still cancels a pending background stop. Unchanged here.
- `WalletViewModel.onRestoreRetry` keeps calling `restartNode` directly. Its caller is activity-scoped, so the cancellation window the debug screen has does not apply, and detaching it there would change cancellation semantics on the restore path for no gain.
- Automated coverage of the issue's full background-during-restart sequence with a real node: not added. Unit tests cover the retry and yield logic with a mocked service.
- `LightningRepo.startNode`: no refactor for the detekt CyclomaticComplexMethod warning (17 on the base branch, 20 now), to avoid behaviour changes on the node start path.
- iOS parity: not checked in `bitkit-ios`; a review of its restart retry is a follow-up.

### Design

N/A — no UI changes.

### Preview

No UI changes. Recordings from the regtest emulator. The transient failure does not happen on demand, so the "simulated" runs used a temporary uncommitted patch that throws "Feerate estimation update timeout" on the first start attempt after a restart.

Before this fix, simulated failure in the foreground: the restart gives up with "Failed to restart node":

https://github.com/user-attachments/assets/615118b0-b4df-41d3-91a4-1af0714d0274

After this fix, same simulated failure in the foreground: the retry succeeds with "Node restarted":

https://github.com/user-attachments/assets/74a254a5-c2e6-4615-a320-8c3f9c1dbb9d

After this fix, simulated failure with Home pressed 1s after Restart: the retry yields to the stop, no toast, and the node is Running after foregrounding:

https://github.com/user-attachments/assets/180bfe29-d9c9-495e-8800-d65d4575f873

| Before (simulated failure) | After (simulated failure) | After, backgrounded then foregrounded |
| --- | --- | --- |
| Failed to restart node toast | Node restarted toast | No failure toast after foreground |

Issue scenario without fault injection (Restart, Home 1s later, foreground after 25s), run 3 of 3; no transient error occurred in any run:

https://github.com/user-attachments/assets/f87fad56-76a2-4fc2-8b99-31b939c09db8

| Restart result | LN payment after restarts |
| --- | --- |
| Node restarted toast | Payment received |

<details>
<summary>Issue scenario runs 1 and 2</summary>

https://github.com/user-attachments/assets/b8ea3295-378f-440c-a550-2dbb334715dc

https://github.com/user-attachments/assets/75e980b1-ed4d-4523-8ca8-793b56f8a6e7

</details>

#### Manual Tests

"Keep Bitkit active in background" must be off for the background tests; while it is on, the foreground service keeps the node alive and no stop is scheduled. Check the app log in `files/logs/` for the lifecycle lines.

- [ ] **1.** Settings → Advanced → Dev Settings → LDK → tap Restart: "Node restarted" toast, log shows "Node restarted successfully".
- [ ] **2a.** Background toggle off → LDK → tap Restart → Home ~1s later → wait 25s → foreground: no "Failed to restart node" toast, node reaches Running.
  - [ ] **2b.** Receive → Spending tab → pay a 1000 sat invoice from the LSP: "Received Instant Bitcoin" sheet shows.
- [ ] **3a.** With a patch that throws AppError("Feerate estimation update timeout") on the first start after restart → LDK → tap Restart: log shows "Start error, will retry...", then "Node restarted successfully" and a "Node restarted" toast.
  - [ ] **3b.** Same patch, background toggle off → tap Restart → Home ~1s later: log shows "Skipped start retry because a stop was requested", no toast; foreground: lifecycle 'Running'.
  - [ ] **3c.** Same patch, app kept in the foreground → tap Restart → Back out of LDK Debug ~1s later: log shows "Start error, will retry..." then "Node restarted successfully", Home shows no health error (no toast, the screen that would show it is gone).
- [ ] **4.** `regression:` Cold launch: node reaches Running, Home shows balances, no "Start error" in the log.
- [ ] **5.** `regression:` Restore flow → start failure → Retry: node reaches Running.

#### Automated Checks

- Unit tests added in `LightningRepoTest.kt`:
  - A transient start failure during restart is retried after 2s and ends Running.
  - Restart ends ErrorStarting when both bounded attempts fail.
  - A stop requested during the failed attempt skips the retry, returns NodeStartYieldedToStopError with the original cause, ends Stopped, runs the debounced stop, and a later foreground start reaches Running.
  - A foreground that cancels the stop before the retry lets the retry run and end Running, with no extra stop.
  - A stop requested inside the retry attempt still runs.
  - `restartNodeDetached` completes the retry and ends Running after the calling scope is cancelled inside the retry delay.
- The first four tests fail with the old `start(shouldRetry = false)` restored. The detached-retry test fails against a plain `restartNode()` call: one start attempt, node left Stopped.
- Ran locally: `just compile`, `just test` (2606 tests), `just lint`.


app/src/main/java/to/bitkit/repositories/LightningRepo.kt
```diff
@@ -2067,8 +2097,12 @@ class LightningRepo @Inject constructor(
             Logger.error("Failed to stop node during restart", it, context = TAG)
             return@withContext Result.failure(it)
         }
-        start(shouldRetry = false).onFailure {
-            Logger.error("Failed to start node during restart", it, context = TAG)
+        startNode(shouldRetryYieldToStop = true).onFailure {
```
jvsena42, line 2112: **LOW, gated** — **Back during the new restart retry delay cancels the retry; the node stays Stopped, no toast, no recovery until the next foreground cycle.**

`LdkDebugViewModel.restartNode` runs in `viewModelScope` (app/src/main/java/to/bitkit/viewmodels/LdkDebugViewModel.kt:160) and the VM is `hiltViewModel()` scoped to the LdkDebug back-stack entry (app/src/main/java/to/bitkit/ui/screens/settings/LdkDebugScreen.kt:48). Trace: tap Restart -> `restartNode` (LightningRepo.kt:2094) -> `stop()` -> `startNode(shouldRetryYieldToStop = true)` (LightningRepo.kt:2100) -> first attempt throws the transient error -> state restored to `initialLifecycleState` = Stopped, `shouldRetryStart = true` -> `delay(2.seconds)` (LightningRepo.kt:472). User presses Back inside that 2s -> VM cleared -> `CancellationException` out of `delay` -> the retry at LightningRepo.kt:478 never runs and `restartNode` never returns a Result. Outcome: `nodeLifecycleState = Stopped` with `lightningService.node != null` (setup done, never started), `HealthState.ERROR` on Home, no "Failed to restart node" toast (the VM that would send it is gone), and nothing restarts the node until ON_START calls `WalletViewModel.start()`. Before this PR the same Back press during `lightningService.start()` also cancelled, but there was no post-failure window; this PR adds a 2s window where the restart is already known to have failed and the bounded recovery it promises is silently dropped. Debug-only route (Settings -> Advanced -> Dev Settings -> LDK), so it affects opted-in users only; `WalletViewModel.onRestoreRetry` is activity-scoped and unaffected.

Fix (narrowest): run the repo call in a scope that survives the screen, e.g. `viewModelScope.launch(bgDispatcher) { withContext(NonCancellable) { lightningRepo.restartNode() } ... }` is too broad for the toast path, so prefer launching `lightningRepo.restartNode()` on the repo's own `appScope` (as `stopDebounced` already does) and only collecting the result for the toast inside `viewModelScope`.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `a655cc43f`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW (gated) — Back from LDK Debug during the 2s retry delay drops the restart and leaves the node Stopped with no signal

Checked: Traced both restartNode callers at the PR head (LdkDebugViewModel.restartNode and WalletViewModel.onRestoreRetry); the Electrum/RGS config-change paths under configChangeMutex call stop()/start() directly and are untouched, and every public start() path (WalletViewModel, LightningNodeService, WakeNodeWorker, AppViewModel) still resolves to shouldRetryYieldToStop=false / shouldCancelPendingStop=true, so foreground and worker starts keep cancelling a pending background stop exactly as before. The yield check reads pendingStopJob after the 2s delay: a stop scheduled during the failed attempt blocks on lifecycleMutex until the attempt releases it, and a stop that already ran leaves its completed Job referenced (nothing nulls it except cancelPendingStop), so the retry still yields; every ON_START calls WalletViewModel.start() which calls cancelPendingStop() before its guards, so no stale completed Job survives into a foreground Restart tap. Concurrent foreground start during the retry window is safe: the retry attempt waits on the mutex, sees Running/Starting and takes skipStartForRunningNode, which succeeds for the null custom-URL case restartNode uses. In the restore-retry flow the yield path ends Stopped rather than ErrorStarting; ContentView keeps the InitializingWalletView until the next ON_START start() reaches Running or ErrorStarting, so the user is not stuck past the background period. This PR is stacked on #1291 (base fix/845-node-start-cancellation) at an older commit than #1291's current head (68e0b8977 adds the keep-live-listener branch in reconcileCancelledStart); that change is inside the cancellation reconcile, not the retry/yield leg, so the rebase is mechanical with no semantic interaction. No iOS twin PR found for restart retry.

app/src/main/java/to/bitkit/repositories/LightningRepo.kt
jvsena42: Confirmed the trace on the branch with a test before fixing it: a caller cancelled inside the 2s retry delay drops the retry, leaving one start attempt and the node Stopped.

Fixed in 6ecd58466 along the lines you suggested. Added `LightningRepo.restartNodeDetached()`, which runs `restartNode()` on the repo scope (`scope.async { restartNode() }.await()`) the way `stopDebounced` already does, and pointed `LdkDebugViewModel.restartNode` at it. The caller still awaits the Result on `viewModelScope`, so a live screen keeps its "Node restarted" / "Failed to restart node" toasts and the `NodeStartYieldedToStopError` suppression unchanged; a cleared ViewModel now only loses the toast, never the retry.

Kept it as a separate method rather than detaching `restartNode` itself so `WalletViewModel.onRestoreRetry` is untouched — that caller is activity-scoped and does not have this window.

Covered by `restartNodeDetached completes the start retry after the caller is cancelled` in `LightningRepoTest.kt`: cancel the calling scope after the first failed attempt, advance past the retry delay, assert two start attempts and Running. It fails against a plain `restartNode()` call. `just compile`, `just test` and `just lint` pass locally.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `6ecd58466`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 6ecd58466: the earlier retry-cancel LOW is fixed. Follow-up at 6ecd58466 resolves the earlier LOW (id 4046143178): LdkDebugViewModel.restartNode now awaits LightningRepo.restartNodeDetached (LightningRepo.kt:2104), which runs restartNode inside scope.async on the repo's appScope (LightningRepo.kt:142, SupervisorJob + logging handler, never cancelled), so a Back press inside the 2s retry delay only cancels the caller's await; the deferred keeps running and the new test cancels a CoroutineScope(testDispatcher) after the first failed attempt and asserts two start calls and Running, which fails with a plain restartNode() call. CancellationException semantics for the caller are preserved: await() rethrows the caller's cancellation, the VM's isLoading/toast path simply stops, and no CancellationException reaches the repo path (stop() is NonCancellable, startNode only sees cancellation if the repo scope is cancelled). The detached retry stays bounded: restartNode is one stop() plus startNode(shouldRetryYieldToStop=true) whose retry call passes shouldRetry=false (LightningRepo.kt:478-489); the graph-reset leg already noted as out of scope in the PR body is unchanged. Stale-retry races traced: a background stop (ON_STOP -> WalletViewModel.stop -> stopDebounced, ContentView.kt:300-308) is still honoured by the pendingStopJob check at LightningRepo.kt:475, and a foreground start cancelPendingStop + Starting/Running state makes the retry take skipStartForRunningNode with null custom URLs; a wallet wipe during the delay is refused by the isWiping check inside lifecycleMutex (LightningRepo.kt:372, WipeWalletUseCase.kt:57-68 set/clear it around stop+wipeStorage), and a retry landing after the wipe finishes only hits setup() on an empty keychain -> ErrorStarting, which the next onboarding start() overrides since ErrorStarting is not runningOrStarting; recovery mode enabled during the delay is re-checked at the top of the recursive startNode call. Interaction with #1291 at f3e3493f5: recoverOnCancellation launches restartWithPreviousConfig on the same repo scope under configChangeMutex; restartNode does not take configChangeMutex, but both builds serialize on lifecycleMutex and the isRunningOrStarting skip, so the worst interleaving (retry's default-config start landing between a config change's stop and start) makes that change fail with NodeConfigNotAppliedError and self-recover without persisting the new server, no double start and no recovery loop; the window is the microseconds between the config change's stop() releasing and start() re-acquiring the mutex, so not raised. Re-entering LDK Debug and tapping Restart again while the detached retry is pending is bounded (each restart max two attempts, the second sees Running and skips). No iOS twin PR found for restart retry.

Issue #804, opened 2026-02-24
bug: app update check only gates the home screen


ovitrif:
The app update check (both critical and non-critical) only evaluates when the user reaches the home screen. It doesn't cover onboarding or other flows, so it can miss showing the update prompt entirely on first launch.

## Repro steps

1. Install an older build
1. Restore a wallet
1. **Expected:** the mandatory update full-screen shows
1. **Actual:** it doesn't show
1. Force-close the app (also from Android app info screen)
1. Reopen the app — now the mandatory update screen shows

## Suggestions

- The update check should also cover the onboarding flow (when no wallet is loaded), scoped at the highest level which includes all screens within the app
- This should apply to the non-critical app update sheet too
- iOS has the same home-screen-only gating for non-critical updates (timed sheets triggered from `HomeView.onAppear` only). Critical updates on iOS are shown at the top level in `AppScene` so they're not affected, but the non-critical path has the same limitation.

The old React Native App does this much better, gating the Onboarding flow screens too:

<img width="200" src="https://github.com/user-attachments/assets/3f11af25-d5a1-49da-8d14-0a92da09ee63" />

PR #1295, opened 2026-09-16
fix: show critical update during onboarding

jvsena42:
Related to #804
Related to #1275

This PR shows the mandatory update screen at the top level of the app, so it also appears during onboarding and restore instead of only after the wallet home screen loads.

### Description

- Keeps the critical update result as state in the app view model instead of a one-shot navigation event, because the event was dropped when no wallet existed and nothing was listening for it.
- Shows the Critical Update screen full screen before the migration, onboarding and wallet branches, so it blocks every flow and is never covered by the PIN screen.
- Removes the Critical Update route from the wallet navigation graph, since the screen is no longer reached through navigation.
- Stops the node when the app goes to the background while the Critical Update screen is shown, using the same keep-alive condition as the wallet screens, so node lifecycle matches master after the switch.
- Blocks the new transaction sheet while a critical update is required, and clears one that is already showing when the check completes, so the sheet cannot cover the Critical Update screen.

#### Out of Scope

- `AppViewModel` / home timed sheets: the non-critical "Update available" sheet still shows only from the home screen. This matches iOS, which also gates the non-critical sheet on the home view. #804 stays open for it.
- iOS: no parity change needed for the critical path, iOS already shows the critical update at the top level in `AppScene`.
- `CriticalUpdateScreen.kt`: layout unchanged. The illustration renders larger on tall devices than in the 375x812 Figma frame; this predates the PR.
- `journeys/app-update/critical-update-onboarding.xml`: the journey is committed, but it is not run here — it needs a non-debug build and a local change to reach a critical release, which the journey now documents as its setup route.

### Design

- Update › Forced update: https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-148422

### Preview

The screenshots come from a temporary, uncommitted build that faked a critical release on an existing wallet, because debug builds skip the release check and reaching onboarding would have meant wiping a funded wallet. The onboarding case itself was not recorded.

| Figma | After |
| - | - |
| Figma Forced update | Critical Update screen after fix |

Launch, splash, Home, then about 6s later the release check completes and the app switches to the Critical Update screen with no PIN overlay:

https://github.com/user-attachments/assets/310d2f72-7b52-44e0-86cc-3eaaec7dbdb1

No "before" capture: on master the navigation event is emitted with no replay and its only listener lives in the wallet content view, which is not composed during onboarding, so there is nothing to show.

#### Manual Tests

Needs a non-debug build and a release feed listing a critical Android build with a higher build number than the installed one.

- [x] **1.** Fresh install, no wallet → launch: Critical Update shows instead of Terms of Use.
  - [x] **1b.** back: app closes to the launcher, onboarding is not revealed.
  - [x] **1c.** relaunch → tap Update Bitkit: Play Store opens the Bitkit listing.
- [x] **2.** Fresh install, slow network → launch → Restore wallet before the release check finishes: Critical Update replaces the restore flow once the check completes.
- [x] **3.** Existing wallet with PIN → launch: Critical Update shows with no PIN overlay and no bottom bar.
- [x] **4a.** Existing wallet, keep-alive off → Critical Update → Home key: node stops.
  - [x] **4b.** Keep-alive on and notifications granted → Critical Update → Home key: node keeps running in the foreground service.
- [x] **5.** `regression:` Release feed with a newer non-critical build → Home: "Update available" sheet still shows, no Critical Update.
- [x] **6.** `regression:` Debug build → launch: no Critical Update screen.
- [x] **7.** Existing wallet, critical release → receive a payment while the Critical Update screen is shown: no transaction sheet appears over it.

The journey below captures test 1 as an agent-run spec. It is committed at `journeys/app-update/critical-update-onboarding.xml` (commits ca846203d, 5b3e5cff8, 87f3ea03d), with its rows added to the suites and cross-platform tables in `journeys/README.md`. It cannot be run from a stock build of this repo: debug builds skip the check, and neither release feed reachable at build time satisfies it, so the journey now gates on that as its first action and spells out the local setup needed:

<details><summary>journeys/app-update/critical-update-onboarding.xml</summary>

```diff
diff --git a/journeys/app-update/critical-update-onboarding.xml b/journeys/app-update/critical-update-onboarding.xml
new file mode 100644
index 000000000..da17bcca2
--- /dev/null
+++ b/journeys/app-update/critical-update-onboarding.xml
@@ -0,0 +1,61 @@
+<journey name="critical update during onboarding">
+  <description>
+    Proves the mandatory update screen blocks the app before a wallet exists, not only on the home
+    screen (synonymdev/bitkit-android#804). MainActivity shows CriticalUpdateScreen ahead of the
+    migration, onboarding and wallet branches, so nothing else is composed while it is up.
+
+    This journey cannot be run against a stock build of this repo, and the first action is the gate
+    that says so. Three things stand in the way, all on the current head:
+    debug builds skip the check entirely (`checkCriticalAppUpdate` returns early on `Env.isDebug`,
+    AppViewModel.kt:5598-5599), so `just run` and `just install` never reach it;
+    `Env.RELEASE_URL` (Env.kt:147-151) is fixed at build time, pointing at the bitkit-e2e-tests feed
+    when `E2E=true` and at the bitkit-android production feed otherwise;
+    and neither feed satisfies the condition — the e2e feed lists android `buildNumber: 0`,
+    `critical: false`, while the production feed lists `buildNumber: 188`, `critical: true` against
+    `versionCode = 188` (app/build.gradle.kts:189), and the check is a strict `>`
+    (AppViewModel.kt:5604), so the flag never rises. Both feeds are release assets on shared repos
+    that a runner must not rewrite.
+
+    Setup, as one local uncommitted change to revert afterwards:
+    1. Confirm the production feed still marks the Android build critical:
+       `curl -sL https://github.com/synonymdev/bitkit-android/releases/download/updater/release.json`.
+       If `critical` is false there, point `Env.RELEASE_URL` at a fixture you serve yourself instead
+       and skip step 2. The fixture has to be that feed's whole object with only `buildNumber` and
+       `critical` edited (`critical: true`, `buildNumber` above the installed `versionCode`):
+       `PlatformDetails` (data/dto/AppUpdaterDTO.kt) declares `version`, `notes`, `pub_date` and
+       `url` with no defaults, and the response is decoded with the default strict `Json`
+       (AppUpdaterService.kt:25), so a fixture that drops a field — or adds one — throws,
+       `checkCriticalAppUpdate` takes `onFailure` and only logs a warning (AppViewModel.kt:5608-5610),
+       and the flag stays false. Serve it over HTTPS, e.g. a raw gist or a release asset on a repo of
+       your own: a release build has no `usesCleartextTraffic` (it is set only in
+       app/src/debug/AndroidManifest.xml) and targetSdk 36 blocks plain `http://`, so a local
+       `http://` server is refused the same silent way.
+    2. Lower `versionCode` in app/build.gradle.kts below the feed's `buildNumber`, e.g. to 1.
+    3. Build non-debug without `E2E`: `just build assembleDevRelease`, then install the APK from
+       `app/build/outputs/bitkit/devRelease/`. This needs a release keystore in keystore.properties,
+       and release builds carry only armeabi-v7a and arm64-v8a, so use an arm device or AVD.
+    4. Fresh install or clear app data, so no wallet exists and onboarding is what launches. Use a
+       throwaway device or AVD: never wipe a funded wallet to reach onboarding.
+
+    If the gate does not hold, report the journey as not run rather than failed — Terms of Use at
+    the third action means the fixture is missing, not that the app regressed. `adb logcat -s APP:V`
+    tells the two apart: "Failure fetching new releases" means the feed did not decode or was not
+    reachable, so the setup is wrong and the journey is again not run.
+    The screen has no testTags, so the steps assert its visible text. The headline is uppercased and
+    carries a line break, so it reads "UPDATE" then "BITKIT NOW". Reverting the local change is the
+    negative check: onboarding then stays on the Terms of Use screen.
+  </description>
+  <actions>
+    <action>Verify the setup above is in place: the installed build is non-debug and built without E2E, and the feed at Env.RELEASE_URL is served over HTTPS and lists a complete Android object with critical true and a buildNumber above the installed versionCode. If not, stop and report the journey as not run</action>
+    <action>Launch the app</action>
+    <action>Wait up to 10 seconds for the release check to finish</action>
+    <action>Verify that "Critical Update", the "UPDATE" / "BITKIT NOW" headline and "Update Bitkit" are visible</action>
+    <action>Verify that the Terms of Use screen (tag "TOS") is not visible</action>
+    <action>Press the system back button</action>
+    <action>Verify that the app closed to the launcher instead of revealing onboarding</action>
+    <action>Launch the app again</action>
+    <action>Verify that "Critical Update" is visible again once the release check finishes</action>
+    <action>Tap "Update Bitkit"</action>
+    <action>Verify that the Play Store (or a browser) opens the Bitkit listing</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `AppViewModelSendFlowTest.kt` covers the critical update check. A newer critical build sets the flag; a newer non-critical build, the same build number marked critical, and a failed fetch do not; debug builds skip the fetch entirely.
- `AppViewModelSendFlowTest.kt` also covers the transaction sheet interaction: raising the flag clears a showing sheet, and a later payment cannot open one.
- Instrumented tests added in `RootDestinationTest.kt`, which renders `RootDestination` — the branch chain, extracted out of `onCreate` into a stateless composable taking `isCriticalUpdateRequired`, `isShowingMigrationLoading`, `walletExists` and `isRecoveryMode` plus a content slot per branch — with a tagged stub per slot. Nine cases, each asserting one slot renders and the other three do not: the critical update wins over migration loading, over onboarding, over the wallet and in recovery mode; migration loading beats onboarding; recovery mode skips both migration loading and onboarding; and the plain onboarding and wallet cases. Moving the critical branch below any other fails four of them.
- The `ON_STOP` observer is still checked by hand: it composes only inside the `criticalUpdate` slot, but asserting the stop itself needs a `TestLifecycleOwner` and `lifecycle-runtime-testing` is not a dependency in this project.
- Device check on an existing wallet with a simulated critical release: screen appears after the check, survives reopening from recents, back closes the app, Update Bitkit opens the Play Store, and the node keeps running with keep-alive on.
- Local verification: `just build`, `just compile`, `just test`, `just lint`. `RootDestinationTest.kt` was compiled only, through `just build compileDevDebugAndroidTestKotlin`; instrumented tests need a device and were not run here, they run in CI's `ui-tests.yml`. The follow-up commits 5b3e5cff8 and 87f3ea03d touch only `journeys/app-update/critical-update-onboarding.xml` and `journeys/README.md`, so they need no Gradle run; the journey XML was checked to parse.


app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -139,7 +145,23 @@ class MainActivity : FragmentActivity() {
...
+                                keepActive &&
+                                appViewModel.isForegroundServiceRunning()
+                            if (walletExists && !isRecoveryMode && !keptAliveByService) {
+                                walletViewModel.stop()
+                            }
+                        }
+                        lifecycle.addObserver(observer)
+                        onDispose { lifecycle.removeObserver(observer) }
+                    }
+                    CriticalUpdateScreen()
```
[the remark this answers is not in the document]
jvsena42: Confirmed. NewTransactionSheet is rendered outside the branch chain in MainActivity, so an incoming payment on an existing wallet (hwWalletRepo.receivedTxs, notifyPaymentReceived, notifyChannelReady) put it over the Critical Update screen, and its Detail button emitted a MainScreenEffect.Navigate that nothing collects there, so it only dismissed the sheet.

Fixed in 09d9eb54b in AppViewModel rather than in the composition: showTransactionSheet now returns early while isCriticalUpdateRequired is set, and checkCriticalAppUpdate calls hideNewTransactionSheet when it raises the flag, which covers a sheet that was already showing when the check completes. The flag never resets in a session, so nothing can repopulate it. Two tests added in AppViewModelSendFlowTest.kt for both directions.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `282f3e485`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Traced the critical-update flag from `checkCriticalAppUpdate` (launched once in `AppViewModel.init`) to the top-level branch in `MainActivity`. The branch comes before the migration, onboarding and wallet branches, so it replaces `OnboardingNav`, the restore flow and `ContentView`. The only surfaces left above it are `ToastOverlay` and `SplashScreen`. With the NavHosts gone, Back finishes the activity, and sheet or deeplink state has nowhere to render, so the screen cannot be bypassed. It also cannot trap a user who has no critical release: the flag is only raised for a critical build with a higher build number, and a failed fetch leaves it false. The greptile transaction-sheet thread is resolved by 282f3e485. `_transactionSheet` is only populated through `showTransactionSheet`, which now returns early while the flag is set, and `checkCriticalAppUpdate` clears a sheet that is already showing. If wallet creation or restore is in progress when the flag rises, the work continues in `WalletViewModel`'s scope: the node, restore and backup observation are not tied to the disposed composables. The new ON_STOP observer uses the same condition as `ContentView`'s observer, so the node still stops in the background. `Routes.CriticalUpdate` has no remaining references, and `docs/screens-map.md` still maps `CriticalUpdateScreen.kt`.

journeys/app-update/critical-update-onboarding.xml
```diff
@@ -0,0 +1,26 @@
+<journey name="critical update during onboarding">
+  <description>
+    Proves the mandatory update screen blocks the app before a wallet exists, not only on the home
+    screen (synonymdev/bitkit-android#804). MainActivity shows CriticalUpdateScreen ahead of the
+    migration, onboarding and wallet branches, so nothing else is composed while it is up.
+    Preconditions: a non-debug build, because debug builds skip the release check; the release.json
```
jvsena42: **LOW** — **The precondition cannot be set up from this repo, so any runner gets a false fail at step 3 (Terms of Use shows instead of Critical Update).**

`Env.RELEASE_URL` (Env.kt:147-151) is hardcoded, with no dev setting or build-time override:
- `just e2e` (the only non-debug build recipe in the Justfile, `E2E=true assembleDevRelease`) reads `https://github.com/synonymdev/bitkit-e2e-tests/releases/download/updater/release.json`, which today lists android `buildNumber: 0`, `critical: false`.
- A plain `assembleDevRelease` reads the production feed `https://github.com/synonymdev/bitkit-android/releases/download/updater/release.json`, which lists `buildNumber: 188`, `critical: true`. Head is `versionCode = 188` (app/build.gradle.kts:189), so `it.buildNumber > BuildConfig.VERSION_CODE` in `checkCriticalAppUpdate` (AppViewModel.kt:5604) is false and the flag never rises.

Both feeds are GitHub release assets the runner cannot rewrite without editing a shared release, and the journey does not say which build or feed to use. The "toggling the release to non-critical" negative check has the same problem. The PR body already says the screenshots came from an uncommitted fake-release build.

Fix (narrowest): add an override the runner can set, e.g. a `RELEASE_URL` build-config field read from an env var the same way `E2E_LOCAL_HOST` is, and name the exact command plus a fixture JSON with a high `buildNumber` and `critical: true` in the journey's preconditions. Alternatively state in the precondition that the run needs a temporary local change to `Env.RELEASE_URL` or `versionCode`, so the runner does not try the stock builds.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `d0b510eb3`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Critical-release precondition cannot be met with repo tooling

Checked: Follow-up on ca846203d (new journey): Reviewed commit ca846203d against MainActivity.kt, CriticalUpdateScreen.kt and AppViewModel.checkCriticalAppUpdate at head d0b510eb3. The asserted text matches strings.xml (Critical Update, Update\n<accent>Bitkit now</accent> uppercased by Display, Update Bitkit), TOS exists on TermsOfUseScreen, the critical branch precedes OnboardingNav so Terms of Use is not composed, Back has no handler and no NavHost so it leaves the app, and Update Bitkit opens Env.PLAY_STORE_URL. The precondition is the problem, see the finding.

journeys/app-update/critical-update-onboarding.xml
jvsena42: Confirmed, all of it. `checkCriticalAppUpdate` returns early on `Env.isDebug` (AppViewModel.kt:5598-5599), `Env.RELEASE_URL` is fixed at build time with no override (Env.kt:147-151), and the check is a strict `>` against `versionCode = 188` (AppViewModel.kt:5604, app/build.gradle.kts:189). I fetched both feeds: the e2e one lists android `buildNumber: 0`, `critical: false`, the production one `buildNumber: 188`, `critical: true` — equal, not greater, so the flag never rises either way.

I took the second option rather than adding a `RELEASE_URL` build-config field, since nothing but this manual journey would read it. The journey now says outright that it cannot be run from a stock build of this repo, names the three reasons with file:line, and gives the setup as one local uncommitted change: check the production feed still marks the build critical, lower `versionCode` below its `buildNumber`, build with `just build assembleDevRelease` (non-debug, no `E2E`), install from `app/build/outputs/bitkit/devRelease/`, and clear app data. It also notes the release ABI filters, so the device or AVD has to be arm. A new first action gates on the precondition and tells the runner to report the journey as not run instead of failed, so Terms of Use at the old step 3 is no longer a false fail. The suite row in `journeys/README.md` says the same.

Journey-only change, so no Gradle run; I checked the XML parses. Commit 5b3e5cff8.

journeys/app-update/critical-update-onboarding.xml
```diff
@@ -0,0 +1,50 @@
...
+    and neither feed satisfies the condition — the e2e feed lists android `buildNumber: 0`,
+    `critical: false`, while the production feed lists `buildNumber: 188`, `critical: true` against
+    `versionCode = 188` (app/build.gradle.kts:189), and the check is a strict `>`
+    (AppViewModel.kt:5604), so the flag never rises. Both feeds are release assets on shared repos
+    that a runner must not rewrite.
+
+    Setup, as one local uncommitted change to revert afterwards:
+    1. Confirm the production feed still marks the Android build critical:
+       `curl -sL https://github.com/synonymdev/bitkit-android/releases/download/updater/release.json`.
+       If `critical` is false there, point `Env.RELEASE_URL` at a fixture you serve yourself instead
```
jvsena42, line 22: **LOW** — **The fallback route as written does not raise the flag, so a runner who takes it gets Terms of Use and a false fail.**

The fallback says to serve a fixture with `critical: true` and a higher `buildNumber`. `PlatformDetails` (AppUpdaterDTO.kt) has no defaults for `version`, `notes`, `pub_date` or `url`, so a fixture with only those two fields fails to deserialize. `checkCriticalAppUpdate` then takes `onFailure`, logs a warning, and leaves the flag false. Release builds also have no `usesCleartextTraffic` (it is set only in app/src/debug/AndroidManifest.xml), so a plain `http://` local server is blocked. In both cases the runner has followed the setup and passes the gate action, then sees Terms of Use.

This matters only once the production feed stops marking the build critical. Fix: say the fixture must be served over HTTPS (for example as a GitHub release asset or gist raw URL) and must mirror the full `platforms.android` object of the production feed, with `buildNumber` and `critical` edited.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `5b3e5cff8`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Fallback self-served fixture needs HTTPS and every PlatformDetails field

Checked: Follow-up: Follow-up on 5b3e5cff8, which answers the critical-release precondition LOW. The primary setup route is real at head. `checkCriticalAppUpdate` returns early on `isDebug` (AppViewModel.kt:5599), and `Env.isDebug = BuildConfig.DEBUG`. Without E2E, `Env.RELEASE_URL` is the production updater feed (Env.kt:147-151). The check is `isCritical && buildNumber > BuildConfig.VERSION_CODE` (AppViewModel.kt:5604), so lowering `versionCode` (build.gradle.kts:189) below the feed's 188 raises the flag. `just build assembleDevRelease` is valid because the `build` recipe takes a task argument. The APK lands in `outputs/bitkit/devRelease` (build.gradle.kts:343). The release build type does need the release signing config (keystore.properties or the KEYSTORE_* env vars) and is filtered to armeabi-v7a/arm64-v8a (:257-261), as the journey says. The new gate action reports a missing fixture as not run, so the old false fail at Terms of Use is gone. One LOW on the fallback route.

journeys/app-update/critical-update-onboarding.xml
jvsena42: Confirmed, both halves. `PlatformDetails` declares `version`, `buildNumber`, `notes`, `pub_date`, `url` and `critical` with no defaults (data/dto/AppUpdaterDTO.kt:23-38), and `AppUpdaterService` decodes with the default strict `Json` (AppUpdaterService.kt:25), which also rejects unknown keys — so a trimmed fixture *and* an embellished one throw, `getReleaseInfo` rethrows as `AppUpdaterError.InvalidResponse` (AppUpdaterService.kt:26-28), and `checkCriticalAppUpdate` takes `onFailure`, logs a warning and leaves the flag false (AppViewModel.kt:5608-5610). Cleartext too: `usesCleartextTraffic` is only in app/src/debug/AndroidManifest.xml:6, there is no network security config, and targetSdk is 36 (app/build.gradle.kts:188), so a plain `http://` fixture server is blocked in the devRelease build the journey asks for.

Fixed the fallback rather than marking it unavailable, since with those two constraints stated it does work. Setup step 1 now says the fixture must be the production feed's whole object with only `buildNumber` and `critical` edited, cites the DTO and the strict decode as the reason, and says to serve it over HTTPS (raw gist or a release asset on your own repo) with the cleartext reason given. The gate action asserts the feed is HTTPS and lists a complete Android object, and the "not run" note now points at `adb logcat -s APP:V`: "Failure fetching new releases" separates a bad fixture from an app regression, since both otherwise look like Terms of Use at the third action. I checked the live production feed still returns `buildNumber: 188`, `critical: true`, so the primary route is intact and the fallback stays the contingency you describe.

Journey-only change, XML parses, no Gradle run. Commit 87f3ea03d; the PR body diff is in step.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `87f3ea03d`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 87f3ea03d: the fixture-route LOW is fixed. The journey now requires the feed to be served over HTTPS (release builds have no `usesCleartextTraffic`), to be the complete object with only `buildNumber` and `critical` edited so strict `Json` decoding of `PlatformDetails` succeeds, and tells the runner to distinguish a bad fixture from a regression via the "Failure fetching new releases" log line.

app/src/main/java/to/bitkit/ui/MainActivity.kt
```diff
@@ -145,7 +151,23 @@ class MainActivity : FragmentActivity() {
                     }
                 }
 
-                if (isShowingMigrationLoading && !isRecoveryMode) {
+                if (isCriticalUpdateRequired) {
```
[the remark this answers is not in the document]
jvsena42: Confirmed. On the previous head the chain lived inline in `onCreate` at MainActivity.kt:154-179, and `AppViewModelSendFlowTest.kt` only asserted `isCriticalUpdateRequired` and the transaction sheet, so reordering or deleting the first branch broke nothing.

Extracted it as `RootDestination` (MainActivity.kt:364-381), a stateless composable taking `isCriticalUpdateRequired`, `isShowingMigrationLoading`, `walletExists` and `isRecoveryMode` plus one content slot per branch. The call site is MainActivity.kt:154-238; the `ON_STOP` `DisposableEffect` moved into the `criticalUpdate` slot (:159-175), so it still composes only on that branch. One incidental normalization: the onboarding branch used to read `walletViewModel.walletExists` directly while the critical branch read the `walletExists` local captured at :131 — both are the same property, and they are now one parameter.

Added `RootDestinationTest.kt` in `app/src/androidTest`, beside the other Compose UI tests. Nine cases, each asserting the expected slot renders and the other three do not: critical update wins over migration, over onboarding, over the wallet, and in recovery mode; migration beats onboarding; recovery mode skips migration and onboarding; and the plain onboarding and wallet cases. Moving the critical branch below any other now fails four of them.

Executed locally: `just build`, `just compile`, `just test` and `just lint` — detekt reported nothing on MainActivity.kt or the new test. `RootDestinationTest.kt` was compiled only, through `compileDevDebugAndroidTestKotlin`; I did not run instrumented tests. Not covered either way: the `ON_STOP` observer itself still has no test — asserting it needs a `TestLifecycleOwner`, and `lifecycle-runtime-testing` is not a dependency here, so that one stays a manual check. Commit 1d524c902.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `1d524c902`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 1d524c902: a behaviour-preserving extraction. `RootDestination` (MainActivity.kt:364-380) keeps the original branch order — critical update, then migration loading (not in recovery), then onboarding (no wallet, not in recovery), then the wallet — and `walletExists` is the same `walletViewModel.walletExists` read in the same composition (MainActivity.kt:131), so the onboarding branch sees the value it read inline before. The critical-update lambda keeps the ON_STOP node-stop observer unchanged, and the wallet lambda keeps ContentView, the auth overlay, the forgot-PIN sheet and the wipe effect. `RootDestinationTest` pins the order, including critical update winning over onboarding and migration.

ovitrif (approved):
utAck

PR #1296, opened 2026-09-17
fix: spread pasted seed fragment

### LINKED ISSUE #896: [Bug]: Pasting partial mnemonic during wallet restore fills first word field and blocks editing
## What happened?

During wallet restore, when entering the mnemonic, pasting part of the sentence (instead of a single word) causes the pasted text to land in the **first word** input. After that, the content in that field **cannot be edited**, leaving the user stuck.

## Expected behavior

Pasted text should either be split across word fields, rejected with clear feedback, or at minimum the affected field(s) should remain fully editable so the user can correct mistakes.

## Steps to Reproduce

1. Start **Restore wallet** and reach the mnemonic entry screen (per-word inputs).
2. Paste a **fragment of the mnemonic sentence** (multiple words or partial phrase), not a single BIP39 word, into the input area (e.g. focus/behavior that routes paste to the first field — exact focus step as in your recording).
3. Observe the text appears in the **first** word field.
4. Try to **edit or clear** that field — editing is not possible (stuck state).

https://github.com/user-attachments/assets/a436017c-00cb-4bb4-b903-1db735f4c6c3

## Bitkit Version

2.2.0

## Device / OS

Android 14

## Reproducibility

Always

## Additional context

- Flow: **Restored wallet** / mnemonic entry during onboarding restore.
- Symptom: paste targets first word slot; field becomes non-editable — likely input handling or focus state bug on the mnemonic composables.

## Related issues (not duplicates)

- synonymdev/bitkit-android#634 — Seed phrase confirmation behaviour
- synonymdev/bitkit-android#633 — Show seed phrase screen breaks line on long word

jvsena42:
Fixes #896
Related to #1275

This PR spreads a pasted partial recovery phrase across the Restore wallet word fields and keeps every field editable with Backspace.

### Description

- Spreads a pasted fragment of any length other than a full phrase across the fields from the one it was pasted into, so a partial paste no longer lands whole in one field.
- Switches the form to 24 words when the pasted words run past field 12, and drops words beyond field 24.
- Revalidates only the fields the paste wrote, then moves focus to the next empty field, or to the first field holding an invalid word when no field is empty, and hides the keyboard only when the phrase is complete and valid.
- Re-requests focus when a paste switches the form between 12 and 24 words, so the field the view model points at is the one that holds the caret even when the focused index does not change.
- Keeps the full replace for a 24-word paste into any field and a 12-word paste into field 1, and spreads 12 words pasted into a later field when other fields already hold words, so pasting the second half of a 24-word phrase keeps the first half.
- Fixes Backspace being swallowed after a paste, because the field checked the saved word instead of its own text.
- Drops a typed space instead of forwarding it as a paste, so editing a filled field to `aban don` no longer writes `don` over the next field; only an insertion longer than one character is spread.
- Serializes word edits, so a word typed while a paste is still validating is no longer undone when the paste finishes.
- Keeps the cursor at the end of a word copied into a field after a paste, so Backspace and typing edit the end of the word.

#### Out of Scope

- iOS: the restore screen there has the same 12/24-only paste guard and needs a matching follow-up.
- `RestoreWalletViewModel`: once a paste switches the form to 24 words, only a 12-word paste into field 1 switches it back.
- `journeys/restore-wallet`: the new journey is committed but its onboarding route has not been run on a device without a wallet.
- Restore wallet subtitle copy differs from Figma ("recovery phrase" vs "12 seed words"); this predates the change.

### Design

- [Restore wallet](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-145923)
- [Restore wallet with suggestions](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-145998)

There is no paste-specific design.

### Preview

Screens use public BIP39 test words. The C3 and C9 buttons at the top right come from a temporary dev harness that copies test words to the clipboard and are not part of the app.

| Figma | Before | After: fragment spread |
| - | - | - |
| figma-restore-wallet | before-fragment-stuck | after-fragment-spread |

| Figma: suggestions | After: Backspace edits previous word | After: phrase completed |
| - | - | - |
| figma-restore-wallet-suggestions | after-backspace | after-phrase-completed |

#### Manual Tests

- [x] **1a.** Onboarding → Restore wallet → paste `abandon abandon abandon` into Word 1: Words 1-3 show `abandon` and Word 4 is focused.
  - [x] **1b.** Backspace in empty Word 4: focus moves to Word 3.
  - [x] **1c.** Backspace again: Word 3 shows `abando`.
- [x] **2.** Restore wallet with Words 1-3 filled → paste the remaining 9 words into Word 4: all 12 fields filled, keyboard hides, Restore enabled.
- [x] **3.** Restore wallet → paste 9 words into Word 12: Words 12-20 filled, form switches to 24 words, Word 21 focused.
- [x] **4.** `regression:` Restore wallet → paste a full 12-word phrase into Word 1: phrase restores as before.
- [x] **5.** `regression:` Restore wallet → paste a full 24-word phrase into Word 1: form switches to 24 words and all fields fill.
- [x] **6.** 24-word layout with Words 1-12 filled → paste the other 12 words into Word 13: Words 1-12 kept, Words 13-24 filled.
- [x] **7.** Restore wallet with Words 1-2 filled → place the cursor inside Word 1 and type a space: Word 1 keeps its word and Word 2 is unchanged.
- [x] **8.** Restore wallet with Words 1-9 filled with valid words → paste `zzzz ability able` into Word 10: Words 10-12 fill, Word 10 shows the error state and keeps focus with the keyboard up, and typing in it still shows suggestions that can be tapped.
- [x] **9.** Restore wallet → paste a full 12-word phrase containing one invalid word into Word 1: the first invalid field takes focus and the keyboard stays up.
- [x] **10.** Restore wallet (12-word layout) → tap Word 7 → paste a full 24-word phrase whose word 7 is invalid: the form switches to 24 words and Word 7 keeps focus with the keyboard up, the caret is not left on Word 13, and tapping a suggestion writes into Word 7 only.

Steps 1-3 were run on an emulator through a temporary dev harness that opens the Restore wallet screen without wiping the wallet. The journey is committed at `journeys/restore-wallet/paste-seed-fragment.xml` and follows the same steps from onboarding. It now fills the clipboard from the host (`pbcopy`/`xclip`, picked up over emulator clipboard sharing) instead of leaving "copy from another app" undefined, since `cmd clipboard` is unimplemented on the emulator image and `input text` drops characters; each paste is followed by a check that reports a short paste as a clipboard setup failure. Its onboarding route has not been run on a device without a wallet.

<details><summary>journeys/restore-wallet/paste-seed-fragment.xml</summary>

```diff
diff --git a/journeys/restore-wallet/paste-seed-fragment.xml b/journeys/restore-wallet/paste-seed-fragment.xml
new file mode 100644
index 000000000..bb12ec8e8
--- /dev/null
+++ b/journeys/restore-wallet/paste-seed-fragment.xml
@@ -0,0 +1,23 @@
+<journey name="paste seed fragment">
+  <description>Proves #896: pasting part of a recovery phrase on Restore wallet spreads the words across the word fields, and every field stays editable with Backspace. Precondition: a throwaway emulator with no wallet (Restore is reachable only from onboarding; never run this on a device holding a wallet you need). Both pasted strings are BIP39 test vectors — "abandon abandon abandon" and "abandon abandon abandon abandon abandon abandon abandon abandon about". Use only these public test words, never a real wallet's phrase. Neither string can be put on the clipboard from the device: `adb shell cmd clipboard` is not implemented on the emulator image and `adb shell input text` drops characters from long strings (journeys/README.md:78-79), and a dropped character would fail the word assertions as if the app were at fault. So the clipboard is filled on the host and the emulator picks it up over clipboard sharing (Extended controls > Settings > General > "Enable clipboard sharing", on by default). Each copy action below gives the host command and is followed by a check that the words arrived whole; a short or empty paste there is a clipboard setup failure, not an app failure — re-copy and repeat the paste before reading it as a journey failure. Paste is sent as KEYCODE_PASTE (279) to the focused field.</description>
+  <actions>
+    <action>On the host, run `printf 'abandon abandon abandon' | pbcopy` (macOS) or `printf 'abandon abandon abandon' | xclip -selection clipboard` (Linux), then give the emulator window focus by clicking its title bar — not the device screen — so the shared clipboard reaches the guest</action>
+    <action>Launch to.bitkit.dev on a device with no wallet</action>
+    <action>Verify that the terms screen shows "Check1" and "Check2", then tap "Continue" — the two blocks are text, not toggles, and "Continue" is always enabled</action>
+    <action>Tap "SkipIntro"</action>
+    <action>Tap "RestoreWallet"</action>
+    <action>Tap "MultipleDevices-button"</action>
+    <action>Verify that the Restore wallet screen shows word fields "Word-0" to "Word-11", all empty</action>
+    <action>Tap "Word-0" and run `adb shell input keyevent 279`</action>
+    <action>Verify that "Word-0", "Word-1" and "Word-2" each show "abandon", "Word-3" is empty and focused, and "Word-0" does not show the whole pasted text — if the three fields are empty, or any of them shows a truncated word such as "abandn", the clipboard did not cross from the host whole: re-copy on the host, tap "Word-0" and paste again, and report a clipboard setup failure rather than an app failure if it repeats</action>
+    <action>Run `adb shell input keyevent 67`</action>
+    <action>Verify that "Word-2" is focused and still shows "abandon"</action>
+    <action>Run `adb shell input keyevent 67`</action>
+    <action>Verify that "Word-2" shows "abando", so Backspace deletes from the end of the pasted word</action>
+    <action>Run `adb shell input text n`</action>
+    <action>On the host, run `printf 'abandon abandon abandon abandon abandon abandon abandon abandon about' | pbcopy` (or the `xclip -selection clipboard` equivalent) and give the emulator window focus by clicking its title bar, then tap "Word-3" and run `adb shell input keyevent 279` — the app stays on the Restore wallet screen throughout, so nothing has to be typed on the device</action>
+    <action>Verify that "Word-0" to "Word-10" show "abandon", "Word-11" shows "about", no word field is focused and the keyboard is hidden — a truncated word or a short field count here means the nine-word copy did not cross whole, which is a clipboard setup failure: re-copy on the host and repeat the paste before reading it as an app failure</action>
+    <action>Verify that "RestoreButton" is enabled, and do not tap it</action>
+    <action>Press the device back button</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `RestoreWalletViewModelTest.kt` covers fragment spread from the first and middle fields, switching to 24 words, dropping words past field 24, invalid word flags, trailing-space words, whitespace-only input, focus and keyboard handling, when a 12-word paste replaces or spreads, that a word typed during a pending paste validation survives, and that a paste leaving an invalid word focuses that field so its suggestions and suggestion taps keep working. `MnemonicInputPasteTest.kt` covers telling a paste from a typed space.
- Not unit tested: re-requesting focus on a 12/24 switch lives in a `LaunchedEffect` in `RestoreWalletScreen`, and `compose.ui.test.junit4` is `androidTestImplementation` only, so nothing in `app/src/test` can observe it. Manual test 10 covers it.
- UI tests added: `MnemonicInputFieldTest.kt` covers whitespace text not kept in the field, a typed space not reaching the view model, showing only the saved word after a paste, Backspace deleting from the end, and Backspace in a field with text not jumping to the previous field. It compiles but has not been run, since connected tests were not run for this PR.
- Local: `just compile`, `just test` and `just lint` pass.


app/src/main/java/to/bitkit/viewmodels/RestoreWalletViewModel.kt
```diff
@@ -103,35 +103,75 @@ class RestoreWalletViewModel @Inject constructor(
...
-                    this[index] = ""
-                }
+        val state = _uiState.value
+        val isFullPhraseTarget = index == 0 || !state.is24Words && state.hasNoWordsExcept(index)
+        when (pastedWords.size) {
+            0 -> return@launch
+            WORDS_MAX -> replaceAllWords(pastedWords)
+            WORDS_MIN if isFullPhraseTarget -> replaceAllWords(pastedWords)
+            else -> spreadWords(index, pastedWords)
+        }
```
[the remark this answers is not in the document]
jvsena42: Confirmed and fixed in 4af0ab9. The field forwarded every whitespace edit to the view model, so a space typed inside `abandon` sent `aban don`, which spread `don` over the next field, and a trailing space moved focus. `MnemonicInputField` now tells a paste from typing by the length of the insertion (accounting for a replaced selection) and drops a typed space instead of forwarding it; a paste still reaches the view model unchanged. Covered by `MnemonicInputPasteTest.kt` and a new case in `MnemonicInputFieldTest.kt`, and the out-of-scope note is off the PR body.

app/src/main/java/to/bitkit/viewmodels/RestoreWalletViewModel.kt
```diff
@@ -103,35 +103,75 @@ class RestoreWalletViewModel @Inject constructor(
...
+    }
+
+    private suspend fun spreadWords(startIndex: Int, pastedWords: List<String>) {
+        val writtenWords = pastedWords.take(WORDS_MAX - startIndex)
+        val writtenValidity = writtenWords.map { bip39Service.isValidWord(it) }
+        val lastWrittenIndex = startIndex + writtenWords.lastIndex
+
+        _uiState.update { state ->
+            val newWords = state.words.toMutableList()
```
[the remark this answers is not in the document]
jvsena42: Confirmed and fixed in 4af0ab9. `Bip39Service.isValidWord` hops to the CORE queue, so each edit really does suspend before it writes state, and a second edit could start and finish in between; `updateWordValidity` also captured the word list before that suspension, so it could resurrect pre-paste words. Word edits and pastes now run under a `Mutex`, and both `updateWordValidity` and `replaceAllWords` build the new list inside `_uiState.update` from the current state. `RestoreWalletViewModelTest.kt` pins it with a gated validation: a word typed while a paste is validating survives, and the test fails on the previous code.

app/src/main/java/to/bitkit/viewmodels/RestoreWalletViewModel.kt
```diff
@@ -103,57 +110,104 @@ class RestoreWalletViewModel @Inject constructor(
...
+            val wordCount = if (is24Words) WORDS_MAX else WORDS_MIN
+            val nextEmptyIndex = (lastWrittenIndex + 1 until wordCount).firstOrNull { newWords[it].isEmpty() }
+                ?: (0 until wordCount).firstOrNull { newWords[it].isEmpty() }
+
+            state.copy(
                 words = newWords.toImmutableList(),
                 invalidWordIndices = newInvalidIndices.toImmutableSet(),
+                is24Words = is24Words,
+                shouldDismissKeyboard = nextEmptyIndex == null && newInvalidIndices.isEmpty(),
+                focusedIndex = nextEmptyIndex,
```
jvsena42: **LOW** — **A fragment that fills the last empty field but contains an invalid word sets `focusedIndex = null` while the pasted field keeps OS focus and the keyboard.**

`spreadWords` (RestoreWalletViewModel.kt:173-181): fields 1-9 filled, paste `zzzz ability able` into field 10 -> `nextEmptyIndex == null`, `newInvalidIndices = {9}`, so `shouldDismissKeyboard = false` and `focusedIndex = null`. `LaunchedEffect(uiState.focusedIndex)` (RestoreWalletScreen.kt:155) does nothing for null, so field 10 stays OS-focused with the keyboard up. Consequences while the user keeps editing field 10 without re-tapping it: `isError` (RestoreWalletScreen.kt:201) shows red on the focused field, `onChangeWord` -> `updateSuggestions(value, null)` (RestoreWalletViewModel.kt:61, 214) never shows suggestions, and `onSelectSuggestion` (line 86) is a no-op. `onFocusChanged` does not fire on a tap of the already-focused field, so only tapping another field resyncs. Mirrors what `replaceAllWords` already did on master for a full paste with invalid words, so the UX is consistent, but the new spread path reaches it from any partial paste.

Fix (narrowest): in `spreadWords`, when `nextEmptyIndex == null && newInvalidIndices.isNotEmpty()`, keep `focusedIndex = startIndex` (or the first invalid written index) instead of null.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `097ab79a9`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — spread that completes the phrase with an invalid word leaves the pasted field focused but focusedIndex null

Checked: Traced MnemonicInputField.onValueChange at the PR head: non-whitespace input updates the local TextFieldValue and the view model together; whitespace input is forwarded only when isPastedInput sees an insertion longer than one character after subtracting a replaced selection, so a typed space is dropped and the controlled field reverts. In the view model, handlePastedWords, replaceAllWords, spreadWords and updateWordValidity all run under wordEditMutex and build the new word list inside _uiState.update, so an older edit cannot resurrect pre-paste words (the two greptile P1 threads are fixed in 4af0ab9 and the four detekt line-length threads are cleared by 097ab79; no line in the changed files exceeds 120 chars). spreadWords writes pastedWords.take(24 - startIndex) at startIndex..lastWrittenIndex, switches to 24 words only when lastWrittenIndex >= 12, and the 12-word layout never holds words at 12..23 because the only 24->12 transition (replaceAllWords with 12 words) clears them, so bip39Mnemonic (subList(0, wordCount)) and hasNoWordsExcept stay consistent. The 12->24 switch re-creates focusRequesters keyed on wordCount in the same state emission as focusedIndex, nextEmptyIndex is always < the new wordCount, and column-two nodes that change index positionally are re-synced by LaunchedEffect(value) because their value key changes. The 12-word full-replace rules (field 1 always; a later field only on an otherwise-empty 12-word form) and the no-way-back from 24 words are explicitly documented in the PR body and not re-raised. Passphrase handling is untouched; RestoreWalletViewModel has no Logger calls and the restore path (WalletViewModel.restoreWallet -> WalletRepo.restoreWallet) logs only a generic error without the mnemonic. No iOS twin PR exists; the PR body lists iOS as a follow-up.

app/src/main/java/to/bitkit/viewmodels/RestoreWalletViewModel.kt
jvsena42: Verified the transition: fields 1-9 filled, pasting "zzzz ability able" into field 10 gives lastWrittenIndex=11, nextEmptyIndex=null and newInvalidIndices={9}, so the old code set focusedIndex=null while field 10 kept OS focus, which left updateSuggestions and onSelectSuggestion dead on the field the user was still editing.

spreadWords now falls back to the first invalid index when no empty field remains, so focus and the keyboard stay on the field that needs fixing; shouldDismissKeyboard is derived from that same index and is still gated on invalidWordIndices being empty, so no new dismissal is introduced. replaceAllWords was a genuine one-liner (focusedIndex = invalidIndices.minOrNull()), so I took it: a full paste with a bad word now focuses the first offender instead of leaving focus null with the keyboard up.

Three tests in RestoreWalletViewModelTest.kt cover it, all red before the change and green after: focus lands on the invalid field when a fragment fills the tail, suggestions resolve and onSelectSuggestion applies on that focused field, and a full 12-word paste focuses its first invalid index. just compile, just test and just lint pass. Commit 9c9d483de.

jvsena42: **LOW (regression from 9c9d483de)** — **The fix covers the reported case, but when the first invalid index equals the field the user pasted into and the paste flips the form from 12 to 24 words out of a column-two field, `focusedIndex` does not change, so the focus effect never re-runs and the view model points at a field that no longer has OS focus: the invalid field is not shown red, and a tapped suggestion writes into it while the user is typing in another field.**

Trace at head: 12-word form, user taps field 7 (index 6; column two renders `wordsPerColumn until wordsPerColumn*2` = 6..11, RestoreWalletScreen.kt:216), so onChangeWordFocus(6, true) sets focusedIndex=6 and `LaunchedEffect(uiState.focusedIndex)` (line 155) has already run for 6. They paste a 24-word phrase whose first invalid word is word 7. handlePastedWords -> replaceAllWords: is24Words=true, focusedIndex=invalidIndices.minOrNull()=6 (RestoreWalletViewModel.kt:151), shouldDismissKeyboard=false. On recomposition the two column loops have no `key()`, so groups are matched positionally: column two now renders 12..23 and the node that holds OS focus (column-two slot 0) becomes index 12 with the new focusRequester from `remember(uiState.wordCount)` (line 131). `FocusChangedElement.update` only swaps the lambda and `focusRequester()` only re-links the requester, so no focus event fires and onChangeWordFocus is never called; the effect on line 155 is keyed on focusedIndex alone, which is still 6, so it does not re-run. Result: OS focus and keyboard on field 13 (index 12), focusedIndex=6. Field 7 is not red (`isError = index in invalid && focusedIndex != index`, lines 201/220) and the red explanation is hidden (line 263, `any { it != focusedIndex }`), while Restore stays disabled with nothing marked. Typing in field 13 calls onChangeWord(12, ...) (correct index) but updateSuggestions(value, 6) (line 61) shows suggestions; tapping one runs onSelectSuggestion -> updateWordValidity(6, suggestion) (lines 86-87), writing into field 7 while field 13 keeps the typed text. Same shape via spreadWords (line 175): fields 1-6 filled, user in index 6 pastes 18 words whose first word is invalid -> lastWrittenIndex=23, nextEmptyIndex=null, first invalid=6, focusedIndex unchanged, form switches to 24. Before 9c9d483de both paths set focusedIndex=null, so field 7 was red and suggestions were dead; the cross-field write and the missing red mark are introduced by this commit. Restore stays gated by invalidWordIndices and then the checksum, and every word stays visible in its field, so this cannot restore a different wallet silently. Recovery is tapping any field. Preconditions are narrow: 12-word form, paste from a column-two field (7-12), the paste ends in 24-word mode with no empty field, and the first invalid word lands at exactly that index.

Fix (narrowest): key the focus effect on the layout too, `LaunchedEffect(uiState.focusedIndex, uiState.wordCount)` at RestoreWalletScreen.kt:155, so a 12/24 switch re-requests focus on the node that now carries focusedIndex; the displaced node then gets onFocusChanged(false) with index 12 (a no-op since focusedIndex != 12) and the new node re-syncs via onChangeWordFocus(6, true). When neither value changed, requestFocus on the already-focused node is a no-op. Do not `key(index)` the loop items instead: a node cannot move between the two Columns, so the focused node would be disposed and onChangeWordFocus(6, false) would null focusedIndex.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `9c9d483de`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 9c9d483de: the reported focus case is fixed. One LOW the new commit introduced (12->24 switching paste from a column-two field) is posted as a reply on the earlier thread; restore stays gated by checksum, so it is not blocking. Follow-up on 9c9d483de (delta over 097ab79a9). The original LOW is fixed on the path it described: spreadWords (RestoreWalletViewModel.kt:175) falls back to the first invalid index within wordCount only when no empty field remains, so `zzzz ability able` into field 10 with fields 1-9 filled now yields focusedIndex=9, scrollToFieldIndex=9 and shouldDismissKeyboard=false; the field stays OS-focused, isError stays hidden on it, updateSuggestions(value, 9) resolves and onSelectSuggestion writes to index 9. shouldDismissKeyboard (line 181) is still true exactly when no empty field and no invalid index remain, so the valid-paste path is unchanged: focusedIndex=null, focus cleared, keyboard hidden, Restore reachable. replaceAllWords (line 151) now focuses invalidIndices.minOrNull(), which is < pastedWords.size <= the new wordCount, so focusRequesters[index] is in range after the 12/24 switch; a paste with all valid words still gives null/dismiss. Typed-space splitting (isPastedInput, RestoreWalletScreen.kt:438) and the wordEditMutex ordering are untouched by the delta. The three new tests fail on the previous code (they assert 9/4 where the old code produced null, and non-empty suggestions where focusedIndex=null cleared them). No Logger calls added; no seed material logged. One narrow over-correction remains and is replied on the original thread.

app/src/main/java/to/bitkit/viewmodels/RestoreWalletViewModel.kt
jvsena42: You are right, and this is ours: 9c9d483de changed `replaceAllWords` and `spreadWords` from `focusedIndex = null` to the first invalid index, which opened the case where the new value equals the old one.

I retraced it at head before changing anything. On a 12-word form with field 7 tapped, `focusedIndex` is already 6; pasting a 24-word phrase whose word 7 is invalid sets `focusedIndex = 6` again, so `LaunchedEffect(uiState.focusedIndex)` never restarts. The two column loops have no `key()`, so the node holding OS focus is reused positionally and becomes index 12 under the new `remember(uiState.wordCount)` requesters, with no focus event to call `onChangeWordFocus`. That leaves OS focus on field 13 while the view model still points at index 6, so field 7 is not marked red and `onSelectSuggestion` writes into it.

Fixed with your narrower suggestion in 66f701936: the effect is now keyed `LaunchedEffect(uiState.focusedIndex, uiState.wordCount)`, so a 12/24 switch re-requests focus on the node that actually carries `focusedIndex`. I did not key the loop items, for the disposal reason you gave. On an ordinary paste where `focusedIndex` also changes both keys change together and the effect still restarts exactly once, and `requestFocus()` on an already-focused node returns early in `FocusTargetNode` without emitting an event, so the no-change case is a no-op. I also checked the reverse transition: `is24Words` only drops to false in `replaceAllWords` with 12 pasted words, which sets `focusedIndex` from indices 0..11 or null in the same `copy()`, so the new trigger cannot index past the requester list.

No unit test covers it, and I would rather say that than add one that passes either way: the regression only exists in composition, and `compose.ui.test.junit4` is `androidTestImplementation` here, so nothing in `app/src/test` can observe the focus effect. The view model state is unchanged by this commit, so every existing `RestoreWalletViewModelTest` assertion still holds. I verified the fix by reading the trace through rather than on a device, since the emulator was not free; I added the 12-to-24 column-two paste to the manual QA list so it gets exercised. `just compile`, `just test` and `just lint` pass locally.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `66f701936`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 66f701936: keying the focus effect on `wordCount` fixes the 12->24 switching-paste LOW from the previous thread. Both state writers set `is24Words` and `focusedIndex` in one `_uiState.update` (`replaceAllWords` at RestoreWalletViewModel.kt:146-152, `spreadWords` at 177-185), and `focusedIndex` is always below the new `wordCount`, so the re-run cannot index past the `focusRequesters` list rebuilt at RestoreWalletScreen.kt:131. The effect runs after the recomposition that attaches the new requesters, the same ordering the `focusedIndex` key already relied on. The rest of the PR is unchanged since the approval at 9c9d483de.

journeys/restore-wallet/paste-seed-fragment.xml
```diff
@@ -0,0 +1,22 @@
+<journey name="paste seed fragment">
+  <description>Proves #896: pasting part of a recovery phrase on Restore wallet spreads the words across the word fields, and every field stays editable with Backspace. Precondition: a throwaway emulator with no wallet (Restore is reachable only from onboarding; never run this on a device holding a wallet you need). The clipboard must hold the BIP39 test words "abandon abandon abandon" before step 6; step 13 copies "abandon abandon abandon abandon abandon abandon abandon abandon about". `adb shell cmd clipboard` is not implemented on the emulator image, so copy the text from another app (for example a note or a Chrome page) first. Use only these public test words. Paste is sent as KEYCODE_PASTE (279) to the focused field.</description>
```
jvsena42: **LOW** — **The journey needs two exact strings on the clipboard but only says to "copy the text from another app", and never says how that text gets into the other app.**

On an emulator the only ways an agent has to get text into a note or a Chrome page are `adb shell input text` and a URI. journeys/README.md:78-79 warns that `input text` drops characters and that `cmd clipboard` is not implemented. The nine-word string used in step 14 is 68 characters and needs `%s` for its spaces. If one character drops (for example `abandn`), step 15 fails on invalid-word or checksum state, which is a false fail. A runner that follows the README rule "if an interaction cannot be performed as written, say so and stop" stops at step 14, or before step 7 for the precondition. Neither case gives a false pass, because a dropped character cannot produce `abandon x11 about`.

Fix (narrowest): give the route in the description, e.g. on the host run `printf 'abandon abandon abandon' | pbcopy`, which the emulator picks up through host clipboard sharing (on by default in Extended controls > Settings), and do the same for the nine-word string before step 14. Also add a verify step that the clipboard text was pasted whole, so a truncated copy is reported as a setup error rather than an app bug.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `5bea120c7`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Clipboard setup gives no route, so runners fall back to input text or stop

Checked: Follow-up on 89920b350 (new journey): Round 4 delta is 89920b350 only: the new journey journeys/restore-wallet/paste-seed-fragment.xml plus two README rows. Every tag it names exists at head and the route matches OnboardingNav: Terms (Check1/Check2 are TermsText blocks, Continue is an always-enabled PrimaryButton) -> Intro SkipIntro -> last slide CreateWalletScreen RestoreWallet -> WarningMultipleDevices MultipleDevices-button -> Restore with Word-0..Word-11 and RestoreButton. Traced each step through RestoreWalletViewModel at head: 3 words into index 0 spread to 0..2 and focus 3; DEL in the empty Word-3 is consumed by onPreviewKeyEvent and moves focus to 2; the second DEL edits from the cursor end set by LaunchedEffect(value), giving abando; 9 words into index 3 end at 11 with no empty or invalid field, so focus clears, the keyboard hides and abandon x11 about passes the checksum, enabling Restore. With the fix reverted, step 8 fails (master's handlePastedWords ignores a 3-word paste, so Word-0 keeps the whole text and Word-1 stays empty), so the journey cannot pass on a regression. The phrase is the public BIP39 abandon x11 about vector, labelled public test words, and the journey never taps Restore, so nothing gets funded. No iOS twin exists, and the cross-platform row records it as not ported because the behaviour does not exist on iOS yet, which matches the SKIP rule.

journeys/restore-wallet/paste-seed-fragment.xml
jvsena42: Confirmed. The description said only "copy the text from another app" and never said how the text got there, and `journeys/README.md:78-79` rules out both device-side routes — `adb shell cmd clipboard` is unimplemented on the emulator image and `adb shell input text` drops characters (it lost 54 of a 397-character invoice). The nine-word string is 69 characters, not 68, so the escaping point stands either way. A dropped character would land on the assertion at the old line 18, which reads like an app failure.

Changed in 4403315af: the copy is now two actions with the host command spelled out — `printf 'abandon abandon abandon' | pbcopy` (or `xclip -selection clipboard` on Linux) before launch, and the nine-word `printf` before the second paste — each followed by giving the emulator window focus by its title bar, not the device screen, so the shared clipboard reaches the guest without a stray tap. The two verify actions now name the truncation failure mode explicitly and tell the runner to re-copy and repeat the paste, reporting a clipboard setup failure rather than an app failure. Both strings stay the public BIP39 vectors. With the copy inside the action list the description no longer carries step numbers, which also drops the off-by-one it had ("before step 6" / "step 13" for what were actions 7 and 14). The app side is unchanged and still matches: `testTag("Word-$index")` at RestoreWalletScreen.kt:425, and whitespace-containing input is treated as a paste (RestoreWalletScreen.kt:80, :436). The PR body diff is in step with the committed file.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `4403315af`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up: the earlier journey finding is fixed. Follow-up on 4403315af: the earlier clipboard-route LOW is fixed. Checked the host route on the running emulator (Pixel_9, API 36 image, emulator 36.6.11, embedded in Android Studio): after `printf '<nine-word vector>' | pbcopy` the guest clipboard did not change while another app was frontmost, and brought to the front the emulator's host window, `dumpsys clipboard` showed a new 69-character text/plain primary clip within a second, matching the string length. So the step's instruction to focus the emulator window after copying is necessary and enough; the host clipboard was restored afterwards and the app was not touched. Neither long string goes through `input text`; the only typed input is the single character `n`. The added verify steps name the truncation and empty-paste symptoms and send them to a clipboard setup failure, so a failed copy cannot read as an app failure, and the copy-before-launch ordering matches the flow. The app-side tags and paste handling are unchanged by this commit.

ovitrif:
@jvsena42 e2e red and conflicts

ovitrif (dismissed):
utAck

jvsena42:
Same as the others: covered centrally in https://github.com/synonymdev/bitkit-android/issues/1275#issuecomment-5732780955. The conflict is the shared `journeys/README.md` row, a merge of current master is running on this branch now, and the red e2e shards failed in setup rather than in an assertion.

I will confirm here once the merge lands and CI re-runs on the new head.

jvsena42:
Test 6 is a real bug and it is fixed in d3ade0843.

**What happened.** The paste did not land in an empty Word 13 — it landed in Word 12, which already
held `about`. `MnemonicInputField` forwarded `newValue.text`, i.e. the *whole field* after the paste,
so the view model received `"aboutabout abandon×11 art"`: 13 tokens, spread from index 11. That is
exactly your screenshot — Word 12 `aboutabout`, Words 13–23 `abandon`, Word 24 `art`. The layout flip
to 24 words follows from the same spread (`lastWrittenIndex >= WORDS_MIN`).

**Fix.** The field now forwards only the text the paste actually inserted, not the word it was dropped
next to:

```kotlin
internal fun insertedText(previous: TextFieldValue, new: TextFieldValue): String {
    val prefixLength = previous.selection.min
    val suffixLength = previous.text.length - previous.selection.max
    ...
}
```

So pasting 12 words onto a filled Word 12 now writes `about` into Word 12 and the rest into 13–23 —
the target word is replaced rather than merged with the first pasted word. Pasting into an empty
field is unchanged, and a single-word paste never reaches this path (no whitespace).

**Proof.** Four cases added to `MnemonicInputPasteTest.kt` (after / before / over a filled word, and
the empty-field regression). Negative control — with `insertedText` stubbed back to `new.text`, the
new test fails with your exact string:

```
expected:<about[] abandon art> but was:<about[about] abandon art>
```

`just test` and `just lint` are green; the only detekt finding is the pre-existing `LargeClass` on
`RestoreWalletViewModelTest`.

piotr-iohk:
QA reviewed on 829f17cd65f162bdcb327b16e3e592ecc18e920e.

Ran on Pixel_6: TOS → Skip intro → Restore. Pasted `abandon abandon abandon` into field 1; fields 1–3 showed abandon and field 4 focused. Backspace moved to field 3, then deleted the last letter; typing `n` restored abandon. Pasted the nine-word test fragment into field 4; fields 1–11 showed abandon, field 12 showed about, Restore was enabled, no field focused. Did not tap Restore. Back returned to the multiple-devices caution.

Unrelated: local e2e `@boost_2` hit a race that drops the RBF boost link on the activity list. Not caused by this PR; filed as [#1317](https://github.com/synonymdev/bitkit-android/issues/1317).

No findings.

Checked and clean
- Fragment paste spreads from the focused field
- Backspace still edits a word that arrived via paste
- Completing 12 words enables Restore and dismisses the keyboard
- Restore IDs `Word-0`…`Word-11` / `RestoreButton` are present

QA LGTM

piotr-iohk (approved):

PR #1298, opened 2026-09-17
fix: require pin when app resumes

### LINKED ISSUE #724: PIN should be required on app restart + resume
### What happened?

Andorid had this:
 - https://github.com/synonymdev/bitkit-android/issues/542
 - https://github.com/synonymdev/bitkit-android/pull/592

In android PIN is only required on app restart. On iOS it is both by default - restart and re-activation.
The behavior should be consistent, so ~the default should be PIN on restart.~ as in iOS.

### Steps to Reproduce

- set PIN
- move app to background
- move back to active

~You are required to put pin, but shouldn't. Pin only on app restart.~
as in iOS

iOS on left:

https://github.com/user-attachments/assets/098ed2e0-db29-4683-844a-1a0ce60b9d87

### Environment

_No response_

jvsena42:
Fixes #724
Related to #1275

This PR locks the app behind the PIN screen whenever it returns from the background, instead of only at launch.

### Description

- Resets the authenticated state when the main activity stops (not on configuration changes), so returning to Bitkit shows the PIN pad when the PIN is enabled, a wallet exists and the node is not in recovery mode. The reset runs without suspending, reading a cached `isPinEnabled` state, so the lock is applied before `onStop` returns and cannot land after the activity has resumed.
- Uses the activity `onStop` instead of `ProcessLifecycleOwner`, because the manifest removes the androidx startup provider, so the process lifecycle never emits `ON_STOP` in this app. Bitkit has a single UI activity, so the result is the same.
- Shows the PIN overlay without a fade-in, so the first frame drawn on resume is the PIN pad and not wallet content.
- Hides alert dialogs while locked, because `AppAlertDialog` is a real `Dialog` and draws in its own window above the PIN overlay. Bottom sheets are left composed: `BottomSheetOverlayHost` lives inside `ContentView`, which is composed before `AuthCheckView` in the same `Surface`, so the opaque PIN pad already covers them and skipping them would only throw away `remember` state such as a half-typed Trezor passphrase.
- Clears focus and hides the keyboard when the lock engages, so an auto-focused field in a sheet left open (for example the hardware passphrase prompt) cannot bring the IME back over the PIN pad on resume.
- Holds the new transaction sheet until unlock, so incoming payment details are not shown over the PIN screen.
- Ignores amount warning confirmations while locked, so a tap on a dialog that has not redrawn yet cannot start a payment.
- Keeps payment URIs opened while locked queued until the PIN is entered, then opens the Send flow. If a sheet was left open when the app was backgrounded, unlocking now closes it so the queued scan can run, instead of leaving the payment queued behind that sheet until the user happens to dismiss it. Send, Gift, Pin and the other high-priority sheets are still left alone, as is the QR scanner sheet, and the scan runs when they close.
- Never auto-pays a scan that crossed the lock. A deferred scan that is still queued when the PIN is accepted skips QuickPay and lands on Send Confirm, because otherwise the unlock gesture authorises a payment the user never confirmed (`isPinForPaymentsEnabled` does not gate `SendQuickPayScreen`). Scanning while already unlocked still uses QuickPay as before.

#### Out of Scope

- `MainActivity`: pressing back on the PIN pad still reaches the hidden UI and can dismiss an open in-window sheet (Send, Hardware) or cancel a hardware connect behind the lock. A `BackHandler` in the overlay is the likely fix; left for a follow-up.
- `MainActivity`: toasts are drawn above the PIN overlay and can show while locked.
- Dev settings, logs and Trezor dev screen raw `AlertDialog`s, the Create Subscription dropdown and tooltip popups are not hidden while locked. None of them move funds.
- No grace period and no exemption for the system share chooser or file pickers. The share chooser only pauses the activity, so it does not lock.
- The recents thumbnail still shows wallet content, as before.
- The Enter PIN screen differs from Figma (no `000` key, logo and title position); this predates the change.
- iOS: already clears PIN verification on entering the background. The journey below should be ported to `bitkit-ios` with this change.

### Design

- [Active user › Enter PIN](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-148340)

There is no design for when the lock triggers; the PIN screen itself is unchanged.

### Preview

Before: HOME then relaunch goes straight to Home with no PIN.

https://github.com/user-attachments/assets/6838b501-9e6e-4b49-9f46-d4c674b21bb3

After: HOME then relaunch shows the PIN pad; unlocking returns to Home.

https://github.com/user-attachments/assets/8db8cf0f-8a80-4a50-ade2-6dd7d9b8f259

| Figma | After: PIN on resume |
| - | - |
| figma-enter-pin | after-pin-on-resume |

| Add Tag sheet while locked | Add Tag sheet after unlock |
| - | - |
| sheet-hidden-while-locked | sheet-back-after-unlock |

| Sign With Device while locked | Sign With Device after unlock |
| - | - |
| hw-sign-locked | hw-sign-after-unlock |

| Payment URI while locked | Payment URI after unlock |
| - | - |
| deeplink-locked | deeplink-after-unlock |

Payment URI opened while backgrounded: PIN first, then Send Amount after unlock.

https://github.com/user-attachments/assets/20b0b06d-9722-46fb-943d-c12c9f30f1d2

#### Manual Tests

- [x] **1a.** PIN enabled → Home → HOME key → reopen Bitkit: PIN pad shows before any wallet content.
  - [x] **1b.** Enter PIN: returns to Home.
- [x] **2.** PIN enabled → toggle dark mode or rotate while unlocked: no PIN prompt.
- [x] **3.** PIN enabled → Receive → Share → back from the system chooser: Receive sheet still open, no PIN prompt.
- [x] **4a.** PIN enabled → Activity Detail → Tag → HOME → reopen: only the PIN pad is drawn, no Add Tag sheet above it.
  - [x] **4b.** Enter PIN: Add Tag sheet is back.
- [x] **5.** PIN enabled → Send with Trezor source → Sign With Device → HOME → reopen → enter PIN: Sign With Device screen unchanged.
- [x] **6.** PIN enabled → HOME → open a `bitcoin:` or `lightning:` URI: PIN pad shows with no Send sheet; after PIN, Send opens with the URI's amount.
- [x] **7.** PIN enabled → Send amount triggering a sanity warning → HOME → reopen: warning dialog hidden until PIN is entered, then shown again.
- [ ] **8.** PIN and biometrics enabled → HOME → reopen: biometric prompt shows; success unlocks.
- [x] **9.** `regression:` PIN disabled → HOME → reopen: Home shows with no lock.
- [x] **10.** `regression:` PIN enabled → kill and relaunch app: PIN pad shows as before.

Steps 1-6 and 9 were run on an emulator (debug build). Step 5 stopped at Sign With Device without a connected Trezor, and step 6 used a `bitcoin:` URI. Step 8 was not run because no fingerprint was enrolled. The journey below covers steps 1, 2, 4 and 6, and is committed at `journeys/security/pin-lock-on-resume.xml`.

<details><summary>journeys/security/pin-lock-on-resume.xml</summary>

```diff
diff --git a/journeys/security/pin-lock-on-resume.xml b/journeys/security/pin-lock-on-resume.xml
new file mode 100644
index 000000000..924308ea5
--- /dev/null
+++ b/journeys/security/pin-lock-on-resume.xml
@@ -0,0 +1,39 @@
+<journey name="pin lock on resume">
+  <description>
+    Verifies that with the app PIN enabled, sending Bitkit to the background locks it: returning
+    to the app shows the PIN pad before any wallet content, a sheet left open is hidden while
+    locked and comes back after unlock, a configuration change does not lock, and a payment URI
+    opened while backgrounded waits for the PIN before the Send flow opens. Matches iOS, which
+    clears the PIN verification when the scene enters the background, with no grace period.
+
+    Precondition: onboarded dev wallet with the app PIN enabled (Settings > Security) and a known
+    PIN, biometrics off, unlocked, on the wallet home screen. The wallet needs a savings balance
+    above 10 000 sats and at least one activity item — with an empty balance the last payment URI
+    is rejected with an insufficient-savings toast instead of opening Send, and with no activity
+    there is no row to open for the Add Tag steps. Never enter a wrong PIN more than once;
+    exhausting attempts leads to the forgot-PIN wallet reset.
+  </description>
+  <actions>
+    <action>Verify the wallet home screen is visible (testTag "Send")</action>
+    <action>Run adb shell input keyevent KEYCODE_HOME and wait 3 seconds</action>
+    <action>Run adb shell am start -n to.bitkit.dev/to.bitkit.ui.MainActivity</action>
+    <action>Verify the PIN pad is visible (testTag "PinPad") with the text "Please enter your PIN code" and keys (testTag "N1")</action>
+    <action>Enter the PIN with the number pad keys (testTags "N0" to "N9")</action>
+    <action>Verify the wallet home screen is visible again (testTag "Send")</action>
+    <action>Run adb shell cmd uimode night yes, wait 5 seconds, then run adb shell cmd uimode night no and wait 5 seconds</action>
+    <action>Verify the wallet home screen is visible and the PIN pad is not shown</action>
+    <action>Tap the first activity item on the home screen (testTag "ActivityShort-0"), then tap Tag (testTag "ActivityTag")</action>
+    <action>Verify the Add Tag sheet is visible (testTag "TagInput")</action>
+    <action>Run adb shell input keyevent KEYCODE_HOME and wait 3 seconds</action>
+    <action>Run adb shell am start -n to.bitkit.dev/to.bitkit.ui.MainActivity</action>
+    <action>Take a screenshot and verify only the PIN pad is drawn, with no Add Tag sheet above it</action>
+    <action>Enter the PIN with the number pad keys</action>
+    <action>Verify the Add Tag sheet is visible again (testTag "TagInput"), then press back until the wallet home screen is visible</action>
+    <action>Run adb shell input keyevent KEYCODE_HOME and wait 3 seconds</action>
+    <action>Run adb shell am start -a android.intent.action.VIEW -d "bitcoin:bcrt1qh6k3khs7qz04a9p5eam6v48nlug9d38x4rgxrn?amount=0.0001" to.bitkit.dev</action>
+    <action>Verify the PIN pad is visible and the Send amount screen is not shown</action>
+    <action>Enter the PIN with the number pad keys</action>
+    <action>Verify the Send amount screen opens (testTag "SendNumberField") showing 10 000 sats</action>
+    <action>Press back to close the Send sheet and verify the wallet home screen is visible</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added in `AppViewModelSendFlowTest.kt`: `lockOnBackground` locks when the PIN is enabled, locks without waiting for a suspension, and does nothing when the PIN is disabled, no wallet exists or recovery mode is on; a `lightning:` URI received after a lock stays queued and opens Send Confirm after unlock; an amount warning confirmation is ignored while locked and accepted after unlock; a URI received with a Receive sheet open closes that sheet on unlock and opens Send Confirm; a URI received with a Pin sheet open leaves it untouched and runs the scan when it is dismissed; a URI deferred by the lock skips QuickPay and opens Send Confirm.
- Unit tests modified in `AppViewModelSendFlowTest.kt`: two warning tests now authenticate first, since confirmations before authentication are rejected; `QuickPay eligible scan remains deferred until authenticated` becomes `QuickPay eligible scan remains deferred and confirms after authenticating`, since a scan that crossed the lock no longer auto-pays.
- Hiding dialogs while locked has no automated test; it was checked on device.
- Ran locally: `just compile`, `just test`, `just lint`.


app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -4837,6 +4837,17 @@ class AppViewModel @Inject constructor(
         }
     }
 
+    /** Requires the PIN again after the app process moves to the background. */
+    fun lockOnBackground() {
+        viewModelScope.launch {
+            if (!settingsStore.data.first().isPinEnabled) return@launch
+            if (!walletRepo.walletExists()) return@launch
+            if (lightningRepo.isRecoveryMode.value) return@launch
+            _isAuthenticated.update { false }
```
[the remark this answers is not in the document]
jvsena42: Verified and fixed in 166d1c89a. `lockOnBackground()` did launch a coroutine whose first statement was a suspending `settingsStore.data.first()`, so the clear of `_isAuthenticated` was ordered after a dispatch hop from `MainActivity.onStop`. `walletExists()` and `isRecoveryMode.value` were already synchronous; only the settings read was not, so it now reads a cached `isPinEnabled` StateFlow (`stateIn(viewModelScope, Eagerly, false)`, same pattern as `isPaykitEnabled`) and the whole function runs without suspending, clearing auth before `onStop` returns. The cached default of false cannot skip a lock in practice: `_isAuthenticated` only becomes true after settings have been read at least once, and it starts false, so an unloaded cache means the app is already locked. Added `lockOnBackground requires auth without waiting for a suspension` in AppViewModelSendFlowTest.kt, which asserts the flag is false with no `advanceUntilIdle()` and fails against the previous implementation. Ran `just compile`, `just test`, `just lint`.

jvsena42 (commented):
**Changes requested** — 1 HIGH/MEDIUM finding(s) inline. (Posted as a comment because GitHub does not allow requesting changes on your own PR.)

Inline: MEDIUM — Payment link opened while locked with a sheet open stays queued after the PIN and fires whenever that sheet is later closed; LOW — BottomSheet early return while locked drops state remembered inside the content lambda; the sheet was already under the PIN overlay

Checked: Traced MainActivity.onStop -> AppViewModel.lockOnBackground: it is synchronous (cached isPinEnabled StateFlow, walletExists(), isRecoveryMode.value), skipped on isChangingConfigurations, and a no-op for PIN-disabled users; there is no separate pin-on-launch setting on Android (SettingsStore has only isPinEnabled, isBiometricEnabled, isPinForPaymentsEnabled). The in-app biometric prompt uses BIOMETRIC_STRONG only (BiometricPrompt.kt:114), so no device-credential activity is launched and the prompt cannot stop MainActivity; the USB permission dialog, camera permission dialog and the share chooser are dialog-themed activities that only pause it, while file pickers do stop it, which the PR body lists as accepted. AuthCheckView (MainActivity.kt:177-189) is composed after ContentView (161-175) inside the same Surface with blockPointerInputPassthrough, so it covers the in-window SheetHost (Send/Hardware/Receive) and the BottomSheetOverlayHost at ContentView.kt:760, and payment/NFC deeplinks are deferred by launchScan (AppViewModel.kt:2223-2232) while locked; a payment in flight keeps running under the overlay and its result sheet is in-window. ForgotPinSheet renders through authSheetOverlayState outside the LocalIsAppLocked provider (MainActivity.kt:191-200), so the early return in BottomSheet does not hide it on the PIN screen. Both greptile threads are resolved at 096175462: the lock no longer suspends, and the new test asserts the flag without advanceUntilIdle. No bitkit-ios twin PR was found by title search; the PR body says iOS already clears PIN verification on background.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
```diff
@@ -4837,6 +4842,19 @@ class AppViewModel @Inject constructor(
...
+    /**
+     * Requires the PIN again after the app process moves to the background.
+     *
+     * Runs without suspending so the lock is applied before the activity can resume.
+     */
+    fun lockOnBackground() {
+        if (!isPinEnabled.value) return
+        if (!walletRepo.walletExists()) return
+        if (lightningRepo.isRecoveryMode.value) return
+        _isAuthenticated.update { false }
```
jvsena42, line 4866: **MEDIUM** — **Claim.** With any non-Send sheet open (Receive, Widgets, Activity tag, Backup), background the app, tap a `bitcoin:`/`lightning:` link in another app, enter the PIN: nothing happens. The scan stays queued and runs the moment the user dismisses that sheet, minutes later if they leave it open, opening Send or, with QuickPay enabled and the amount under the threshold, paying without another confirmation.

**Trace at 096175462.**
- HOME with Receive open: `lockOnBackground()` clears `_isAuthenticated` (4854); `_currentSheet` stays `Sheet.Receive`.
- Link -> `handleDeeplinkIntent` (5383) -> `launchScan` -> `!_isAuthenticated.value` -> `enqueueDeferredScan` (2223-2232).
- PIN -> `setIsAuthenticated(true)` -> `flushDeferredScan()` (409-412) -> `if (_currentSheet.value != null) return` (2411), scan still queued. The remaining flush points are `hideSheet(shouldFlushDeferredScan = true)` (4741-4776), the timed-sheet/pair-code collectors (485, 503) and scan-job completion (2274).
- Receive dismissed -> `hideSheet()` -> `flushDeferredScan()` -> `launchScan(preserveUntilComplete = true)` -> `handleScan` -> `isMainScanner` (1971) -> `showSheet(Sheet.Send(Confirm))`, or `handleQuickPayIfApplicable` (3658) -> `SendRoute.QuickPay`, where `SendQuickPayScreen` pays in a `LaunchedEffect` on entry (SendQuickPayScreen.kt:54-58). `isPinForPaymentsEnabled` is only read on SendConfirmScreen, so it does not gate this path.

Before this PR the app resumed unlocked and the same link ran `launchScan` -> `showSheet(Sheet.Send)` immediately, replacing Receive (4716-4724). Master only locked at process start (443), where `_currentSheet` is null, so this sequence was practically unreachable. The new test `payment deeplink received after background lock flushes after unlock` starts with `currentSheet == null` and does not cover it.

**Fix (narrowest).** In `setIsAuthenticated(true)`, when a deferred scan is queued and `_currentSheet.value` is non-null and not `isHighPrioritySheet`, call `hideSheet()` (it clears the sheet then flushes) instead of `flushDeferredScan()`. That restores the pre-PR behaviour (link replaces the open sheet right after unlock) and leaves Send/Gift/Pin sheets uninterrupted. Add a test variant with `showSheet(Sheet.Receive())` before the lock.

app/src/main/java/to/bitkit/ui/components/BottomSheet.kt
```diff
@@ -72,6 +73,7 @@ fun BottomSheet(
     sheetState: SheetState = rememberModalBottomSheetState(skipPartiallyExpanded = true),
     content: @Composable ColumnScope.() -> Unit,
 ) {
+    if (LocalIsAppLocked.current) return
```
jvsena42: **LOW** — **Claim.** Returning from `BottomSheet` while locked removes the overlay entry, so `remember` state inside the `content` lambda is discarded and rebuilt after unlock. The sheet never drew above the PIN pad, so the return costs state without adding coverage.

**Trace at 096175462.**
- `BottomSheet` registers an entry (89-92) rendered by `BottomSheetOverlayHost` (97) via `SheetHost` -> `BottomSheetScaffold` (SheetHost.kt:165), in-window. The host is `BottomSheetOverlayHost(state = bottomSheetOverlayState)` at ContentView.kt:760, inside ContentView, which MainActivity composes (161-175) before `AuthCheckView` (177-189) in the same Surface. `AppAlertDialog` is a real `Dialog` window, so its return is justified; this one is not.
- On lock the early return (76) skips the `DisposableEffect`, which leaves composition -> `overlayState.hide(sheetKey)` (91) -> the entry and its content are disposed. `HwPassphrasePromptSheet` keeps `var passphrase by remember { mutableStateOf("") }` inside `Content` (HwPassphrasePromptSheet.kt:114), composed inside the lambda (85-98): a Trezor passphrase typed before switching to a password manager is empty after the PIN. The other `BottomSheet` callers (AddTag, AddLink, AddContact, Boost, ContactDetail, WatchOnly rename at line 194, HardwareWalletsSettings) hoist their state outside the lambda and survive.

**Fix (narrowest).** Either drop the `LocalIsAppLocked` check from `BottomSheet` and keep it in `AppAlertDialog`, or, if the return is wanted to release IME focus under the overlay, hoist `passphrase` out of `Content` in `HwPassphrasePromptSheet` so it survives the re-show.

app/src/main/java/to/bitkit/viewmodels/AppViewModel.kt
jvsena42: Confirmed the trace at 096175462: lockOnBackground() clears _isAuthenticated without touching _currentSheet, the deeplink lands in launchScan -> enqueueDeferredScan, and flushDeferredScan() on unlock returns early at `if (_currentSheet.value != null) return`, so the scan sat queued until the sheet was dismissed.

Fixed in 6a1c0c825. setIsAuthenticated(true) now checks for a queued deferred scan and calls hideSheet() when a sheet is open, which clears it and then flushes. Two deviations from the suggestion. High-priority sheets are left alone as you proposed, and I also excluded Sheet.QrScanner: the camera sheet is an active capture the user opened, and `scanner result does not discard a queued locked scan` / `custom scanner result flushes a queued locked scan` already pin that behaviour, so the queued scan still waits for it.

I also concluded the flush should not auto-pay. QuickPay assumes the user just acted and is looking at the screen; with the PIN pad interposed the unlock gesture ends up authorising a payment the user never confirmed, and isPinForPaymentsEnabled does not gate SendQuickPayScreen. So a scan that is still queued when setIsAuthenticated(true) runs is marked and skips QuickPay, landing on Send Confirm instead. The mark is applied at unlock, not at enqueue, so scans queued only because settings had not loaded yet are unaffected; a normal scan while already unlocked still QuickPays. This changes the pre-existing expectation in `QuickPay eligible scan remains deferred until authenticated`, which I renamed and updated, and I noted it in the PR body.

New tests in AppViewModelSendFlowTest.kt: `payment deeplink received with a sheet open closes it on unlock` (Receive open before the lock), `payment deeplink received with a high priority sheet open keeps it on unlock` (Pin stays, scan runs on dismiss), and `payment deeplink deferred by the lock skips QuickPay on unlock`. just compile, just test (2612 passing) and just lint are green.

app/src/main/java/to/bitkit/ui/components/BottomSheet.kt
jvsena42: Verified the composition order before touching anything. ContentView is composed at MainActivity.kt:161 inside the LocalIsAppLocked provider and hosts BottomSheetOverlayHost(bottomSheetOverlayState) at ContentView.kt:760; AuthCheckView comes after it in the same Material3 Surface, which stacks children in a Box, and it is fillMaxSize().background(Colors.Black).blockPointerInputPassthrough(). SheetHost also uses BottomSheetScaffold, not ModalBottomSheet, so no sheet gets its own window. The PIN pad therefore already covers every app sheet opaquely and swallows touches. The second host, authSheetOverlayState, is composed after AuthCheckView and sits outside the provider, so ForgotPinSheet is unaffected either way.

So I dropped the check from BottomSheet and kept it in AppAlertDialog, which is a real Dialog window. HwPassphrasePromptSheet now keeps its typed passphrase across the lock without hoisting anything.

One thing the guard was doing incidentally: it removed focused TextFields from the composition, so nothing could bring the IME back over the PIN pad on resume. HwPassphrasePromptSheet auto-focuses its input, and the PIN pad is a custom NumberPad with no IME of its own, so a restored keyboard would have covered the digits. I added a LaunchedEffect in MainActivity that clears focus and hides the keyboard when isAuthenticated goes false, which keeps that closed without discarding state.

Covered by the existing suite only; there is no Compose rendering harness in app/src/test, so this rests on the composition-order argument above rather than a new test. just compile, just test and just lint are green on 6a1c0c825.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `6a1c0c825`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 6a1c0c825: both earlier findings are fixed (lock-deferred link now closes the open sheet on unlock and skips QuickPay; passphrase no longer discarded). Follow-up at 6a1c0c825. MEDIUM (4046216525) is fixed: setIsAuthenticated(true) (AppViewModel.kt:409-422) marks the queued scan, then calls hideSheet() when _currentSheet is non-null, has a deferred scan and is neither high-priority nor QrScanner (425); hideSheet (4753-4791) clears the sheet and flushes, and flushDeferredScan (2406-2427) launches with preserveUntilComplete and the stored suppressQuickPay. Traced Receive/Widgets/Backup/Hardware/TimedSheet open at lock (closed, scan runs to Send Confirm), Send/Gift/Pin open (kept, scan runs on dismiss to Confirm, the option offered in the thread), QrScanner open (kept, existing scanner-result tests pin the flush), two links queued (last wins at enqueueDeferredScan 2364-2389, mark applied to the survivor), link arriving with the PIN pad up (launchScan 2254 enqueues, same path), a job suspended at lock time (deferLockedScan 2337 re-enqueues, mark at unlock covers it), process death (intent re-delivered in onCreate, queued at auth false, flushed by the PIN), and PIN disabled (lockOnBackground 4862 no-ops, scans run unlocked with suppressQuickPay=false). QuickPay: handleQuickPayIfApplicable (3665-3676) returns false only when suppressQuickPay is true, which is set solely by markDeferredScanUnlocked (427-433) or propagated from a queued scan; every user-initiated entry (onScanResult, scanner sheet, clipboard, address continue) passes the default false, so normal QuickPay is unchanged; the only unmarked flush is resetIsAuthenticatedStateInternal with PIN disabled at cold start/post-migration, which is the master behaviour. LOW (4046216534) is fixed by dropping the check from BottomSheet.kt (72-90); all BottomSheet entries render in ContentView's BottomSheetOverlayHost (ContentView.kt:760) through SheetHost/BottomSheetScaffold in-window, ContentView is composed before AuthCheckView inside the same Material3 Surface (MainActivity.kt:161-189), AuthCheckView is fillMaxSize/black/blockPointerInputPassthrough (AuthCheckView.kt:90-95), no Dialog/Popup/ModalBottomSheet hosts exist, and AppAlertDialog keeps its window guard (AppAlertDialog.kt:75); the new LaunchedEffect (MainActivity.kt:165-170) clears focus and hides the IME on lock so a focused TextInput in AddTag or HwPassphrasePromptSheet cannot restore the keyboard over the NumberPad. Widening of the accepted back-press limitation: BottomSheet-based sheets now also stay in composition while locked, so back on the PIN pad reaches their SheetHost BackHandler as it already did for Send/Hardware (PR body lists this as out of scope). Tests: `payment deeplink received with a sheet open closes it on unlock` and `payment deeplink deferred by the lock skips QuickPay on unlock` fail at 096175462 (Receive stays open / quickPayData set); the renamed QuickPay test asserts Confirm and fails on the old assertion; the high-priority test passes on both and only pins intended behaviour. Locals.kt is a doc comment. Earlier clean items (synchronous lockOnBackground, isChangingConfigurations skip, amount-warning gate at 4935, transaction sheet gated on isAuthenticated at MainActivity.kt:220) are untouched.

journeys/security/pin-lock-on-resume.xml
```diff
@@ -0,0 +1,39 @@
...
+    <action>Verify the wallet home screen is visible and the PIN pad is not shown</action>
+    <action>Tap the first activity item on the home screen (testTag "ActivityShort-0"), then tap Tag (testTag "ActivityTag")</action>
+    <action>Verify the Add Tag sheet is visible (testTag "TagInput")</action>
+    <action>Run adb shell input keyevent KEYCODE_HOME and wait 3 seconds</action>
+    <action>Run adb shell am start -n to.bitkit.dev/to.bitkit.ui.MainActivity</action>
+    <action>Take a screenshot and verify only the PIN pad is drawn, with no Add Tag sheet above it</action>
+    <action>Enter the PIN with the number pad keys</action>
+    <action>Verify the Add Tag sheet is visible again (testTag "TagInput"), then press back until the wallet home screen is visible</action>
+    <action>Run adb shell input keyevent KEYCODE_HOME and wait 3 seconds</action>
+    <action>Run adb shell am start -a android.intent.action.VIEW -d "bitcoin:bcrt1qh6k3khs7qz04a9p5eam6v48nlug9d38x4rgxrn?amount=0.0001" to.bitkit.dev</action>
```
jvsena42, line 42: **LOW** — **The payment-URI section does not exercise either fix from 6a1c0c825, so it passes at 096175462 as well.**

The step before this one presses back to Home, so `_currentSheet` is null at unlock. `setIsAuthenticated(true)` (AppViewModel.kt:409-422) then goes straight to `flushDeferredScan()`, which is also what it did before the fix. The case that was stuck, a scan queued behind a sheet left open, never happens here. Keeping the Add Tag sheet open would not help: it is a local `BottomSheet` inside ActivityDetailScreen.kt:263 and is not tracked in `_currentSheet`. The URI is a plain on-chain `bitcoin:` with no `lightning=` param, and `onScanOnchain` only calls `handleQuickPayIfApplicable` when `lnAmountSats > 0u` (AppViewModel.kt:3285-3301), so the QuickPay suppression is never reached either.

Fix (narrowest): after returning to Home, tap Receive (testTag "Receive"), verify "ReceiveScreen", then HOME and fire the URI. After the PIN, assert "SendNumberField" is visible and "ReceiveScreen" is gone. That step fails if the unlock stops closing the open sheet. QuickPay coverage would need a lightning invoice under the threshold with QuickPay enabled in the precondition, asserting Send Confirm rather than a sent payment. That can be left out if the fixture cost is too high.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `5fb363325`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Deeplink step would pass with the deferred-scan fix reverted

Checked: Round 3 (5fb363325, new journey): Round 3 covers only the new journeys/security/pin-lock-on-resume.xml and the two README rows. Every testTag the journey names exists at head: Send (TabBar.kt:135), PinPad (AuthCheckView.kt:143), N0-N9 via NumberPad's default "N$text" tag, ActivityShort-0 (ActivityListSimple.kt:52, rendered on Home), ActivityTag (ActivityDetailScreen.kt:648), TagInput (ActivityAddTagSheet.kt:68) and SendNumberField (SendAmountScreen.kt:260); "Please enter your PIN code" matches security__pin_enter. The preconditions match the lock conditions in lockOnBackground (PIN enabled, wallet exists, not recovery mode), and biometrics off keeps AuthCheckView on the PinPad branch. The uimode toggle really does exercise the isChangingConfigurations skip in MainActivity.onStop, because MainActivity declares no configChanges and is recreated. The payment URI uses a valid regtest bech32 address; with savings above 10 000 sats, onScanOnchain goes to SendRoute.Amount while unlocked, and launchScan defers it while locked. No bitkit-ios twin PR exists, and the cross-platform table records the journey as not ported yet.

journeys/security/pin-lock-on-resume.xml
jvsena42: Confirmed on both counts. The Add Tag sheet is local `showAddTagSheet` state inside ActivityDetailScreen.kt:262-267, not a tracked `Sheet`, and the step before pressed back to Home anyway, so `_currentSheet` was null and setIsAuthenticated (AppViewModel.kt:409-423) fell straight through to `flushDeferredScan()`. The plain `bitcoin:` URI also never reaches the QuickPay branch, which is gated on `lnAmountSats > 0u` (AppViewModel.kt:3286-3291).

Applied the suggested fix: the journey now taps Receive (testTag "Receive"), verifies "ReceiveScreen", then backgrounds and fires the URI, and after the PIN asserts "SendNumberField" is visible and "ReceiveScreen" is gone. `Sheet.Receive` (SheetHost.kt:70) is neither high priority (AppViewModel.kt:5466-5471) nor the QR scanner, so `canCloseSheetForDeferredScan` returns true and the step fails if the unlock stops closing the sheet.

Left QuickPay out and said so in the description: it needs a fresh under-threshold lightning invoice per run plus a funded channel and QuickPay enabled, and it asserts Send confirm rather than anything observable on-chain. AppViewModelSendFlowTest.kt covers both paths ("payment deeplink received with a sheet open closes it on unlock", "payment deeplink deferred by the lock skips QuickPay on unlock"). 510f4b098

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `510f4b098`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Round 4 (510f4b098): the journey LOW is fixed. Round 4 (510f4b098) covers only journeys/security/pin-lock-on-resume.xml. It answers the earlier LOW (4047397642): after returning Home the journey now taps Receive (testTag "Receive", TabBar.kt:156, which calls showSheet(Sheet.Receive) at ContentView.kt:729-731) and verifies "ReceiveScreen" (ReceiveSheet.kt:144) before backgrounding and firing the bitcoin: URI. After the PIN it asserts "SendNumberField" is shown and "ReceiveScreen" is gone. At head, setIsAuthenticated(true) (AppViewModel.kt:409-423) sees a tracked Receive sheet with a deferred scan, canCloseSheetForDeferredScan (425) returns true because Receive is neither high priority nor QrScanner, and hideSheet() closes it and flushes the scan into Send Amount. With the fix reverted, the scan stays queued behind the open Receive sheet, so both halves of that assertion fail. The description now states why the Add Tag sheet does not count and why the QuickPay path is left to AppViewModelSendFlowTest.kt. No new findings.

journeys/README.md
```diff
@@ -129,6 +129,7 @@ fixtures, push notifications) live in each suite's README.
 | [receive](receive) | 1 | Receive sheet tab selection; needs a spending channel, no README |
 | [security](security) | 1 | PIN result sheet layout at a long locale and font scale; no README |
```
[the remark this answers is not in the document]
jvsena42: Right, and fixed in 8220d326e: the stale `[security] | 1` row is gone and the combined `[security] | 2` row now sits between `[receive]` and `[shop]`. The conflict resolution had kept both sides of the row instead of taking the one that matches the folder. The same mistake had hit the suite table on eight other open PRs from this batch, so I fixed those too and checked each result against its `journeys/` folder: one row per suite, alphabetical, and each count matching the folder's `.xml` files.

jvsena42:
Test 8 was a real bug, and this PR introduced it. Fixed in 4931a3b7b.

**Root cause:** the lock now resets `isAuthenticated` in `MainActivity.onStop`, and the flow is collected with `collectAsState`, so `AuthCheckView` → `BiometricsView` is composed while the activity is stopped. `LaunchedEffect(Unit)` in `BiometricPrompt.kt` then called `BiometricPrompt.authenticate()` right away. With the fragment state already saved, androidx biometric logs `Unable to start authentication. Called after onSaveInstanceState()` and returns without calling back, so the screen waited forever. On master the auth overlay is only composed at cold start, when the activity resumes before the effect runs, so master does not have this bug.

**Fix:** `BiometricPrompt` waits for `lifecycle.currentStateFlow` to reach `RESUMED` before it calls `verifyBiometric`. The wait is inside the same `LaunchedEffect(Unit)`, so the prompt still starts once per composition, i.e. once per lock (and once per tap-to-retry). It does not start again on every resume. This also covers the other callers (`SendConfirmScreen`, `PinBiometricsScreen`, `PubkyAuthApprovalSheet`).

**Device (Pixel_9 emulator, dev build of 4931a3b7b, PIN + biometrics on):**
- HOME → reopen, 4 times in a row: the system prompt (`biometric_icon`, "Use PIN code") shows every time, `emu finger touch 1` unlocks it (`Biometric auth succeeded`), and logcat has no `onSaveInstanceState` line.
- HOME while the prompt is up → reopen: the system cancels the prompt, and the app falls back to the PIN pad with "Use Biometrics", as before. PIN unlocks.
- Cold start (force-stop → launch): the prompt shows and the fingerprint unlocks.

`screencap` does not capture the system prompt window (it is secure), so the shots show the state after unlocking and the PIN fallback.

| After resume + fingerprint | Backgrounded during prompt → PIN fallback |
| - | - |
| resume-unlocked | bg-during-prompt |

`just compile`, `just test` and `just lint` pass.

ovitrif (dismissed):
utAck

ovitrif (approved):
utAck

Issue #809, opened 2026-02-26
android: transfer to savings completes on spending details screen


piotr-iohk:
## What happened?
On Android, after completing the "transfer to savings" flow, the app navigates to the **spending balance details** screen instead of the main wallet/home screen.

This feels counterintuitive because spending balance is typically `0` after transferring to savings.

## Expected behavior
After transfer to savings completes, app should return to the main wallet/home screen (matching iOS behavior), not to spending balance details.

## Steps to Reproduce
1. Open Bitkit Android with funds available for transfer.
2. Start a transfer to savings flow.
3. Complete the flow.
4. Observe post-success navigation destination.
5. App lands on spending balance details screen instead of main wallet screen.

## Logs / Screenshots / Recordings

iOS left / Android right:

https://github.com/user-attachments/assets/55a6ad72-88fd-4cbb-abcb-408b565340f7

## Bitkit Version
2.0.5 (`master` revision `b57417d6f7cb6af8fe4cde41d01264f7060285a9`)

## Device / OS
Android simulator (`sdk_gphone64_arm64`), Android 13 (SDK 33)

## Reproducibility
Always

## Additional context
- iOS flow reportedly returns to the main screen after completion.
- This issue focuses on post-transfer navigation behavior and UX consistency across platforms.
- Related broader tracker: #599.

PR #1299, opened 2026-09-17
fix: notify confirmed-only onchain receives

### LINKED ISSUE #797: android: no received transaction notification when tx confirms without being seen in mempool
## What happened?

When an on-chain transaction is sent to the wallet and a block is mined before the next LDK sync picks up the unconfirmed tx, the "received transaction" sheet/notification does not appear. The balance updates correctly and the activity list shows the transaction, but the user gets no visual feedback that funds arrived.

This happens because LDK Node only fires `OnchainTransactionConfirmed` (skipping `OnchainTransactionReceived`) when the tx goes straight to confirmed. The notification is only wired to `OnchainTransactionReceived` in `AppViewModel.handleOnchainTransactionReceived()`, while `handleOnchainTransactionConfirmed()` only updates the activity — it never checks whether the user was already notified.

## Expected behavior

The "received transaction" sheet should appear whenever a new incoming on-chain transaction is detected, regardless of whether it was first seen in the mempool or directly as confirmed.

## Steps to Reproduce

1. Create a new wallet on regtest
2. Copy the wallet's on-chain address
3. Send BTC to that address from an external wallet (e.g. via bitcoin-cli)
4. Immediately mine a block (before the app's next sync cycle)
5. Observe: balance updates, activity list shows the tx, but no "received" notification/sheet appears

Android log confirms zero `OnchainTransactionReceived` events. The tx appears only as `OnchainTransactionConfirmed`:

```
12:28:33.796 LDK event fired: {"type":"OnchainTransactionConfirmed","txid":"bedcc68c...","blockHeight":69235,...}
12:28:33.879 AppViewModel: LDK-node event received: OnchainTransactionConfirmed
12:28:33.954 LDK event fired: {"type":"BalanceChanged","oldSpendableOnchainBalanceSats":0,"newSpendableOnchainBalanceSats":100000000,...}
```

No `OnchainTransactionReceived` event appears anywhere in the log session.

[2026-02-20-missing-received-notification-logs.zip](https://github.com/user-attachments/files/25442547/2026-02-20-missing-received-notification-logs.zip)

## Bitkit Version

master (98fd89c14)

## Device / OS

Android Emulator, API 35

## Reproducibility

Always

## Additional context

- New wallet on regtest
- On-chain related
- Environment: Regtest (network electrum)
- On mainnet this is very unlikely because the ~10 min block interval gives periodic sync time to see the unconfirmed tx first, but it's theoretically possible if the app is backgrounded/offline during the brief window between broadcast and confirmation
- The same issue exists on iOS (filed separately)
- Relevant code: `AppViewModel.kt` lines ~501-512 — `handleOnchainTransactionConfirmed` only updates activity, while `handleOnchainTransactionReceived` (lines ~510-512) triggers `notifyPaymentReceived()`
- Fix: `handleOnchainTransactionConfirmed` should also check whether the tx was already notified and, if not, trigger the same notification flow
- same as: https://github.com/synonymdev/bitkit-ios/issues/455

jvsena42:
Fixes #797
Related to #1275
E2E companion: synonymdev/bitkit-e2e-tests#231 (same branch name, so CI here runs against it)

This PR shows the received sheet and notification for onchain deposits that the wallet first sees already confirmed, without a prior mempool event.

### Description

- Routes confirmed onchain transaction events to the received-payment handler, in the foreground and in the background node service, so a deposit mined before any wallet sync saw it in the mempool is no longer silent.
- Shows a confirmed-only receive only when its block timestamp is within one hour of the device clock, so confirmations replayed by a full scan after a restore stay silent.
- Skips confirmed-only receives while a backup restore or migration is running.
- Keeps a single sheet per deposit: a transaction already presented from its mempool event does not show again when it confirms, because the existing txid claim and persisted seen state still apply.
- Skips the confirmed activity update when the activity already carries the confirmation, so the foreground path, which updates it before calling the handler, does not queue the same Core update twice.
- Adds unit tests for the confirmed-only path, the dedupe with the mempool path, the background notification, and the restore, migration and age guards.

#### Out of Scope

- `NotifyPaymentReceivedHandler`: a confirmed-only deposit found more than one hour after its block (for example, the app was killed while it confirmed and reopened later) stays silent, while a deposit still in the mempool at reopen always shows. A replay discriminator that does not depend on recency, such as a persisted flag set after the first post-restore full scan, is left for a follow-up.
- `AppViewModel`: other receive side effects keyed to the mempool event, such as closing an open Receive sheet and the Paykit contact attachment and private address rotation, still run only on `OnchainTransactionReceived`. On device the Savings address did rotate after a confirmed-only receive.
- RBF: if an already presented incoming tx is replaced and the replacement confirms before any mempool sync sees it, a second sheet can show when the replacement event has not been processed yet. The same race exists today for `OnchainTransactionReceived`.
- `LightningNodeService`: after a process restart, a recent confirmation of a tx already seen through its mempool event still runs the handler's short retry delay (about 1.4 s) before skipping.
- `@boost_1 - Can do CPFP` in the local e2e suite restores from seed and then mines, so the boosted receive becomes a confirmed-only receive and the sheet covers `HeaderMenu`. The sheet is correct there, so the spec dismisses it: synonymdev/bitkit-e2e-tests#231.
- `NotifyPaymentReceivedHandler`: the one-hour window is measured against the device clock, so a device whose clock is off by more than an hour never shows a confirmed-only receive, and ldk-node emits `OnchainTransactionConfirmed` once, so it cannot be recovered later. The mempool path is unaffected. A clock-independent reference, such as the timestamp of the tip block from the sync that produced the event, is left for the same follow-up as the replay discriminator above.
- iOS parity: bitkit-ios#455 tracks the same bug; the follow-up should use the same one-hour block-timestamp rule instead of a block-depth rule, because the node's best block can lag behind the onchain wallet sync.

### Design

Reuses the existing received sheet with no visual change: [Payment received event](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-148491)

### Preview

| Figma | Confirmed-only, foreground | Confirmed-only, background |
| --- | --- | --- |
| Figma payment received event | Received sheet for confirmed-only deposit | Payment Received notification for confirmed-only deposit |

The wallet's primary display is USD, so fiat leads on device; the button label is chosen at random. Both are existing behaviour.

Deposit and mine in one command while on Home; the received sheet appears after the next sync:

https://github.com/user-attachments/assets/ced8234e-228f-43f3-ba0f-7be0192364fb

#### Manual Tests

- [x] **1.** Regtest, app on Home → Receive → Savings address → deposit and mine 1 block in one command: received sheet shows once after the next sync and the log has no `OnchainTransactionReceived` for the txid.
- [x] **2a.** `regression:` Home → deposit to a Savings address without mining: received sheet shows; dismiss it.
  - [x] **2b.** mine 1 block: no second sheet and no "Payment Received" notification.
- [x] **3a.** Background payments on → app in background → deposit and mine 1 block in one command: exactly one "Payment Received" notification.
  - [x] **3b.** tap the notification: app opens on the received sheet; dismiss shows Home with no second sheet.
- [x] **4.** `regression:` Restore a wallet with old onchain deposits: no received sheets or notifications for them.

The device run used the `journeys/onchain-receive` suite, committed in this PR at
`journeys/onchain-receive/` (`confirmed-only-received-sheet.xml`,
`mempool-then-confirmed-single-sheet.xml`, `confirmed-only-background-notification.xml`, plus the
suite `README.md` and the `journeys/README.md` rows). The same diff inline:

<details><summary>journeys/onchain-receive</summary>

```diff
diff --git a/journeys/README.md b/journeys/README.md
index 403c5e230..63807b479 100644
--- a/journeys/README.md
+++ b/journeys/README.md
@@ -119,6 +119,7 @@ fixtures, push notifications) live in each suite's README.
 | [deeplinks](deeplinks) | 2 | `bitkit://screen/…` and sheet routing behind the dev-mode gate; no README |
 | [hardware-wallet](hardware-wallet) | 17 | Trezor over USB; needs the Trezor emulator |
 | [notification-permission](notification-permission) | 4 | Background-setup toggles |
+| [onchain-receive](onchain-receive) | 3 | Received sheet and notification for mempool-first and confirmed-only deposits |
 | [payment-requests](payment-requests) | 2 | Requires a linked fixture issuer; rejected shapes are unit fixtures |
 | [pubky-marketplace](pubky-marketplace) | 1 | Two-wallet Paykit marketplace payment; integration fixture required |
 | [widgets](widgets) | 2 | Needs no backend — the quickest way to see the loop work; no README |
@@ -141,6 +142,7 @@ Known differences in the corpus, as of the iOS port (synonymdev/bitkit-ios#691):
 | `hardware-wallet/usb-reconnect.xml` | `reconnect.xml` — over Bridge, since iOS cannot do WebUSB |
 | `hardware-wallet/receive-onchain.xml`, `hardware-wallet/send-onchain.xml` | not ported |
 | `payment-requests/requested-resolution-failure.xml` | not ported |
+| `onchain-receive/*` | not ported — bitkit-ios#455 tracks the same confirmed-only bug |
 | `deeplinks/*` | not ported — iOS registers the `bitkit` scheme but has no screen or sheet router |
 | — | `hardware-wallet/transfer-to-spending-over-max.xml` exists only on iOS |
 
diff --git a/journeys/onchain-receive/README.md b/journeys/onchain-receive/README.md
new file mode 100644
index 000000000..6aa704af7
--- /dev/null
+++ b/journeys/onchain-receive/README.md
@@ -0,0 +1,25 @@
+# Onchain receive journeys
+
+These journeys cover the received sheet and notification for onchain deposits (issue #797).
+
+ldk-node emits `OnchainTransactionReceived` when the wallet sync finds a transaction in the mempool
+and `OnchainTransactionConfirmed` when it confirms. A transaction that is mined before any sync sees
+it in the mempool produces only the confirmed event. Both events go through
+`NotifyPaymentReceivedHandler`, from `AppViewModel` in the foreground and `LightningNodeService` in
+the background.
+
+A confirmed-only receive is shown only when its block timestamp is within one hour of the device
+clock and no restore or migration is running. A full scan after a restore replays old confirmations
+and stays silent; that case cannot be driven on a funded device and is covered by
+`NotifyPaymentReceivedHandlerTest.kt`.
+
+## Preconditions
+
+- Onboarded regtest wallet with the node running. Fund and mine with the `lsp` helper at the repo root.
+- Wallet sync runs every 10s. For the confirmed-only journeys, run the deposit and the mine in one
+  shell command, then check the log: an `OnchainTransactionReceived` line for the txid means the
+  sync saw the mempool first and the run tested the other path.
+- The background journey needs background payments enabled (Settings > Notifications).
+- Every event the node emits is logged by `LightningService` as `LDK event fired: <json>` under the
+  `APP` logcat tag, so `adb logcat -d -s APP:V` is enough to tell the two paths apart.
+- The Receive sheet's tabs carry no test tag; select the Savings tab by its label.
diff --git a/journeys/onchain-receive/confirmed-only-background-notification.xml b/journeys/onchain-receive/confirmed-only-background-notification.xml
new file mode 100644
index 000000000..58bf496a1
--- /dev/null
+++ b/journeys/onchain-receive/confirmed-only-background-notification.xml
@@ -0,0 +1,22 @@
+<journey name="confirmed-only onchain receive in background posts one notification">
+  <description>
+    Covers issue #797 on the LightningNodeService path. With background payments enabled and the app
+    in the background, a deposit first seen already confirmed must post exactly one "Payment
+    Received" notification.
+
+    Precondition: onboarded regtest wallet with background payments enabled, so the "Bitkit is
+    running in background" foreground service notification is present.
+  </description>
+  <actions>
+    <action>Tap Receive (testTag "Receive") and verify the Receive sheet opens (testTag "ReceiveScreen")</action>
+    <action>Tap the "Savings" receive tab (the tab row carries no test tag), tap "Show Details" (testTag "ShowDetails") and read the address from testTag "ReceiveOnchainAddress"</action>
+    <action>Press back, then send the app to the background: adb shell input keyevent KEYCODE_HOME</action>
+    <action>Run in one command: ./lsp POST /regtest/chain/deposit '{"address":"&lt;savings addr&gt;","amountSat":7970}' &amp;&amp; ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Wait 40s</action>
+    <action>Run: adb shell dumpsys notification --noredact | grep -A3 "pkg=to.bitkit.dev"</action>
+    <action>Verify exactly one notification titled "Payment Received" is posted, with body "Received &lt;amount&gt;" carrying the fiat and BTC amounts in the order set by the primary display setting</action>
+    <action>Open the notification shade and tap the "Payment Received" notification</action>
+    <action>Verify the app opens with the received sheet (testTag "ReceivedTransaction") showing the deposited amount</action>
+    <action>Tap the sheet button (testTag "ReceivedTransactionButton") and verify the home screen shows with no second sheet</action>
+  </actions>
+</journey>
diff --git a/journeys/onchain-receive/confirmed-only-received-sheet.xml b/journeys/onchain-receive/confirmed-only-received-sheet.xml
new file mode 100644
index 000000000..7f4dbc286
--- /dev/null
+++ b/journeys/onchain-receive/confirmed-only-received-sheet.xml
@@ -0,0 +1,25 @@
+<journey name="confirmed-only onchain receive shows the received sheet">
+  <description>
+    Covers issue #797. An onchain deposit the wallet first sees already confirmed, with no prior
+    mempool event, must show the received sheet once. ldk-node emits OnchainTransactionConfirmed
+    without OnchainTransactionReceived in that case.
+
+    Precondition: onboarded regtest wallet, node running, app in the foreground on the home screen.
+    The deposit and the mine must run in one shell command so the 10s wallet sync does not see the
+    transaction in the mempool first. If the log shows OnchainTransactionReceived for the txid, the
+    run tested the mempool path instead; repeat with a new address.
+  </description>
+  <actions>
+    <action>Tap Receive (testTag "Receive") and verify the Receive sheet opens (testTag "ReceiveScreen")</action>
+    <action>Tap the "Savings" receive tab (the tab row carries no test tag), tap "Show Details" (testTag "ShowDetails") and read the address from testTag "ReceiveOnchainAddress"</action>
+    <action>Press back to return to the home screen</action>
+    <action>Run: adb logcat -c</action>
+    <action>Run in one command: ./lsp POST /regtest/chain/deposit '{"address":"&lt;savings addr&gt;","amountSat":21797}' &amp;&amp; ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Wait up to 30s for the next wallet sync</action>
+    <action>Run: adb logcat -d -s APP:V | grep &lt;txid&gt;</action>
+    <action>Verify the log shows an "LDK event fired" line with OnchainTransactionConfirmed for the txid and no OnchainTransactionReceived for it</action>
+    <action>Verify the received sheet (testTag "ReceivedTransaction") is visible with the deposited amount (testTag "MoneyText")</action>
+    <action>Tap the sheet button (testTag "ReceivedTransactionButton")</action>
+    <action>Wait 10s and verify the received sheet does not appear again</action>
+  </actions>
+</journey>
diff --git a/journeys/onchain-receive/mempool-then-confirmed-single-sheet.xml b/journeys/onchain-receive/mempool-then-confirmed-single-sheet.xml
new file mode 100644
index 000000000..6f912d135
--- /dev/null
+++ b/journeys/onchain-receive/mempool-then-confirmed-single-sheet.xml
@@ -0,0 +1,21 @@
+<journey name="mempool-first onchain receive shows one sheet after confirmation">
+  <description>
+    Covers issue #797. Confirmed events now reach the received-payment handler, so a deposit seen
+    in the mempool first must not show a second sheet when it confirms. The handler dedupes on the
+    txid and on the persisted seen state.
+
+    Precondition: onboarded regtest wallet, node running, app in the foreground on the home screen.
+  </description>
+  <actions>
+    <action>Tap Receive (testTag "Receive") and verify the Receive sheet opens (testTag "ReceiveScreen")</action>
+    <action>Tap the "Savings" receive tab (the tab row carries no test tag), tap "Show Details" (testTag "ShowDetails") and read the address from testTag "ReceiveOnchainAddress"</action>
+    <action>Press back to return to the home screen</action>
+    <action>Run: ./lsp POST /regtest/chain/deposit '{"address":"&lt;savings addr&gt;","amountSat":14797}'</action>
+    <action>Wait up to 30s and verify the received sheet (testTag "ReceivedTransaction") is visible with the deposited amount</action>
+    <action>Tap the sheet button (testTag "ReceivedTransactionButton")</action>
+    <action>Run: ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Run: adb logcat -d -s APP:V | grep &lt;txid&gt; and wait until an "LDK event fired" line shows OnchainTransactionConfirmed</action>
+    <action>Wait 30s after that event and verify the received sheet does not appear again</action>
+    <action>Verify no "Payment Received" notification is posted: adb shell dumpsys notification --noredact | grep "Payment Received"</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `NotifyPaymentReceivedHandlerTest.kt` covers a recent confirmed-only receive showing the sheet and the notification, timestamps inside and outside the one-hour window in both directions, received-then-confirmed showing once, negative amounts, the restore and migration skips, and the confirmed activity update being applied once and skipped when it is already stored.
- Unit tests modified: `NotifyPaymentReceivedHandlerTest.kt` moves the existing onchain cases to the new handler dependencies.
- Unit tests added: `LightningNodeServiceTest.kt` covers a background confirmed event reaching the handler and posting exactly one notification, and a handler skip posting none.
- Unit tests added: `AppViewModelSendFlowTest.kt` covers the confirmed event updating the activity and then showing the sheet, and no handler call during a migration.
- Local verification: `just compile`, `just test`, `just lint`.


jvsena42:
### CI: `@boost_1 - Can do CPFP` fails on this branch

`e2e-tests-local - onchain_boost_receive_widgets` fails here across three runs, and the last run had a healthy regtest stack, so it is not the infra flake the first two runs hit (`electrs failed / bitcoind RPC polling failed`).

Only one spec fails; the other 8 in the shard pass:

```
» test/specs/boost.e2e.ts
   ✖ @boost_1 - Can do CPFP
   ✓ @boost_2 - Can do RBF
Error: element ("...resourceId("HeaderMenu")") still not displayed after 30000ms
    at async doNavigationClose (test/helpers/navigation.ts:76:3)
    at async <anonymous> (test/specs/boost.e2e.ts:123:5)
```

The step order in that test is: receive unconfirmed → CPFP boost → restore the wallet from seed → reinstall → mine 1 block → `doNavigationClose`.

The reinstalled wallet never saw the boosted transaction in the mempool, so the mined block makes it a confirmed-only receive. That is exactly the path this PR adds: `Event.OnchainTransactionConfirmed` now reaches `NotifyPaymentReceivedHandler`, and neither guard stops it — the restore has finished, so `backupRepo.isRestoring` is false, and the block was mined seconds earlier, so it is well inside `MAX_CONFIRMED_ONLY_AGE`. The received sheet opens over Home and `HeaderMenu` is not tappable.

The same shard, including `@boost_1`, passes on #1171 in the same time window, so the test is not broken on master.

The sheet is arguably correct here — the wallet really did receive a confirmed transaction it had not seen before — in which case the fix belongs in `bitkit-e2e-tests` (dismiss the sheet after the restore-and-mine step), the way #809 shipped synonymdev/bitkit-e2e-tests#225. The alternative is to widen the guard so a confirmation that lands shortly after a restore stays silent. Leaving the call to the reviewer rather than guessing.

app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
```diff
@@ -96,25 +108,52 @@ class NotifyPaymentReceivedHandler @Inject constructor(
     }
 
     private suspend fun shouldShowOnchain(command: NotifyPaymentReceived.Command.Onchain): Boolean {
-        activityRepo.handleOnchainTransactionReceived(command.event.txid, command.event.details)
-        if (command.event.details.amountSats <= 0) return false
+        if (command.isConfirmedOnly) {
+            if (command.details.amountSats <= 0) return false
+            if (!canShowConfirmedOnly(command)) return false
+            activityRepo.handleOnchainTransactionConfirmed(command.txid, command.details)
```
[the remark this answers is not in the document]
jvsena42: Confirmed for the foreground path: `AppViewModel.handleOnchainTransactionConfirmed` awaits `activityRepo.handleOnchainTransactionConfirmed` and then calls the handler, which ran the same update again for an eligible confirmed-only receive, and the background node service adds a third when it is registered too. The update is idempotent in Core, but each call goes through `ServiceQueue.CORE`, re-lists payments and emits another activity-change notification.

Fixed in 24ab459b9: the handler now reads the activity first and applies the confirmation only when it is not already stored, so the caller that got there first keeps ownership and the background-only process still gets the update it needs. Removing the update from `AppViewModel` instead would have dropped it for the cases the handler skips - old confirmations, outgoing txs, restores, and a mempool-first tx whose presentation is already claimed - so the write stays where it is unconditional.

`NotifyPaymentReceivedHandlerTest.kt` covers both branches: `confirmed-only onchain receive does not reapply a confirmation already stored` and `confirmed-only onchain receive applies the confirmation when the activity is still unconfirmed`.

jvsena42:
Settled the `@boost_1 - Can do CPFP` failure on the e2e side: synonymdev/bitkit-e2e-tests#230.

Checked the spec against the branch. The step order is receive unconfirmed, CPFP boost, restore from seed, reinstall, `mineBlocks(1)`, `doNavigationClose()`. The restored wallet never saw the boosted tx in the mempool, the restore has finished by then so `backupRepo.isRestoring` is false, and the block is seconds old so it is inside `MAX_CONFIRMED_ONLY_AGE`. The sheet is the intended behaviour - it really is a receive the wallet had not seen - so the guard stays as it is and the spec dismisses the sheet between the mine and `doNavigationClose()`.

The e2e PR adds `dismissReceivedPaymentIfShown()`, a tolerant wrapper around `acknowledgeReceivedPayment()` that does not throw when no sheet appears, because the same spec has to pass on iOS and on Android master where no sheet shows. `acknowledgeReceivedPaymentIfPresent()` could not be reused: it is strict on the `local` backend this shard runs on.

`@boost_2 - Can do RBF` also restores and mines and is left alone: that block confirms the outgoing RBF tx, and the incoming 100k was already confirmed before the restore, which matches it passing here today.

jvsena42:
Correction on the e2e link: the companion PR is now synonymdev/bitkit-e2e-tests#231. Same commit, on a branch named `fix/797-notify-confirmed-onchain-receive` so `determine-e2e-branch` resolves to it and this PR's `e2e-tests-local` shards run against the fixed spec. #230 is closed.

jvsena42:
CI is green on 24ab459b9. The rerun resolved the e2e branch to `fix/797-notify-confirmed-onchain-receive` (synonymdev/bitkit-e2e-tests#231) and `e2e-tests-local - onchain_boost_receive_widgets` passed in 23m, along with the other six local shards, build, lint and detekt.

Merge order: #231 first (or together), otherwise this PR is green only while the matching e2e branch exists.

app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
```diff
@@ -96,25 +108,60 @@ class NotifyPaymentReceivedHandler @Inject constructor(
...
+        if (backupRepo.isRestoring.value) {
+            Logger.debug("Skipping confirmed-only receive '${command.txid}' during restore", context = TAG)
+            return false
+        }
+        if (migrationService.isShowingMigrationLoading.value || migrationService.needsPostMigrationSync()) {
+            Logger.debug("Skipping confirmed-only receive '${command.txid}' during migration", context = TAG)
+            return false
+        }
+        val age = nowMillis(clock).milliseconds - confirmationTime.toLong().seconds
+        if (age.absoluteValue > MAX_CONFIRMED_ONLY_AGE) {
```
jvsena42, line 147: **LOW** — **A device whose clock is more than one hour off never shows a confirmed-only receive, and the event cannot be recovered later.**

`canShowConfirmedOnly` compares the block timestamp against `nowMillis(clock)` (NotifyPaymentReceivedHandler.kt:146-147). With automatic time off and the clock set two hours wrong, `age.absoluteValue > MAX_CONFIRMED_ONLY_AGE` is true for a block mined seconds ago, so the handler returns Skip without marking anything. ldk-node emits `OnchainTransactionConfirmed` once (eventHandled runs before onEvent in LightningService.listenForEvents), so the deposit stays silent for that device, which is the #797 behaviour this PR fixes. The mempool path is unaffected.

Not a regression; it narrows who the fix reaches. If a clock-independent reference is available (for example the timestamp of the wallet's tip block from the sync that produced the event), compare against that instead of the device clock; otherwise document the limitation next to the out-of-scope note on stale confirmations.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `24ab459b9`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Device clock off by more than an hour keeps every confirmed-only receive silent

Checked: Traced both entry paths at the PR head: LightningRepo invokes the service handler sequentially and only then emits to AppViewModel, so in the background the service claims first and posts one notification while AppViewModel's later call hits isPresentationClaimed and skips; in the foreground the service's canPresent (App.currentActivity == null) fails and AppViewModel claims, so one sheet. Dedupe across process death does not depend on the in-memory presentationClaims set: markOnchainActivityAsSeen writes seenAt into the Core activity DB via present -> recordPresentation, and CoreService.shouldShowReceivedSheet returns false on seenAt != null, so a mempool-presented tx that confirms after a restart (foreground or background path) resolves to Skip; ldk-node calls eventHandled before onEvent so events are not replayed either. Sends, CPFP children and channel-funding/order payments carry a negative net amount_sats (ldk-node get_tx_details = received - sent) and exit on amountSats <= 0 before any activity write; coop-close and sweep transactions run the same findClosedChannelForTransaction / isTransfer / channelId checks as the existing mempool path, with lightningService.balances read live from node.listBalances(). ldk-node rc.66 updates the payment store before returning wallet events, so handleOnchainTransactionConfirmed finds the payment and creates the activity before shouldShowReceivedSheet runs. Restore, migration and the one-hour block-timestamp guard are evaluated before any state is written, and the greptile thread on the duplicate confirmed update is resolved by applyConfirmationIfMissing. iOS twin https://github.com/synonymdev/bitkit-ios/pull/588 is open.

app/src/main/java/to/bitkit/domain/commands/NotifyPaymentReceivedHandler.kt
jvsena42: Confirmed, and taking the documentation option rather than a code change.

The clock dependency is real: `canShowConfirmedOnly` compares the block timestamp against `nowMillis(clock)`, and ldk-node emits `OnchainTransactionConfirmed` once, so a device more than an hour off never shows that deposit. The rule replaced an earlier block-depth check for a reason: `Node.status().current_best_block` tracks the *lightning* sync, which can lag the onchain sync that produced the event, so the depth check made the sheet depend on which sync ran first.

The tip-block timestamp you suggest is the right reference, but the handler has no clock-independent path to it today — `getStatus()` gives the lightning best block, which is the value the depth rule already failed on. Rather than trade one wrong reference for another in this PR, the limitation is now recorded under Out of Scope next to the existing replay-discriminator follow-up, which is where a persisted post-restore flag would also remove the need for a recency test at all.

journeys/onchain-receive/confirmed-only-background-notification.xml
```diff
@@ -0,0 +1,22 @@
...
+    Precondition: onboarded regtest wallet with background payments enabled, so the "Bitkit is
+    running in background" foreground service notification is present.
+  </description>
+  <actions>
+    <action>Tap Receive (testTag "Receive") and verify the Receive sheet opens (testTag "ReceiveScreen")</action>
+    <action>Tap the "Savings" receive tab (the tab row carries no test tag), tap "Show Details" (testTag "ShowDetails") and read the address from testTag "ReceiveOnchainAddress"</action>
+    <action>Press back, then send the app to the background: adb shell input keyevent KEYCODE_HOME</action>
+    <action>Run in one command: ./lsp POST /regtest/chain/deposit '{"address":"&lt;savings addr&gt;","amountSat":7970}' &amp;&amp; ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Wait 40s</action>
+    <action>Run: adb shell dumpsys notification --noredact | grep -A3 "pkg=to.bitkit.dev"</action>
```
jvsena42: **LOW** — **The command in this step cannot show what step 17 asserts.**

`dumpsys notification --noredact` prints each record as a `NotificationRecord(... pkg=to.bitkit.dev ...)` header followed by `uid`, `opPkg`, `icon`, `flags`, `pri`, `key` and so on; `android.title=String (Payment Received)` and `android.text=...` sit inside the `extras={` block much further down. Three lines of context after each `pkg=to.bitkit.dev` match (which also hits `opPkg=` and `icon=Icon(... pkg=to.bitkit.dev ...)`) never reach it, and the same output includes the foreground-service notification, so the runner cannot count "exactly one notification titled Payment Received" from it. `journeys/README.md` tells the runner to run commands as written and stop rather than improvise, so this is a false fail on the only journey covering the `LightningNodeService` path.

Fix: grep for the title and body directly, e.g. `adb shell dumpsys notification --noredact | grep -E "android.title=String \(Payment Received\)|android.text=String \(Received"`, and count the title matches.

journeys/README.md
```diff
@@ -141,6 +142,7 @@ Known differences in the corpus, as of the iOS port (synonymdev/bitkit-ios#691):
 | `hardware-wallet/usb-reconnect.xml` | `reconnect.xml` — over Bridge, since iOS cannot do WebUSB |
 | `hardware-wallet/receive-onchain.xml`, `hardware-wallet/send-onchain.xml` | not ported |
 | `payment-requests/requested-resolution-failure.xml` | not ported |
+| `onchain-receive/*` | not ported — bitkit-ios#455 tracks the same confirmed-only bug |
```
jvsena42: **LOW** — **The row records a skip the repo rule does not allow here.**

CLAUDE.md allows skipping a journey only when the feature does not exist on the other side. https://github.com/synonymdev/bitkit-ios/pull/588 (`Fixes #455`) wires `onchainTransactionConfirmed` into the received-sheet flow on iOS, so the behaviour these journeys describe is being shipped there, yet #588 carries no `journeys/` files and this row points only at the issue. Whoever reads the table concludes iOS lacks the feature.

Fix: port the three journeys to bitkit-ios#588 with the same file and journey names (noting any adaptation, for example if iOS has no one-hour block-timestamp guard or no background service path), or change the row to say the port is pending in https://github.com/synonymdev/bitkit-ios/pull/588 and what differs.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `2a6b9dd68`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — `grep -A3 "pkg=to.bitkit.dev"` never prints the notification title the next step verifies; LOW — Cross-platform row says not ported, but the iOS twin fixes the same behaviour without the journeys

Checked: Round 3 (2a6b9dd68, new journeys): Round 3 delta is journeys only (three onchain-receive journeys, the suite README and two rows in journeys/README.md). Every testTag the journeys name exists at head: Receive (TabBar.kt:156), ReceiveScreen (ReceiveSheet.kt:144), ShowDetails and ReceiveOnchainAddress (ReceiveQrScreen.kt:427/658), ReceivedTransaction and ReceivedTransactionButton (NewTransactionSheet.kt:161/194), MoneyText (Money.kt). The preconditions are reachable with repo tooling: `./lsp` at the root wraps the blocktank-api script, `deposit` returns the txid the log grep needs, `Env.walletSyncIntervalSecs` is 10, and `LightningService` logs every event as `LDK event fired: <json>` at debug level under the `APP` tag with the event class name as `type`. The staging regtest tip timestamp was checked against the wall clock (about 5 minutes apart), so the one-hour `MAX_CONFIRMED_ONLY_AGE` guard does not make the confirmed-only journeys false-fail today. The background flow matches the code: `App.currentActivity` clears on stop, `LightningNodeService` claims and posts `Payment Received` / `Received %s` and caches the sheet, `AppViewModel`'s later call skips on the claim, and `ContentView` ON_START calls `consumePaymentReceivedInBackground`, so tapping the auto-cancel notification opens the received sheet once. The mempool-then-confirmed journey would catch a second sheet if the txid claim or seenAt dedupe regressed.

journeys/onchain-receive/confirmed-only-background-notification.xml
jvsena42: Confirmed. The title and body sit in the record extras, far below the `pkg=` header, so `grep -A3 "pkg=to.bitkit.dev"` never reaches them, and the foreground-service notification is in the same dump.

Checked the exact strings the app posts before rewriting the command: ReceivedNotificationContent.kt:30 uses `notification__received__title` = "Payment Received" (strings.xml:446) and line 47 uses `notification__received__body_amount` = "Received %s" (strings.xml:442), pushed through `pushNotification` (Notifications.kt:85-86). The foreground-service notification uses `app_name`, which is "Bitkit Regtest" on the dev flavour (LightningNodeService.kt:176, build.gradle.kts:205), so it does not collide with either pattern.

The step now runs `adb shell dumpsys notification --noredact | grep -E "android\.title=String \(Payment Received\)|android\.text=String \(Received "` and the assertion counts the title lines and reads the body line next to it. e2be5d905

Not changed here: the three `cjit-notifications` journeys use the same `grep -A3` shape and have the same blind spot, but they are outside this PR.

journeys/README.md
jvsena42: Confirmed against the local bitkit-ios checkout. synonymdev/bitkit-ios#588 ("Fixes #455") routes `onchainTransactionConfirmed` into the same received-sheet flow (AppViewModel.swift, `presentReceivedSheetForOnchainTransaction` called from both the received and confirmed cases), so iOS is not missing the feature, and the PR carries no `journeys/` files.

Took the second option and reworded the row rather than porting into someone else's open PR. Two real differences, both from reading #588:

- Historical replays: iOS suppresses them with a persisted `pendingRestoreActivitySeen` flag set at Get Started and cleared by the first post-restore on-chain sync (SettingsViewModel.swift:407-415, AppViewModel.swift syncCompleted handler). Here it is the one-hour block-timestamp guard, `MAX_CONFIRMED_ONLY_AGE` in NotifyPaymentReceivedHandler.kt. A stale-confirmation step has to drive a restore on iOS instead of a clock.
- `confirmed-only-background-notification.xml` has no counterpart: it covers the LightningNodeService foreground-service path (LightningNodeService.kt:104-107), which iOS does not have; #588 only presents an in-app sheet.

The row now says the port is pending in synonymdev/bitkit-ios#588 and lists both. c67782eb6

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `c67782eb6`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Round 4 (e2be5d905, c67782eb6): both journey LOWs are fixed. Round 4 covers two journey-only commits that answer my earlier LOWs. e2be5d905: the new grep matches what AOSP NotificationRecord prints inside the extras block under --noredact (`android.title=String (Payment Received)`, `android.text=String (Received ...)`). The receive notification is posted through Context.pushNotification (Notifications.kt:84-86) with plain Strings from ReceivedNotificationContent.build (R.string.notification__received__title = "Payment Received", notification__received__body_amount = "Received %s"), so the extras class is String and both patterns hit. The foreground-service notification uses app_name ("Bitkit Regtest" on dev, build.gradle.kts:205) and notification__service__body, so it matches neither pattern and the title count is exact. c67782eb6: the cross-platform row now says the port is pending in synonymdev/bitkit-ios#588 and records both adaptations (restore-flag replay guard instead of the one-hour window, no foreground-service counterpart for the background notification journey), which is what I asked for. No new findings.

ovitrif:
@jvsena42 ci failed

jvsena42:
Looked at it: infra, not the change. The `settings` shard failed in its `before all` hook waiting for `TotalBalance-primary` for 60 s, and the same job log has `Failed to load snapshot 'default_boot'`, three `adb ... failed with exit code 1` lines and `Failed to find ColorBuffer: 209` — the app never reached Home on that runner. Every other shard passed, including `onchain_boost_receive_widgets`, which is the one this PR actually affects.

That shard has failed the same way on several PRs in this batch this afternoon. Re-running the failed jobs now. The real e2e issue this PR did cause is already handled separately in synonymdev/bitkit-e2e-tests#231, which dismisses the received sheet in `@boost_1`.

ovitrif:
@jvsena42 maybe we can sync with base branch to retrigger ci

ovitrif (dismissed):
utAck

piotr-iohk:
QA reviewed on d23597d0cd581236ba98d58ba9c18037e956a22a.

Reviewed the confirmed-only receive path against `@boost_1` in [run 35210047449](https://github.com/synonymdev/bitkit-android/actions/runs/35210047449); no device run this round. Twin [bitkit-ios#588](https://github.com/synonymdev/bitkit-ios/pull/588) is still a draft.

## Findings

**HIGH**
- [twin] After seed restore, mining the same 100k already on the list opens Received Bitcoin over Boost Fee ([run 35210047449](https://github.com/synonymdev/bitkit-android/actions/runs/35210047449)). [bitkit-ios#588](https://github.com/synonymdev/bitkit-ios/pull/588) suppresses that path (`pendingRestoreActivitySeen` + mark unseen seen) and left `@boost_1` alone; this PR plus [bitkit-e2e-tests#231](https://github.com/synonymdev/bitkit-e2e-tests/pull/231) treats the sheet as product. Pick one restore rule before merge — and do not land [bitkit-e2e-tests#231](https://github.com/synonymdev/bitkit-e2e-tests/pull/231) until the twin matches. See note https://github.com/synonymdev/bitkit-ios/pull/588#issuecomment-4686118396 that classified this as a false behavior rather than the e2e issue.

jvsena42:
> QA reviewed on [d23597d](https://github.com/synonymdev/bitkit-android/commit/d23597d0cd581236ba98d58ba9c18037e956a22a).
> 
> Reviewed the confirmed-only receive path against `@boost_1` in [run 35210047449](https://github.com/synonymdev/bitkit-android/actions/runs/35210047449); no device run this round. Twin [bitkit-ios#588](https://github.com/synonymdev/bitkit-ios/pull/588) is still a draft.
> 
> ## Findings
> **HIGH**
> 
> * [twin] After seed restore, mining the same 100k already on the list opens Received Bitcoin over Boost Fee ([run 35210047449](https://github.com/synonymdev/bitkit-android/actions/runs/35210047449)). [bitkit-ios#588](https://github.com/synonymdev/bitkit-ios/pull/588) suppresses that path (`pendingRestoreActivitySeen` + mark unseen seen) and left `@boost_1` alone; this PR plus [bitkit-e2e-tests#231](https://github.com/synonymdev/bitkit-e2e-tests/pull/231) treats the sheet as product. Pick one restore rule before merge — and do not land bitkit-e2e-tests#231]([test: dismiss confirmed-only receive sheet in boost_1 bitkit-e2e-tests#231](https://github.com/synonymdev/bitkit-e2e-tests/pull/231)) until the twin matches. See note [fix: notify on-chain receives that skip mempool bitkit-ios#588 (comment)](https://github.com/synonymdev/bitkit-ios/pull/588#issuecomment-4686118396) that classified this as a false behavior rather than the e2e issue.

@piotr-iohk could you update your skill for using a "require change" comment type for cases like this?

piotr-iohk (changes requested):
See comment. https://github.com/synonymdev/bitkit-android/pull/1299#issuecomment-5759595316

jvsena42:
Agreed, this was a real bug, and the fix now matches bitkit-ios#588.

**Rule (same as iOS):**
1. Get Started on restore success sets a persisted `pendingRestoreActivitySeen` flag.
2. While it is set, no on-chain Received sheet is shown on either the mempool or the confirmed-only path. The activity rows are still written.
3. The first `SyncCompleted(ONCHAIN_WALLET)` after restore marks all unseen activities as seen, then clears the flag. The existing seen check then stops restored transactions from notifying when they confirm, and new receives notify as before.

Commits: 83e20fa25 (the rule) and ee50a0c1d. The second one clears the flag only if marking the activities seen succeeds. Otherwise a failed write would leave the whole restored history unseen with the hold already lifted, which is this same bug. iOS clears the flag without checking; Android retries on the next on-chain sync. The other Android-only guards (restore in progress, migration, 1-hour age window) are unchanged.

Tests are in `NotifyPaymentReceivedHandlerTest.kt`, `AppViewModelSendFlowTest.kt` and `WalletViewModelTest.kt`: the hold on both paths, the clear on the first on-chain sync, no clear on a lightning sync, a restored transaction that has already been seen staying silent, new receives notifying again, and the flag being kept when marking seen fails. Each key test was checked against the code with the fix reverted, and it fails.

On e2e: bitkit-e2e-tests#231 is closed without merging, as you asked. `@boost_1` no longer gets the sheet, so the spec on `main` is right as it is, same as iOS.

I haven't done a device run of your restore → mine scenario on this revision yet. It's listed unchecked in the PR's QA notes.

jvsena42:
Re: ovi-reviewer Test 5 (received sheet over the activity detail when the restored deposit confirms).

Real bug. The flag timing was not the cause.

**Root cause:** `ActivityService.markActivityAsSeen` wrote `seenAt` by calling bitkit-core `updateActivity` with a copied activity. In bitkit-core, `update_onchain_activity_by_id` and `update_lightning_activity_by_id` only write `tx_type`, `timestamp` and `contact` to the `activities` table, so `seen_at` was never persisted. Only the dedicated `mark_activity_as_seen` call writes it, and iOS already uses that one. As a result, `markAllUnseenActivitiesAsSeen()` after restore logged `Marked activity … as seen` for every row and changed nothing. The `seenAt != null` check in `shouldShowReceivedSheet` never blocked anything, and the only dedupe left was the in-memory presentation claim.

On device, before the fix: after `Marked activity 15b53a5f… as seen` was logged twice, `activity.db` still had `seen_at = NULL` for all three onchain rows.

That explains the run: in your restore, ldk-node loaded the unconfirmed tx from the VSS wallet state and emitted no `OnchainTransactionReceived`, so nothing claimed the txid in memory. The first onchain sync after Get Started lifted the hold, the "mark seen" was a no-op, and `OnchainTransactionConfirmed` passed every check. This affects master too: it is the same code, and the post-migration `markAllUnseenActivitiesAsSeen()` has also been a no-op.

**Fix** (d853eb2f2): `markActivityAsSeen` now calls bitkit-core `markActivityAsSeen(walletId, activityId, seenAt)` instead of `updateActivity`. No change to the restore hold.

**Tests:** `ActivityServiceTest.kt` (new) checks that the core seen call gets the wallet id, activity id and timestamp and that `updateActivity` is not called, plus the missing-activity case. It fails on the previous code (`Wanted but not invoked: markActivityAsSeen`, with `updateActivity` invoked instead). `just compile`, `just test`, `just lint` pass.

**Device** (RestoreTest AVD, dev debug build with local Electrum so mining is manual):
1. Deposited 27 182 sats without mining, then wiped the app and restored from seed. The log shows no `OnchainTransactionReceived` for the txid. Tapped Get Started.
2. The first onchain sync logged `Marking activities replayed by the first sync after restore as seen`. `activity.db` now has `seen_at = 1790084825` on all rows, including the unconfirmed one.
3. Opened the pending activity and mined 1 block. The status turned Confirmed, there was no received sheet for 50 s, and the log shows `Skipping received sheet for transaction 80db50aa… - already seen`.
4. Fresh deposit of 16 180 sats: the received sheet showed after 10 s. Dismissed it and mined 1 block: no second sheet.

| Detail after the block | Fresh receive after restore |
| --- | --- |
| Activity detail confirmed with no sheet | Received sheet for a new deposit |

**Separate observation, not changed here:** in an earlier run (before the fix), ldk-node did emit `OnchainTransactionReceived` for the restored mempool tx while the restore screen was still up. The hold only starts at Get Started, so that sheet was queued and opened right after Get Started. This fix does not change it. Moving the flag to restore start would cover it, but iOS sets it on Get Started too, so I left it for a decision.

<details><summary>Sheet after Get Started for a mempool tx seen during restore</summary>

Received sheet right after Get Started

</details>

jvsena42:
On the open point in https://github.com/synonymdev/bitkit-android/pull/1299#issuecomment-5777782485: the restore hold keeps starting at Get Started, matching iOS. A mempool event emitted while the restore screen is still showing is out of scope for this PR.

piotr-iohk:
QA reviewed on d853eb2f22db32bdb77ad8ad9df135b7f70b83a4.

Since the review on 21 Sep, Get Started holds on-chain received sheets until the first on-chain sync, and that sync marks activities seen through the core call before it clears the hold. [bitkit-e2e-tests#231](https://github.com/synonymdev/bitkit-e2e-tests/pull/231) is closed and was not merged.

Ran on Pixel_6. A deposit mined before the wallet saw it (21 797 sats) opened Received Bitcoin once, with only OnchainTransactionConfirmed in the log. A mempool deposit (14 797 sats) opened the sheet once; after the block there was no second sheet and no Payment Received notification. Sending the app home while a 9 970 sat sheet was up, then coming back, still showed that sheet.

Twin [bitkit-ios#588](https://github.com/synonymdev/bitkit-ios/pull/588) is still a draft, so it was not reviewed.

## Findings

No findings.

**Checked and clean**
- The hold blocks both the mempool and the confirmed-only on-chain paths, and a lightning sync does not clear it.
- The flag stays set when marking activities seen fails.
- `markActivityAsSeen` calls core `markActivityAsSeen` instead of `updateActivity`, which never wrote `seen_at`.
- The next confirm reuses the seen row, so a mempool receive does not sheet again.

QA LGTM

piotr-iohk (dismissed):

piotr-iohk:
Pls resolve conflicts.

piotr-iohk:
🐞 potential bug · [5e83e77](https://github.com/synonymdev/bitkit-android/commit/5e83e77149e286ccd987b1e5b27ef14753cecdb7) · https://github.com/synonymdev/bitkit-android/actions/runs/35903670272/job/107352366182

E2E failure triaged as potential bug · @boost_1 failed 3/3: Payment Received sheet for 100000 sats covers HeaderMenu after restore and mine; the same spec passed on 2926fd71 before the restore

app/src/main/java/to/bitkit/viewmodels/WalletViewModel.kt
```diff
@@ -564,6 +566,9 @@ class WalletViewModel @Inject constructor(
         // The node starts and syncs long before the backup is read, so ordinary uploads are held from
         // here rather than from the restore itself, which would upload over the backup it has not read.
         backupRepo.setRestorePending(true)
+        // Same reason for the received-sheet hold: by the time the user taps continue the node has
+        // already been replaying historical transactions for a while.
+        settingsStore.update { it.copy(pendingRestoreActivitySeenSince = nowTimestamp().epochSecond) }
```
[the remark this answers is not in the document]
jvsena42: **Correct, and already fixed in `f1c2e26d3`** — this landed against `73bd3bbe5`, one commit before the fix. Your diagnosis matches what I found from the other end: CI went red on `73bd3bbe5` with `@boost_1 - Can do CPFP` failing on `element ("HeaderMenu") still not displayed after 30000ms`, three attempts, right after the test's restore step. The sheet that should have been held was opening over the wallet and covering the header.

The fix is the first option you suggested — keep the current value when the backup is applied:

```kotlin
store.updateData { current ->
    // The received-sheet hold is armed before the backup is read, so it has to survive the
    // settings the backup brings with it - otherwise the replayed history raises sheets.
    data.copy(pendingRestoreActivitySeenSince = current.pendingRestoreActivitySeenSince)
}
```

And the test you asked for, in `SettingsStoreTest.kt`:

```kotlin
@Test
fun `restoring settings keeps the received sheet hold armed by the restore in flight`() = test {
    val restoreStartedAt = 1_700_000_000L
    sut.update { it.copy(pendingRestoreActivitySeenSince = restoreStartedAt) }
    val backup = SettingsBackupV1(createdAt = 0L, settings = SettingsData())

    assertTrue(sut.restoreFromBackup(backup).isSuccess)

    assertEquals(restoreStartedAt, sut.data.first().pendingRestoreActivitySeenSince)
    assertTrue(sut.data.first().pendingRestoreActivitySeen)
}
```

It goes red on the old `store.updateData { data }`:

```
SettingsStoreTest > restoring settings keeps the received sheet hold armed by the restore in flight FAILED
```

E2E is green on the current head `3ebff42b3`, including the boost shard that caught it, so the Test 5 case you name is covered end to end now.

I checked the iOS twin (synonymdev/bitkit-ios#588) for the same hazard and it does not have it: `restoreSettingsDictionary` writes only the keys listed in `SettingsBackupConfig.settingsKeyTypes`, and `pendingRestoreActivitySeenSince` is not one of them, so the iOS restore leaves it alone.

piotr-iohk:
QA reviewed on 3ebff42b.

Android device, Pixel 6 emulator, local regtest, CI APK bitkit-e2e-apk_5029 from run 35983947634.

Passed
- Confirmed-only deposit (d44a20de, 21 797 sats): one Received Bitcoin sheet. Log had OnchainTransactionConfirmed and no OnchainTransactionReceived.
- Mempool then mine (fc051b99, 14 797 sats): sheet once, nothing on confirm.
- Background, foreground service running (411e373b, 44 444 sats): one Payment Received notification. Tap opened the sheet. Dismiss returned home with no second sheet.
- Restore of that wallet: no received sheet for the existing deposits. Mining the restored unconfirmed 55 555 marked it Confirmed and did not open a received sheet over the activity.

Head is now 57da4891. Those commits are mnemonic wrapping, not this receive path, so they were not in this run.

iOS twin bitkit-ios#588 must be undrafted and brought to parity with this PR before the pair is approved. Android still has the one-hour block-time guard and the background notification. iOS shows the in-app sheet only.

ovitrif:
@ovi-reviewer test

ovitrif (approved):
utAck

PR #1303, opened 2026-09-17
fix: settle transfers for closed channels

### LINKED ISSUE #614: Transfer in Progress banner never hides
### Repro

> [!NOTE]
> Not sure what I did but I suspect these were the steps.

1. Start a Transfer to Spending
2. Wipe wallet before it completes
3. Uninstall + reinstall and restore from same seed
4. Either: do another transfer or wait for the transfer order to fulfill and the channel to open
5. Certainly I have no pending channels (as seen in LN Connections list)
   but I have "TRANSFER IN PROGRESS" banner on home screen.
   > idea: probably this could be a good optimisation there, sync the banners with the list of pending channels

### Media

https://github.com/user-attachments/assets/73dad4d4-6168-4891-a4e6-c8063f02a497

### Extra Issues
- [ ] Predictive back gesture broken on LN Connections List & Detail Screens

jvsena42:
Related to #614
Related to #1275

This PR settles Transfer to Spending entries whose channel was closed or whose LSP order ended, so the "TRANSFER IN PROGRESS" banner no longer stays up after the channel is gone.

### Description

- Skips transfer syncing while the node's channel list is unavailable, because a channel missing from an empty list was treated as not open and could leave or settle transfers on bad data.
- Settles a transfer whose resolved channel is no longer open when the channel is recorded in the local closed channels or when its LSP order reports the channel closed (state closed, closing transaction or close info). A channel counts as open only while the node reports it ready, so a channel still listed while it negotiates a shutdown is checked for closure evidence too.
- Settles a transfer with no resolved channel when its LSP order has expired (unchanged) or when the order's channel is closed.
- Fetches the orders of unsettled transfers by id from the LSP, so orders missing from the local cache after a restore, or cached with a stale open channel, can resolve their channel or show closure; one batched fetch per 2 minute LSP poll interval.
- Never settles without evidence: an unknown order, a failed fetch or a failed closed-channels read leaves the transfer active.

#### Out of Scope

- Restore before the channel opens (the literal repro in #614): restore does not bring back the paid orders list, so a restored PAID order is never opened by the app and the banner stays until the LSP expires the order. This PR keeps that transfer active because there is no closure evidence; it needs a follow-up issue, likely rebuilding paid orders from active transfers on restore. For that reason this PR does not close #614.
- LN Connections List & Detail: the predictive back gesture sub-item of #614 belongs in its own issue.
- `BlocktankRepo.fetchOrders`: the by-id fetch goes through the core service call that also refreshes active orders, so each fetch makes an extra round trip and does not recompute the paid orders list.
- Balance derivation: the banner and balance logic are unchanged; settled transfers drop out of the active transfers flow as before.
- iOS: parity of the closed-channel and order-closure settle rules (and the journey port) is a follow-up.

### Design

N/A — no UI changes.

### Preview

Pending transfer banner survives a force-stop and relaunch while the node starts; the by-id order fetch runs and does not settle an unpaid order.

| Pending transfer | Node starting | Node running |
| - | - | - |
| Pending transfer banner | Splash while node starts | Banner after node start |

https://github.com/user-attachments/assets/b449d1bf-ac8a-4bec-aee9-6f14ce134a1a

Closed-channel settle, on a temporary uncommitted build where the channel-ready settle was disabled so the transfer stayed active with a persisted channel id (zero-conf orders otherwise settle on channel ready before any close):

| Channel open, transfer active | After LSP close, transfer settled |
| - | - |
| Transfer active with open channel | Transfer settled after close |

https://github.com/user-attachments/assets/539770d8-71a0-4285-aa54-3a0a1d13866c

Branch build after the tests, with no active transfers:

| Lightning Connections | Home |
| - | - |
| Lightning Connections after | Home final |

#### Manual Tests

- [ ] **1a.** Savings → Transfer To Spending → confirm, order unpaid: Home shows TRANSFER IN PROGRESS.
  - [ ] **1b.** force-stop → relaunch → wait for node start: banner still shows.
- [ ] **2.** `regression:` Transfer To Spending → mine 1 block → channel ready: banner clears and Spending increases.
- [ ] **3.** Build with channel-ready settle disabled → Transfer To Spending → channel open → LSP closes channel → mine 3 blocks: Connection Closed sheet, banner gone, Spending back to its prior value, log shows the transfer settled for a closed channel.
- [ ] **4.** `regression:` Settings → Advanced → Lightning Connections after the above: no pending connection left behind.

Tests 1 to 4 were run on a Pixel_9 emulator against staging regtest. In test 3 the transfer settled 59 ms after the closed channel was registered. Gap (b), gap (c) and a to-self-delayed force close were not device-tested; unit tests cover (b) and (c). The journey `journeys/transfers/closed-channel-transfer-settles.xml`, committed in this PR, walks tests 1 and 2 plus the close path as a no-regression check; its on-screen assertions also pass on master, and the fixed build is told apart by the "orders for active transfers" log line it writes while the transfer is pending. Test 3 needs the build with the channel-ready settle disabled and is not reachable from the journey. The journey is inlined below.

<details><summary>journeys/transfers/closed-channel-transfer-settles.xml</summary>

```diff
diff --git a/journeys/transfers/closed-channel-transfer-settles.xml b/journeys/transfers/closed-channel-transfer-settles.xml
new file mode 100644
index 000000000..7b0daa4ee
--- /dev/null
+++ b/journeys/transfers/closed-channel-transfer-settles.xml
@@ -0,0 +1,58 @@
+<journey name="transfer to spending banner clears when the lsp closes the channel">
+  <description>
+    Covers issue #614. A pending Transfer to Spending keeps its "TRANSFER IN PROGRESS" banner across
+    an app restart while the node starts, and a transfer whose channel the LSP later closes does not
+    leave a banner behind or inflate the Spending balance.
+
+    Precondition: onboarded dev wallet (staging regtest) with a positive on-chain Savings balance and
+    a running node connected to the LSP. Start on the wallet home screen. The LSP only opens the
+    channel after the order payment has one confirmation, so the transfer stays pending until a block
+    is mined. The dev build mines that block itself: the SettingUp screen auto-mines one block 5 s
+    after it appears on regtest, and leaving the screen cancels it. Tap "Continue Using Bitkit" within
+    those 5 s so the transfer is still pending for the restart steps; the block is then mined by hand
+    later in the journey.
+
+    Transfer orders are zero-conf, so on an unmodified build the transfer settles as soon as the
+    channel is ready and the closed-channel settle path is not reached on its own. This journey then
+    checks that nothing is left behind after the close. Reaching the closed-channel path on a device
+    needs a temporary local change that skips the channel-ready settle; it was verified that way for
+    issue #614 and is covered by TransferRepoTest.kt.
+
+    On-screen, this journey is a no-regression check: the banner, balance and close-sheet steps pass
+    on master too, because both versions settle the transfer when the zero-conf channel becomes ready.
+    The one thing that tells the two builds apart on a device is the app log. Only the fixed build
+    fetches LSP orders for still-pending transfers, so only it writes "orders for active transfers"
+    while the transfer is pending; master never calls the LSP there. The fixed build may also write
+    "Skipped syncing transfers because channels are unavailable" while the node is starting after the
+    restart, which master does not write either, but that line is not guaranteed on every run.
+
+    The close moves the transfer amount back to Savings minus fees, so the wallet ends close to its
+    starting balances.
+  </description>
+  <actions>
+    <action>Note the Spending balance shown on the home screen (testTag "ActivitySpending")</action>
+    <action>Tap the Savings balance card (testTag "ActivitySavings")</action>
+    <action>Tap "Transfer To Spending" (testTag "TransferToSpending")</action>
+    <action>If the spending intro screen appears, tap its button (testTag "SpendingIntro-button")</action>
+    <action>Verify the spending amount screen (testTag "SpendingAmount") is visible</action>
+    <action>Wait until the available amount (testTag "SpendingAmountUnit") shows a positive value</action>
+    <action>Tap the "5" key (testTag "N5") enough times to enter an amount the LSP accepts (50000 sats is a safe value), then tap Continue (testTag "SpendingAmountContinue")</action>
+    <action>Swipe "Swipe To Transfer" to confirm</action>
+    <action>As soon as the setting up screen (testTag "LightningSettingUp") appears, tap "Continue Using Bitkit" (testTag "TransferSuccess-button") within 5 s, before the screen auto-mines a block; do not dump the layout first</action>
+    <action>Verify the purple "TRANSFER IN PROGRESS" banner is visible on the home screen (screenshot; the banner carries no testTag). If the banner is already gone, the auto-mine won the race: report the run as inconclusive and start over</action>
+    <action>Run adb shell am force-stop to.bitkit.dev, then launch the app again</action>
+    <action>Wait for the home screen and for the node to start (app log line "Node started")</action>
+    <action>Verify the purple "TRANSFER IN PROGRESS" banner is still visible (screenshot)</action>
+    <action>Read the order id from the app log: the active TO_SPENDING transfer's "lspOrderId" in the latest "Active transfers at block height" line</action>
+    <action>Verify the app log has a line ending in "orders for active transfers" written after the transfer was created; master never writes it, so its absence means the build under test does not carry the fix</action>
+    <action>Run ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Poll ./lsp GET /channels/&lt;lspOrderId&gt; until channel.state is "open", and note channel.fundingTx.id and channel.fundingTx.vout</action>
+    <action>Tap the OK button (testTag "TransferSuccess-button") if the Transfer Successful screen is shown; its label is randomized ("Great!", "Nice!", …)</action>
+    <action>Run ./lsp POST /regtest/channel/close '{"fundingTxId":"&lt;fundingTx.id&gt;","vout":&lt;fundingTx.vout&gt;,"forceCloseAfterSec":0}'</action>
+    <action>Run ./lsp POST /regtest/chain/mine '{"count":3}'</action>
+    <action>Wait for the "Connection Closed" sheet and tap OK (testTag "ConnectionClosedButton")</action>
+    <action>Verify no "TRANSFER IN PROGRESS" banner is visible on the home screen (screenshot)</action>
+    <action>Verify the Spending balance (testTag "ActivitySpending") matches the value noted in the first step</action>
+    <action>Verify the app log has a "ready, settled transfer:" line for the transfer (the build writes "Channel &lt;channelId&gt; ready, settled transfer: &lt;id&gt;", with no quotes around the ids), and that the latest "Active transfers at block height" line lists no TO_SPENDING transfer. The closed-channel settle lines ("Settled transfer '&lt;id&gt;' for closed channel …") only appear on a build that skips the channel-ready settle, so do not expect them here</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added in `TransferRepoTest.kt`: settle on a closed-channel record or an LSP-closed order for a resolved channel; settle on an executed order with a closed channel, close info or an expired order; no sync when channels are unavailable; a restored transfer with an uncached order settles once the fetched order resolves a ready channel; stale cached orders reported closed by the fetch settle; unknown orders, failed fetches and a PAID order with an opening channel stay active; a listed but not ready channel settles on cached or fetched order closure and stays active while it is still opening; no fetch when the channel is ready or the cached order already shows closure; one batched fetch, throttled until the interval passes.
- Unit tests added in `BlocktankRepoTest.kt`: `fetchOrders` caches an uncached order and replaces a stale one; on server error it fails and keeps the cache.
- Negative control for the closed-channels fallback: a case where `activityService.closedChannels` throws asserts `syncTransferStates` still succeeds, the TO_SPENDING transfer stays active, and the to-savings settle still runs. Dropping the `runSuspendCatching { … }.getOrDefault(emptySet())` in `loadClosedChannelIds` makes it fail on `result.isSuccess`, along with one pre-existing case, 2 of 58 in the file.
- No balance use case test: settlement is proven in `TransferRepoTest.kt`, and a mock-flow test in `DeriveBalanceStateUseCaseTest.kt` was dropped because it proved nothing.
- Local runs: `just compile`, `just test` and `just lint` pass; detekt findings reported are in files this PR does not touch.


app/src/main/java/to/bitkit/repositories/TransferRepo.kt
```diff
@@ -163,46 +168,137 @@ class TransferRepo @Inject constructor(
...
-                    if (order?.state2 == BtOrderState2.EXPIRED) {
-                        markSettled(transfer.id)
-                        Logger.info(
-                            "Order ${transfer.lspOrderId} expired, settled transfer: ${transfer.id}",
-                            context = TAG,
-                        )
-                    }
                 }
+
+                channelId != null -> settleIfChannelClosed(transfer, channelId, closedChannelIds)
```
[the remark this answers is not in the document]
jvsena42: Confirmed and fixed in 75dbcbd3c. LDK sets is_channel_ready false while a channel negotiates a shutdown and still lists it, so a transfer whose channel started closing before the app ever saw it ready hit the `channel != null` branch and did nothing. The settle path now keys on `channel?.isChannelReady == true` and otherwise falls through to `settleIfChannelClosed`, and `fetchOrdersAwaitingEvidence` only skips a listed channel when it is ready, so the order is refreshed while the channel is not ready. Nothing settles without evidence: an opening channel with a PAID order stays active. The local closed-channels read is still gated on a missing channel, since that record is only written on the ChannelClosed event after the channel leaves the list, so the listed-but-closing case is settled by the order. New tests in TransferRepoTest.kt: settle on a cached order closure for a listed non-ready channel, settle when the fetched order reports it closed, and no settle while it is still opening.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `75dbcbd3c`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Traced settleToSpendingTransfers at the PR head (app/src/main/java/to/bitkit/repositories/TransferRepo.kt:193-301) against every TO_SPENDING/MANUAL_SETUP creation site and every syncTransferStates caller (LDK event handlers, debounced sync, ActivityRepo.syncActivities, TransferViewModel.watchOrder): a settle now requires the LDK channel to be ready, the persisted channel id to appear in the local closed-channel records (written only on Event.ChannelClosed by LightningRepo.registerClosedChannel), or the LSP order to carry closure evidence (state CLOSED, closingTxId or close), so a PAID order with no channel, an OPENING channel, an unknown order, a failed by-id fetch and a failed closed-channels read all leave the transfer active and the banner up. Confirmed in bitkit-core v0.5.14 (src/lib.rs get_orders, src/modules/blocktank/api.rs refresh_orders) that orders(orderIds, refresh=true) fetches those ids from the LSP and upserts them, so restored transfers whose order was never cached do get fresh evidence; BlocktankRepo.fetchOrders merges by id and never drops cached orders. Traced DeriveBalanceStateUseCase: a TO_SPENDING transfer settled for a closed channel stops contributing to balanceInTransferToSpending and to the on-chain subtraction, and the closing funds are already covered by the COOP_CLOSE/FORCE_CLOSE transfer that handleChannelClosed creates for counterparty closes, so to-spending vs to-savings attribution is unchanged; the only window where a listed-but-closing channel's balance shows as Spending instead of in-transfer is bounded by the ChannelClosed event and matches existing LSP-initiated close behaviour. The by-id fetch is throttled with a CAS on an AtomicLong so concurrent syncs make one request per 2 minutes, EXPIRED settlement is unchanged, and the runSuspendCatching switch propagates cancellation to callers that already rethrow it. No fund-safety, key-material or authorization surface is touched.

journeys/transfers/closed-channel-transfer-settles.xml
```diff
@@ -0,0 +1,46 @@
...
+  <actions>
+    <action>Note the Spending balance shown on the home screen (testTag "ActivitySpending")</action>
+    <action>Tap the Savings balance card (testTag "ActivitySavings")</action>
+    <action>Tap "Transfer To Spending" (testTag "TransferToSpending")</action>
+    <action>If the spending intro screen appears, tap its button (testTag "SpendingIntro-button")</action>
+    <action>Verify the spending amount screen (testTag "SpendingAmount") is visible</action>
+    <action>Wait until the available amount (testTag "SpendingAmountUnit") shows a positive value</action>
+    <action>Tap the "5" key (testTag "N5") enough times to enter an amount the LSP accepts (50000 sats is a safe value), then tap Continue (testTag "SpendingAmountContinue")</action>
+    <action>Swipe "Swipe To Transfer" to confirm</action>
+    <action>Tap "Continue Using Bitkit" (testTag "TransferSuccess-button")</action>
```
jvsena42: **MEDIUM, journey, non-blocking** — **The journey assumes the transfer stays pending until a block is mined, but the app mines that block itself.**

`SettingUpScreen` (app/src/main/java/to/bitkit/ui/screens/transfer/SettingUpScreen.kt:56-68) runs `regtestMine(1u)` 5 s after it appears when `Env.network == Network.REGTEST`, which is the case on the dev build this journey targets. An agent that swipes, dumps the layout and then taps "Continue Using Bitkit" (line 30) usually spends more than 5 s on that screen. The block confirms the order payment, the LSP opens the zero-conf channel, the channel becomes ready and `settleToSpendingTransfers` settles the transfer (TransferRepo.kt:212-214). The banner is then gone at line 31 or line 34, before the force-stop/relaunch check it was meant to exercise, and the manual mine at line 35 is redundant. The precondition text (lines 8-10) mentions only other staging users mining the block.

Fix: tell the runner to tap "Continue Using Bitkit" as soon as the SettingUp screen appears (within 5 s, since leaving the screen cancels the LaunchedEffect), and add the auto-mine to the precondition paragraph. If the banner is already gone, the runner should report the run as inconclusive instead of failing it.

journeys/transfers/closed-channel-transfer-settles.xml
```diff
@@ -0,0 +1,46 @@
...
+    <action>Run ./lsp POST /regtest/chain/mine '{"count":1}'</action>
+    <action>Poll ./lsp GET /channels/&lt;lspOrderId&gt; until channel.state is "open", and note channel.fundingTx.id and channel.fundingTx.vout</action>
+    <action>Tap the OK button (testTag "TransferSuccess-button") if the Transfer Successful screen is shown; its label is randomized ("Great!", "Nice!", …)</action>
+    <action>Run ./lsp POST /regtest/channel/close '{"fundingTxId":"&lt;fundingTx.id&gt;","vout":&lt;fundingTx.vout&gt;,"forceCloseAfterSec":0}'</action>
+    <action>Run ./lsp POST /regtest/chain/mine '{"count":3}'</action>
+    <action>Wait for the "Connection Closed" sheet and tap OK (testTag "ConnectionClosedButton")</action>
+    <action>Verify no "TRANSFER IN PROGRESS" banner is visible on the home screen (screenshot)</action>
+    <action>Verify the Spending balance (testTag "ActivitySpending") matches the value noted in the first step</action>
+    <action>Verify the app log has a "Settled transfer" line naming the closed channel, and that the latest "Active transfers at block height" line lists no TO_SPENDING transfer</action>
```
jvsena42: **LOW, journey, non-blocking** — **Every assertion also passes on master, and the last step looks for a closed-channel settle line the head build does not emit.**

On master's `syncTransferStates`, a pending transfer with no resolved channel settles only on an EXPIRED order, so the banner survives the restart (lines 32-34). After the mine, the zero-conf channel becomes ready and settles the transfer in both versions, so the banner is gone and Spending is back before the close. With the fix reverted the journey passes. The description admits the closed-channel path is not reached, but the PR body says the journey "covers tests 1 to 3", and test 3 needs a modified build.

For the same reason, line 44's "`Settled transfer` line naming the closed channel" cannot be met by the closed-channel settle logs (TransferRepo.kt:261-263, 273-275, 290-292). The build only writes `Settled transfer: <id>` (no channel, TransferRepo.kt:107) and `Channel <cid> ready, settled transfer: <id>` (TransferRepo.kt:214). A strict runner false-fails; a lenient one passes on the ready line, which master also writes.

Fix: change line 44 to check for the `Channel '<cid>' ready, settled transfer` line and the absence of TO_SPENDING in the latest "Active transfers" line. Also say in the description, or in the PR body's QA Notes, that the journey is a no-regression check for tests 1-2 and the close path, and that test 3 needs the modified build.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `d1ed9bd2b`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: MEDIUM (journey, non-blocking) — SettingUp auto-mines a block after 5 s on regtest, so the restart checks can false-fail; LOW (journey, non-blocking) — Passes with the fix reverted, and the final log check asks for a line an unmodified build never writes

Checked: Follow-up on 26e3e355c (new journey; the auto-mine claim was re-checked against SettingUpScreen.kt at head): Reviewed the new journey commit 26e3e355c against the PR head. Every testTag it names exists at head: ActivitySpending/ActivitySavings (HomeScreen.kt), TransferToSpending (SavingsWalletScreen.kt:168), SpendingIntro-button, SpendingAmount, SpendingAmountUnit, SpendingAmountContinue, N5 (NumberPad default `N$text`), TransferSuccess-button (SettingUpScreen.kt:177, label "Continue Using Bitkit" while in progress and a randomized OK after), and ConnectionClosedButton. The screen order SpendingAmount -> SpendingConfirm ("Swipe To Transfer") -> SettingUp -> Home matches ContentView.kt. The ./lsp endpoints (POST /regtest/chain/mine, GET /channels/:id, POST /regtest/channel/close with fundingTxId/vout/forceCloseAfterSec) match the blocktank-api reference, and an LSP force close is a counterparty close, so AppViewModel.handleChannelClosed shows Sheet.ConnectionClosed. The banner assertions match HomeViewModel (SPENDING banner iff balanceInTransferToSpending > 0) and DeriveBalanceStateUseCase; the "Node started" and "Active transfers at block height" log lines exist and verbose logs are written to the log file. The README suite row and the cross-platform "not ported" row are present, and there is no bitkit-ios twin PR.

journeys/transfers/closed-channel-transfer-settles.xml
jvsena42: Confirmed and fixed in 79f3fd46c. `SettingUpScreen.kt:56-72` runs `regtestMine(1u)` after a 5 s delay when `Env.network == Network.REGTEST`, and the dev flavor maps to REGTEST (`app/build.gradle.kts:87`), so the dev build this journey targets mines its own block. The progress screen and the success screen are the same composable (`SettingUpScreen.kt:91-180`), tagged "LightningSettingUp" while `lightningSetupStep < 3` and sharing the "TransferSuccess-button" tag, so leaving it by tapping that button cancels the `LaunchedEffect` before the mine fires.

The journey now tells the runner to tap "Continue Using Bitkit" within 5 s of the "LightningSettingUp" screen appearing and to skip the layout dump there, records the auto-mine in the precondition paragraph, and says to report the run inconclusive (not failed) if the banner is already gone at the next step. The later manual `./lsp POST /regtest/chain/mine` is no longer redundant, since it is now what confirms the order payment.

journeys/transfers/closed-channel-transfer-settles.xml
jvsena42: Mostly holds; fixed in 79f3fd46c, with two corrections.

The "also passes on master" part is right. Before the mine, a pending transfer has no resolved channel, so master reaches the EXPIRED check only and leaves it active; after the mine both versions settle it at `TransferRepo.kt:213-215` when the zero-conf channel is ready, so the banner, balance and Connection Closed steps are a no-regression check on both builds. The journey now says that in its description rather than implying the close path discriminates.

Correction 1: the head build does emit closed-channel settle lines (`TransferRepo.kt:260-263`, `272-275`, `289-292`) — they are just unreachable on this journey's path, since the channel-ready branch settles first. Line 44 now asserts the line the run really produces, `Channel <channelId> ready, settled transfer: <id>` (`TransferRepo.kt:214`, no quotes around the ids — the suggested `Channel '<cid>'` form would never match), plus no TO_SPENDING in the latest "Active transfers at block height" line (`DeriveBalanceStateUseCase.kt:73`), and it states explicitly that the closed-channel lines are not expected here.

Correction 2: there is an observable discriminator, so the journey does not have to settle for "no check possible". Only this branch fetches orders for still-pending transfers, so only it writes `Fetched '<n>' of '<m>' orders for active transfers` (`TransferRepo.kt:249-251`); neither that string nor `BlocktankRepo.fetchOrders` exists on master. A new step after the restart asserts that line. `Skipped syncing transfers because channels are unavailable` (`TransferRepo.kt:177`) is also head-only, but it depends on a sync landing while the node is still starting, so it is mentioned in the description as a possible extra signal rather than asserted.

The PR body's QA note and inline diff are updated to match: the journey walks tests 1 and 2 plus the close path, and test 3 still needs the build with the channel-ready settle disabled.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `79f3fd46c`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up: the earlier journey finding is fixed. Follow-up on 79f3fd46c: both earlier journey findings are fixed. Auto-mine race: `SettingUpScreen` mines one block 5 s after its `LaunchedEffect(Unit)` starts on REGTEST (SettingUpScreen.kt:56-71), the in-progress screen carries testTag `LightningSettingUp` and the button `TransferSuccess-button` (:97, :177), and `onContinueClick` calls `navController.navigateToHome()` (ContentView.kt:1053-1055), which takes the screen out of composition and cancels the effect before the mine fires; the journey now requires that tap within 5 s and marks a lost race as inconclusive rather than failed. Discrimination: `Fetched '<n>' of '<m>' orders for active transfers` (TransferRepo.kt:250) is head-only (`fetchOrders` and the string do not exist on origin/master), and `lastOrdersFetchMs` starts at 0 (:47), so the first sync with a pending transfer that has no ready channel fetches and logs it; the new step therefore fails on a reverted build. The final assertion now matches the real `Channel $channelId ready, settled transfer: ${transfer.id}` line (:214), and the description states correctly that the closed-channel settle lines are unreachable on this path.

app/src/test/java/to/bitkit/repositories/TransferRepoTest.kt
```diff
@@ -715,6 +726,409 @@ class TransferRepoTest : BaseUnitTest() {
...
+        val settledAt = setupClockNowMock()
+        val transfer = transferEntity(id = ID_TRANSFER, fundingTxId = null, channelId = ID_CHANNEL)
+        val closedChannel = mock<ClosedChannelDetails> {
+            on { channelId } doReturn ID_CHANNEL
+        }
+
+        whenever(transferDao.getActiveTransfers()).thenReturn(flowOf(listOf(transfer)))
+        whenever(lightningRepo.getChannels()).thenReturn(emptyList())
+        whenever(lightningRepo.getBalancesAsync()).thenReturn(Result.success(mock()))
+        whenever(activityService.closedChannels(SortDirection.DESC)).thenReturn(listOf(closedChannel))
```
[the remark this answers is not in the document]
jvsena42: Verified: `loadClosedChannelIds` swallows the failure at `TransferRepo.kt:297-301` with `runSuspendCatching { … }.getOrDefault(emptySet())`, it is called from `settleToSpendingTransfers` at `TransferRepo.kt:203`, and that call sits inside the `runSuspendCatching` of `syncTransferStates` at `TransferRepo.kt:172-186` — so an escaping exception would fail the whole pass and skip `settleToSavingsTransfers` at `TransferRepo.kt:185`. All four existing closed-channel cases stubbed `closedChannels` with successful returns only (`TransferRepoTest.kt:742, 764, 788, 808`).

Added `syncTransferStates keeps transfer active and settles to-savings when closed channels fail`: `closedChannels(SortDirection.DESC)` throws, one TO_SPENDING transfer with a channel missing from the LDK list, one TO_SAVINGS transfer whose balance is gone. It asserts the read was attempted, `syncTransferStates` still succeeds, the TO_SPENDING transfer is never settled, and the to-savings settle still runs.

Negative control: replacing the body of `loadClosedChannelIds` with the bare `coreService.activity.closedChannels(...)` call (no `runSuspendCatching`, no default) makes the new test fail on `assertTrue(result.isSuccess)` — the pass now returns failure, exactly the regression described. The pre-existing `does not settle TO_SPENDING transfer when channel not found` case fails with it too, 2 of 58. Restored the fallback and re-ran: `just compile`, `just test` and `just lint` all pass.

Commit: f2a884101.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `f2a884101`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on f2a884101: test-only. The new `TransferRepoTest` case stubs `activityService.closedChannels` to fail and asserts `syncTransferStates` still succeeds, keeps the TO_SPENDING transfer active (no closed-channel evidence) and settles the TO_SAVINGS one, which pins the failure branch reviewed earlier.

ovitrif (approved):
utAck

PR #1304, opened 2026-09-17
fix: clear migration data on wipe

### LINKED ISSUE #869: [Bug]: Network status icon keeps showing sync issues
### What happened?

It keeps showing UI for network connection issues for me, yellow sync icon pops up repeatedly. Not sure if there is actually any problems or just UI.

### Additional Context

- I did not allow the app to run in background
- It is the same with and without connecting over VPN, otherwise stable WiFi

https://github.com/user-attachments/assets/eaaf70db-b797-45a6-99a6-827f5c7d8b25

[bitkit_2026-03-26_11-16-44.log](https://github.com/user-attachments/files/26272073/bitkit_2026-03-26_11-16-44.log)

### Bitkit Version

2.1.2

### Device / OS

Pixel 8

### Reproducibility

Always

jvsena42:
Fixes #869
Related to #1275

This PR stops the post-migration sync from retrying forever, which made the network status icon keep flashing sync issues for wallets migrated from the previous Bitkit app.

### Description

- Clears pending migration data and the post-migration sync flag when a wallet is wiped, so a new or restored wallet no longer retries a sync for the old wallet's Blocktank orders, metadata, transfers and boosts. Both steps run right after the LDK storage wipe and before the keychain wipe, so the migration store can only survive a wipe that is interrupted while the wallet itself also survives. Neither step depends on the keychain or on any later wipe step: `rn_migration` is its own DataStore, untouched by the keychain, core, database, settings, cache and widgets wipes. The sync flag is cleared in its own wipe step, so a failure in either step still leaves the other attempted. If they fail, the error is logged and the wipe continues, like the other wipe steps, and the wipe's overall result still depends only on the keychain wipe.
- Fetches Blocktank orders in chunks of 20 IDs in all four migration fetch paths (RN Blocktank apply, invalid transfer cleanup, metadata reapply retry, remote paid orders), because a request with every pending order ID can exceed the server's array limit and fail on each sync.
- Refreshes active orders only on the first chunk. `CoreService.orders` gained `refreshActive: Boolean = refresh`, so every existing caller is unchanged, while the chunked migration fetch runs one blanket active-order refresh per fetch instead of one per chunk. The per-ID `refresh = true` stays, because with `refresh = false` bitkit-core reads only the local DB and would return nothing for migrated order IDs it has never seen.
- Keeps the migration checked and completed markers when wiping, so the RN migration does not run again from leftover files.

#### Out of Scope

- A device that wiped on an earlier released build with pending migration data still imports the old wallet's Blocktank orders once on its first sync after upgrading: the stale store left behind by that earlier wipe still has `needsPostMigrationSync` true, and the chunked fetch in this PR makes the request succeed where master's single request failed. The imported orders show in Settings → Advanced → Channel Orders. No channel is opened — `openChannelWithPaidOrders` reads `cacheStore.paidOrders`, which the wipe reset and this path never repopulates. Master already imports for wallets with 20 or fewer stale IDs, so this only widens an existing population, and no proportionate fix exists for devices already in that state.
- `MigrationService`: in-memory restore state (the remote backup restoring and loading flags, pending channel migration and pending remote activity data) is not reset on wipe. It already survives a wipe on master, and this only matters when an RN remote backup restore stalls and the wallet is wiped in the same process. Planned as a follow-up.
- `MigrationService.reapplyMetadataAfterSync`: the retry still has no limit if an order fetch keeps failing for another reason, such as the server rejecting a purged order ID. Chunking only removes the array limit cause.
- bitkit-core `refresh_active_orders` still sends every active order ID in a single request, so a wallet with more than the server's array limit of active orders can still fail that call. Fixing it belongs in bitkit-core; this PR only stops that refresh from running once per chunk.
- Network status icon debounce and health semantics are unchanged.
- iOS parity: check whether `bitkit-ios` keeps RN migration pending data after a wipe and sends order IDs in one request.

### Design

N/A — no UI changes.

### Preview

N/A. There is no UI change, and the fix needs RN migration data that cannot be created on the test emulator without writing DataStore entries by hand.

#### Manual Tests

- [x] **1.** Wallet migrated from RN Bitkit with more than 20 Blocktank orders → open app and wait for sync: network status icon stops showing sync issues and logs no longer repeat "Post-migration sync incomplete, will retry on next sync".
- [x] **2a.** RN-migrated wallet with pending migration data → Settings → Backup → Reset And Restore → Reset Wallet → create new wallet: no repeated post-migration sync retries and the status icon stays healthy.
  - [x] **2b.** Restore the wallet instead of creating a new one: same result, and the RN migration does not run again.
- [x] **3.** `regression:` Native wallet (not migrated) → Settings → Backup → Reset And Restore → Reset Wallet → create or restore wallet: wipe completes and the app starts normally.
- [x] **4.** `regression:` Any wallet with Blocktank orders → open the Transfer flow and pull the order list: orders still load and refresh, confirming the `refreshActive` default did not change normal order fetching.

#### Automated Checks

- Unit tests modified: `WipeWalletUseCaseTest.kt` checks that migration data is cleared after the LDK storage wipe and before the keychain wipe, that a failed clear still marks the migration checked and finishes the wipe, that the post-migration sync flag is still cleared when the migration data clear fails, and that migration data is still cleared when the flag clear fails.
- Unit tests added: `MigrationServiceTest.kt` checks that 25 order IDs are fetched in two requests of 20 and 5 with results joined in order, that active orders are refreshed only on the first chunk, and that an empty list makes no request.
- Ran locally: `just compile`, `just test` (2601 passed), `just lint`.


app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
```diff
@@ -103,6 +103,7 @@ class WipeWalletUseCase @Inject constructor(
         activityRepo.resetState()
         hwWalletRepo.resetState()
         resetWalletState()
+        step("clear migration data") { migrationService.cleanupAfterMigration() }
```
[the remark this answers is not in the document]
jvsena42: Confirmed and fixed in 42e4804. `cleanupAfterMigration()` runs `clearPersistedMigrationData()` before `setNeedsPostMigrationSync(false)`, so a throw in the first edit skipped the flag clear while the wipe step swallowed the error and still marked the migration checked. The wipe now clears the sync flag in its own step before `clear migration data`, so either step failing leaves the other attempted.

Covered by two new tests in `WipeWalletUseCaseTest.kt`: the flag is still cleared when `cleanupAfterMigration()` throws, and migration data is still cleared and the wipe still finishes when the flag clear throws. `AppViewModel` call sites are unchanged; there `cleanupAfterMigration()` only runs when `canCleanupAfterMigration` is true and keeping the flag on failure is the correct retry behaviour.

app/src/main/java/to/bitkit/services/MigrationService.kt
```diff
@@ -1410,6 +1410,9 @@ class MigrationService @Inject constructor(
         Logger.info("Migration transfer cleanup completed", context = TAG)
     }
 
+    private suspend fun fetchOrdersChunked(orderIds: List<String>): List<IBtOrder> =
+        fetchOrdersInChunks(orderIds) { coreService.blocktank.orders(orderIds = it, filter = null, refresh = true) }
```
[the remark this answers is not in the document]
jvsena42: Confirmed against bitkit-core and fixed in 42e4804. `CoreService.orders(refresh = true)` calls `refreshActiveOrders()` and then `getOrders(ids, refresh = true)`; in Rust those are `db.refresh_active_orders()` (which fetches every active order from the LSP and upserts it) and `db.refresh_orders(ids)`. Chunking therefore ran ceil(N/20) full active-order refreshes. Worse for this PR: `refresh_active_orders` sends all active order IDs in one `get_orders` request, so on a migrated wallet with many active orders it can hit the same array limit the chunking is meant to avoid, on every chunk.

`CoreService.orders` now takes `refreshActive: Boolean = refresh`, so every existing caller behaves exactly as before, and `MigrationService.fetchOrdersChunked` passes `refreshActive = true` only for the first chunk. That keeps one blanket refresh per fetch, as on master, and still fetches the IDs in chunks. Dropping `refresh = true` on the per-ID call was not an option: with `refresh = false` bitkit-core reads only the local DB, which would return nothing for migrated order IDs the fresh core DB has never seen. `MigrationServiceTest.kt` has a new test asserting the refresh flag is true only on the first of three chunks.

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
```diff
@@ -103,6 +103,8 @@ class WipeWalletUseCase @Inject constructor(
         activityRepo.resetState()
         hwWalletRepo.resetState()
         resetWalletState()
+        step("clear post-migration sync flag") { migrationService.setNeedsPostMigrationSync(false) }
```
jvsena42: **LOW** — **If the wipe coroutine is cancelled or the process dies after `keychain.wipe()` (line 94) and before lines 106-107, the device has no mnemonic but still has `needsPostMigrationSync` and the pending order IDs, so the next wallet runs the old wallet's post-migration pass once on its first sync.**

`WalletViewModel.wipeWallet` runs in the activity-scoped `viewModelScope` (MainActivity.kt:84, WalletViewModel.kt:543); `step()` uses `runSuspendCatching`, which rethrows `CancellationException` (ext/Coroutines.kt:11-12), so finishing the activity mid-wipe aborts `wipeLocal` at the current step. The window is small (95-105 are local resets; `deleteToken()` returns a Task that is not awaited), and it is the same slot master used for `markMigrationChecked`, so this is pre-existing. With this PR's chunking the survivor no longer loops (the fetch now succeeds and clears the store), it imports the old wallet's orders once (see the `MigrationService.kt:1732` finding).

Fix: move the two migration steps up to directly after `lightningRepo.wipeStorage` (line 91), before `keychain.wipe()`, so the store can only survive when the wallet itself also survives.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `42e480427`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Wipe cancelled between keychain wipe and migration cleanup keeps the stale store for the next wallet; LOW — Upgrade path: a device already in the #869 state imports the old wallet's orders into the new wallet once

Checked: Traced clearPersistedMigrationData (MigrationService.kt:305-319): it removes only the five rn_migration keys for pending Blocktank order IDs, paid orders, metadata, transfers and boosts, plus the in-memory copies; RN_MIGRATION_CHECKED/COMPLETED, channel-recovery and peer-recovery markers, RN LDK/MMKV files and the RN keychain are untouched, so the RN migration cannot re-run after a wipe and nothing needed to restore funds is deleted (wipe-then-restore re-derives channel data, paid orders, metadata, transfers and boosts from the RN remote backup via restoreFromRNRemoteBackup at 1290-1304 and 1559-1623, and a VSS restore does not depend on this store). Ordering: both migration steps run after coreService.wipeData/db.clearAllTables/cacheStore.reset, so a later sync can no longer find pending data or the sync flag; markMigrationChecked stays last; each step is isolated by step() so a DataStore failure in one leaves the other attempted, and WipeIncomplete still surfaces when the keychain wipe fails. Chunking: fetchOrdersInChunks (2181-2188) runs sequentially through ServiceQueue.CORE, chunk size 20 stays under the qs arrayLimit and the server's Zod max(50); bitkit-core get_orders(refresh=true) only calls db.refresh_orders(ids) (lib.rs:799-803), which upserts each fetched order itself, so refreshActive=true on the first chunk only is exactly one active-order refresh per fetch as on master, and an existing caller with refresh=false still skips it. A chunk failure throws through the outer runCatching in all four call sites, which is the same all-or-nothing outcome as master's single request; applyRNBlocktank re-persists all IDs for retry and cleanupInvalidMigrationTransfers leaves its done flag unset, so retries stay idempotent. Tests would fail on master (missing verify calls and the ordering assertion).

app/src/main/java/to/bitkit/services/MigrationService.kt
```diff
@@ -1719,11 +1729,7 @@ class MigrationService @Inject constructor(
             if (orderIds.isNotEmpty()) {
                 Logger.info("Retrying ${orderIds.size} pending Blocktank orders", context = TAG)
                 runCatching {
-                    val fetchedOrders = coreService.blocktank.orders(
-                        orderIds = orderIds,
-                        filter = null,
-                        refresh = true,
-                    )
+                    val fetchedOrders = fetchOrdersChunked(orderIds)
```
jvsena42, line 1732: **LOW** — **For a device that wiped on a released build with the stale store still present (the #869 report: 34 order IDs), the first sync after upgrading now succeeds and upserts the old wallet's Blocktank orders into the new wallet's DB.**

Nothing clears the store on upgrade, so `needsPostMigrationSync` is still true; `handleSyncCompleted` (AppViewModel.kt:1437) runs `completeMigration` -> `reapplyMetadataAfterSync` (1535 -> 1671); `loadPersistedMigrationData` (1672) loads the old IDs; the chunked fetch at 1732 succeeds where master's single request failed; bitkit-core `refresh_orders` and `upsertOrderList` (1734) write every fetched order into the current wallet's DB. Visible result: the old orders appear in Settings > Advanced > Channel Orders (ui/settings/ChannelOrdersScreen.kt:71). A `TO_SPENDING` transfer is only created for an order still in state `PAID` (1638-1646); that would add `clientBalanceSat` to the Home to-spending amount (DeriveBalanceStateUseCase.kt:87-99) until `syncTransferStates` settles it on `EXPIRED` (TransferRepo.kt:186-194), but a stale device's paid orders have expired long before the upgrade. No channel is opened: `openChannelWithPaidOrders` (BlocktankRepo.kt:475-479) reads `cacheStore.paidOrders`, which the wipe reset and this path never repopulates. Master already does this import for <=20 stale IDs; the PR extends it to >20 as the price of ending the loop, and only wipes performed on this build are covered. Population is RN-migrated, wiped pre-fix, different new wallet, >20 IDs.

Fix: no persisted marker separates a stale store from a legitimate pending migration, and comparing the RN keychain mnemonic against the native one is disproportionate for this population (and must still allow wipe-then-restore-same-seed). Add it to Out of Scope: "A device that wiped on an earlier build with pending migration data still imports the old wallet's orders once on its first sync after upgrading."

app/src/main/java/to/bitkit/usecases/WipeWalletUseCase.kt
jvsena42: Verified the ordering claim before acting on it. `WalletViewModel.wipeWallet` runs in `viewModelScope`, and `step()` uses `runSuspendCatching`, which rethrows `CancellationException`, so finishing the activity mid-wipe really does abort at the current step and can leave the mnemonic gone with the migration store intact.

Both migration steps only edit `rn_migration`, which is its own `preferencesDataStore` (`MigrationService.kt:2309`) plus the in-memory pending fields, so nothing in them depends on the keychain, `coreService.wipeData()`, `db.clearAllTables()` or any store reset, and nothing later depends on them. Moved `clear post-migration sync flag` and `clear migration data` to directly after `lightningRepo.wipeStorage`, before the keychain wipe. `mark migration checked` stays where it was. Failure handling is unchanged: both are still `step()` calls that log and continue, and the result still depends only on `keychainWiped`.

Test: added `invoke should clear migration data before wiping the keychain` to `WipeWalletUseCaseTest.kt`, an `inOrder(migrationService, keychain)` verification, and updated the full-order test to expect the new slot. 16/16 in that class pass, `just test` is 2601 passed, `just compile` and `just lint` clean. Commit 9295975c0.

app/src/main/java/to/bitkit/services/MigrationService.kt
jvsena42: Recorded under Out of Scope in the PR body: a device that wiped on an earlier released build with pending migration data still imports the old wallet's Blocktank orders once on its first sync after upgrading, because the stale store still has `needsPostMigrationSync` true and the chunked fetch makes the request succeed where the single request failed. The orders show in Settings > Advanced > Channel Orders. No channel is opened: `openChannelWithPaidOrders` reads `cacheStore.paidOrders`, which the wipe reset and this path never repopulates.

No code change here is proportionate. The state predates this PR, master already imports for wallets with 20 or fewer stale IDs, and the only fix that would reach the affected devices is a heuristic that discards migration data on upgrade without being able to tell a legitimately pending sync from a leftover one — which risks dropping a real post-migration pass. The wipe-path fix in this PR stops any new device from entering that state.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `9295975c0`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 9295975c0: the earlier wipe-ordering LOW is fixed. Follow-up at 9295975c0: the two migration steps now sit at WipeWalletUseCase.kt:92-93, directly after lightningRepo.wipeStorage (91) and before keychain.wipe (96), so the earlier LOW (wipe cancelled after the keychain wipe leaving the old wallet's pending order IDs and sync flag for the next wallet) no longer has a window: once the mnemonic is gone the rn_migration store has already been cleared, and the only way it can survive a wiped keychain is a swallowed DataStore failure in a step() at 92-93, which is logged and was equally possible at 42e480427. Stressed the inverse window (cancel or process death after 93, before 96): the wallet keeps its mnemonic and the migration store is gone, but that same window already deletes local LDK storage at 91 on master (LightningService.kt:505-511; the node rebuilds from VSS at LightningService.kt:298), Room, core data, settings and cache are still intact, and what is lost is only the post-migration reapply (metadata tags, transfer/boost markers, old Blocktank order history via reapplyMetadataAfterSync at MigrationService.kt:1671 gated by needsPostMigrationSync at AppViewModel.kt:1433-1437). No fund or channel state depends on the cleared keys: clearPersistedMigrationData (305-319) removes only the five pending-data keys, paid orders were already written to cacheStore during migration (1264-1266, 1612-1615) and cacheStore is not reset until line 101, pendingChannelMigration is in-memory and the channel/peer recovery markers are untouched. The user asked for that data to be destroyed, the window is milliseconds between two local DataStore edits, and the wallet survives only if they then abandon the reset, so this is a smaller trade-off than the original cross-wallet leak and not a finding. The non-cancel failure path is unchanged: every step after 91 swallows, a keychain.wipe failure still returns WipeIncomplete after all other steps ran (109) and WalletRepo.kt:482 restores walletExists. No path re-persists migration data after the clear except the restore/migration flows (MigrationService.kt:495, 1244, 1278-1280, 1334, 1562-1598), which are the explicitly accepted out-of-scope stalled-restore case, so moving the clear earlier does not open a race with an in-flight completeMigration beyond what master already had. The new test `invoke should clear migration data before wiping the keychain` uses inOrder(migrationService, keychain) with setNeedsPostMigrationSync -> cleanupAfterMigration -> keychain.wipe; it fails at 42e480427 (keychain.wipe verified first) and on master (calls absent). No bitkit-ios twin PR found.

ovitrif (approved):
utAck

PR #1270, opened 2026-09-15
fix: polish wallet ui

pwltr:
### Description

This PR polishes the current wallet UI stack:

- Aligns button text styles, small button sizing, scanner/send icon button backgrounds, selected tab indicator color, and money symbol styling.
- Tightens amount screen, transfer flow, wallet activity, invoice tag, receive confirmation, boost sheet, and activity detail spacing/typography.
- Updates onboarding/settings intro copy, translations, sheet intro layout behavior, and restore wallet mnemonic emphasis.
- Keeps a changelog fragment for the wallet polish work and restores the existing incoming payment request fragment from `master`.

### Design

Mapped screen references are listed in `docs/screens-map.md` for the touched screens, including:

- `HomeScreen.kt` → Wallet On-chain › Wallet BTC
- `SavingsWalletScreen.kt` → Savings, Spending, Hardware Balances › Savings Populated
- `SpendingWalletScreen.kt` → Savings, Spending, Hardware Balances › Spending Ready
- `SpendingAmountScreen.kt` / `SpendingAdvancedScreen.kt` / `SpendingConfirmScreen.kt` → Transfer to Spending Balance From Savings frames
- `EditInvoiceScreen.kt` / `ReceiveConfirmScreen.kt` / `ReceiveQrScreen.kt` → receive/send invoice frames
- `ActivityDetailScreen.kt` / `AllActivityScreen.kt` → transaction/activity frames
- `RestoreWalletScreen.kt` / `OnboardingSlidesScreen.kt` / `CreateWalletScreen.kt` → onboarding frames
- `BackupIntroScreen.kt` / `QuickPayIntroScreen.kt` → settings intro frames

Out-of-map sheets and shared components were cross-checked against the existing iOS implementation and user-provided screenshots during implementation.

#### Manual Tests

- Open the wallet home screen and verify the section spacing between balances, activity, and action areas remains visually consistent.
- Open Savings and Spending wallet screens and verify the transfer button has the intended space before the first activity row.
- Walk through the transfer-to-spending amount, advanced, and confirm screens; verify title/amount spacing, number pad spacing, amount symbols, and coin image tilt on slider drag.
- Open receive invoice editing and verify tags sit in the same row as the add tag button, note/add-tag input styling matches, and selected tag spacing is correct.
- Open transaction details and the boost sheet; verify the boost icon, fee amount symbol, long duration text, outline suggested-fee button, and single top toast behavior.
[...]

PR #1308, opened 2026-09-17
fix: avoid animated switch to auto tab

### LINKED ISSUE #876: Tab always transitions Spending → Auto when there are LN channel every time Receive Sheet opens
Can you also please fix the issue with the transition from Spending to Auto always showing when there are lightning channel the moment user opens the Receive Sheet?!
there should be a way to "not transition"/ start on "auto" from the start.

Thx in advance 🙏🏻

_Originally posted by @ovitrif in https://github.com/synonymdev/bitkit-android/issues/875#issuecomment-4157730696_

jvsena42:
Fixes #876
Related to #1275

This PR:
1. Stops the Receive sheet from overriding a tab the user picked when the Auto tab becomes available
2. Replaces the animated slide to Auto with an instant jump when the user has not touched the tabs
3. Clears a leftover Edit Invoice amount before the Receive sheet opens, so it no longer starts on Savings and then jumps to Auto

### Description

- Tracks whether the user has picked a tab by tap or by dragging the QR pager, and skips the switch to Auto once they have, because the switch used to fire whenever Lightning became ready and took the user off the tab they chose.
- Skips the switch when an initial tab was requested or Auto is already selected, and moves the decision into a small pure function so it can be unit tested.
- Uses an instant scroll instead of an animated one when the sheet still moves to Auto after opening, so it no longer visibly slides across tabs.
- Clears the Receive amount when the Receive sheet is requested, before it is shown, because an amount above inbound capacity left over from an earlier Edit Invoice session made the next open start on Savings and then jump to Auto once the receive state refreshed.

On a warm open with the node running, master already starts on Auto from the first frame on the emulator; the slide on open did not reproduce in that case. On a cold start, master opens on Savings while the node starts, then moves to Auto and overrides a tab the user picked in the meantime. That override is the part reproduced and fixed here.

#### Out of Scope

- `ReceiveQrScreen.kt`: when the sheet is opened while the Lightning node is still starting, it still opens on Savings (no Auto tab yet) and then jumps to Auto once Lightning is ready. The jump is now instant, but the Auto QR needs a bolt11 from the running node, so it cannot be shown from the first frame.
- `ReceiveQrScreen.kt`: the user-selection flag records that the tabs were touched, not which tab was picked. If the user stays on Auto, Lightning drops (node restart) and comes back, the sheet stays on Savings instead of returning to Auto.
- `AppViewModel.kt`: only the leftover amount is cleared on open; the old bip21/bolt11 strings are replaced when the receive state refresh finishes, as before.
- The CJIT switch to Spending and the other tab effects are unchanged.
- iOS parity: not checked in this PR; bitkit-ios may need the same guard if its Receive sheet auto-switches to Auto.
- #1231 renames the Trezor tab in `ReceiveQrScreen.kt` and #1270 edits the tab row there; expect a small text conflict with whichever lands first.

### Design

- [Receive Lightning Active › Receive Auto (Unified QR)](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-146064)

### Preview

Recorded on a Pixel 9 emulator with a paired hardware wallet (hence the extra Trezor tab). Differences from Figma that already exist on master: the Trezor tab, no contacts icon in the top bar, and no "Scan or tap to pay" hint.

| Figma | Before (master) | After |
| - | - | - |
| Figma Receive Auto | Before: user picked Savings, moved to Auto | After: user picked Savings, stays on Savings |

Before (master): cold start, tap Trezor then Savings before the Auto tab exists; when Auto appears the sheet moves to it.

https://github.com/user-attachments/assets/f39d8edc-dd48-414f-9b51-3a59dbabd47d

After: same flow; Savings stays selected after the Auto tab appears.

https://github.com/user-attachments/assets/ce8a391f-aae1-4a5b-8600-2e7066284b19

Before (master): cold start, Receive opened without touching the tabs; opens on Savings, then moves to Auto once Lightning is ready.

https://github.com/user-attachments/assets/d99bd715-e658-40cd-8b3e-fe78a3f618ad

After: same flow; the move to Auto is an instant jump with no slide frames.

https://github.com/user-attachments/assets/ff013fd1-f32f-429a-9043-58df791b46b7

After: five warm opens, each on Auto from the first frame.

https://github.com/user-attachments/assets/69484f76-0f59-42dd-90c3-0f1dfda127f1

After: warm open on Auto, swipe to Savings; Savings stays selected.

https://github.com/user-attachments/assets/f953c163-7f70-4b19-9514-f5171883dcfa

After: reopening Receive after a $3,000 Edit Invoice amount (above inbound capacity) opens directly on Auto.

https://github.com/user-attachments/assets/c911f138-e473-4621-b964-21c7f59c6047

#### Manual Tests

Needs a wallet with a usable spending channel.

- [x] **1.** Node running → Home → Receive: Auto tab selected on the first frame, no slide.
- [x] **2.** Receive → swipe QR to Savings → wait a few seconds: Savings stays selected.
- [x] **3a.** Force-stop app → relaunch → Receive right away (no Auto tab yet) → tap Savings → wait until the Auto tab appears: Savings stays selected.
  - [x] **3b.** Force-stop app → relaunch → Receive right away without touching tabs → wait until the Auto tab appears: sheet jumps to Auto without sliding.
- [x] **4.** Receive → Edit → amount above inbound LN capacity → Continue → QR Code (no Auto tab) → close → Receive again: opens on Auto.
- [x] **5.** `regression:` No channels → Receive → Edit → amount → CJIT flow: still switches to Spending with the CJIT invoice.

A journey covering tests 1 to 4 is committed at `journeys/receive/receive-auto-tab-selection.xml`;
the same patch inline:

<details>
<summary>receive-auto-tab-selection.xml</summary>

```diff
diff --git a/journeys/receive/receive-auto-tab-selection.xml b/journeys/receive/receive-auto-tab-selection.xml
new file mode 100644
index 000000000..c70f76736
--- /dev/null
+++ b/journeys/receive/receive-auto-tab-selection.xml
@@ -0,0 +1,87 @@
+<journey name="receive auto tab selection">
+  <description>
+    Verifies that the Receive sheet opens on the Auto tab without sliding to it, that a tab the user
+    picks is not overridden by the Auto tab, and that a leftover invoice amount from an earlier Edit
+    Invoice session does not open the sheet on Savings. Requires a wallet with a usable spending
+    channel so the Auto tab is available once the Lightning node is running. The selected tab is
+    shown only by its underline, which `android layout` does not expose, so tab assertions need a
+    screenshot; the Auto tab's presence can be read from the testTag "Tab-auto". The underline itself
+    crossfades over 200 ms even when the switch is instant, so the two steps that check for an
+    animation read the QR pager out of a screen recording instead of the underline. Each of those
+    steps first requires the recording to contain the frames it reasons about; a recording that ended
+    before the node started, or before the sheet opened, is a rerun with a longer `--time-limit`, not
+    a pass.
+  </description>
+  <actions>
+    <action>
+      Launch the Bitkit app and go to the wallet home screen
+    </action>
+    <action>
+      Tap the Receive button (testTag "Receive"), take a screenshot within one second, and verify the
+      Auto tab (testTag "Tab-auto") is selected on the first frame of the sheet
+    </action>
+    <action>
+      Swipe the QR code area from left to right once and verify the Savings tab (testTag
+      "Tab-savings") is selected
+    </action>
+    <action>
+      Wait 3 seconds, take a screenshot and verify the Savings tab is still selected
+    </action>
+    <action>
+      Close the sheet, then run `adb shell am force-stop to.bitkit.dev` and launch the app again
+    </action>
+    <action>
+      As soon as the home screen shows, tap the Receive button and verify the tab row has no Auto tab
+      (testTag "Tab-auto" absent) while the Lightning node starts
+    </action>
+    <action>
+      Tap the Trezor tab (testTag "Tab-trezor") if present, then tap the Savings tab (testTag
+      "Tab-savings")
+    </action>
+    <action>
+      Wait until the Auto tab (testTag "Tab-auto") appears, wait 5 more seconds, take a screenshot and
+      verify the Savings tab is still selected
+    </action>
+    <action>
+      Close the sheet, force-stop and relaunch the app, and start `adb shell screenrecord
+      --time-limit 20 /sdcard/receive-auto.mp4`. As soon as the home screen shows, tap the Receive
+      button without touching the tabs and verify the Auto tab (testTag "Tab-auto") is absent; do not
+      wait for the node, the tab row must still be Savings-first when the sheet opens
+    </action>
+    <action>
+      Wait until the Auto tab (testTag "Tab-auto") appears, and check that it appeared while the
+      recording was still running. The node has to finish `lightningService.start` and load its
+      channels before the Auto tab exists, which on a cold process can outlast the 20 second window;
+      if the Auto tab is still absent when the recording ends, the jump happened off-tape and the run
+      holds no evidence, so repeat the previous step with a longer `--time-limit` rather than reading
+      the empty recording as a pass
+    </action>
+    <action>
+      Let the recording finish, pull it (`adb pull /sdcard/receive-auto.mp4`), extract frames (`ffmpeg
+      -i receive-auto.mp4 -vf fps=30 frames/%03d.png`) and verify first that the recording contains
+      the jump at all: a frame showing the Savings page centred in the QR pager, followed by a later
+      frame showing the Auto page centred. If either frame is missing, the step did not capture the
+      switch and has to be rerun; it is not a pass. Then verify that no frame between those two shows
+      the QR pager part-way between the Savings and the Auto page: the sheet jumps to Auto in one
+      frame instead of sliding
+    </action>
+    <action>
+      Tap "Edit" (testTag "SpecifyInvoiceButton"), tap the amount field (testTag
+      "ReceiveNumberPadTextField"), enter an amount above the inbound Lightning capacity (for example
+      N3 N0 N0 N0 in USD), tap Continue (testTag "ReceiveNumberPadSubmit") and then "QR Code" (testTag
+      "ShowQrReceive"); verify the Auto tab (testTag "Tab-auto") is absent
+    </action>
+    <action>
+      Start `adb shell screenrecord --time-limit 20 /sdcard/receive-reopen.mp4`, close the sheet and
+      tap the Receive button again, then verify the Auto tab (testTag "Tab-auto") is present
+    </action>
+    <action>
+      Let the recording finish, pull it (`adb pull /sdcard/receive-reopen.mp4`), extract frames
+      (`ffmpeg -i receive-reopen.mp4 -vf fps=30 frames/%03d.png`) and verify that the recording holds
+      frames in which the sheet is visible — if it holds none, rerun the step, it is not a pass — then
+      that the first such frame already shows the Auto tab in the tab row with the selected underline,
+      and that no frame shows Savings selected. A sheet that opens on Savings and only then gains the
+      Auto tab means the leftover invoice amount was not cleared
+    </action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `ReceiveAutoTabSwitchTest.kt` covers when the sheet should switch to Auto (initial tab, user selection, Auto already selected, Lightning availability, CJIT invoice).
- Unit tests added: `AppViewModelSendFlowTest.kt` checks that opening Receive clears the leftover amount before the sheet is set, and that opening another sheet keeps it.
- Compose test added: `ReceiveAutoTabSelectionTest.kt` renders `ReceiveQrScreen`, flips Lightning to ready, and checks the tab row wiring: a tab picked by tap survives the Auto tab appearing, and with no tap the sheet lands on Auto.
- Local: `just compile`, `just test`, `just lint` and `compileDevDebugAndroidTestKotlin` passed; the instrumented test itself runs in the `UI Tests` workflow.


app/src/test/java/to/bitkit/ui/screens/wallets/receive/ReceiveAutoTabSwitchTest.kt
```diff
@@ -0,0 +1,57 @@
+package to.bitkit.ui.screens.wallets.receive
+
+import org.junit.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class ReceiveAutoTabSwitchTest {
+
+    @Test
```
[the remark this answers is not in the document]
jvsena42: Fair point, the predicate tests alone would stay green if the wiring stopped feeding it. Added `ReceiveAutoTabSelectionTest.kt` in `androidTest` (fbc1dc2), next to `ReceiveHardwareFlowTest.kt`: it renders `ReceiveQrScreen` with the node running and no channels, then flips `LightningState` to a ready channel with inbound capacity. One case taps Tab-spending then Tab-savings and asserts the sheet is still on Savings after the Auto tab appears; the other touches nothing and asserts the sheet ends on Auto. The selected page is read back through `onClickEditInvoice`, since the selection is only drawn as an underline.

Left uncovered on purpose: the pager-drag path and the "no animation" part. A swipe assertion here depends on snap-fling settling and on which neighbouring page the LazyRow has composed, which is flaky in CI, and the drag only sets the same flag the tap sets before the same predicate. The instant-jump behaviour stays covered by the recorded videos in the PR body and by the journey in the QA notes.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `fbc1dc281`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Traced the tab state machine in ReceiveQrScreen.kt at the PR head (effects at lines 194, 213, 221, 232, 239, 255) for cold open (no Auto tab yet), warm open (Auto from the first frame), tap and drag before Auto appears, Auto appearing with a Trezor tab in front (LazyRow keys by tab name so the Savings index is preserved and the jump targets the right index), channel drop and return, CJIT confirm, and return from Edit Invoice or the Paykit contact picker (fresh composition, initialTab set by ReceiveInvoiceEditState so the auto-switch is skipped exactly as on master). hasUserSelectedTab is only set by the tab row click (CustomTabRowWithSpacing invokes onTabChange solely from clickableAlpha) and by DragInteraction.Start on the LazyRow interaction source, which programmatic scrollToItem/animateScrollToItem never emit, so the guard cannot be tripped by the screen's own scrolls. The early return on firstVisibleItemIndex == autoIndex cannot suppress a needed jump: the only way to sit on a non-Auto page equal to the new Auto index without a user pick is a CJIT invoice, which hides the Auto tab. invoicesByTab is unchanged, so every page still renders the URI for its own tab; the transient on-chain-only URI under Auto while the old bip21 string waits for refreshBip21 (behind syncState and the geoblock check in WalletViewModel.refreshReceiveState) is the limitation the author already listed as out of scope and existed on master on the Savings page. The showSheet change in AppViewModel.kt:4727 is an in-memory StateFlow update after the transition delay and before the sheet is published; no caller sets bip21AmountSats before showSheet(Sheet.Receive) (all writers are inside the open sheet), cancel by a competing showSheet, hideSheet, or process death leaves nothing half-applied, and refreshBip21 clears the same field on open anyway, so Paykit request routes and the ReceiveRoute.Amount funding entry are unaffected.

journeys/receive/receive-auto-tab-selection.xml
```diff
@@ -0,0 +1,57 @@
...
+    <action>
+      Tap the Trezor tab (testTag "Tab-trezor") if present, then tap the Savings tab (testTag
+      "Tab-savings")
+    </action>
+    <action>
+      Wait until the Auto tab (testTag "Tab-auto") appears, wait 5 more seconds, take a screenshot and
+      verify the Savings tab is still selected
+    </action>
+    <action>
+      Close the sheet, force-stop and relaunch the app, wait for the node to start, and tap the Receive
```
jvsena42: **MEDIUM, journey, non-blocking** — **This step waits for the node to start before tapping Receive. By then the Auto tab already exists, so the sheet opens on Auto and the "Savings tab is selected first" check fails every time. The jump this step is meant to test never happens.**

`LightningRepo.start` sets `NodeLifecycleState.Running` and calls `syncState()` in the same coroutine with no suspension in between (LightningRepo.kt:390-395). `syncState()` loads the persisted channels, and `isChannelReady` does not need a peer connection. So by the time a runner can see that the node is running, `canCreateLightningInvoice` (ReceiveQrScreen.kt:118) is already true. The first composition then includes AUTO in `visibleTabs`, and `defaultTab` resolves to AUTO (ReceiveQrScreen.kt:144). That is the warm open from step 2. The only window with Savings and no Auto tab is the one step 6 uses: tap Receive before the node reaches Running.

"Record the screen to check" also has no tool and no pass criterion. The README documents screenshots and `android layout` only. The tab underline crossfades over 200 ms (`animateColorAsState`, CustomTabRowWithSpacing.kt:65) even after an instant `scrollToItem`, so the underline alone does not show whether the pager slid. The QR pager does: on master, `animateScrollToItem` leaves frames where the pager sits between the Savings and Auto pages.

Fix: "Close the sheet, force-stop and relaunch the app, start `adb shell screenrecord --time-limit 20 /sdcard/receive-auto.mp4`, and as soon as the home screen shows, tap the Receive button without touching the tabs. Verify Tab-auto is absent. Wait until Tab-auto appears, pull the recording, extract frames (`ffmpeg -i receive-auto.mp4 -vf fps=30 frames/%03d.png`) and verify that no frame shows the QR pager part-way between the Savings and Auto pages."

journeys/receive/receive-auto-tab-selection.xml
```diff
@@ -0,0 +1,57 @@
...
+    </action>
+    <action>
+      Tap "Edit" (testTag "SpecifyInvoiceButton"), tap the amount field (testTag
+      "ReceiveNumberPadTextField"), enter an amount above the inbound Lightning capacity (for example
+      N3 N0 N0 N0 in USD), tap Continue (testTag "ReceiveNumberPadSubmit") and then "QR Code" (testTag
+      "ShowQrReceive"); verify the Auto tab (testTag "Tab-auto") is absent
+    </action>
+    <action>
+      Close the sheet, tap the Receive button again and verify the Auto tab (testTag "Tab-auto") is
+      present and selected on the first frame of the sheet
```
jvsena42: **LOW, journey, non-blocking** — **Step 11 checks "selected on the first frame" with a single screenshot. It still passes if the `bip21AmountSats` clear in `AppViewModel.showSheet` is reverted.**

Without the clear, the reopened sheet starts with the leftover amount from step 10. That amount is above inbound capacity, so there is no Auto tab and the sheet opens on Savings. `WalletViewModel.refreshReceiveState` then runs `syncState()`, `updateGeoBlockState()` and `refreshBip21()`, and `refreshBip21()` clears the amount through `clearBip21State` (WalletRepo.kt:219). The Auto tab appears and, with no user pick, the sheet jumps to Auto. This PR made that jump an instant `scrollToItem`. A screenshot taken after the tap and capture latency shows Auto selected whether or not the fix is present. Step 2 has the same weakness but does not guard any change in this PR.

Fix: in step 11, start `adb shell screenrecord` before tapping Receive. Extract the frames, and verify that Tab-auto is present and selected in the first frame where the sheet is visible.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `007e933a6`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: MEDIUM (journey, non-blocking) — Instant-jump step waits for the node first, so Savings never shows and the jump cannot be observed; LOW (journey, non-blocking) — First-frame checks rely on one screenshot and pass if the leftover-amount fix is reverted

Checked: Round 4 (7ae8457ea, new journey): Reviewed the journey commit 7ae8457ea only. Every testTag it names exists at head: Receive (TabBar.kt:156), Tab-auto/Tab-savings/Tab-trezor (CustomTabRowWithSpacing.kt:62, `Tab-${tab.name.lowercase()}`), SpecifyInvoiceButton, ReceiveNumberPadTextField, ReceiveNumberPadSubmit, ShowQrReceive and the N<digit> keys. The Edit Invoice route in the last two steps matches the code: source AUTO always yields ReceiveAdditionalLiquidityAction.None (ReceiveLiquidityDecision.kt:42), so Show QR updates the invoice and pops back to the QR screen with no Auto tab, and reopening hits the new bip21AmountSats clear in AppViewModel.showSheet. The cold-start window with no Auto tab is real: channels are only populated by syncState() after lightningService.start returns (LightningRepo.kt:390-395). The README index row and the cross-platform not-ported row are in place.

journeys/receive/receive-auto-tab-selection.xml
jvsena42: Confirmed, the step could not fail. `LightningRepo.start` sets `NodeLifecycleState.Running` at LightningRepo.kt:392 and calls `syncState()` at LightningRepo.kt:395 in the same coroutine with nothing suspending in between, so the persisted ready channels are loaded by the time the state is observably Running. `canCreateLightningInvoice` (ReceiveQrScreen.kt:118-130) is then true on the first composition, `visibleTabs` includes AUTO (ReceiveQrScreen.kt:132-143) and `defaultTab` resolves to AUTO (ReceiveQrScreen.kt:144-146). Waiting for the node before tapping Receive is the warm open of step 2, not the asynchronous flip.

The underline point holds too: the indicator colour is an `animateColorAsState` 200 ms tween at CustomTabRowWithSpacing.kt:65-72, so it fades even when the scroll is instant.

Changed in 049c57fdf: the step no longer waits for the node. It now starts `adb shell screenrecord` before the relaunch, taps Receive as soon as the home screen shows and asserts `Tab-auto` is absent at that point, and a following step pulls the recording, extracts frames with ffmpeg and checks that no frame shows the QR pager part-way between the Savings and the Auto page. The description says why the pager is read instead of the underline.

journeys/receive/receive-auto-tab-selection.xml
jvsena42: Confirmed. Without the clear in `AppViewModel.showSheet`, `WalletViewModel.refreshReceiveState` (WalletViewModel.kt:536-541) runs `syncState()`, `updateGeoBlockState()` and `refreshBip21()`, and `refreshBip21` calls `clearBip21State` at WalletRepo.kt:219, which nulls `bip21AmountSats` (WalletRepo.kt:652-661). The Auto tab therefore appears either way, and with no user pick the sheet lands on Auto, so a screenshot taken after the tap and capture latency passes with or without the fix.

Changed in 049c57fdf: step 11 is now two steps. The first starts `adb shell screenrecord`, reopens the sheet and asserts `Tab-auto` is present live. The second pulls the recording, extracts frames and requires that the first frame in which the sheet is visible already shows Auto selected and that no frame shows Savings selected, which is exactly the state the missing clear would produce.

Step 2 keeps its single screenshot as you noted; it guards nothing in this PR, and the frame check added here covers the same path from a cold state.

journeys/receive/receive-auto-tab-selection.xml
jvsena42: **LOW (journey, non-blocking)** — **The cold-start frame check only asserts that no frame is part-way between pages. Nothing requires the recording to contain the jump, so if the node reaches Running after the recording stops, zero frames show the pager moving and the step passes on master too.**

The step at line 43 starts `screenrecord --time-limit 20` after the relaunch, and step at line 49 says "Wait until Tab-auto appears, let the recording finish". Tab-auto appears only after `lightningService.start` returns and `syncState()` loads channels (LightningRepo.kt:390-395), which includes node setup and the VSS-backed build on a cold process. When that takes longer than the time remaining in the 20 s window, the jump happens off-tape, and "no frame shows the QR pager part-way" is trivially true. My earlier suggested fix had the same gap.

Fix: add a positive check to line 49-52, e.g. "verify that the recording contains a frame with the Savings page centred followed by a frame with the Auto page centred; if the last frame still shows Savings, rerun with a longer `--time-limit`".

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `049c57fdf`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up on 049c57fdf: the journey MEDIUM and the first-frame LOW are fixed. One remaining LOW (nothing asserts the recording actually contains the jump) is a reply on the existing thread. Reviewed 049c57fdf only. The cold-start step no longer waits for the node: it taps Receive as soon as home shows and asserts Tab-auto absent, which is the reachable window before `LightningRepo.start` sets Running and runs `syncState()`. AUTO is inserted after SAVINGS in `visibleTabs` (ReceiveQrScreen.kt:132-143), so Savings stays index 0 and the jump targets index 1; on master `animateScrollToItem` leaves intermediate pager frames that the ffmpeg frame check catches. The reopen step now records and requires the first visible sheet frame to already show Auto selected and no Savings frame, which fails if the `bip21AmountSats` clear in `showSheet` is reverted; the underline has no initial crossfade on a first composition, so that check is sound.

journeys/receive/receive-auto-tab-selection.xml
jvsena42: Confirmed. At 049c57fdf the cold-start check was negative only — "no frame shows the QR pager part-way" — and nothing required the recording to hold the switch. `LightningRepo.start` calls `lightningService.start` at LightningRepo.kt:390 and only then sets Running (:392) and runs `syncState()` (:395), so `canCreateLightningInvoice` (ReceiveQrScreen.kt:118) flips, AUTO enters `visibleTabs` (ReceiveQrScreen.kt:132-143) and the jump at ReceiveQrScreen.kt:250 fires only after node setup finishes. On a cold process that can outlast the 20 s window, and with zero frames of the transition the assertion holds on master too, where the line was `animateScrollToItem`.

Changed in c5b492353: the step is split. The first half now waits for Tab-auto and requires it to appear while the recording is still running — if the tab is still absent when the recording ends, the jump happened off-tape and the step is rerun with a longer `--time-limit`, not passed. The second half requires the extracted frames to contain the switch before reasoning about it: a frame with the Savings page centred in the QR pager followed by a later frame with the Auto page centred, and only then the part-way check between those two. A missing frame is an explicit rerun.

The reopen check already carried a positive requirement, so it was not vacuous in the same way, but it had the same edge if the recording held no frame of the sheet at all; it now rejects that case too. The description says both frame checks require their evidence. XML parses (`xmllint --noout`), and the PR body diff is in step with the committed file.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `c5b492353`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up: Follow-up on c5b492353, which answers the LOW that nothing required the recording to contain the jump. The cold-start step is now split into two. The first requires Tab-auto to appear while the recording is still running, and tells the runner to rerun with a longer `--time-limit` otherwise. The second requires the extracted frames to hold a Savings-centred frame followed by a later Auto-centred frame before it runs the part-way check, and treats a missing frame as a rerun, not a pass. This is the positive requirement asked for. It matches the code: AUTO is inserted after SAVINGS, and the `scrollToItem(autoIndex)` jump in ReceiveQrScreen.kt:239-251 fires only after the node is ready. The reopen step now also rejects a recording with no sheet frames. Journey-only commit.

ovitrif:
@jvsena42 conflicts, I am approving

ovitrif (dismissed):
utAck

ovitrif (approved):

PR #1288, opened 2026-09-16
fix: add retry to coin selection loading

jvsena42:
Fixes #616
Related to #1275

This PR adds a bounded retry and an inline error with a Retry button to the coin selection load, and stops activity updates from reloading the list.

### Description

- Retries loading spendable outputs up to 3 times (1s, then 2s apart) when the node is not set up or not running, so a load that races the node start does not fail on the first attempt.
- Leaves other failures unretried, including the node run timeout (the repo already waits up to 1 minute for the node) and fee calculation errors, so the worst-case wait stays the same as before.
- Replaces the raw English error toast with an inline, localized error message and a Retry button when no coins could be loaded; the message stays visible and the button shows its loading state while a retry runs.
- Centers the error message and its Retry button vertically in the space the UTXO list would occupy, instead of pinning them under the title.
- Shows a spinner while the first load is in progress and nothing is listed yet, centered in that same space so it does not jump when the error replaces it.
- Loads UTXOs once per required amount and address, and passes activity updates to the view model separately, so a new activity no longer reloads the list or resets a manual selection.
- Adds a changelog fragment.

The race that originally caused the "Failed to load utxos" toast was fixed by c32193f13 and 27d4f0cfb; #616 is now about the error and retry handling, and could be retitled to match.

#### Out of Scope

- `SendCoinSelectionScreen.kt`: going back from Send Confirm to Coin Selection still re-runs the load. A successful reload replaces a manual selection with all UTXOs, and a failed reload with coins already listed leaves the stale list without an error, because the error block only renders on an empty list. The same applies to a toggle made during the up-to-3s retry window. Deferred to a follow-up.
- `SendCoinSelectionViewModel.kt`: `ServiceError.NodeNotStarted` and `NodeRunTimeoutError` are not retried.
- `LightningRepo`: it already logs each failed `listSpendableOutputs` attempt, so a persistent failure logs once per attempt plus the final view model error. Repo logging is unchanged.
- `changelog.d/next/616.fixed.md`: the fragment says coin selection "retries while the Lightning node is starting", but the retry covers the node not being set up or not running. A node that is still starting is already handled by the repo's wait. The wording can be tightened before release.
- Pre-existing differences from the Figma frame (no ₿ prefix on amounts, no divider above the first row, no inline tag chips) are left as is.
[...]

PR #1309, opened 2026-09-17
fix: fit long mnemonic words on one line

### LINKED ISSUE #633: [Bug]: GUI: Show seed phrase screen breaks line on long word
In particular the word `awesome`. 

App prevents me from screenshotting.

jvsena42:
Fixes #633
Related to #1275

This PR fits long recovery phrase words on one line in the mnemonic grid instead of breaking them mid-word.

### Description

- Shrinks all words in the grid to one shared font size, from 17sp down to 12sp in 0.5sp steps, so that long words such as `mushroom` or `mountain` next to the two-digit labels no longer wrap at larger font scales.
- Keeps the number labels at full size and aligns labels and words by baseline, so rows stay aligned when words render smaller.
- Computes the size for both the revealed words and the hidden placeholders and uses the smaller one, so the card does not change height when the phrase is revealed.
- Falls back to the previous wrapping behaviour when a word does not fit even at 12sp (for example font scale 2.0), so words stay fully visible inside their column instead of overflowing or being clipped.
- Leaves the default font scale unchanged: short words keep 17sp and the grid matches the design.
- Applies to both the backup and the recovery mnemonic screens through the shared grid component.

#### Out of Scope

- `MnemonicWordsGrid.kt`: no shrinking below 12sp; at extreme font scales words still wrap inside their column.
- `MnemonicWordsGrid.kt`: no Compose UI instrumented test for the composable wiring; sizing logic is covered by unit tests and was checked on device.
- `MnemonicWordsGrid.kt`: font size measurement runs during composition; not optimised for the worst case (24 words at large font scales).
- iOS parity: check whether `bitkit-ios` wraps long mnemonic words at larger text sizes; not addressed here.

### Design

- [Back up wallet › Backup Recovery Phrase Displayed](https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-147925)

The design only shows short words at the default font scale. At font scale 1.0 the build matches it.

### Preview

The screen sets `FLAG_SECURE`, so these captures come from temporary uncommitted builds with screenshot blocking removed and BIP39 test words (`abstract awesome category mushroom mosquito document multiply mechanic marriage mountain awesome mushroom`).

| Figma | Before, font scale 1.3 | After, font scale 1.3 |
| - | - | - |
| Figma Backup Recovery Phrase Displayed | Before font scale 1.3 revealed | After font scale 1.3 revealed |

| Before, font scale 1.3, hidden | After, font scale 1.3, hidden |
| - | - |
| Before font scale 1.3 hidden | After font scale 1.3 hidden |

| Before, font scale 1.0 | After, font scale 1.0 | After, font scale 2.0 (wrap fallback) |
| - | - | - |
| Before font scale 1.0 revealed | After font scale 1.0 revealed | After font scale 2.0 revealed |

Reveal transition at font scale 1.3, no layout jump:

https://github.com/user-attachments/assets/3d46ff76-0aa4-4ddb-9435-2d082d497dbd

#### Manual Tests

- [x] **1a.** System font scale 1.3 → Settings → Security → Back up wallet → Show Mnemonic → Tap To Reveal: every word stays on one line, including words next to 10., 11. and 12.
  - [x] **1b.** Hidden vs revealed: card height does not change on reveal.
- [x] **2.** `regression:` System font scale 1.0 → Show Mnemonic → reveal: words render at the default size, matching the design.
- [x] **3.** System font scale 2.0 → Show Mnemonic → reveal: words too wide for the column wrap inside it, none overflow into the next column or get clipped.
- [x] **4.** `regression:` Recovery → Recovery Mnemonic: phrase grid renders correctly.

The steps of test 1 were run on a Pixel 9 emulator with `journeys/backup/show-mnemonic-long-words.xml`, committed in this PR and listed in `journeys/README.md`. The journey restores a fixed public BIP39 test phrase first, so the long words it checks are guaranteed to be there rather than left to whatever phrase the device happens to hold, and it wipes the app to do so — run it on a throwaway emulator. Its contents are inlined below.

<details><summary>show-mnemonic-long-words.xml</summary>

```diff
diff --git a/journeys/backup/show-mnemonic-long-words.xml b/journeys/backup/show-mnemonic-long-words.xml
new file mode 100644
index 000000000..259422230
--- /dev/null
+++ b/journeys/backup/show-mnemonic-long-words.xml
@@ -0,0 +1,54 @@
+<journey name="show mnemonic long words">
+  <description>
+    Verifies that every recovery phrase word on the backup screen renders on a single line at a
+    larger system font scale, including words next to the two-digit labels 10., 11. and 12.
+    (synonymdev/bitkit-android#633). The grid shrinks all words to one shared size, from 17sp down
+    to 12sp; only a word that is still too wide at 12sp wraps, which happens at extreme font scales
+    such as 2.0.
+
+    The screen sets FLAG_SECURE, so screenshots and recordings come out black, and the SeedContainer
+    content description holds the phrase. Never print, log, save or quote the words: the word
+    elements carry no testTag, so read rows from the number labels ("1." to "12.") and from the
+    bounds of the word elements, never their text or content-desc.
+
+    Precondition: the phrase must be known to contain long words, or the check passes for the wrong
+    reason. A random 12-word phrase has an 8-letter word next to a two-digit label only about one
+    time in eight, and on an unfixed build every other phrase renders on one line anyway. So the
+    journey restores a fixed phrase rather than revealing whatever the device already holds:
+
+      abstract awesome category mushroom mosquito document multiply mechanic marriage mountain
+      awesome mushroom
+
+    That is a public BIP39 test vector, not anyone's wallet. Its checksum is valid, so it restores.
+    It puts 8-letter words in slots 10 and 12, the two-digit positions this journey is about. It is
+    public, so anyone can spend from it: dev flavor (to.bitkit.dev, regtest) only, and never send it
+    funds.
+
+    Restoring replaces the wallet on the device — run this on a throwaway emulator, or on a device
+    whose wallet you are willing to lose. Steps 1-8 do the restore; skip them only if the device
+    already holds this exact phrase. No PIN is set after a fresh restore; if one is set, enter the
+    correct PIN when prompted and never a wrong one. Restore font_scale to 1.0 when done.
+  </description>
+  <actions>
+    <action>Run adb shell pm clear to.bitkit.dev</action>
+    <action>Run adb shell monkey -p to.bitkit.dev -c android.intent.category.LAUNCHER 1</action>
+    <action>On the terms screen, tap both checkboxes (testTags "Check1" and "Check2"), then tap "Continue" (testTag "Continue")</action>
+    <action>Tap "Skip" (testTag "SkipIntro") to reach the last onboarding slide</action>
+    <action>Tap "Restore" (testTag "RestoreWallet"), then confirm the multiple-devices warning (testTag "MultipleDevices-button")</action>
+    <action>Enter the 12 words of the phrase above one field at a time, tapping each field (testTags "Word-0" through "Word-11") and typing only that one word — never paste or type the whole phrase in one go, adb drops characters from long strings</action>
+    <action>Verify no field is marked invalid and no checksum error is shown, then tap "Restore" (testTag "RestoreButton")</action>
+    <action>Wait for the restore to finish and tap "Get Started" (testTag "GetStartedButton"); if the backup restore fails instead, the failure screen offers only "Try Again" (testTag "TryAgainButton") — tap it until "Proceed Without Backup" (testTag "ProceedWithoutBackupButton") also appears, which takes two taps, then tap it and confirm with "Yes, Proceed" (testTag "DialogConfirm") — the phrase has no backup and the on-chain wallet is what this journey needs</action>
+    <action>Run adb shell settings put system font_scale 1.3</action>
+    <action>Tap the menu icon (testTag "HeaderMenu")</action>
+    <action>Tap "Settings" (testTag "DrawerSettings")</action>
+    <action>Tap the Security tab (testTag "Tab-security")</action>
+    <action>Tap "Back up your wallet" (testTag "BackupWallet")</action>
+    <action>Verify the recovery phrase screen shows "Tap To Reveal" (testTag "TapToReveal") inside the reveal overlay (testTag "SeedContainer") covering the words box (testTag "backup_mnemonic_words_box")</action>
+    <action>Tap "Tap To Reveal"</action>
+    <action>Verify the words box lists 12 number labels, "1." to "12.", grouped in 6 rows of 2 with the same vertical center per row</action>
+    <action>Verify every word element inside the words box is no taller than the number label beside it. A word element is the element to the right of a number label in the same row; identify it by position and compare bounds heights only, never its text. Labels always render one line at the full size while words shrink, so a word element taller than its label has wrapped onto a second line. This is the check that matters: it catches a wrap in any slot, including 6 and 12</action>
+    <action>Verify the vertical distance between consecutive rows is equal. This is a weaker, secondary check — the two columns lay out independently and the number label sits on the word's first-line baseline, so a wrap in the last slot of a column (6 or 12) shifts nothing below it and leaves every label position unchanged. Do not treat equal pitch on its own as proof that no word wrapped</action>
+    <action>Press back to close the sheet</action>
+    <action>Run adb shell settings put system font_scale 1.0</action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: cover the per-list font size selection (short words keep 17sp, longest word sets the grid size, two-digit labels reduce the width, exact fit at 12sp stays on one line, 1px over 12sp falls back to wrapping), the size shared between the revealed words and the hidden placeholders (either list can set the size, either list failing at 12sp wraps both), and the row width arithmetic in `MnemonicWordsGridTest.kt`.
- Ran locally: `just compile`, `just test` (2608 passed), `just lint`.


app/src/test/java/to/bitkit/ui/components/MnemonicWordsGridTest.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    @Test
+    fun `longest word picks the largest step that fits for the whole grid`() {
+        val result = fitMnemonicFontSize(
+            words = listOf("cat", "abstract", "dog"),
+            wordBudgetPx = { 110 },
+            measureWordPx = measureWord,
+        )
+
+        assertEquals(MnemonicFontFit(13.5.sp, fits = true), result)
```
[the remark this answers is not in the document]
jvsena42: Right, the two-list selection in `rememberWordFontFit` was only reachable from a composable and had no coverage. Extracted it into `fitSharedMnemonicFontSize(wordLists, wordBudgetPx, measureWordPx)` in 39373f24c and added five tests: actual words setting the shared size, placeholders setting it, an actual word that never fits at 12sp, a placeholder that never fits, and the empty-list guard. `just test` passes (2608).

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `39373f24c`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Traced the fit path end to end against Compose 1.12.0 sources (foundation LayoutUtils.finalMaxWidth, ui-text TextMeasurer/AndroidParagraph, material3 Text). On the fits=true path WordItem renders the word with softWrap=false, maxLines=1, overflow=Visible and weight(1f) (MnemonicWordsGrid.kt:226-235); with softWrap=false Compose gives the paragraph width max(ceil(maxIntrinsicWidth), minWidth), so a word is never broken or dropped, and a measurement mismatch could only overflow visibly into the 32dp column gap, never truncate. The measurement in rememberWordFontFit uses the same TextStyle (BodyMSB.copy(fontSize)), the same LocalDensity (font scale included) and the same font resolver as the rendered Text; material3 Text does not merge LocalTextStyle when a style is passed, InterFontFamily is a blocking resource font, so measured widths equal laid-out widths. Budget arithmetic is conservative: (gridWidth - gap)/2 floors while Row weight distribution gives each column floor or ceil, and per-number label widths are measured for 1..24 so two-digit labels in 24-word grids reduce the budget of the right words. On the fits=false fallback (font scale 2.0 or narrower) softWrap=true with maxLines=MAX and overflow=Visible keeps every character inside its column across lines, identical to master's behaviour, and the parent Box clip in ShowMnemonicScreen/RecoveryMnemonicScreen sits 32dp outside the grid. BoxWithConstraints always sees bounded width in both callers (verticalScroll Column, no IntrinsicSize ancestors), remember keys cover words, width and density, and both callers (backup sheet and recovery screen) go through the shared grid.

journeys/backup/show-mnemonic-long-words.xml
```diff
@@ -0,0 +1,31 @@
...
+    (synonymdev/bitkit-android#633). The grid shrinks all words to one shared size, from 17sp down
+    to 12sp; only a word that is still too wide at 12sp wraps, which happens at extreme font scales
+    such as 2.0.
+
+    The screen sets FLAG_SECURE, so screenshots and recordings come out black, and the SeedContainer
+    content description holds the phrase. Never print, log, save or quote the words: the word
+    elements carry no testTag, so read rows from the number labels ("1." to "12.") and from the
+    bounds of the word elements, never their text or content-desc.
+
+    Precondition: onboarded dev wallet with a 12-word phrase and no PIN (with a PIN, enter the
```
jvsena42: **MEDIUM, journey, non-blocking** — **The precondition accepts any 12-word dev wallet, so a phrase without a long word passes on master too.**

The journey never restores a known phrase. It reveals whatever mnemonic the current wallet has. On master, `WordItem` wraps only when a word is wider than its column at font scale 1.3, and the PR names 8-letter words such as `mushroom` and `mountain` next to the two-digit labels as the failing case. Taking 8 letters as the threshold, 88 of the 2048 BIP39 words qualify, so a random phrase has one in slots 10-12 in only about 12% of wallets. For every other wallet the equal-pitch check at lines 26-27 also holds on master, and the agent reports a pass on the unfixed build. The agent cannot tell which case it is in, because the description forbids it from reading the words.

Fix: require the phrase in the precondition. Restore the dev wallet from the phrase already in the PR body (`abstract awesome category mushroom mosquito document multiply mechanic marriage mountain awesome mushroom`) and label it a public test vector in the description (regtest only, never fund). Its BIP39 checksum is valid, so it restores. It puts long words at 10 and 12.

journeys/backup/show-mnemonic-long-words.xml
```diff
@@ -0,0 +1,31 @@
...
+  <actions>
+    <action>Run adb shell settings put system font_scale 1.3</action>
+    <action>Tap the menu icon (testTag "HeaderMenu")</action>
+    <action>Tap "Settings" (testTag "DrawerSettings")</action>
+    <action>Tap the Security tab (testTag "Tab-security")</action>
+    <action>Tap "Back up your wallet" (testTag "BackupWallet")</action>
+    <action>Verify the recovery phrase screen shows "Tap To Reveal" (testTag "TapToReveal") inside the reveal overlay (testTag "SeedContainer") covering the words box (testTag "backup_mnemonic_words_box")</action>
+    <action>Tap "Tap To Reveal"</action>
+    <action>Verify the words box lists 12 number labels, "1." to "12.", grouped in 6 rows of 2 with the same vertical center per row</action>
+    <action>Verify the vertical distance between consecutive rows is equal, so no word wraps onto a second line</action>
```
jvsena42: **LOW, journey, non-blocking** — **At head, a wrapped word in the last slot of either column leaves every label position unchanged, so lines 26-27 still pass.**

`MnemonicWordsGrid` lays out two independent `Column`s, 1-6 and 7-12 (MnemonicWordsGrid.kt:88-111). The PR switched `WordItem` from `CenterVertically` to `alignByBaseline()` (MnemonicWordsGrid.kt:223/234), so a label now sits on the first line's baseline. If word 6 or word 12 wraps, its item only grows downward and nothing below it shifts, so the same-center-per-row check and the equal-pitch check both pass. A future fit bug that makes word 12 wrap would go unnoticed, and 12 is one of the two-digit slots this journey covers. On master the label is vertically centred, so a full revert still moves label 12 and is caught.

Fix: add one verify step on the word elements' bounds, which the description already allows. All 12 word elements inside `backup_mnemonic_words_box` should have the same height, one line. That catches a wrap in any slot without reading the text.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `535f6c022`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: MEDIUM (journey, non-blocking) — Journey runs on whatever phrase the wallet has, so it can pass with the bug reverted; LOW (journey, non-blocking) — Label-position checks cannot see a wrap in words 6 and 12 at head

Checked: Round 3 (535f6c022, new journey): Reviewed the round-3 delta: the new journeys/backup/show-mnemonic-long-words.xml and the two journeys/README.md rows. Every testTag the journey names exists at head: HeaderMenu (HomeScreen.kt:1402), DrawerSettings (DrawerMenu.kt:311), Tab-security (CustomTabRowWithSpacing.kt:58 via SettingsTab.Security), BackupWallet (SettingsScreen.kt:419), backup_mnemonic_words_box, SeedContainer and TapToReveal (ShowMnemonicScreen.kt:165/178/189). The flow order matches the code: BackupWalletClick opens Sheet.Backup, whose default route is BackupRoute.ShowMnemonic (SheetHost.kt:81), so the reveal screen is the first screen of the sheet. The visible texts "Back up your wallet" and "Tap To Reveal" match strings.xml, and the FLAG_SECURE note and the SeedContainer content-description warning match ShowMnemonicScreen.kt. No bitkit-ios twin exists, and the skip is recorded in the cross-platform table.

journeys/backup/show-mnemonic-long-words.xml
jvsena42: Confirmed. The precondition at show-mnemonic-long-words.xml:14-16 asked only for "an onboarded dev wallet with a 12-word phrase", and nothing in the journey restored a known one, so the phrase was whatever the device held. On master `WordItem` is `Alignment.CenterVertically` with a plain `BodyMSB(text = word)` and no `maxLines` (MnemonicWordsGrid.kt:86-97 at origin/master), so it wraps only when a word is wider than its column — a phrase without a long word next to a two-digit label renders on one line on master too and the journey reports a pass on the unfixed build.

Fixed as suggested. The journey now wipes the app and restores the fixed phrase `abstract awesome category mushroom mosquito document multiply mechanic marriage mountain awesome mushroom` through the onboarding restore flow (`RestoreWallet`, `MultipleDevices-button`, `Word-0`..`Word-11` from RestoreWalletScreen.kt:416, `RestoreButton`) before touching font_scale. I recomputed its BIP39 checksum independently and it is valid, so it restores; the description labels it a public test vector, dev flavor and regtest only, never to be funded, and warns that the restore replaces the wallet on the device. Words are entered one field at a time rather than pasted, since the README notes `adb shell input text` drops characters from long strings.

24117c6e4b0906955e011a678f0b3d7b5cce7968

journeys/backup/show-mnemonic-long-words.xml
jvsena42: Confirmed. The two columns are independent (MnemonicWordsGrid.kt:88-99 and 100-111), and this PR moved both the label and the word to `alignByBaseline()` (MnemonicWordsGrid.kt:223 and 232-234), so a wrap in slot 6 or 12 only grows the item downward with nothing below it to shift. Both the row-center check and the equal-pitch check still passed. On master the label is `Alignment.CenterVertically`, so a wrap there does move the label and the old check caught it — the blind spot is new in this PR.

Added the bounds check. A new action verifies that every word element is no taller than the number label beside it, identified by position within its row and compared on bounds only, never on text. The label always renders one line at the full size while words shrink to a shared size between 17sp and 12sp, so a single-line word is never taller than its label and a wrapped one always is. That holds in any slot and also in the case where every word wraps, which a word-to-word height comparison would miss. The equal-pitch action is kept but demoted, with the last-slot limitation spelled out in the step so a runner does not read equal pitch as proof on its own.

24117c6e4b0906955e011a678f0b3d7b5cce7968

journeys/backup/show-mnemonic-long-words.xml
```diff
@@ -0,0 +1,54 @@
...
+    <action>Run adb shell monkey -p to.bitkit.dev -c android.intent.category.LAUNCHER 1</action>
+    <action>On the terms screen, tap both checkboxes (testTags "Check1" and "Check2"), then tap "Continue" (testTag "Continue")</action>
+    <action>Tap "Skip" (testTag "SkipIntro") to reach the last onboarding slide</action>
+    <action>Tap "Restore" (testTag "RestoreWallet"), then confirm the multiple-devices warning (testTag "MultipleDevices-button")</action>
+    <action>Enter the 12 words of the phrase above one field at a time, tapping each field (testTags "Word-0" through "Word-11") and typing only that one word — never paste or type the whole phrase in one go, adb drops characters from long strings</action>
+    <action>Verify no field is marked invalid and no checksum error is shown, then tap "Restore" (testTag "RestoreButton")</action>
+    <action>Wait for the restore to finish and tap "Get Started" (testTag "GetStartedButton"); if the backup restore fails instead, tap "Proceed Without Backup" (testTag "ProceedWithoutBackupButton") and confirm — the phrase has no backup and the on-chain wallet is what this journey needs</action>
```
jvsena42: **LOW** — **The fallback in this step points to a button that is not on screen after the first failure.**

In `WalletRestoreErrorView.kt:78-84`, `ProceedWithoutBackupButton` renders only when `retryCount > 1`. On the first failure the only button is `TryAgainButton`. If the backup restore fails, a runner following this step does not find the tag and can report a false fail before reaching the grid.

Fix: "if the backup restore fails, tap 'Try Again' (testTag `TryAgainButton`) until 'Proceed Without Backup' (testTag `ProceedWithoutBackupButton`) appears, tap it and confirm with 'Yes, Proceed'."

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `24117c6e4`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Proceed Without Backup is not shown on the first restore failure

Checked: Round 4 (24117c6e4): both earlier journey findings are fixed. Round 4 (24117c6e4): both earlier journey findings are fixed. The journey now wipes to.bitkit.dev and restores the fixed phrase `abstract awesome category mushroom mosquito document multiply mechanic marriage mountain awesome mushroom`. I recomputed its BIP39 checksum against the English wordlist (checksum bits 1110 = 1110), so it restores. Every slot except 2 and 11 holds an 8-letter word, including 10 and 12, and the description labels it a public test vector for the dev flavor and regtest only, never to be funded. The restore steps follow MainActivity's startup graph (Terms -> Intro skip -> Slides(LAST_SLIDE_INDEX) -> WarningMultipleDevices -> Restore), and every tag they name exists at head: Check1/Check2/Continue (TermsOfUseScreen.kt:91/99/109), SkipIntro (IntroScreen.kt:92), RestoreWallet (CreateWalletScreen.kt:90), MultipleDevices-button (WarningMultipleDevicesScreen.kt:70), Word-$index (RestoreWalletScreen.kt:416), RestoreButton (:301), GetStartedButton (WalletRestoreSuccessView.kt:59). The new height check is sound. In WordItem (MnemonicWordsGrid.kt:218-240) the label is BodyMSB at full size with maxLines=1. The word uses the same style at a size of 17sp or less, so one line is never taller than the label and any wrap is, in every slot including 6 and 12.

journeys/backup/show-mnemonic-long-words.xml
jvsena42: Verified. `WalletRestoreErrorView.kt:78-84` renders `ProceedWithoutBackupButton` only when `retryCount > 1`, and the count starts at 0: `ContentView.kt:406-407` passes `restoreState.retryCount()`, which is 0 for any state other than `Retry` (`WalletViewModel.kt:621`), and only `onRestoreRetry` advances it via `countRetry()` (`WalletViewModel.kt:263-264`, `622`). So the first failure shows `TryAgainButton` alone, and it takes two Try Again taps before Proceed Without Backup appears, not one.

Rewrote the fallback accordingly: tap `TryAgainButton` until `ProceedWithoutBackupButton` also appears, which takes two taps, then tap it and confirm with "Yes, Proceed". Used the confirm tag `DialogConfirm` from `AppAlertDialog.kt:79` rather than the label alone, since the dialog buttons are the tagged elements.

Commit d2149c174; the PR body diff is back in step with the file.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `d2149c174`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Follow-up: Reviewed d2149c174 only. The fallback now matches the restore error flow: WalletRestoreErrorView renders only when the node is ErrorStarting (ContentView.kt:405-407), `ProceedWithoutBackupButton` shows only when `retryCount > 1` (WalletRestoreErrorView.kt:78-84), and `onRestoreRetry` advances Initial -> Retry(1) -> Retry(2) via `countRetry()` (WalletViewModel.kt:263-264, 622), so two failed Try Again taps are needed before Proceed appears. `TryAgainButton` (WalletRestoreErrorView.kt:75) and the confirm tag `DialogConfirm` (AppAlertDialog.kt:79) exist at head, and the dialog's confirm invokes `onProceedWithoutRestore`, which settles the restore state and lands on the wallet.

ovitrif (approved):
utAck

ovitrif (dismissed):
utAck

ovitrif:
@jvsena42 e2e red + conflicts

jvsena42:
Approved, thanks. Both points:

Conflicts: the shared `journeys/README.md` row again — #1306 merged a few minutes ago and re-conflicted the batch. A merge of current master is running on this branch now, with the two stack bases included so #1288 and #1310 clear as well.

e2e: `onchain_boost_receive_widgets` is the shard whose genuine failure belongs to #1299 (the confirmed-only receive sheet covering the header), fixed on the e2e side in synonymdev/bitkit-e2e-tests#231. On this branch it is the usual setup flake. I will confirm from the re-run on the merged head rather than from the current commit.

Context on the churn is in https://github.com/synonymdev/bitkit-android/issues/1275#issuecomment-5732780955 — including the option of pulling the README rows out of every PR into one follow-up, if the team would rather stop the re-merges.

ovitrif (dismissed):
utAck

ovitrif (approved):
reapprove

Issue #1193, opened 2026-08-28
fix: flaky unit tests from virtual-time timeouts racing real datastore i/o


jvsena42:
Three unit tests fail intermittently on CI and never locally. They share one root cause: a `withTimeoutOrNull` measured on `runTest`'s **virtual** clock, racing work that crosses a **real** dispatcher.

## Affected tests

| Test | Location |
|---|---|
| `QuickPayRepoTest > stale failure event does not settle a dispatch the ldk row shows pending` | `QuickPayRepoTest.kt:714` |
| `QuickPayRepoTest > reconcile during live dispatched op does not steal completion` | `QuickPayRepoTest.kt:1008` |
| `LightningServiceTest > stop destroys the node handle when it is already not running` | `LightningServiceTest.kt:126` |

## Observed occurrences

Not tied to any one branch — all three have failed on `master` itself:

- CI `33082462001` (master, 2026-08-27) — `reconcile during live dispatched op`
- CI `32989065879` (master, 2026-08-26) — `stop destroys the node handle when it is already not running`
- CI `33170079347` (PR #1192, commit `bd66a2142`) — failed on three consecutive attempts, each with a *different* subset:
  - attempt 1: `stale failure event does not settle a dispatch...`
  - attempt 2: `reconcile during live dispatched op...` **and** `stop destroys the node handle...`
  - attempt 3: `reconcile during live dispatched op...`

Roughly 6 of the last 20 CI runs across branches failed this way.

## Root cause

`QuickPayRepoTest` builds a real, file-backed DataStore:

```kotlin
private val cacheStore = CacheStore(context)            // QuickPayRepoTest.kt:110

private val Context.appCacheDataStore by dataStore(     // CacheStore.kt:25
    fileName = "app_cache.json",
    serializer = AppCacheSerializer,
)
```

The AndroidX `dataStore` delegate defaults its scope to `Dispatchers.IO`, so every `cacheStore.data.first()` and `cacheStore.update {}` is real file I/O on a real dispatcher, executed inside a virtual-time `runTest`.

While the test coroutine is parked on that I/O, the test scheduler has no runnable task, so `runTest` fast-forwards virtual time to the next scheduled task — which is the armed timeout in `QuickPayRepo.awaitCompletionOrPending`:

```kotlin
withTimeoutOrNull(LightningRepo.SEND_LN_TIMEOUT) { current.settled.await() }   // QuickPayRepo.kt:350
[...]

PR #1310, opened 2026-09-17
fix: keep lnurl pay comment as activity note

### LINKED ISSUE #417: LNURL-pay comment not displayed on activity
When processing LNURL-pay request one can add a comment to the activity. The comment then should be present on the activity list and currently it isn't.

See video:

https://github.com/user-attachments/assets/205f1cf1-c1c3-4b46-a623-9fe13a55fe8a

jvsena42:
Fixes #417
Related to #1275

This PR keeps the comment typed on an LNURL-pay payment as the note of the sent Lightning activity.

### Description

- Saves the LNURL-pay comment before sending, because the app sent it to the server and then dropped it, so the activity only showed the invoice's description hash.
- Applies the saved comment to the activity once the payment succeeds or goes pending, and again when the payment sync creates or updates the row, so it survives settlement and relaunch.
- Replaces the stored message only when it is empty or equals the description LDK reports (the hash hex), so an invoice's own description or an existing note is never overwritten.
- Clears the saved comment when the payment definitely fails.
- Reads the payment list before reading the activity row, so the row write cannot race the payment sync and write back a stale pending row.

#### Out of Scope

- `AppViewModel`: the comment is stored only for description-hash invoices (LUD-06). LNURL-pay servers that return a plain-description invoice still send the comment without showing it in the activity, to protect invoice descriptions.
- `CacheStore`: pending comments have no pruning. An entry is left behind only if the app dies or the send coroutine is cancelled between saving and completing the send.
- `CoreService`: the pending comments are read once per outbound Bolt11 payment during sync rather than once per sync. DataStore caches the value, so the cost is small.
- A failed payment row created before the comment is cleared may still show the comment.
- iOS parity: bitkit-ios needs the same check of whether it keeps the LNURL-pay comment on the activity.

### Design

The comment appears in the existing Invoice note slot of the Lightning transaction detail. No layout changes.

- TX Lightning: https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-146974
- LNURL Pay Confirm (comment input, unchanged): https://www.figma.com/design/ltqvnKiejWj0JQiqtDf2JJ/?node-id=46920-148562

### Preview

Recorded on a simulation build: staging regtest has no LNURL-pay endpoint that allows comments. The comment "Thanks for the channel" was hardcoded in place of the LNURL checks, and a Blocktank order invoice with a plain description stood in for the LDK description hash. Everything after that (saving, sending, applying on pending and on sync) ran the branch code.

| Figma TX Lightning | Confirm | Pending |
| - | - | - |
| Figma TX Lightning | Confirm | Pending note |

| Succeeded | Home after relaunch | Detail after relaunch |
| - | - | - |
| Succeeded note | Home after relaunch | Detail after relaunch |

#### Manual Tests

Needs an LNURL-pay endpoint that issues description-hash invoices (LUD-06) and allows comments, on the same Lightning network as the app: an lnurl-node `LNURL.createServer` backed by a local LND, as the bitkit-e2e-tests lnurl suite sets up in `test/specs/lnurl.e2e.ts`. It needs minSendable below maxSendable with 1,000 sats in range, since a fixed amount skips the amount screen. The bitkit-docker lnurl-server does not work for this: it builds the invoice with a plain memo containing the comment, so the note reads `LNURL Payment <id> - thanks` on this branch and on master alike. The comment input and the LNURL and invoice-description checks were not driven on a device; they are covered by unit tests. The bitkit-e2e-tests lnurl suite has a commented-out InvoiceComment assertion that could be re-enabled against the `InvoiceNote` tag.

- [x] **1a.** Scan LNURL-pay → Send Amount → 1000 → Send Confirm → type comment "thanks" → swipe to pay: success or pending screen shows.
  - [x] **1b.** Close → open newest activity → Activity Detail: Invoice note shows "thanks", not the description hash.
  - [x] **1c.** Force-stop → relaunch: Home row subtitle and Activity Detail Invoice note still show "thanks".
- [x] **2.** LNURL-pay → pay with LSP holding the HTLC (pending) → wait for settlement → Activity Detail: note stays "thanks" after status turns Successful.
- [x] **3.** LNURL-pay → leave comment empty → pay: Activity Detail shows the same note as before this change.
- [x] **4.** `regression:` Pay a regular LN invoice with a description: Invoice note shows the invoice description.
- [x] **5.** `regression:` LNURL-pay with comment → payment fails (e.g. no route): no stale comment applied to later payments.

A new journey covers flow 1, committed at `journeys/lnurl/lnurl-pay-comment-note.xml` and
listed in `journeys/README.md` (suites table and cross-platform table). It was not run as written
because staging has no LNURL-pay endpoint; the simulated run above covered its payment, note and
relaunch steps.

<details><summary>lnurl-pay-comment-note.xml</summary>

```diff
diff --git a/journeys/lnurl/lnurl-pay-comment-note.xml b/journeys/lnurl/lnurl-pay-comment-note.xml
new file mode 100644
--- /dev/null
+++ b/journeys/lnurl/lnurl-pay-comment-note.xml
@@ -0,0 +1,60 @@
+<journey name="LNURL Pay Comment Note">
+  <description>
+    Verifies that the comment typed when paying an LNURL-pay request is saved as the Invoice note of
+    the sent Lightning activity, and that it survives an app restart.
+
+    Requires an LNURL-pay endpoint that issues description-hash invoices (LUD-06) and allows comments
+    (commentAllowed of at least 12), reachable from the same Lightning network as the app: an
+    lnurl-node `LNURL.createServer` backed by a local LND, as the bitkit-e2e-tests lnurl suite sets up
+    in test/specs/lnurl.e2e.ts. The endpoint also needs minSendable below maxSendable with 1,000 sats
+    (1,000,000 msat) in range: a fixed amount goes straight to Confirm and skips the amount screen, so
+    the amount steps would fail, and the e2e suite's own pay ranges (149,500-200,999 msat) do not
+    cover 1,000 sats.
+
+    The bitkit-docker lnurl-server is not usable here. It builds the invoice with a plain memo that
+    already contains the comment, so the note reads "LNURL Payment &lt;id&gt; - thanks" on this branch
+    and on master alike, and the run says nothing about the fix. A note in that shape means the
+    endpoint issued a memo invoice rather than a description-hash one.
+
+    The staging regtest backend the dev flavor uses has no LNURL-pay endpoint. Also requires a
+    spending balance of at least 1,000 sats plus routing fees.
+  </description>
+  <actions>
+    <action>
+      Launch the Bitkit app and go to the wallet home screen
+    </action>
+    <action>
+      Run adb shell am start -a android.intent.action.VIEW -d "lightning:&lt;lnurl-pay&gt;" to.bitkit.dev with the LNURL-pay string
+    </action>
+    <action>
+      Verify the send amount screen (testTag "SendNumberField") is visible
+    </action>
+    <action>
+      Enter 1000 on the number pad (testTags "N1", "N0", "N0", "N0") and tap Continue (testTag "ContinueAmount")
+    </action>
+    <action>
+      Verify the comment field (testTag "CommentInput") is visible on the confirm screen
+    </action>
+    <action>
+      Type "thanks" into the comment field (testTag "CommentInput")
+    </action>
+    <action>
+      Swipe the pay slider (testTag "GRAB") to the right
+    </action>
+    <action>
+      Verify the payment success screen (testTag "SendSuccess") or the "Payment Pending" screen is visible — the pending screen carries no testTag, so assert it by its title
+    </action>
+    <action>
+      Close the result screen and open the newest activity row (testTag "ActivityShort-0")
+    </action>
+    <action>
+      Verify the Invoice note (testTag "InvoiceNote") is exactly "thanks", not a memo containing it and not the invoice description hash
+    </action>
+    <action>
+      Run adb shell am force-stop to.bitkit.dev, relaunch the app and open the newest activity row (testTag "ActivityShort-0")
+    </action>
+    <action>
+      Verify the Invoice note (testTag "InvoiceNote") is still exactly "thanks"
+    </action>
+  </actions>
+</journey>
```

</details>

#### Automated Checks

- Unit tests added: `CoreServiceTest.kt` covers filling an empty message, replacing the LDK description hash, keeping a different stored message, and ignoring a null or blank comment.
- Unit tests added: `AppViewModelSendFlowTest.kt` covers saving the comment before paying and applying it on success and on pending, skipping invoices with their own description, and clearing it on failure.
- Unit tests added: `ActivityRepoTest.kt` covers the new repository methods passing through to the cache and activity service, and a service failure returning a failure result.
- Unit tests added: `ActivityServiceLightningMessageTest.kt` runs the real activity service with mocked bitkit-core calls. It covers the stale-row race with the payment sync, the no-row case, and `processBolt11` creating or updating a row with the comment, including a comment saved mid-read. The two race tests fail against the previous implementation.
- Ran locally: `just compile`, `just test` (2613 passed), `just lint`.


app/src/main/java/to/bitkit/services/CoreService.kt
```diff
@@ -787,13 +806,35 @@ class ActivityService(
                 contact = contact,
                 seenAt = null,
             )
-        }
+        }.withPendingMessage(pendingMessage = pendingMessage, description = kind.description)
```
jvsena42, line 809: **LOW** — **Pre-existing, adjacent: an LNURL-pay (or any h-tagged) invoice paid with an empty comment keeps the description hash hex as the activity message, so the Home row subtitle and the Activity Detail "Invoice note" show a 64-character hex string.**

processBolt11 creates the row with `message = kind.description.orEmpty()` (app/src/main/java/to/bitkit/services/CoreService.kt:804); ldk-node fills `description` with the hash hex for `Bolt11InvoiceDescriptionRef::Hash`. This PR only replaces that string when a pending comment exists (`withPendingMessage`, line 809), so with `state.comment` blank `savePendingLnurlComment` returns null (AppViewModel.kt:4052) and the hash stays. ActivityRow.kt:138 renders `message.ifEmpty { time }` and ActivityDetailScreen.kt:558 renders any non-empty message as the note, so the hex is user-visible. Same on master; the PR's QA step 3 describes this as unchanged.

Narrowest fix, if wanted here: in processBolt11 pass `message = ""` when `kind.description` is a 64-char hex that is not a direct description (or have bitkit-core expose whether the invoice is description-hash), so the row falls back to the time subtitle and shows no note. Otherwise track separately.

jvsena42 (commented):
**Approved** — no HIGH/MEDIUM findings at head `05e18eabe`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: LOW — Description-hash invoice paid without a comment still shows the 64-hex hash as the Invoice note

Checked: Traced the comment from SendUiState.comment through savePendingLnurlComment (AppViewModel.kt:4050-4056) to CacheStore.pendingLightningMessages and back out via ActivityService.setLightningMessageIfEmpty and processBolt11. The key is decodedInvoice.paymentHash.toHex(), which equals ldk-node's PaymentId/event paymentHash for bolt11 (sendLightning already matches events on it), so the entry cannot attach to another payment; a retry re-fetches a new LNURL invoice, so it cannot inherit a stale hash. The description gate is correct: bitkit-core v0.5.14 scanner returns description=None for h-tagged invoices while ldk-node reports the hash hex as PaymentKind.Bolt11.description, so withPendingMessage replaces exactly the hash-hex placeholder and never a real description or user note; withPaymentUpdate from the base branch never touches message. Concurrency between setLightningMessageIfEmpty and processBolt11 (full sync, handlePaymentEvent, async chunks) is safe: both run on the single-threaded CORE queue and for outbound payments processBolt11 has no suspension point between reading the row and writing it (the contact resolver returns synchronously for OUTBOUND), so every interleaving ends with the comment on the row or still in the cache. Upgrade from the released build reads app_cache.json without the new key into the default empty map (kotlinx defaults, ignoreUnknownKeys=true for downgrade); the comment is persisted on the activity row, which ActivityBackupV1 serializes whole, and restoreFromBackup upserts it back with the message intact. Definite-failure, pending (10s timeout / held HTLC), process-death-while-pending and dismissed-sheet paths all leave the comment either applied or in the cache for the next sync. iOS twin absent: bitkit-ios LnurlPayConfirm.swift sends the comment but nothing stores it, as the PR body already notes.

app/src/main/java/to/bitkit/services/CoreService.kt
jvsena42: Confirmed, and confirmed as pre-existing: `processBolt11` sets `message = kind.description.orEmpty()` and ldk-node puts the hash hex in `description` for a description-hash invoice, so the hex reaches the Home subtitle and the Activity Detail note on master too. This PR only swaps that string in when a pending comment exists, so it neither causes nor worsens the case.

Leaving it out of this PR: the narrow version needs a way to tell a description-hash invoice from a real description, and hex-sniffing the 64-character string in `processBolt11` is a guess that a genuine 64-character hex description would trip. The clean fix is bitkit-core exposing the description ref kind, which is a separate change.

Tracked in #1312 with the trace and both options.

journeys/lnurl/lnurl-pay-comment-note.xml
```diff
@@ -0,0 +1,48 @@
+<journey name="LNURL Pay Comment Note">
+  <description>
+    Verifies that the comment typed when paying an LNURL-pay request is saved as the Invoice note of
+    the sent Lightning activity, and that it survives an app restart. Requires an LNURL-pay endpoint
+    that returns a description-hash invoice and allows comments (commentAllowed of at least 12), such
+    as the bitkit-docker lnurl-server, reachable from the same Lightning network as the app. The
```
jvsena42: **MEDIUM, journey, non-blocking** — **The journey names the bitkit-docker lnurl-server as an example of an endpoint that returns a description-hash invoice. It does not: its invoices carry a plain memo, so a run against it never stores the comment, and the note check cannot tell this branch from master.**

The lnurl-server pay callback builds the invoice with a plain memo that already contains the comment: `lndService.createInvoiceMsat(amountMsat, comment ? `LNURL Payment ${paymentId} - ${comment}` : `LNURL Payment ${paymentId}`, 3600)` (bitkit-docker `lnurl-server/routes/pay.js:54`). `createInvoiceMsat` sends only `value_msat`, `memo` and `expiry` to LND (`lnurl-server/services/lnd.js:114`), with no `description_hash`. On this branch `savePendingLnurlComment` returns early when `invoice.description` is not empty (AppViewModel.kt:4053), so no comment is cached and `withPendingMessage` never runs. The activity message is the memo, `LNURL Payment <id> - thanks`, on this branch and on master alike. The step "InvoiceNote shows \"thanks\"" then fails on an exact match or passes on a substring match, and neither result says anything about the fix. The PR body's QA Notes give the same example.

Fix: drop the lnurl-server example and name an endpoint that issues description-hash invoices, e.g. an lnurl-node `LNURL.createServer` backed by a local LND as in the bitkit-e2e-tests lnurl suite (`test/specs/lnurl.e2e.ts:69`). Change the note check to require exactly "thanks". Also state that the endpoint needs minSendable < maxSendable with 1,000 sats in range. A fixed amount skips the amount screen (AppViewModel.kt:3508-3519), so step 3 would fail, and the e2e suite's own pay ranges (149,500-200,999 msat) do not include 1,000 sats.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `2a28e7575`. (Posted as a comment because GitHub does not allow approving your own PR.)

Inline: MEDIUM (journey, non-blocking) — The example endpoint, the bitkit-docker lnurl-server, returns plain-description invoices, so the journey never exercises the fix

Checked: Round 3 (2a28e7575, new journey): Round 3 delta is the new journey and two README rows. Every testTag the journey names exists at head: SendNumberField and ContinueAmount (SendAmountScreen.kt:260, :334), CommentInput (SendConfirmScreen.kt:456, rendered only when commentAllowed > 0 at :850), GRAB (SwipeToConfirm.kt:171), SendSuccess (SendSheet.kt:348), ActivityShort-N (ActivityListSimple.kt:52) and InvoiceNote (ActivityDetailScreen.kt:588). The flow order matches the code: a variable-amount LNURL-pay goes to SendRoute.Amount (AppViewModel.kt:3523), onAmountContinue does not apply QuickPay, the pending screen title is "Payment Pending" and has no tag, and InvoiceNote is the Invoice note block on Activity Detail shown for any Lightning activity with a non-empty message. `lightning:lnurl1...` is accepted by the bitkit-core scanner's LNURL regex. No bitkit-ios twin exists, and the journey is recorded as not ported in the cross-platform table.

journeys/lnurl/lnurl-pay-comment-note.xml
jvsena42: Confirmed in bitkit-docker. The pay callback builds the invoice with a plain memo that already carries the comment (lnurl-server/routes/pay.js:54-58), and `createInvoiceMsat` sends only `value_msat`, `memo` and `expiry` to LND with no `description_hash` (lnurl-server/services/lnd.js:114-120) — the only `description_hash` reference in that server is in the decoder. On this branch `savePendingLnurlComment` returns null when `invoice.description` is non-empty (AppViewModel.kt:4050-4056), so a run against it is identical on master.

Changes in 7a011ab52:
- Dropped the bitkit-docker example and named an lnurl-node `LNURL.createServer` backed by a local LND, as in bitkit-e2e-tests test/specs/lnurl.e2e.ts:69-81.
- Added the range requirement: minSendable below maxSendable with 1,000 sats (1,000,000 msat) in range, since `isFixed` sends the flow straight to Confirm (AppViewModel.kt:3508-3519) and the e2e suite's own pay ranges (149,500-200,999 msat, lnurl.e2e.ts:150-154) do not cover 1,000 sats.
- The note steps now require exactly "thanks", not a memo containing it.
- Kept a short note on the failure shape: a note reading "LNURL Payment &lt;id&gt; - thanks" means the endpoint issued a memo invoice, so the run proves nothing.

Also updated the PR body's QA Notes and the embedded copy of the journey with the same example and requirements.

One thing I could not verify locally: bitkit-e2e-tests has no `node_modules` installed here, so I did not read lnurl-node's invoice construction. The journey states the requirement as description-hash (LUD-06) and gives the runner the memo-shaped note as the tell if an endpoint does not meet it.

jvsena42 (commented):
**Approved** — no blocking HIGH/MEDIUM findings at head `7a011ab52`. (Posted as a comment because GitHub does not allow approving your own PR.)

Checked: Round 4 (7a011ab52): the journey MEDIUM is fixed. Round 4 covers 7a011ab52, which only edits journeys/lnurl/lnurl-pay-comment-note.xml in answer to the memo-only endpoint finding. The journey now points at an lnurl-node LNURL.createServer backed by LND, as bitkit-e2e-tests test/specs/lnurl.e2e.ts sets up. lnurl@0.26.2 (the version in that repo's lockfile) passes descriptionHash = sha256(metadata) to the backend, and lightning-backends@1.6.5 lnd.js sets description_hash on POST /v1/invoices, so those invoices are description-hash invoices with no comment in them. The journey also says the bitkit-docker lnurl-server cannot be used, and that is correct: lnurl-server/routes/pay.js builds the invoice memo as `LNURL Payment <id> - <comment>`. It requires minSendable < maxSendable with 1,000,000 msat in range and states that the e2e suite's payRequest1 range (149,500-200,999 msat) does not cover it, so the runner has to generate its own payRequest. The InvoiceNote assertions now require the note to equal "thanks" exactly, which fails on a memo-shaped note or the description hash. Since the comment cannot reach the note through the invoice, a pass can only come from the persisted comment. testTags and step order are unchanged from the approved round.

ovitrif (dismissed):
utAck

jvsena42:
The red `build` here was the known `QuickPayRepoTest > reconcile during live dispatched op does not steal completion` flake — the one #1193 exists to remove. #1306 landed that fix in master, but this branch had not picked it up: `QuickPayRepoTest.kt` on the old head had no in-memory cache store at all.

Merged current master in; the file now carries the fix and `just test` is green on the merged head. The remaining `settings` and `onchain_boost_receive_widgets` failures are the e2e side, tracked with the companion branch synonymdev/bitkit-e2e-tests#239.

ovitrif (approved):
utAck

